PR #1, opened 2018-08-09
Add different default paths for config and css files

BlueGone:


src/client.cpp
```diff
@@ -36,7 +36,37 @@ waybar::Client::Client(int argc, char* argv[])
...
+          std::string result = p.we_wordv[0];
+          wordfree(&p);
+          return result;
+        } else {
+          wordfree(&p);
+        }
+      }
+    }
+
+    return fallbackPath;
```
Alexays: fallbackPath is undefined cause compile error

PR #4, opened 2018-08-09
Fix formula used to compute RAM used and available

BlueGone:


BlueGone:
I supposed the percentage refer to the amount of free RAM (lower is better, like CPU charge). I probably did a wrong supposition (I just read the variable name 'available', so I make no sense anymore). But the tooltip was showing the amount of RAM free instead of the used one, which is for sure a real error.

PR #53, opened 2018-10-22
fix custom intervals causing cpu to go mental

### LINKED ISSUE #47: Abnormal CPU usage of the CPU module
For some reason, even with a 30 second interval, the CPU module makes Waybar consume a lot of cycles (~75% of a core). As soon as I disable it and restart Waybar, it remains idle. I also have the RAM module enabled, but no problems there. Persists between reboots.

dangerousdan:
```
clock {
  "interval": 1,
  "format": "{:%H:%M:%M"},
  ...
}
```
Caused the Waybar to use 100% CPU.

This was caused by `interval` not being set properly due to a typo.

Possibly also fixes #47

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -4,7 +4,7 @@ waybar::modules::Clock::Clock(const Json::Value& config)
   : ALabel(config, "{:%H:%M}")
 {
   label_.set_name("clock");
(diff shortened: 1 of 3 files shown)


Alexays:
Fuck, good catch thx :)

PR #68, opened 2018-10-29
Add configuration options for widgets on mouse events

harishkrupo:
Subscribe for mouse scroll events on the pulseaudio widget
and change volume when event is received.
Scroll up increments the volume and scroll down decrements it.

Signed-off-by: Harish Krupo <harishkrupo@gmail.com>

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -1,4 +1,5 @@
 #include "ALabel.hpp"
+#include <util/command.hpp>
 
 #include <iostream>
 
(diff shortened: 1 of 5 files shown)


harishkrupo:
Hi, 
I like and use waybar in my sway setup. 
The only thing that it felt was missing was this. So, I added it :)
Thanks

Alexays:
Thanks, i'll test it today !

harishkrupo:
I see that the commit contains a few changes that weren't added by me. They got added when I ran clang-fromat-diff (for style Google) on the patch. Let me know if you find some of the changes unneeded. I will remove them.

src/factory.cpp
```diff
@@ -4,48 +4,65 @@ waybar::Factory::Factory(Bar& bar, const Json::Value& config)
...
     }
-    #endif
+#endif
     if (ref.compare(0, 7, "custom/") == 0 && ref.size() > 7) {
-      return new waybar::modules::Custom(ref.substr(7), config_[name]);
+      module = new waybar::modules::Custom(ref.substr(7), config_[name]);
     }
+
+    // Now configure events' user commands
+    const Json::Value& module_config = config_[name];
```
Alexays: Why not doing this in the ALabel constructor? ;)

harishkrupo: Makes sense. Will move it there.

src/ALabel.cpp
```diff
@@ -14,30 +15,129 @@ waybar::ALabel::ALabel(const Json::Value& config, const std::string format)
...
+    else
+      ret = forkExec(scroll_down_cmd_);
+
+    dp.emit();
+  }
+
+  return ret;
+}
+
+bool waybar::ALabel::setEventCmd(waybar::EventType evType, std::string cmd) {
```
Alexays: I don't understand why this function and these 3 variables is needed, why not bind event directly line 23 ?

src/ALabel.cpp
```diff
@@ -14,30 +15,129 @@ waybar::ALabel::ALabel(const Json::Value& config, const std::string format)
...
+      gdk_event_get_scroll_deltas(reinterpret_cast<const GdkEvent*>(e),
+                                  &delta_x, &delta_y);
+      if (delta_y < 0) {
+        direction_up = true;
+      } else if (delta_y > 0) {
+        direction_up = false;
+      }
+    }
+
+    if (direction_up)
```
Alexays: Check if user provided both up and scroll cmds :)

src/ALabel.cpp
```diff
@@ -14,30 +15,129 @@ waybar::ALabel::ALabel(const Json::Value& config, const std::string format)
...
+      format_ = default_format_;
+    }
+    ret = true;
   }
+
   dp.emit();
+  return ret;
+}
+
+bool waybar::ALabel::forkExec(std::string cmd) {
```
Alexays: Moves this function in the util header file, it will always be useful.

src/ALabel.cpp
```diff
@@ -14,30 +15,129 @@ waybar::ALabel::ALabel(const Json::Value& config, const std::string format)
...
-  alt = !alt;
-  if (alt) {
-    format_ = config_["format-alt"].asString();
-  } else {
-    format_ = default_format_;
+bool waybar::ALabel::handleToggle(GdkEventButton* const& e) {
+  bool ret = false;
+  if (e->button == 1) {
+    ret = forkExec(button_press_cmd_);
+  } else if (e->button == 2) {
```
Alexays: Maybe if the user has not providing a click event we want the alt label change also with button1

harishkrupo: I thought that would lead to confusion as for some widgets it would be button1 and some button2. I felt that standardizing (and documenting) it would be better.

src/ALabel.cpp
harishkrupo: Sure, will do that.

src/ALabel.cpp
harishkrupo: Sorry, didn't understand. At one time we would be processing only one direction, right?

src/ALabel.cpp
harishkrupo: Initially, the idea was to set the event commands in factory.cpp. After moving it to the constructor, I thought that it would be better as a separate function as we might need to add more events in the future?
I actually don't have a good justification for this. Will change it to however you see fit :)

src/ALabel.cpp
Alexays: If the user provided only scroll up the scroll down will fail :?

harishkrupo: in that case, scroll down would be called but since the cmd for it would be "", it would act as a no-op.
Probably we could add a check in forkexec for null commands.

Alexays: Ok my bad ;)

harishkrupo:
@Alexays modified the code as per the comments.
For the format-alt toggle, made the change as follows:
when the on-click event is configured, format-alt is toggled on any mouse click other than left click. When on-click is not defined, any mouse button would toggle format-alt.
Does this sound good?

Alexays:
Thanks!

PR #70, opened 2018-10-30
Add module to show sway binding mode 

Robinhuett:
This PR adds a module which shows the current binding mode of sway (eg. "resize").

Also 668b7b7 adds a default config for the module.

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -4,7 +4,7 @@
     // "height": 30, // Waybar height
     // "width": 1280, // Waybar width
     // Choose the order of the modules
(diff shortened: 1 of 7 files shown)


resources/config
```diff
@@ -23,6 +23,9 @@
     //         "default": ""
     //     }
     // },
+    "sway/mode": {
```
Alexays: it's not really useful since the default format is {} :?

Robinhuett: I'd be happy to remove this. I have only included this to let people know `format` is available.

PR #72, opened 2018-10-30
Custom modules can control tooltip and CSS class

Robinhuett:
This PR enables custom modules to control their own tooltip and CSS class.

Modules can return data similar to [i3blocks](https://github.com/vivien/i3blocks#format):
Line 1: Content (if it's the only line also tooltip)
Line 2: Tooltip
Line 3: CSS class

Every aditional line will be ignored.

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -15,8 +15,13 @@ class Custom : public ALabel {
   private:
     void delayWorker();
     void continuousWorker();
+    void parseOutput();
 
     const std::string name_;
+    std::string text_;
+    std::string tooltip_;
+    std::string class_;
+    std::string prevclass_;
     waybar::util::SleeperThread thread_;
     waybar::util::command::res output_;
 };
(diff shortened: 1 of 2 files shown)


Alexays:
It may be better to use a JSON output rather than multi lines?

Robinhuett:
I thought about this too, but didn't wanted to add a breaking change.
Maybe a config parameter could enable JSON. But I can look into that change later, if you'd rather have JSON :)

Alexays:
I think with a continuous script it can be buggy, no?

Robinhuett:
Technically it should work. But it would be definitely more anoying to write something like this:
```bash
echo -e "{\"text\":\""$text"\", \"tooltip\":\""$tooltip"\", \"class\":\" \""$class1"\"}"
```

instead of:

```bash
echo -e $text"\n"$tooltip"\n"$class
```

Ultimatly both could be implemented. Multiline for easy use and i3blocks compatibility and JSON for more complex returns like mutliple CSS classes.

Robinhuett:
When a custom module is configured with `"return-type": "json"` JSON returns get parsed.
With this also continuous scripts work.
Currently it supports `text`, `tooltip`, `class`, but more can be added quiet easy.
The only limitation is, that the json has to be a one-liner.

Alexays:
Nice work! Thank you 😄

PR #76, opened 2018-11-02
Add class for full battery and give option to interpret unknown as full

mithodin:
Full battery status should also bet set as a class.
Additionally, users can configure the battery module such that an "Unknown" status is interpreted as "Full", because that's how some notebooks behave (e.g. Thinkpad T430s. Yup, that's the one I use ^^).

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -74,7 +74,7 @@ std::tuple<uint16_t, std::string> waybar::modules::Battery::getInfos()
       std::ifstream(bat / "status") >> _status;
       if (_status != "Unknown") {
         status = _status;
-      }else if (config_["full-is-unknown"].isString() && config_["full-is-unknown"] == "true") {
+      } else if (config_["full-is-unknown"].isString() && config_["full-is-unknown"] == "true") {
         status = "Full"; //Some notebooks (e.g. Thinkpad T430s) report a full battery as "Unknown".
       }
       total += capacity;
@@ -127,12 +127,12 @@ auto waybar::modules::Battery::update() -> void
     }
   } else {
     label_.get_style_context()->remove_class("charging");
-    if (status == "Full"){
+    if (status == "Full") {
       label_.get_style_context()->add_class("full");
       if (config_["format-full"].isString()) {
         format = config_["format-full"].asString();
       }
-    }else{
+    } else {
       label_.get_style_context()->remove_class("full");
     }
   }
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -74,8 +74,6 @@ std::tuple<uint16_t, std::string> waybar::modules::Battery::getInfos()
       std::ifstream(bat / "status") >> _status;
       if (_status != "Unknown") {
         status = _status;
-      } else if (config_["full-is-unknown"].isString() && config_["full-is-unknown"] == "true") {
-        status = "Full"; //Some notebooks (e.g. Thinkpad T430s) report a full battery as "Unknown".
       }
       total += capacity;
     }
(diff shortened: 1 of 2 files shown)


src/modules/battery.cpp
```diff
@@ -74,6 +74,8 @@ std::tuple<uint16_t, std::string> waybar::modules::Battery::getInfos()
       std::ifstream(bat / "status") >> _status;
       if (_status != "Unknown") {
         status = _status;
+      }else if (config_["full-is-unknown"].isString() && config_["full-is-unknown"] == "true") {
```
Alexays: I don't really like this kind of option, I think it's better to do that for example "full-capacity": 98, what do you think?
PS: also a space is missing next to the bracket

src/modules/battery.cpp
```diff
@@ -125,8 +127,13 @@ auto waybar::modules::Battery::update() -> void
...
   } else {
     label_.get_style_context()->remove_class("charging");
-    if (status == "Full" && config_["format-full"].isString()) {
-      format = config_["format-full"].asString();
+    if (status == "Full"){
+      label_.get_style_context()->add_class("full");
+      if (config_["format-full"].isString()) {
+        format = config_["format-full"].asString();
+      }
+    }else{
```
Alexays: here too

src/modules/battery.cpp
mithodin: Sorry for the formatting errors.
Problem is, full-capacity does not allow me to change the format, no? So I would have the charging format, since that is the default if status is Unknown.

mithodin: Okay, how's this:
state and status are now completely independent. A class is set for both (if configured).
There's the possibility for format-[status]-[state], format-[status] and format-[state] (and the default format, of course).

mithodin: Which allows people to do with unknown what they want.

mithodin: Here's an example config:
```
    "battery": {
        "bat": "BAT0",
        "states": {
            "full": 100,
            "good": 90,
            "warning": 30,
            "critical": 15
        },
        "format-charging-full": "🔌[full]",
        "format-unknown-full": "🔌[full]",
        "format-charging": "🔌{capacity}% {icon}",
        "format-discharging": "🔋{capacity}% v {icon}",
        "format": "🔋{capacity}% {icon}",
        "format-icons": [" ","▁","▂","▃","▄","▅","▆","▇","█"]
    },
```

Alexays: Love it!

src/modules/battery.cpp
```diff
@@ -34,6 +34,7 @@ waybar::modules::Battery::Battery(const Json::Value& config)
   for (auto const& bat : batteries_) {
     inotify_add_watch(fd_, (bat / "uevent").c_str(), IN_ACCESS);
   }
+  old_status_ = "";
```
Alexays: Not needed, default string constructor is empty

src/modules/battery.cpp
```diff
@@ -116,21 +117,17 @@ auto waybar::modules::Battery::update() -> void
...
-  }
-  auto state = getState(capacity, charging);
-  if (!state.empty() && config_["format-" + state].isString()) {
+  auto state = getState(capacity);
+  label_.get_style_context()->remove_class(old_status_);
+  label_.get_style_context()->add_class(status);
+  old_status_ = status;
+  if (!state.empty() && config_["format-" + status + "-" + state].isString()) {
+    format = config_["format-" + status + "-" + state].asString();
+  }else if (config_["format-" + status].isString()) {
```
Alexays: Space

src/modules/battery.cpp
```diff
@@ -116,21 +117,17 @@ auto waybar::modules::Battery::update() -> void
...
+    format = config_["format-" + status].asString();
+  }else if (!state.empty() && config_["format-" + state].isString()) {
```
Alexays: Space

mithodin: It's hard to break old habits. Fixed. (+Initialization for status in getInfos(), so we don't get an empty status on Unknown).

PR #77, opened 2018-11-02
fix compilation on systems with old gdbus-codegen

mithodin:
Older versions of gdbus-codegen do not support --body and --header call. On those systems, the old synax --generate-c-code must be used. A simple call to gdbus-codegen find out whether --body and --header are supported and falls back to the old syntax if they are not.

### DIFF
--- a/protocol/meson.build
+++ b/protocol/meson.build
@@ -38,26 +38,58 @@ endforeach
 
 gdbus_codegen = find_program('gdbus-codegen')
 
-gdbus_code = generator(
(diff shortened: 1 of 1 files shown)


Alexays:
Is the old syntax still present in recent versions if so, keep the old one instead of doing an if else no?

mithodin:
I'm not sure if --generate-c-code might be deprecated and be removed at some point? Didn't want to break compilation for someone else :-)

PR #78, opened 2018-11-02
add option for when filesystem still lives in the experimental namespace

mithodin:
For gcc-7, filesystem still lives in the experimental namespace, thus compilation will fail. I have added an option to meson to fix the header includes and namespaces accordingly.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -83,11 +83,9 @@ if dbusmenu_gtk.found()
     )
 endif
 
-test_fs = '''#include <filesystem>
-int main(void){}'''
 compiler = meson.get_compiler('cpp')
 
-if not compiler.compiles(test_fs, name : 'is filesystem still in experimental?')
+if not compiler.has_header('filesystem')
     add_project_arguments('-DFILESYSTEM_EXPERIMENTAL', language: 'cpp')
 endif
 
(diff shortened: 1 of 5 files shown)


Alexays:
I'm looking for if there's no better option than a meson option, because it's not very friendly to the end user

mithodin:
I've looked around a bit for some kind of auto-detection method, but I've not been able to find something. Possibly one could write a small test file and see whether it compiles or not.

mithodin:
I've figured it out :-)

meson.build
```diff
@@ -83,7 +83,11 @@ if dbusmenu_gtk.found()
     )
 endif
 
-if get_option('filesystem-experimental')
+test_fs = '''#include <filesystem>
+int main(void){}'''
+compiler = meson.get_compiler('cpp')
+
+if not compiler.compiles(test_fs, name : 'is filesystem still in experimental?')
```
Alexays: if not compiler.has_header('filesystem')

mithodin: nice, that is even simpler. I've updated the pr.

colemickens:
Thanks for this fix, I now don't have to patch Waybar to package it for Nix: https://github.com/colemickens/nixpkgs-wayland/commit/4a8d7a5ccd6104366a8d5663040af6352ff580be !

mithodin:
@colemickens Yup, part of why I created this pr is to avoid having to write a patch for my gentoo ebuild ^^

PR #79, opened 2018-11-03
Don't call layer_surface.set_size on configure

vberger:
Calling `set_exclusive_zone` and `set_size` with the values provided by the compositor in a configure will actually freeze the bar in the new configuration it requested, as calling those with non-zero values opts-out of autoconfiguration by the compositor.

By removing these calls, waybar can now adapt to dynamic changes in the output dimensions, which can happen for example if the monitor's transform is changed (like for a tablet-like device auto-detecting its orientation).

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -119,8 +119,6 @@ height: {} required by the modules", o->height_, min_height) << std::endl;
       "Bar configured (width: {}, height: {}) for output: {}",
       o->width_, o->height_, o->output_name) << std::endl;
 
-    zwlr_layer_surface_v1_set_exclusive_zone(surface, o->height_);
-    zwlr_layer_surface_v1_set_size(surface, o->width_, o->height_);
     wl_surface_commit(o->surface);
   }
 }


Alexays:
Thanks!

ForTheReallys:
Is removing `set_exclusive_zone` necessary?  Removing this line breaks #75.

vberger:
There should not be any need to set it again every time the bar receive a configure, as it has been set once at the start and there is no reason this would change.

Unless maybe sway configures the bar to a different height than the one it requested?

ForTheReallys:
No there isn't anything wrong with sway.  The problem is that the gtk window is not created until the exclusive zone is already set.  When it is finally created, it has it's own minimum height which is what can cause an overlap between regular windows and waybar.  I suppose we can just set the exclusive zone as soon as the gtk window is created.

Alexays:
Maybe we need to use wl_display_roundtrip to wait

vberger:
>  When it is finally created, it has it's own minimum height which is what can cause an overlap between regular windows and waybar.

In that case, I'd argue that the real issue is that the gtk window may have a different height than the one specified in the configuration file... But yeah, setting the exclusive zone as soon as the gtk window knows its size should work.

ForTheReallys:
What if we set the exclusive zone ONLY if the requested height exceeds the minimum height (still in layerSurfaceHandleConfigure)?

Alexays:
@ForTheReallys Fixed with 311c34ecbc61a3dd4f1f9b4954693bf134e17604

PR #85, opened 2018-11-04
modules: Keyboard Layout

harishkrupo:
This patch adds a new module to show the current keyboard layout.

Signed-off-by: Harish Krupo <harishkrupo@gmail.com>

### DIFF
--- a/include/IModule.hpp
+++ b/include/IModule.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <gtkmm.h>
+#include <wayland-client.h>
 
 namespace waybar {
 
@@ -9,6 +10,7 @@ class IModule {
     virtual ~IModule() = default;
     virtual auto update() -> void = 0;
     virtual operator Gtk::Widget &() = 0;
+  virtual void handleSeat(struct wl_seat*, uint32_t) {};
     Glib::Dispatcher dp; // Hmmm Maybe I should create an abstract class ?
 };
 
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -17,6 +17,7 @@ class Bar {
     Bar(const Bar&) = delete;
 
     auto toggle() -> void;
+  void handleSeat(struct wl_seat*, uint32_t);
 
     const Client& client;
     Gtk::Window window;
--- a/include/client.hpp
+++ b/include/client.hpp
@@ -34,6 +34,8 @@ class Client {
       uint32_t name, const char *interface, uint32_t version);
     static void handleGlobalRemove(void *data,
       struct wl_registry *registry, uint32_t name);
+  static void seatName(void *data, struct wl_seat *wl_seat, const char *name) {}
+  static void seatCapabilities(void *data, struct wl_seat *wl_seat, uint32_t caps);
 };
 
 }
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -19,6 +19,7 @@
 #ifdef HAVE_LIBPULSE
 #include "modules/pulseaudio.hpp"
 #endif
+#include "modules/kbdlayout.hpp"
 #include "modules/custom.hpp"
 
 namespace waybar {
--- a/include/modules/kbdlayout.hpp
+++ b/include/modules/kbdlayout.hpp
@@ -1,0 +1,44 @@
+#pragma once
+
+#include <fmt/format.h>
+#include <algorithm>
+#include "ALabel.hpp"
+#include <xkbcommon/xkbcommon.h>
+#include <wayland-client.h>
+#include <sys/mman.h>
+
+namespace waybar::modules {
+
+class KbdLayout : public ALabel {
+  public:
+    KbdLayout(const Json::Value&);
+    ~KbdLayout();
+    auto update() -> void;
+  void handleSeat(struct wl_seat* seat, uint32_t caps);
+  struct wl_keyboard *wl_kbd_ = nullptr;
+  struct xkb_context *xkb_ctx_ = nullptr;
+  struct xkb_keymap *keymap_ = nullptr;
+  struct xkb_state *xkb_state_ = nullptr;
+  uint32_t current_group_ = 0;
+
+private:
+
+static void kbdKeymap(void *data, struct wl_keyboard *wl_kbd, uint32_t format,
+                       int fd, uint32_t size);
+static void kbdEnter(void *data, struct wl_keyboard *wl_kbd, uint32_t serial,
+          struct wl_surface *surf, struct wl_array *keys) {}
+static void kbdLeave(void *data, struct wl_keyboard *wl_kbd, uint32_t serial, struct wl_surface *surf) {}
+
+static void kbdKey(void *data, struct wl_keyboard *wl_kbd, uint32_t serial, uint32_t time,
+        uint32_t key, uint32_t state) {}
+
+static void kbdModifiers(void *data, struct wl_keyboard *wl_kbd, uint32_t serial,
+                         uint32_t mods_depressed, uint32_t mods_latched,
+                         uint32_t mods_locked, uint32_t group);
+
+static void kbdRepeatInfo(void *data, struct wl_keyboard *wl_kbd, int32_t rate,
+                int32_t delay) {}
+  std::string layout_ = "EN";
+};
+
+}  // namespace waybar::modules
--- a/meson.build
+++ b/meson.build
@@ -39,6 +39,7 @@ sigcpp = dependency('sigc++-2.0')
 libnl = dependency('libnl-3.0', required: get_option('libnl'))
 libnlgen = dependency('libnl-genl-3.0', required: get_option('libnl'))
 libpulse = dependency('libpulse', required: get_option('pulseaudio'))
+xkbcommon = dependency('xkbcommon')
 
 src_files = files(
     'src/factory.cpp',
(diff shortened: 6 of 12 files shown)


eternal-sorrow:
Doesn't work for me. When I hit keyboard layout switch shortcut, label of module does not change (input layout changes though).

Also, would be nice to set it to output only short language code as 'en' or 'de' instead of full layout name.

harishkrupo:
> Doesn't work for me. When I hit keyboard layout switch shortcut, label of module does not change (input layout changes though).

Okay, I am a bit confused here. I don't change keyboard layouts, so I am not very familiar with the topic. On the issues page, I was referred to https://github.com/swaywm/sway/wiki#input-configuration which contains documentation of how to change the input configuration. Could you give me more information? Which DE do you use and how do you change the layout? ( we need to figure out which command runs when you hit the layout switch key).

> Also, would be nice to set it to output only short language code as 'en' or 'de' instead of full layout name.

Okay, will check if the library provides short names.

harishkrupo:
@eternal-sorrow Could you also paste the `bindsym Super+Space ...` line?

harishkrupo:
also, by layout are you also referring to the qwerty/dvorak schemes?

eternal-sorrow:
> @eternal-sorrow Could you also paste the `bindsym Super+Space ...` line?

No such line in my config. ` xkb_options grp:win_space_toggle` does that.

> also, by layout are you also referring to the qwerty/dvorak schemes?

no, just Russian/English layout

harishkrupo:
Thanks, got it. let me check.

include/IModule.hpp
```diff
@@ -9,6 +10,7 @@ class IModule {
     virtual ~IModule() = default;
     virtual auto update() -> void = 0;
     virtual operator Gtk::Widget &() = 0;
+  virtual void handleSeat(struct wl_seat*, uint32_t) {};
```
Alexays: I don't like this way of doing it, why can't we use the existing dp.emit when we get the seat?

harishkrupo: Yes, agreed. This is a RFC patch. I was just trying out to see if we can get the keyboard layout.
Could you please explain how we can use the dp.emit() call to send the seat details?

Alexays: You can store the seat in the bar and pass bar as parameters to the module like sway workspaces and so when emit is called in the module, you just need to access the variable

Alexays: I'll look into it!

harishkrupo: I searched for a solution. looks like this can be achieved only with a new protocol (or maybe there is an existing one?). 
The compositor currently sends modifier events only to the client which has the focus. The keyboard layout group is sent as part of the modifier event. We would need a new protocol to register for such modifier events so that we can change the layout based on the received event.

@Alexays Were you able to find something?

jeremejevs: I guess the way to go for now is polling then? Kind of sucks to have a delay.

harishkrupo: poll what? unless we receive the modifiers from the compositor we will never know of the layout change.

harishkrupo: There is one request in layer shell interface to set keyboard interactivity. Unfortunately, on setting it, all the keyboard events are sent only to the bar as it is the topmost :(

harishkrupo:
I have added a new request to wlroots's layer shell v1 interface (patch available [here](https://github.com/harishkrupo/wlroots/commit/e8efe2575a18a750e5e7d0f6656bcdb874b0b3c9).) This request informs the compositor of our intent to receive keyboard modifiers. When we receive the modifiers, we update our xkb state (specifically the group state) and then call emit. When we query for the layout in the update, it will show the new layout.
The wlroot's patch is a POC. I will clean up the patches and send it to the sway community for review. Hopefully, they will suggest how to properly implement it or provide an alternate solution to receive the keyboard modifiers. 
@Alexays In the mean time, please feel free to review it. :)

harishkrupo:
Also, I am yet to incorporate your reviews in the patch. Will do that once I have a proper plan on the solution (either via the protocol or something different)

Alexays:
Wow such a contribution thank you <3

include/IModule.hpp
jeremejevs: > poll what? unless we receive the modifiers from the compositor we will never know of the layout change.

Ah, okay, my assumption was that the information about the current layout could be queried somewhere, but upon further reading it appears that it's managed by Sway and, right now, isn't exposed at all. Bummer.

harishkrupo:
@eternal-sorrow If you need it now, then you can use this script in your custom module:
```
swaymsg -t get_inputs | jq '.[]  | select(.identifier == "1241:5890:USB_Keyboard") | .xkb_active_layout_name'
```
Replace "1241:5890:USB_Keyboard" with your keyboard identifier and install jq.

jeremejevs:
Adjusted the script slightly:

```bash
#!/usr/bin/env bash

while true; do
  x="`swaymsg -t get_inputs | jq --raw-output '[.[] | select(.type == "keyboard") | .xkb_active_layout_name | select(contains("English (US)") | not)] | first'`"
  if [[ "${x}" == "null" ]]; then
    echo English
  else
    echo "${x}"
  fi

  sleep 1
done
```

Drops the need to hardcode your keyboard ID, and also a bit more efficient than having Waybar launch a shell every N seconds. Change "English (US)" to your default. Used like so:

```json
"custom/layout": {
  "exec": "path/to/script/above"
}
```

No interval. Sharing this for copy-pasting.

jeremejevs:
Sway 1.2 (released 5 hours ago) brings [this goodie](https://github.com/swaywm/sway/pull/4342) (thanks, @RedSoxFan!), allowing us to finally get rid of polling (props to @asahaf for spotting this [here](https://github.com/greshake/i3status-rust/issues/404#issuecomment-524374112)). It makes the script slightly more complicated:

```bash
#!/usr/bin/env bash

swaymsg \
  --type get_inputs | \
  jq \
    --raw-output \ '
      [
        .[] |
          select(.type == "keyboard") |
          .xkb_active_layout_name |
          select(contains("English (US)") | not)
      ] |
        first // "English"
    '

swaymsg \
  --type subscribe \
  --monitor \
  --raw \
  '["input"]' | \
  jq \
    --raw-output \
    --unbuffered \ '
      select(.change == "xkb_layout") |
        .input.xkb_active_layout_name |
        sub(" \\(US\\)"; "")
    '
```

... but is so worth it! No more electronic ink feel.

As before, adjust accordingly if `English (US)` is not your default.

eternal-sorrow:
here's my python variant (you'll need i3ipc-python from git)
```python
#!/usr/bin/env -S python3 -u
# vi:syntax=python

import i3ipc

sway = i3ipc.Connection()

# this script works with only one keyboard. we find it's ID.
# it should have more than one layout
for input_dev in sway.get_inputs():
	if input_dev.type == 'keyboard' and len(input_dev.xkb_layout_names) > 1:
		keyboard = input_dev.identifier
		print(input_dev.xkb_active_layout_name[:2].upper())
		break

def on_input(sway, event):
	if event.input.identifier == keyboard and event.change == "xkb_layout":
		print(event.input.xkb_active_layout_name[:2].upper())

sway.on(i3ipc.Event.INPUT, on_input)
sway.main()
```

m1s4k1:
Thanks @jeremejevs. Here is my variant:
- supports multiple keyboards (via identifier)
- independ of layout name
- output is 2 first letters of layout name (ex `EN`)

**Module** (Note: you can replace `ascii_upcase` with `ascii_downcase` to get output in LC):
```
#!/bin/bash

swaymsg -t get_inputs | jq -r \
    "first(.[]|select(.identifier == \"$1\" and .type == \"keyboard\")) \
    | .xkb_active_layout_name \
    | .[0:2] \
    | ascii_upcase"

swaymsg -mrt subscribe '["input"]' | jq -r --unbuffered \
    "select(.change == \"xkb_layout\")
    | .input
    | select(.identifier == \"$1\" and .type == \"keyboard\") \
    | .xkb_active_layout_name \
    | .[0:2] \
    | ascii_upcase"
```

**Usage** (pass identifier as argument):
```
    "custom/layout": {
        "format": " {}",
        "exec": "~/.config/waybar/modules/kblayout '1:1:AT_Translated_Set_2_keyboard'",
        "tooltip": false
    },
    "custom/layout_ext": {
        "format": " {}",
        "exec": "~/.config/waybar/modules/kblayout '3141:30219:SONiX_USB_Keyboard'",
        "exec-if": "swaymsg -t get_inputs | grep '3141:30219:SONiX_USB_Keyboard'",
        "tooltip": false
    }
```

vyachkonovalov:
Thanks @jeremejevs and @m1s4k1. Made a little change to show emoji flags
```fish
#!/usr/bin/env fish

set icons '["🇺🇸", "🇷🇺"]'

swaymsg -t get_inputs | jq -rcM --argjson icons $icons \
    "first(.[] | select(.identifier == \"$argv[1]\" and .type == \"keyboard\")) \
    | \$icons[.xkb_active_layout_index]"

swaymsg -mrt subscribe '["input"]' | jq -rcM --unbuffered --argjson icons $icons \
    "select(.change == \"xkb_layout\")
    | .input
    | select(.identifier == \"$argv[1]\" and .type == \"keyboard\") \
    | \$icons[.xkb_active_layout_index]"

```

I also tried to return '{"percentage": 0}' (and 1) from jq with config:
```config
"custom/kblayout": {
    "format": "{icon}",
    "exec": "kblayout.fish '1:1:AT_Translated_Set_2_keyboard'",
    "return-type": "json",
    "format-icons": ["🇺🇸", "🇷🇺"],
    "tooltip": false
}
```
but it doesn't get updates from `swaymsg -mrt subscribe ...`, maybe a 🐞 in waybar.

Alexays:
Superseded by #659

PR #86, opened 2018-11-05
Fix clicking and scrolling through workspaces

David96:
The way waybar used the workspace "num", clicking a workspace called "1:
something" resulted in going to a newly created workspace called "1",
because the workspace ipc command expects the workspace name, not its number.


Alexays:
Oh! my bad thx :)

PR #87, opened 2018-11-05
Fix workspaces not being removed from bar when moved to another output

David96:
And another one :D 
Now only the ordering of the workspaces has to be fixed if all-outputs == true.


Alexays:
Thanks!

PR #91, opened 2018-11-08
Use /proc/meminfo for Memory module

### LINKED ISSUE #6: RAM free/used
screenshot_2018-08-09-22 07 44

@BlueGone You agree you have same problem on your computer ?

Robinhuett:
This parses `/proc/meminfo` as suggested in #6 

Closes #6

### DIFF
--- a/src/modules/memory.cpp
+++ b/src/modules/memory.cpp
@@ -14,28 +14,56 @@ waybar::modules::Memory::Memory(const Json::Value& config)
 auto waybar::modules::Memory::update() -> void
 {
   parseMeminfo();
-  int used_ram_percentage = 100 * (memtotal_ - memfree_) / memtotal_;
-  label_.set_text(fmt::format(format_, used_ram_percentage));
-  auto used_ram_gigabytes = (memtotal_ - memfree_) / std::pow(1024, 2);
-  label_.set_tooltip_text(fmt::format("{:.{}f}Gb used", used_ram_gigabytes, 1));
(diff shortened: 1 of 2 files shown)


src/modules/memory.cpp
```diff
@@ -13,13 +13,30 @@ waybar::modules::Memory::Memory(const Json::Value& config)
...
+  label_.set_tooltip_text(fmt::format("{:.{}f}Gb used", used_ram_gigabytes, 1));
+}
+
+void waybar::modules::Memory::parseMeminfo()
+{
+  int memtotal, memfree, memavail, membuffer, memcache;
+  FILE* info = fopen("/proc/meminfo","r");
+  if(fscanf (info, "MemTotal: %d kB MemFree: %d kB Buffers: %d kB Cached: %d kB",&memtotal, &memfree, &membuffer, &memcache) < 4) { // Old meminfo format 
+    fclose(info);
+    info = fopen("/proc/meminfo","r");
```
Alexays: Why close and re-open?

Robinhuett: You're right, after looking at it again, this can be replaced by `rewind(info);`

Alexays:
There is no way to do that in modern cpp?
Or simply with ifstream :)

Robinhuett:
The most cpp way, would be using `sysinfo`, which unfortunatly lacks `MemAvailable`.
But thanks for pointing me to `ifstream`, this looks a bit cleaner. I'll changed it later :)
And sorry, cpp is not my "native language"

PR #93, opened 2018-11-09
Add config option for volume change scroll step size

### LINKED ISSUE #92: PulseAudio volume change via scoll is super slow
As mentioned [here](https://github.com/Alexays/Waybar/blob/2d2fb8804068770c604293e9128fc59b6b962e43/src/modules/pulseaudio.cpp#L77-L78), the change value doesn't seem to be the same for everyone.
And since PulseAudio doesn't seem to provide min & max values the easiest would be to let the user set a step size.

This could be done by replacing line 78 
https://github.com/Alexays/Waybar/blob/2d2fb8804068770c604293e9128fc59b6b962e43/src/modules/pulseaudio.cpp#L77-L78
with something like this: 
```cpp
uint16_t change = config_["scroll-resolution"].isUInt() ? config_["scroll-resolution"].asUInt() : 100;
```

I can open a PR for this, but wanated to open an issue fist, in case there are better ideas how to haandle this.

Robinhuett:
Closes #92

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -63,6 +63,7 @@
         "format-disconnected": "Disconnected ⚠"
     },
     "pulseaudio": {
+        //"scroll-step": 1,
         "format": "{volume}% {icon}",
         "format-bluetooth": "{volume}% {icon}",
         "format-muted": "",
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks!

PR #94, opened 2018-11-11
Use /proc/stat for cpu load

Robinhuett:
This PR fixes the cpu load readings, which are way off in `sysinfo`
Also the individual cores get displayed in the tooltip

### DIFF
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -1,7 +1,9 @@
 #pragma once
 
 #include <fmt/format.h>
-#include <sys/sysinfo.h>
+#include <fstream>
+#include <vector>
+#include <numeric>
 #include "util/chrono.hpp"
 #include "ALabel.hpp"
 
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks!

jeremejevs:
The readings were correct, [load != utilization](https://serverfault.com/a/667250/101998). This should've been a new module.

Robinhuett:
@jeremejevs Thanks for the heads up. The information on the Wiki also already states that this module displays the utilization. So it's just the naming in the code that's off.
But the pervious reading were off none the less, I had somtimes readings of 130% an the like o.O

@Alexays  If you like, I can make in PR to rename the variables

jeremejevs:
That wiki page is 1 day younger than this PR :stuck_out_tongue: But not a big deal, as long as it's correctly documented.

About 130% - the load can go above 100%, that means there's more stuff queued up than what the system is capable of processing. At least that's my understanding.

Alexays:
@Robinhuett @jeremejevs I added both load and usage on latest commit :)

jeremejevs:
Great, thanks! :+1:

PR #96, opened 2018-11-13
Module network ipaddr

Robinhuett:
This PR adds the ability to display the IP address, as well as the subnet mask (also in CIDR notation).

The following format replacements will get added:

| string      | replacement |
| --- | --- |
| `{ipaddr}`  | The first IP of the interface. |
| `{netmask}` | The subnetmask corresponding to the IP. |
| `{cidr}`    | The subnetmask corresponding to the IP in CIDR notation. |

And while we are at it, the following format replacement is missing from network:

| string      | replacement |
| --- | --- |
| `{signaldBm}`  | Signal strenth of the wireless network in dBm. |

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -269,11 +269,11 @@ void waybar::modules::Network::getInterfaceAddress() {
   int success = getifaddrs(&ifaddr);
   if (success == 0) {
     ifa = ifaddr;
-    while(ifa != NULL && ipaddr_.empty() && netmask_.empty()) {
-      if(ifa->ifa_addr->sa_family == family_) {
-        if(strcmp(ifa->ifa_name, ifname_.c_str()) == 0){
-          ipaddr_=inet_ntoa(((struct sockaddr_in*)ifa->ifa_addr)->sin_addr);
-          netmask_=inet_ntoa(((struct sockaddr_in*)ifa->ifa_netmask)->sin_addr);
+    while (ifa != NULL && ipaddr_.empty() && netmask_.empty()) {
+      if (ifa->ifa_addr->sa_family == family_) {
+        if (strcmp(ifa->ifa_name, ifname_.c_str()) == 0) {
+          ipaddr_ = inet_ntoa(((struct sockaddr_in*)ifa->ifa_addr)->sin_addr);
+          netmask_ = inet_ntoa(((struct sockaddr_in*)ifa->ifa_netmask)->sin_addr);
           cidrRaw = ((struct sockaddr_in *)(ifa->ifa_netmask))->sin_addr.s_addr;
           unsigned int cidr = 0;
           while (cidrRaw) {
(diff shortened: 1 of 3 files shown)


src/modules/network.cpp
```diff
@@ -255,6 +263,32 @@ int waybar::modules::Network::getExternalInterface()
...
+              cidr += cidrRaw & 1;
+              cidrRaw >>= 1;
+          }
+          cidr_ = cidr;
+        }
+      }
+      ifa = ifa->ifa_next;
+    }
+  }
+  freeifaddrs(ifaddr);
```
Alexays: Include this in the success scope, otherwise it will fail on unsuccess :/

src/modules/network.cpp
```diff
@@ -255,6 +263,32 @@ int waybar::modules::Network::getExternalInterface()
   return ifidx;
 }
 
+void waybar::modules::Network::getInterfaceAddress() {
+  unsigned int cidrRaw;
+  struct ifaddrs *ifaddr, *ifa;
+  int success = getifaddrs(&ifaddr);
+  if (success == 0) {
+    ifa = ifaddr;
+    while(ifa != NULL && ipaddr_.empty() && netmask_.empty()) {
```
Alexays: Space before (

src/modules/network.cpp
```diff
@@ -255,6 +263,32 @@ int waybar::modules::Network::getExternalInterface()
...
+      if(ifa->ifa_addr->sa_family == family_) {
```
Alexays: same

src/modules/network.cpp
```diff
@@ -255,6 +263,32 @@ int waybar::modules::Network::getExternalInterface()
...
+        if(strcmp(ifa->ifa_name, ifname_.c_str()) == 0){
```
Alexays: same and space before {

src/modules/network.cpp
```diff
@@ -255,6 +263,32 @@ int waybar::modules::Network::getExternalInterface()
...
+          ipaddr_=inet_ntoa(((struct sockaddr_in*)ifa->ifa_addr)->sin_addr);
```
Alexays: space around =

src/modules/network.cpp
```diff
@@ -255,6 +263,32 @@ int waybar::modules::Network::getExternalInterface()
...
+          netmask_=inet_ntoa(((struct sockaddr_in*)ifa->ifa_netmask)->sin_addr);
```
Alexays: space around =

Alexays:
Nice addition :)

Robinhuett:
Sorry, for missing the code styles. I'll be watching out more carefully for those in the future :)
I've also added that the vars get cleared if `getifaddrs(&ifaddr)` fails.

Alexays:
Thanks!

PR #103, opened 2018-11-22
Event based mediaplayer script

SibrenVasse:
For personal use I created an event based python script based on playerctl to use with the custom module. Currently it's specifically for Spotify.
I thought this might be useful for other users, and an improvement over the current delayWorker script.

Is this something we should work towards merging?

Current requirements:
playerctl > 2.0
python-gobject

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -78,8 +78,6 @@
     "custom/spotify": {
         "format": " {}",
         "max-length": 40,
-        "interval": 30, // Remove this if your script is endless and write in loop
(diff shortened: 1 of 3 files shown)


Alexays:
Nice improvement!
But why only spotify, playerctl manages other in the same way no?

Alexays:
If you can, it would help more people ;)

SibrenVasse:
Alright, this should work for all media players supported by playerctl.

Alexays:
Thanks!

PR #105, opened 2018-11-24
Allow custom plugins that don't execute anything

David96:
This can be quite useful to simply add a button to the bar that executes some script on click, for example to control a mediaplayer.


Alexays:
Thanks! yup nice idea!

PR #110, opened 2018-11-28
Update to new Sway IPC protocol

### LINKED ISSUE #109: Sway IPC protocol has changed
As seen [here](https://github.com/swaywm/sway/pull/3206/files#diff-bd22061da2e0c700a03433ee2e560a79R708), instead of `{\"success\": true}` it now responds with `[{\"success\": true}]` :smile: So this check doesn't work anymore:

https://github.com/Alexays/Waybar/blob/d0933ab50fd9a5638170ce0cd67e7e33c860831e/src/modules/sway/ipc/client.cpp#L125

I think the best fix is to compare `res.payload` against both variants for now, since some people may remain on an older Sway for a little while.

tokyovigilante:
Fixes #109 by checking for old and new IPC response formats. Works with swaywm/sway@9d18d81

### DIFF
--- a/src/modules/sway/ipc/client.cpp
+++ b/src/modules/sway/ipc/client.cpp
@@ -122,7 +122,7 @@ struct waybar::modules::sway::Ipc::ipc_response
 void waybar::modules::sway::Ipc::subscribe(const std::string& payload) const
 {
   auto res = send(fd_event_, IPC_SUBSCRIBE, payload);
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

jeremejevs:
Aaand they have reverted the change :smile: https://github.com/swaywm/sway/pull/3218

Alexays:
@jeremejevs Nice! x) i'll revert

PR #124, opened 2018-12-19
feat(pulseaudio): do not restrict format-icon list

chep:
Some ports are not in the list, I wanted to add an icon for
"analog-output-lineout" and I had to modify
waybar::modules::Pulseaudio::getPortIcon().

I tried to remove the function and use port_name_ variable directly and it seems
to work. If the string is not in the config file, "default" icon is used. If
there is no "default" entry, no icon is shown.

### DIFF
--- a/include/modules/pulseaudio.hpp
+++ b/include/modules/pulseaudio.hpp
@@ -22,8 +22,6 @@ class Pulseaudio : public ALabel {
     static void volumeModifyCb(pa_context*, int, void*);
     bool handleScroll(GdkEventScroll* e);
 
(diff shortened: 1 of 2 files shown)


Alexays:
I did this because often the names were headphone followed by numbers per example, but what we can do is if it does not find the port instead of returning an empty string we return `port_name_` what do you think?

chep:
I think it is a good idea.

Alexays:
@chep Do you make the changes or do I?

chep:
I can do, I'll make another pull request

PR #128, opened 2018-12-25
Use SignalStrength for format icons

Robinhuett:
This adds the ability to set format icons similar to the pulseaudio module.

A config for this could look like this (Needs FontAwesome Pro):
```
"format": "{icon}",
"format-icons": {
  "wifi": ["", "" ,""],
  "ethernet": [""],
  "disconnected": [""] 
}
```

Also if fixes getIcon() in ALabel.

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -90,7 +90,7 @@ std::string waybar::ALabel::getIcon(uint16_t percentage,
                                     const std::string& alt) {
   auto format_icons = config_["format-icons"];
   if (format_icons.isObject()) {
-    if (!alt.empty() && format_icons[alt].isArray()) {
+    if (!alt.empty() && (format_icons[alt].isString() || format_icons[alt].isArray())) {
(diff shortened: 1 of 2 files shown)


src/ALabel.cpp
```diff
@@ -90,7 +90,7 @@ std::string waybar::ALabel::getIcon(uint16_t percentage,
                                     const std::string& alt) {
   auto format_icons = config_["format-icons"];
   if (format_icons.isObject()) {
-    if (!alt.empty() && format_icons[alt].isString()) {
+    if (!alt.empty() && format_icons[alt].isArray()) {
```
Alexays: it's better to do `if (!alt.empty() && (format_icons[alt].isString() || format_icons[alt].isArray())) {`

Robinhuett: Fixed in 34df2b0

Alexays:
Thanks!

PR #129, opened 2018-12-26
Add format-icons to custom module

Robinhuett:
This commit allows custom modules (json only) to set a percentage. This can be displayed either by using `{percentage}` or by using `{icon}` with `format-icons` set.

## Doku:
```
#### Format replacements:

| option           | typeof  | default | description |
| `format`         | string  | `{}`    | The format, how information should be displayed. On `{}` data gets inserted. |
| `format-icons`   | array   |         | Based on the set percentage, the corresponding icon gets selected.<br>The order is *low* to *high*. |

| string | replacement |
| --- | --- |
| `{}` | Output of the script. |
| `{percentage}` | Percentage wich can be set via a json return-type. |
| `{icon}` | An icon from 'format-icons' according to percentage. |
```

### DIFF
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -132,19 +132,6 @@ void waybar::modules::Custom::parseOutputRaw()
   }
 }
 
-bool waybar::modules::Custom::isInteger(const std::string& n)
-{
-  if (std::isdigit(n[0]) || (n.size() > 1 && (n[0] == '-' || n[0] == '+'))) {
-    for (std::string::size_type i{ 1 }; i < n.size(); ++i) {
-      if (!std::isdigit(n[i])) {
-        return false;
-      }
-    }
-    return true;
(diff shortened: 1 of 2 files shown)


src/modules/custom.cpp
```diff
@@ -130,6 +132,19 @@ void waybar::modules::Custom::parseOutputRaw()
   }
 }
 
+bool waybar::modules::Custom::isInteger(const std::string& n)
```
Alexays: It is not necessary the JSON format supports numbers so just do isUInt and AsUInt :)

Robinhuett: I have tried this but it nerver triggered, but after having a second look at it, my json was not vorrectly formated. I'll push a commit in a sec.

Alexays:
Thank you! I'll update the doc

PR #137, opened 2018-12-28
Add custom text module and update makefile

cjbassi:


Alexays:
I don't really want to add a module for that, since the "custom" module does it perfectly if you don't put "exec" in it's config, but using a format it will display this format, maybe we need to document it better :?

PR #139, opened 2018-12-28
Remove ws index from sway ws names

cjbassi:


Alexays:
Maybe the user doesn't want that, can you make it optional with an option?

cjbassi:
You can already configure that by adding `{index}` to the sway/workspace format field I'm pretty sure.

Alexays:
Yep, you're right, I'm fine with that.
Thanks!

PR #140, opened 2019-01-08
refactor: replace all gtkmm.h includes

Robinhuett:
Replaces all occurences of `#include <gtkmm.h>` with only ne neccesary headers


Alexays:
Thanks a lot!

PR #152, opened 2019-01-25
Add fmt dependency to README.md

jubalh:


Alexays:
Even if fmt is compiled if it is not found, it's fine to specify it.
Thanks!

Issue #148, opened 2019-01-17
Feature request: hide tooltip


trobjo:
I'd would like to be able to hide the tooltip as only find it annoying. 

E.g. if I accidentally move my cursor to the top center where the title of the current active window is, it will change to 'waybar'. This new title is too short to be under the cursor, thus it becomes inactive and focus switches to my previous window. Now the focus of my cursor is the title of waybar again, and it will switch back to 'waybar'. This creates a rapid back-and-forth in active window state, which manifests itself as a flickering screen.

Arnaudv6:
#172 duplicate?

Alexays:
Done in 83a6475510df2c3db325c6ba24b377858b487ca5
Just set in each module `"tooltip": false`

PR #154, opened 2019-01-26
fix(window): Ignore waybar

Robinhuett:
With this, the window title `waybar` will be ignored so it won't flicker as decribed in #148.

Regarding the hiding of tooltips, i think this belongs in another pull (together with formating tooltips like module text).

### DIFF
--- a/src/modules/sway/window.cpp
+++ b/src/modules/sway/window.cpp
@@ -23,9 +23,11 @@ void waybar::modules::sway::Window::worker()
     try {
       auto res = ipc_.handleEvent();
       auto parsed = parser_.parse(res.payload);
+      // Check for waybar prevents flicker when hovering window module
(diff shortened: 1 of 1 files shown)


Alexays:
Could you add a todo next to the line to keep track of this workaround?

Robinhuett:
Actually I haven't  thought of this as a workaround since I don't think waybar should be displayed in the window module.
The things I mentioned for the other PR was more a notice, that this wont close the Issue.

But I can certainly add the comment if you like :)

Alexays:
Thanks!

Alexays:
@JohnathanFL Oh yeah before the force push, it was in the if condition
@Robinhuett

Robinhuett:
Ohh yeah, sorry. I messed up there. :/ I'm fixing it in a second.

PR #178, opened 2019-02-17
Strip workspace number from name

steevel:
We already have access to the workspace number as {index} in the format string.

I refactored the duplicated code for creating the formatted string into a new method called `getWorkspaceName`. The stripping of workspace number is made in the same was as `strip_workspace_numbers` in swaybar.

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -25,7 +25,8 @@ class Workspaces : public IModule {
     std::string getPrevWorkspace();
     std::string getNextWorkspace();
     uint16_t getWorkspaceIndex(const std::string &name);
-    std::string trimWorkspaceName(std::string);
+    std::string getWorkspaceName(const Json::Value&);
+    std::string trimWorkspaceName(const Json::Value&);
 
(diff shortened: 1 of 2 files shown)


coredotbin:
Sway let's you define and identify workspaces with arbitrary names if that is what you are looking for. You can take a look at [my dotfiles](https://github.com/coredotbin/dotfiles/blob/master/.config/sway/workspaces) for an example

Alexays:
I don't really understand what this pr solves, it's just an code refactoring?

steevel:
It removes the workspace number from the name. But yeah I give up. I will just keep changes to myself.

Have a nice day!

ndrewtl:
This might be important! It seems to be the only solution to [swaywm#2837](https://github.com/swaywm/sway/issues/2837), on the display order of workspaces. If you have two workspaces called `one` and `two` and you always want them to appear in that order, then you need this stripping feature for alphabetization.

Alexays:
@ndrewtl Oh i got it, so the best is to also strip number when `strip_workspace_numbers` is defined

PR #181, opened 2019-02-17
Add backlight module

ianhattendorf:
Monitor the backlight level via udev. Poll every `interval` as well,
in case backlight udev events aren't supported.

- Should this be enabled in the default config?
- Should this be `#ifdef`'d depending on `libudev` availability?
- My laptop doesn't send udev events for backlight change, so it would be nice if someone with one that does could test that.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -16,10 +16,12 @@
 #ifdef HAVE_LIBNL
 #include "modules/network.hpp"
 #endif
+#ifdef HAVE_LIBUDEV
+#include "modules/backlight.hpp"
+#endif
 #ifdef HAVE_LIBPULSE
 #include "modules/pulseaudio.hpp"
 #endif
-#include "modules/backlight.hpp"
 #include "modules/custom.hpp"
 
 namespace waybar {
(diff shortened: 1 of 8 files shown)


Alexays:
Nice work!
Yeah could you make udev a optional dependencies like some other modules?

ianhattendorf:
@AlessioDP  So, I completely forgot that writing to `/sys/class/backlight/*/brightness` requires root permissions. `light` gets around this by being `setuid root`, which we can't (and shouldn't anyway) do due to gtk. It looks like our options are:

- Not support changing brightness via scrolling
- Shell out to `light` to set the brightness if available
- Create a helper binary of our own to change the brightness to avoid a dependency on `light`

Personally, I'm leaning towards either not implementing setting the brightness or optionally enabling setting brightness if `light` is available. @alexays, what do you think?

Alexays:
I think we can let the user define `on-scroll-up` with `light` cmd no?

Alexays:
Thanks!

Alexays:
You'll have time to do the wiki?

Issue #177, opened 2019-02-16
New module for brightness (light based)


AlessioDP:
I would like to see a module for changing brightness thanks to `light`.

Commands are quite simple and we just need to run:
`light -G` to see current brightness
`light -A 10` to increase by 10 the brightness level
`light -U 10` to decrease by 10 the brightness level

Further ideas:
* Scroll support
* `scroll-change` to choose how much increase/decrease brightness
* Update every time you change brightness with click/scroll events

I would develop it by myself but I don't dev in C++ :S

PR #183, opened 2019-02-18
Added light module (brightness)

AlessioDP:
Reference to #177 

I have added a new module called 'light', this supports [light](https://github.com/haikarainen/light) brightness tool.

Features of light module:
- [x] Build option to enable light module
- [x] Show/increase/decrease brightness
- [x] Scroll support
- [x] `scroll-step` to choose brightness change
- [x] Update brightness value on scroll
- [x] Styling
- [ ] Upper and lower brightness limits
- [ ] Format icons
- [ ] On click events
- [ ] Wiki page

<br>

Further improvements:
- Now modules inside `config` are in alphabetic order (except for sway and customs), like `style.css`
- Changed cpu style color to `black` instead of `#000000`

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -19,6 +19,9 @@
 #ifdef HAVE_LIBPULSE
 #include "modules/pulseaudio.hpp"
 #endif
+#ifdef HAVE_LIGHT
+#include "modules/light.hpp"
+#endif
 #include "modules/custom.hpp"
 
 namespace waybar {
--- a/include/modules/light.hpp
+++ b/include/modules/light.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+#include <fmt/format.h>
+#include <iostream>
+#include "util/sleeper_thread.hpp"
+#include "util/command.hpp"
+#include "ALabel.hpp"
+
+namespace waybar::modules {
+
+class Light : public ALabel {
+  public:
+    Light(const std::string&, const Json::Value&);
+    ~Light() = default;
+    auto update() -> void;
+  private:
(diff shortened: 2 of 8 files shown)


SethBarberee:
Would it be beneficial to allow the user to redefine the backlight commands? Mainly wondering if it would be possible to use other tools besides `light`.

AlessioDP:
Lets see what thinks @Alexays about that :)

It will be quite similar to a custom module but this will allow you to set a limit for brightness and is strictly related on how to increase/decrease brightness.

Alexays:
This module is pretty similar to a custom module, otherwise the #181 implementation is more lower level compared to this one, so lower memory footprint and lower cpu impact.

AlessioDP:
That implementations works on my system so I am okay with that, but missing some features like scroll to change brightness and the code is quite different from yours.

ianhattendorf:
I like the scrolling! I've been thinking about a way to update brightness as a 2nd pass for my module, it's easy enough to do through `udev` via `udev_device_set_sysattr_value` but I wasn't sure about the UX. I went down this path to try and keep the dependencies and resource usage to a minimum and avoid shelling out to another process.

@AlessioDP would you be interested in merging your scroll functionality into mine? I can also give it a shot as I was already planning on looking into it later tonight or tomorrow when I get the time.

AlessioDP:
Sure, I will.
I'm working on merge it into custom module too ;)

Just I'm not familiar with C++ and I am strictly following the structure that @Alexays used, yours is a little bit different so I need some time to do it :P

Alexays:
Thanks anyway @AlessioDP :smiley:

PR #187, opened 2019-02-22
Add module Idle inhibitor

jonfin:
This patch adds a module to simulate user action, so the lockscreen won't be started.

I tested it only with swayidle, but it uses the idle inhibit protocol and should work with others programs.


Alexays:
Nice work! :heart: 
Do you have time to do the wiki?

PR #189, opened 2019-02-24
Added clock tooltip with optional alternate format

DutchMofo:
I wanted the ability to see the current date when I hover my mouse over the clock module.

Adds the following settings:
- clock/tooltip
- clock/tooltip-format (default is same as normal format)

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -38,7 +38,7 @@
         "spacing": 10
     },
     "clock": {
-        "tooltip-format": "{:%Y-%m-%d | %H:%S}"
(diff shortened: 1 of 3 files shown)


resources/config
```diff
@@ -38,6 +38,8 @@
         "spacing": 10
     },
     "clock": {
+        "tooltip": true,
```
Alexays: This is not needed since the default value is `true` :)

Alexays:
Thanks!

PR #191, opened 2019-02-25
fix: Escape user controlled input

### LINKED ISSUE #162: external widget does not update if & in text
I'm using `playerctl metadata -f '{{emoji(status)}} {{xesam:artist}} - {{xesam:title}}' -F` for a spotify widget. However, when there is an & in the song or artist name, it fails to set this. The tooltip is still correct.

```
Failed to set text ' ▶️ The Whistles & The Bells - Canary Cage' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

Robinhuett:
Fixes #162 

This should escape all needed user controlled inputs.

### DIFF
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -123,7 +123,7 @@ void waybar::modules::Custom::parseOutputRaw()
   int i = 0;
   while (getline(output, line)) {
     if (i == 0) {
-      text_ = line;
+      text_ = Glib::Markup::escape_text(line);
       tooltip_ = line;
       class_ = "";
     } else if (i == 1) {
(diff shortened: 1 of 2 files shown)


Alexays:
Nice job, I hadn't thought about it while i had already used this function :smile:

mnussbaum:
Would you be amenable to an option to turn off the escaping of custom module output? I've been relying on a custom module that returns markup with changing background and foreground colors based on machine state.

If you'd be amenable to this option I'm happy to work on a PR for it.

Robinhuett:
@mnussbaum This change wouldn't be hard to implement. 
On the otherhand it may be better to apply an css class if you don't have to many states. For an example you can look in my dotfiles ([This module](https://github.com/Robinhuett/dotfiles/blob/eca9add5852166622402e51863a7b95a04bd68ac/.config/waybar/modules/storage#L13-L20) and [This Style](https://github.com/Robinhuett/dotfiles/blob/eca9add5852166622402e51863a7b95a04bd68ac/.config/waybar/style.css#L57-L63)).
If it's not suitable I happily make the change :)

PR #192, opened 2019-02-25
Fix tooltip format

DutchMofo:
Messed up the example format... I'm an idiot.


Alexays:
Thanks! I hadn't seen it too :smile:

Issue #100, opened 2018-11-15
Be able to create dynamic fmt args array


Alexays:
In order to avoid any unnecessary logic.
e.g. the CPU module without label {usage}, still calculates the usage when the argument is not provided

Alexays:
Done for the CPU module in 2d9bcb1a2df916e4a6846659ce91c5cd8038097e

Organic-Code:
I'll try to work on that

Organic-Code:
Alright, so my idea was to implement some kind of lazy argument support, so that you could be able to do something like
```cpp
fmt::format(str, fmt::arg("data1", lazy(computeData1)), …)
```
where computeData1 would be called only when `"data1"` is present is the `str` format specifier
That part is very simple to do, but I find it quite cumbersome to use, especially when parameters are to be taken into account, leading to code looking like this:
```cpp
fmt::format(str, fmt::arg("data1", lazy([&] () mutable { return this->computeData1(param1, param2); }), …)
```
which I personally find pretty awful, especially where there would be several named arguments.


So I figured out something like this would be much nicer to use (taking the same function as above):
```cpp
fmt::format(str, lazy::arg("data1", this, computeData1, param1, param2), …)
```
This is however mostly not feasible with fmt as it is due to life time management issues.
If you agree with it, I'd like to modify fmt so as to be able to do this.
If however you disagree with that, then I'll try to go and improve the first method usage

Alexays:
Yeah thats why 2d9bcb1a2df916e4a6846659ce91c5cd8038097e sigsev cause of the lifetime 
 of arguments.
I don't really want to have to modify fmt to implement this knowing that they will probably one day propose a way to make dynamic arguments.
Maybe we can override fmt::arg to do this in a cleaner way?

Organic-Code:
I've tried various things, but unfortunately that's not possible AFAIK
Fmt::arg relies on the type `named_arg` to detect the use of named args, and  `named_arg` is the type storing a reference, which in turn causes the lifetime issues

Organic-Code:
I just realized we could achieve a similar syntax without changing Fmt:
```cpp
fmt::format(str, fmt::arg("data1", lazy(this, computeData1, param1, param2)) , …)
```
Which is slightly worse, but is still OK IMHO

Well, using macros would actually allow the syntax I proposed with  `lazy::arg` by expanding to this code, but I don't think that would be a really good idea

Alexays:
[...]

PR #193, opened 2019-02-25
feat(custom): Add field for additional data to json

Robinhuett:
Just a small change that adds an additional string field to the custom module, so `format-alt` can be utilized better.

Maybe this should be touched again when tackeling #100


Alexays:
Thanks!

PR #199, opened 2019-03-01
Disable backlight module if no backlights found

### LINKED ISSUE #196: Backlight module hangs instead of disabeling
I share my config between my notebook and my workstation.
On my notebook everything works fine but on my workstation the backlight modules hangs because of the missing backlight.
Since it's outputting a empty string there is a gap in my bar.
I think the best way to handle this, would be to disable the module (like the battery module).

ianhattendorf:
Also ignore other potential build directories

Fixes #196. @Robinhuett, could you confirm?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ vgcore.*
 *.swp
 packagecache
 /subprojects/**/
-/build
+/build*
 /dist
 /meson.egg-info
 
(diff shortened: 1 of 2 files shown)


Robinhuett:
@ianhattendorf Works beautiful :+1:

Alexays:
Thansk!

PR #204, opened 2019-03-02
Improving mouse buttons support

Organic-Code:
Adds support for on-click-[backward/forward/middle]

Also fixing format-alt, which was making modules to behave a bit erratically (was not sure if I was to make a separated pr for this issue, but I decided against it in the end, sry if that's a bother)

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -40,11 +40,19 @@ auto waybar::ALabel::update() -> void {
 bool waybar::ALabel::handleToggle(GdkEventButton* const& e) {
   if (config_["on-click"].isString() && e->button == 1) {
     waybar::util::command::forkExec(config_["on-click"].asString());
+  } else if (config_["on-click-middle"].isString() && e->button == 2) {
+    waybar::util::command::forkExec(config_["on-click-middle"].asString());
   } else if (config_["on-click-right"].isString() && e->button == 3) {
     waybar::util::command::forkExec(config_["on-click-right"].asString());
-  } else {
+  } else if (config_["on-click-forward"].isString() && e->button == 8) {
+    waybar::util::command::forkExec(config_["on-click-backward"].asString());
+  } else if (config_["on-click-backward"].isString() && e->button == 9) {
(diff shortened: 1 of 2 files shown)


src/ALabel.cpp
```diff
@@ -40,11 +40,19 @@ auto waybar::ALabel::update() -> void {
...
   } else if (config_["on-click-right"].isString() && e->button == 3) {
     waybar::util::command::forkExec(config_["on-click-right"].asString());
-  } else {
+  } else if (config_["on-click-forward"].isString() && e->button == 8) {
+    waybar::util::command::forkExec(config_["on-click-backward"].asString());
+  } else if (config_["on-click-backward"].isString() && e->button == 9) {
+    waybar::util::command::forkExec(config_["on-click-forward"].asString());
+
+
+  } else if (config_["format-alt-click"].isUInt() && e->button == config_["format-alt-click"].asUInt()) {
```
Alexays: This is a kind of breaking change, no?

Organic-Code: It can be considered as one, but I did not think it was the intended behavior.
What happened before was that after registering a left/right click action on a module, clicking on it with any other mouse button would "fold" it ([turning the former into the latter](https://i.imgur.com/o3TSYn2.png)).
With this pr, this behavior can be partially achieved by adding '"format-alt": " "' in the config for the concerned module, the difference being that only one type of click will trigger the change, as opposed to any type of click that was not registered

jonfin: At least it would help the module which are overriding the function, as they could call it with out checking if the event was handled and relaying on the default behavior. (See #208)

Second point: I would not do it as an exclusive or, because we could have now a mouse button event which can trigger two actions. We can specify it in the config and some one could wonder why it won't work.

```suggestion
  }
  if (config_["format-alt-click"].isUInt() && e->button == config_["format-alt-click"].asUInt()) {
```

Organic-Code: That's a good point, it's true it will probably the more expected behavior

Alexays:
Nice! thanks <3

PR #206, opened 2019-03-03
resolves #205

### LINKED ISSUE #205: Add option to show only current workspace in sway/workspaces module
Add option `current-only` so when it is `true`, `sway/workspaces` module only shows current workspace, similar to the way `xworkspace` module from polybar behaves.

dikeert:
Hello!

This one resolves #205 that I've just created.

It simply hides the all the buttons but the focused one.


dikeert:
Do I need to do anything to pass CI?

Alexays:
Thanks!

PR #207, opened 2019-03-03
Add custom tooltip format for network module

jonfin:
Adds the possibility to set different formats in the tooltip. Mainly the same as the normal format just with the prefix `tooltip`. If non are defined, but the tooltip enabled, it will use the same output as the bar.
```
"network" {
    "interface": "wlp2s0",
    "format": "{ifname}",
    "format-wifi": "{essid} ({signalStrength}%) ",
    "format-ethernet": "{ifname} ",
    "format-disconnected": "", //An empty format will hide the module.
    "tooltip-format": "{ifname}",
    "tooltip-format-wifi": "{essid} ({signalStrength}%) ",
    "tooltip-format-ethernet": "{ifname} ",
    "tooltip-format-disconnected": "Disconnected",
    "max-length": 50
}
```

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -55,8 +55,8 @@ void waybar::modules::Network::createInfoSocket()
 {
   info_sock_ = nl_socket_alloc();
   if (nl_connect(info_sock_, NETLINK_ROUTE) != 0) {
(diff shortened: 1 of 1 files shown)


Alexays:
Nice! could you update the Wiki?

PR #208, opened 2019-03-03
[bugfix] idle_inhibitor: handle click events correctly

jonfin:
The idle inhibitor module won't work if any click event is specified in the config. Because the click event handler of the idle inhibitor module is not activated. 

- Declare event handler in ALabel virtual so the idle_inhibitor can overriding them
- Handle the left click event in idle_inhibitor and call ALabel handler if needed

It should not conflict with #204.


Alexays:
Thanks!

PR #221, opened 2019-03-15
 fix(sway): ipc client crash when compiled with -D_GLIBCXX_ASSERTIONS 

alebastr:
Waybar from my Fedora package has been crashing with following assertion:
`/usr/include/c++/8/bits/basic_string.h:1067: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator[](std::__cxx11:      :basic_string<_CharT, _Traits, _Alloc>::size_type) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _All      oc>::reference = char&; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type = long unsigned int]: Assertion '__pos <= size()' failed.`

The crash happens because std::string.reserve() does not change string size, leaving it as 0. And all further modifications of string.data() are actually happening out of range.
-D_GLIBCXX_ASSERTIONS enforces strict out-of-bounds access checks in stl containers and makes the issue visible.

### DIFF
--- a/src/modules/sway/ipc/client.cpp
+++ b/src/modules/sway/ipc/client.cpp
@@ -64,7 +64,7 @@ struct waybar::modules::sway::Ipc::ipc_response
   waybar::modules::sway::Ipc::recv(int fd) const
 {
   std::string header;
-  header.reserve(ipc_header_size_);
+  header.resize(ipc_header_size_);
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #222, opened 2019-03-15
fix(pulseaudio) use case-insensitive comparison for icon lookup

alebastr:
Pulseaudio ports names in my setup are starting with capital letter:
```
% pacmd list-sinks |grep '\[Out\]'
                [Out] Speaker: Speaker playback (priority 100, latency offset 0 usec, available: no)
                [Out] Headphones: Headphones playback (priority 100, latency offset 0 usec, available: yes)
        active port: <[Out] Headphones>
```
Because of that waybar::modules::Pulseaudio::getPortIcon() fails to select "headphones" icon for "Headphones" port and returns "phone" as the only matching icon.

The second commit is a small optimization: it saves a few CPU cycles per call on allocation of vector and all these strings.

### DIFF
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -187,8 +187,10 @@ const std::string waybar::modules::Pulseaudio::getPortIcon() const
     "hifi",
     "phone",
   };
+  std::string nameLC = port_name_;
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks for the improvement!

Issue #170, opened 2019-02-14
Send a pkill signal to a module


AlessioDP:
Is it possible to send a pkill real time signal to waybar and update a specific module?

I can do a thing like that thanks to i3blocks:
`pkill -RTMIN+10 i3blocks`
This will update a specific i3block module:
```
[example]
...
signal=10
```

If not, is there a way to update a specific module when I execute a custom command?

Alexays:
Done in #228

PR #228, opened 2019-03-18
Add pkill signals to custom modules

hoellen:
This implements pkill signals for custom modules (see #170). You can define it with `"signal"` in the config and test it with `pkill -RTMIN+8 waybar` (for example signal 8). 
In addition there is now also the setting `"interval": "once"`, to load the module only once at startup.

edit: I already updated the [wiki](https://github.com/Alexays/Waybar/wiki/Module:-Custom) for this PR. I didn't know they were directly live. If you need time for the review, I can reverse it. Just let me know.

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -22,6 +22,7 @@ class Bar {
     ~Bar() = default;
 
     auto toggle() -> void;
+    void handleSignal(int);
 
     const Client& client;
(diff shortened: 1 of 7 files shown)


Alexays:
I don't think it's the best way for the `once' interval but for now i'll leave it like that, Thanks!

PR #241, opened 2019-03-23
Fix #240: Escape ESSID text before interpreting as pango

### LINKED ISSUE #240: ESSID not escaped before interpretation as pango markup
ESSID is currently not escaped before being formatted/interpreted as Pango markup. This leads to the network widget breaking when network SSIDs contain characters like `&`:

```
Failed to set text '&' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

Related issues and PRs: #106, #162 and #191.

jlbribeiro:
As detailed in #240, ESSIDs with characters like `&` were causing the network widget to break when connected to those networks. This PR fixes it by using `Glib::Markup::escape_text` before interpreting as pango.


Alexays:
Thanks!

PR #245, opened 2019-03-25
[custom] Allow icon selection based on json alt attribute

SibrenVasse:
Because getIcon() has logic to test if format-icon is an array or a dictionary, it is possible to also pass alt_ to this function.


Alexays:
Thanks!

PR #251, opened 2019-03-27
Feature select player

DanielVoogsgerd:
And here is the feature itself. I verified that it work in my usecases. But yours might be different. If you don't mind testing it first that would be awesome.

This branch is based on my branch from my other PRs (#249, #250) as I needed command line arguments to implement this.

### DIFF
--- a/resources/custom_modules/mediaplayer.py
+++ b/resources/custom_modules/mediaplayer.py
@@ -38,6 +38,13 @@ def on_metadata(player, metadata, manager):
     sys.stdout.flush()
 
 
+def on_player_appeared(manager, player, selected_player=None):
+    if player is not None and player.name == selected_player:
+        init_player(manager, player)
+    else:
+        logger.debug('New player appeared, but it\'s not the selected player, skipping')
+
+
 def on_player_vanished(manager, player):
     logger.info("Player has vanished")
     sys.stdout.write("\n")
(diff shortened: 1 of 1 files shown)


Alexays:
Nice!
Do we also want to change the default config with `--player spotify`?

DanielVoogsgerd:
I'd prefer not too. Right now, if the --player is left empty it will update the widget depending on the last player that had an event. Considering nothing is provided I think that is ideal behavior. I would suggest maybe adding ```--player spotify``` to the documentation, to showcase the functionality.

Now I'm thinking of it, it might be a good idea to change it in the example config as well, but I'll let you decide on that matter.

Alexays:
For me, it'll better to specify `--player spotify` since it's a spotify module in the default conf

DanielVoogsgerd:
How do you suggest implementing dynamic player recognition in that case then? I'm fine with a default but do like the dynamic possible behaviour. On top of that changing the default to ```--player spotify``` is not backwards compatible, which current behaviour is.

May I suggest we add the ```--player spotify``` to the default config. To me, it makes more sense that way.

Alexays:
A custom module can return a json with an icon and/or a class specified, so we can specify an icon for each supported players and also a class with a background color, what do you think?

DanielVoogsgerd:
Yeah sure, I suspect that that is more in the backend compartment. 
Do you mind finishing off that part?

Alexays:
@DanielVoogsgerd Sure, i'll do it tomorrow ;)

PR #252, opened 2019-03-28
Process tray icon updates

alebastr:
A functional but not quite finished implementation of update signals subscription for SNI::Item. Also changes most of dbus code in the class to use giomm wrappers because raw gdbus API is a pain to use.

I'm still not happy with the way updateImage could be invoked multiple times for each NewIcon signal. Ideally there should be some kind of batch update, and I'm considering if it's better to call GetAll and emulate PropertiesChanged signal. Going to give it a try on the weekend.

### DIFF
--- a/include/modules/sni/item.hpp
+++ b/include/modules/sni/item.hpp
@@ -50,6 +50,10 @@ public:
 private:
   void proxyReady(Glib::RefPtr<Gio::AsyncResult>& result);
   void setProperty(const Glib::ustring& name, Glib::VariantBase& value);
+  void getUpdatedProperties();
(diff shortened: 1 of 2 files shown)


alebastr:
I looked at plasma-workspace SNI implementation and their approach seems reasonable to me. Therefore, the code was changed to debounce signals for 10ms, fetch all properties and check for updated ones.
Not checking signal type for now, but that could wait until tooltip implementation.

Also, no more lambdas. They don't work with sigc::trackable.

Alexays:
A huge thank you for taking the time to finish the implementation correctly <3

PR #256, opened 2019-03-31
Convert spotify module to generic json version

SibrenVasse:
- Class is set via json class attribute 
- Different background colors for classes (examples 'custom-spotify' and 'custom-vlc' in style.css)
- Icon is chosen based on json alt attribute (default is musical note icon)


Alexays:
Faster than me, thanks! 😍

PR #265, opened 2019-04-15
Custom: Ability to add multiple classes

### LINKED ISSUE #263: Multiple classes for custom modules?
A class can represent a state, though multiple states may be available at once. For example a music player custom module may say whether the video is playing or not and whether the song has been added to your library on Spotify. It would be neat if the json return-type also accepted a list of classes instead of a string so multiple classes can be applied. A workaround is possible by just appending them in the custom script, but if many states are available, it becomes cumbersome.

Alexays:
Resolve #263
@AndreasBackx Can you confirm that it doesn't break anything and solve your problem correctly :)

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -26,8 +26,7 @@ class Custom : public ALabel {
     std::string text_;
     std::string alt_;
     std::string tooltip_;
-    std::string class_;
-    std::string prevclass_;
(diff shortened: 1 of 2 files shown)


AndreasBackx:
Just tested it and it works beautifully. Thank you for getting a PR out so quickly. Both strings and lists seem to be supported.

Issue #59, opened 2018-10-26
Huge tray icons


xajler:
Thanks for this great bar for `sway`, I really enjoy using it, much more than `polybar`...
I have problem with non electron apps showing in tray they are huge:

![waybar-02](https://user-images.githubusercontent.com/40003/47566031-ada62200-d92a-11e8-9778-eb79b7af3c78.png)

 Combined electron and normal tray icon:

![waybar-04](https://user-images.githubusercontent.com/40003/47566038-b26ad600-d92a-11e8-832d-e72e4e7c1a62.png)

Only electron tray icon:

![waybar-03](https://user-images.githubusercontent.com/40003/47566046-b8f94d80-d92a-11e8-9a30-24b6eeed9dec.png)

And full photo, using your latest [commit: 07d8dfb](https://github.com/Alexays/Waybar/commit/07d8dfb3d68b95d46641c57559a3baeb433d6c2c) with added tray space...

![waybar-01](https://user-images.githubusercontent.com/40003/47566127-ffe74300-d92a-11e8-9419-831c57eca4e2.png)

If you can do something with it, it would be great. I've used `nitrokey-app` and `lxqt-powermanagement` for normal tray icons.

And if you plan a `mpd` integration, not sure how much people use it, `polybar` was having it. Currently I've hacked it up to behave as it looked in `polybar`...

xajler:
And both of them are QT apps, maybe there is something about QT and tray icons.

Alexays:
it's okay it's fixed and as a bonus we can choose a size in the tray configuration 🎉

xajler:
Great, thank you!

PR #268, opened 2019-04-16
Add MPD support

minijackson:
This adds a basic MPD module, somewhat based on Polybar's implementation.

Also adds a minimal [EditorConfig](https://editorconfig.org/) so editors picks up the codebase's style. ~~(Maybe a `.clang_format` would be nice too but I don't think I should be the one choosing the style. See [here](https://zed0.co.uk/clang-format-configurator/) if you're interested)~~

Clang format configuration added!

**Don't merge now** I'm currently trying to also add support for options (shuffle, repeat, etc.). I'm putting it here now so I pick up on recommendations/reviews as soon as possible ^^

Features:
- artist, album-artist, album, title, and date in the label
- format, format-stopped, and format-disconnected depending on this module's state
- Event driven updates (so fast!)
- Automatic reconnect

Implementation details:
- Uses the official libmpdclient
- Use the `Deleter` feature of `unique_ptr`s to avoid the [rule of 5](https://cpppatterns.com/patterns/rule-of-five.html)

Todo:
- [x] Add support for options (shuffle, repeat, etc.)
- [x] Allow specifying the reconnect interval
- [x] Add support for icons
- [x] Add support for pause/play when clicking
- [x] Add a better sample config with icons in `resources/config`
- [ ] Update the wiki?
- [x] Check error handling ~~(the current implementation can throw an exception in a thread which won't be handled, and can make Waybar crash)~~
- [x] Allow setting a label when disconnected
- [x] Periodically fetch elapsed time when a song is playing

Thanks for the great work on Waybar ^^

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,6 @@
 ---
 BasedOnStyle: Google
-ColumnLimit: '100'
-
+AlignConsecutiveDeclarations: true
+BinPackArguments: false
+ColumnLimit: 100
 ...
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# EditorConfig configuration for Waybar
+# http://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+[*.{build,css}]
+indent_style = space
+indent_size = 4
+
+[*.{hpp,cpp}]
+indent_style = space
+indent_size = 2
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 - Memory
 - Cpu load average
 - Temperature
+- MPD
 - Custom scripts
 - Multiple output configuration
 - And much more customizations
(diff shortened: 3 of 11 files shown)


minijackson:
This was very briefly mentioned on #59, so @xajler might be interested (sorry if you didn't want to be tagged)

Alexays:
Nice work!
I wanted to do the clang format based on google rules but I forgot about it, if you have time to do it before me do not hesitate :heart:

xajler:
> This was very briefly mentioned on #59, so @xajler might be interested (sorry if you didn't want to be tagged)

No problem for tagging, thank you very much @minijackson for mpd support, I'm very eager to try it, because I currently have some nasty shell script  displaying song and elapsed time of song in waybar.

minijackson:
> I currently have some nasty shell script displaying song

I'm exactly the same!

Unfortunately, I'm not really sure how to handle the elapsed time, since the current implementation is event based, and we have to have periodic updates to get the elapsed time. Maybe we should do this in a separated thread, and synchronize the elapsed time only when playing using a mutex? Seems a bit much for that :-/
It would be nice nonetheless, so I'm going to add it to the todolist.

minijackson:
Okay, I think it's pretty much good to merge, now! :tada: 

I just need to rebase it on master, which currently have a small conflict on the .clang-format file

So @Alexays should I keep yours or mine ? ^^

Alexays:
Nice work! 🎉
You can keep yours, but i just don't really like `AlignConsecutiveAssignments: true`
Also if you can update the wiki it'll be perfect ❤️

minijackson:
Rebase done! You can merge it, I'll update the wiki in the meantime

Issue #273, opened 2019-04-19
Certain songs crash waybar with the new mpd module


cole-h:
On waybar version `Waybar v0.5.1-33-ga14b933 (Apr 18 2019, branch 'master')`, sway version `sway version 1.0-rc1-243-g35ddd9aa (Apr 18 2019, branch 'master')`. You can find my config and stylesheet here: https://gist.github.com/cole-h/33d75033b027ae59f177c2a0dd419330

This only occurs on a handful of songs for me. I've only gone through a few looking for this behavior, but the songs so far are "Puzzle" by Beats Antique, "Everything In Its Right Place" by Radiohead, any song by Sabaton (so far). They all exhibited the exact same backtraces:

```
#0  0x00007ffff5f056e5 in __strlen_avx2 () at /usr/lib/libc.so.6
#1  0x00005555555b2754 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign(char const*) ()
#2  0x00005555555bfd25 in waybar::modules::MPD::setLabel() ()
#3  0x00005555555c0002 in waybar::modules::MPD::update() ()
#4  0x000055555559a169 in sigc::internal::slot_call0<waybar::Bar::getModules(waybar::Factory const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)::{lambda()#1}, void>::call_it(sigc::internal::slot_rep*) ()
#5  0x00007ffff751edea in Glib::DispatchNotifier::pipe_io_handler(Glib::IOCondition) () at /usr/lib/libglibmm-2.4.so.1
#6  0x00007ffff7520ecc in Glib::IOSource::dispatch(sigc::slot_base*) () at /usr/lib/libglibmm-2.4.so.1
#7  0x00007ffff75210cf in Glib::Source::dispatch_vfunc(_GSource*, int (*)(void*), void*) () at /usr/lib/libglibmm-2.4.so.1
#8  0x00007ffff69f8661 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#9  0x00007ffff69fa739 in  () at /usr/lib/libglib-2.0.so.0
#10 0x00007ffff69fa77e in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#11 0x00007ffff6bb2f5e in g_application_run () at /usr/lib/libgio-2.0.so.0
#12 0x000055555557acac in waybar::Client::main(int, char**) ()
[...]

PR #274, opened 2019-04-19
fix(mpd): Escape MPD values in the label

minijackson:
The idea comes from #273 but doesn't fix it (I haven't managed to crash the bar yet)

This still doesn't support the `escape` configuration, the rationale is that:
- Users probably won't modify the song metadata to change the style of Waybar's MPD module
- The rest of the format replacements comes from the configuration and can be escaped there directly


Alexays:
Thanks!

PR #278, opened 2019-04-21
fix(mpd): regularly timeout the event listener to prevent timeout

### LINKED ISSUE #277: `mpd_status_get_state` segfaults on null `state_.get()`
From my last comment at #273:

---

Update on the `mpd_status_get_state` segfault: it occurs due to `state_.get()` occasionally being `nullptr`. I don't know of a nice way to handle this because my rudimentary modifications (just checking `if (state_.get() != nullptr)` before get state) would update the module to say "Disconnected", but wouldn't ever refresh state. Just `mpd: Connection to MPD closed` in the logs.

First thing we probably want to look at is why `tryConnect()` doesn't connect after this occurs, and the `connection_.reset()` and `alternate_connection_.reset()` lines don't help in `checkErrors()`. I have 0 experience with threaded programs in this regard.

## STR (on my system, at least)

1. Start up the bar
2. Play music for a second or two
3. Pause it
4. Do something for a minute ~~or so~~ (for me, always happens after 1 minute, exactly: `mpc pause;sleep 60;mpc play`)
5. Play music again
6. Observe segfault

N.B. Everywhere else that uses `state_.get()` has a chance to segfault -- I saw a few in the `elapsedTime` and `totalTime` variables, as well as the icons just below that while messing around.

cc: @minijackson @Alexays

minijackson:
Finally got to the bottom of this!

Didn't manage to find a way to "unblock" the event listener, so here's what I propose (from the commit message):

The MPD server has a `connection_timeout` that defaults to 60. If no data
is transferred in this time span, the connection is closed. If the
connection is closed while the event listener thread is listening for events,
it will block forever. By timing out the event listening and
re-connecting regularly, we prevent this issue. An option "timeout" has
been added for users that have a lower server `connection_timeout` than
default. Fixes #277

### DIFF
--- a/include/modules/mpd.hpp
+++ b/include/modules/mpd.hpp
@@ -48,6 +48,8 @@ class MPD : public ALabel {
   const char*    server_;
   const unsigned port_;
 
+  unsigned timeout_;
+
   // We need a mutex here because we can trigger updates from multiple thread:
   // the event based updates, the periodic updates needed for the elapsed time,
   // and the click play/pause feature
--- a/src/modules/mpd.cpp
+++ b/src/modules/mpd.cpp
@@ -7,17 +7,29 @@ waybar::modules::MPD::MPD(const std::string& id, const Json::Value& config)
     : ALabel(config, "{album} - {artist} - {title}", 5),
       module_name_(id.empty() ? "mpd" : "mpd#" + id),
       server_(nullptr),
-      port_(config["port"].asUInt()),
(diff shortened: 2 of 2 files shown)


cole-h:
Works for me so far. My test case of `mpc toggle;sleep 60;mpc toggle` no longer segfaults!

I'll be pausing throughout the day to see if `sleep`ing longer reproduces, but for now LGTM! :+1:

Alexays:
Nice job, thanks for taking time for that!

PR #284, opened 2019-04-23
Update fmt build dependency

### LINKED ISSUE #279: Compilation fails after requiring a missing fmt/chrono.h
The fmt.wrap file included at the subprojects folder is requiring the 5.2.1 version of the fmt library which doesn't include the chrono header support. This causes the build to fail:
```
../src/modules/mpd.cpp:3:10: fatal error: fmt/chrono.h: No such file or directory
 #include <fmt/chrono.h>
```

The correct version should be >=5.3.0. This version included the experimental chrono support:
https://github.com/fmtlib/fmt/releases/tag/5.3.0

Updated wrap file:
https://wrapdb.mesonbuild.com/v1/projects/fmt/5.3.0/1/get_wrap

Rikorose:
Closes #279

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -42,7 +42,7 @@ add_global_link_arguments(cpp_link_args, language : 'cpp')
 
 thread_dep = dependency('threads')
 libinput = dependency('libinput')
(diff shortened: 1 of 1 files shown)


Alexays:
Oh thanks, forgot about that!

Issue #223, opened 2019-03-15
timezone on clock module


spk:
Hi,
Nice project thanks ! It would be nice to be able to set the timezone on the config file so we don't have to bother to set the `TZ` env var and reload the session, wdyt ?
Cheers

rif:
Also, the possibility of adding multiple clocks in different timezones would be a plus!

layus:
I misunderstood this one as being about displaying the current timezone in the clock widget, which can already be done with `%Z` of course.

At the moment, waybar does not detect system timezone changes, but this should be fixed soon (see https://github.com/Alexays/Waybar/issues/479). By default, the clock should reflect the system's timezone.

So this feature chiefly adds
> [...] the possibility of adding multiple clocks in different timezones [...].

mjec:
I'd be happy to work on this (specifically, being able to specify the TZ for a particular clock module), and I think the easiest way to make this happen would be to bring in [Howard Hinnant's `tz.h`](https://github.com/howardhinnant/date) -- is adding a new dependency for this ok?

ZeiP:
Adding multiple clocks with different time zones is the most important functionality I'm missing when changing from py3status. Is there something I can do to help get this done?

layus:
@mjec @ZeiP Since Alexays [seems okay](https://github.com/Alexays/Waybar/pull/285#issuecomment-486000055) with adding a dependency on tz.h, there remains to make the actual pull request.
That would entail 
1. Updade clock.cpp to actually use tz.h
2. Augment the clock module config to support a timezone argument.
3. Check that it works correctly with two modules (i.e. the timezones are actually different).

Most of this should be straightforward, given that clock.cpp is 31 lines :-). A potential unknown pain point will be the formatting, but I guess there are ways to convert a tz.h time to a more standard value for formatting with fmt.

mjec:
PR #560 created which adds timzeone support -- feedback extremely welcome, as I don't normally write C++.

PR #285, opened 2019-04-23
Add timezone config to clock module

spk:
Hi, not sure its a clean way to do that but it work and fallback to the OS TZ when timezone is invalid, wdyt ?
ref https://github.com/Alexays/Waybar/issues/223

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -43,6 +43,9 @@ add_global_link_arguments(cpp_link_args, language : 'cpp')
 thread_dep = dependency('threads')
 libinput = dependency('libinput')
 fmt = dependency('fmt', version : ['>=5.3.0'], fallback : ['fmt', 'fmt_dep'])
+date_dep = dependency('date',
+        default_options : [ 'use_system_tzdb=true' ],
+        fallback: ['date', 'date_dep'])
 wayland_client = dependency('wayland-client')
 wayland_cursor = dependency('wayland-cursor')
 wayland_protos = dependency('wayland-protocols')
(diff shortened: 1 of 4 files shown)


Alexays:
Thanks! But that's not compatible with multiple clock modules :/
Maybe we can give a shot to https://howardhinnant.github.io/date/tz.html

spk:
I'm having issue using the lib with just [this](https://github.com/Alexays/Waybar/pull/285/commits/0af3c2260b24fef44c5542fcddd36290d726ae86) I'm having this error on ninja build:

```
/usr/bin/ld: waybar@exe/src_modules_clock.cpp.o: in function `waybar::modules::Clock::update()':
clock.cpp:(.text+0x84): undefined reference to `date::locate_zone(std::basic_string_view<char, std::char_traits<char> >)'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

I will need some help @Alexays if you can, I'm not a cpp expert and maybe I did some misconfiguration with the meson build system that is new to me, thanks

Alexays:
@spk Just tested and seems a pb with the wrap file :/
I'll test something

spk:
Look like related with this issue https://github.com/HowardHinnant/date/issues/398 tried to set `HAS_STRING_VIEW=0` on the meson.build of the subproject but had same issue ¯\_(ツ)_/¯

nerddan:
> Thanks! But that's not compatible with multiple clock modules :/
> Maybe we can give a shot to https://howardhinnant.github.io/date/tz.html

Having multiple clock modules with different tz is impossible? That's quite a shame!

Alexays:
Done in #560

PR #287, opened 2019-04-25
Allow rotating label contents by specifying a new "rotate" property 

jonvaldes:
This change allows rotation of label contents, which can be useful when using a vertical bar.
See:
waybar

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -18,7 +18,7 @@ waybar::ALabel::ALabel(const Json::Value& config, const std::string& format, uin
   }
 
   if (config_["rotate"].isUInt()) {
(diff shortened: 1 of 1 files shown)


jonvaldes:
Oops, just realized indentation issue with tabs vs spaces. Correcting in next commit

Alexays:
Thansk! also can you update the wiki?

jonvaldes:
Wiki updated :)

jonvaldes:
By the way, and so that I don't forget, I tried to make this work by only modifying the CSS, as text orientation is really a styling thing and not a configuration thing. But unfortunately it seems like that property hasn't been exposed in the CSS system by the GTK team, so it doesn't seem to be possible to do it there :disappointed:

PR #297, opened 2019-05-02
feat(percent): adds a percent class to numeric modules

dlasky:
not sure if you are interested in this as a feature or not, but it basically enables the use of additional classes along the lines of:

`
#backlight.percent95 {
      background: linear-gradient(to right, rgba(200,200,200,0.2) 95%, transparent 5%)
  }
`

definitely needs more effort but I wanted to get a sentiment check before i put more work into it.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -34,7 +34,6 @@ class Battery : public ALabel {
   void                                   worker();
   const std::string                      getAdapterStatus(uint8_t capacity) const;
   const std::tuple<uint8_t, std::string> getInfos() const;
-  // const std::string                      getState(uint8_t) const;
 
   util::SleeperThread   thread_;
   util::SleeperThread   thread_timer_;
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -34,7 +34,7 @@ class Battery : public ALabel {
   void                                   worker();
   const std::string                      getAdapterStatus(uint8_t capacity) const;
   const std::tuple<uint8_t, std::string> getInfos() const;
-  const std::string                      getState(uint8_t) const;
+  // const std::string                      getState(uint8_t) const;
 
   util::SleeperThread   thread_;
   util::SleeperThread   thread_timer_;
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -43,6 +43,7 @@ class Battery : public ALabel {
   int                   fd_;
   std::vector<int>      wds_;
   std::string           old_status_;
+  std::string           old_capacity_;
 };
 
 }  // namespace waybar::modules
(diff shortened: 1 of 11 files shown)


Alexays:
What do you think about defining some ranges in the backlight config and so if a range match the context it'll add the defined class?

dlasky:
that's actually how i did it for battery, and was going to copy the states config into a few other places when i realized it wouldn't work. So im definitely not opposed to that, I just thought it would be lighter to output the numbers directly.  The ideal would probably be to have some sort of CSS variable or maybe the GTK @ vars that could dynamically update the styling, but i have no clue how to do that.

I may need some help with the range code if thats the way you want me to do it, cause TBH i don't have a very good understanding of how that piece works.

I think also having that much stuff in the json (like ~20 entries per module) is causing issues, cause i've gotten a few crashes with that config:

`
"battery": {
    "bat": "BAT1",
    "states": {
        "p0":0,
        "p5":5,
        "p10":10,
        "p15":15,
        "p20":20,
        "p25":25,
        "p30":30,
        "p35":35,
        "p40":40,
        "p45":45,
        "p50":50,
        "p55":55,
        "p60":60,
        "p65":65,
        "p70":70,
        "p75":75,
        "p80":80,
        "p85":85,
        "p90":90,
        "p95":95,
        "p100":100
    },
`

dlasky:
@Alexays easier that i thought. good suggestion.

include/modules/battery.hpp
```diff
@@ -34,7 +34,7 @@ class Battery : public ALabel {
   void                                   worker();
   const std::string                      getAdapterStatus(uint8_t capacity) const;
   const std::tuple<uint8_t, std::string> getInfos() const;
-  const std::string                      getState(uint8_t) const;
+  // const std::string                      getState(uint8_t) const;
```
Alexays: Remove this line

src/modules/battery.cpp
```diff
@@ -109,29 +109,7 @@ const std::string waybar::modules::Battery::getAdapterStatus(uint8_t capacity) c
...
-    if (capacity <= state.second && valid_state.empty()) {
-      label_.get_style_context()->add_class(state.first);
-      valid_state = state.first;
-    } else {
-      label_.get_style_context()->remove_class(state.first);
-    }
-  }
-  return valid_state;
-}
+
```
Alexays: Same

Alexays:
Nice work!

src/modules/battery.cpp
dlasky: oh did you mean the now redundant method here? i deleted the space but you probably meant the extra getState implementation?

Alexays: Nope that's it

Alexays:
Can you update the wiki? :)

dlasky:
absolutely

PR #313, opened 2019-05-11
remove empty and solo classes when they do not apply

apiote:


Alexays:
How I missed that, thx!

Issue #301, opened 2019-05-04
Pulseaudio scroll-step does not correlate to volume percentage


cjbassi:
i.e. scroll-step = 1 does not mean that scrolling one tick will change volume by 1 percentage point.

I have to scroll several times to change the percentage by 1 point. Changing the scroll-step multiplier to about 650 fixes this for me, but I don't think it's exact. Does anyone else have this problem? Or is it something on my end?

varac:
Same here, weird, but not a big issue for me

PR #316, opened 2019-05-14
Change scroll-step unit to percent

jeremejevs:
Fixes https://github.com/Alexays/Waybar/issues/301

~~`scroll-step` is probably not what most people want, but kept it for backwards compatibility. `scroll-step-percent = 1` is now effectively the same as `scroll-step = 6.5535`.~~

`scroll-step = 1` ≈ 1% now.

Better alignment with the percentage ticks could be achieved (by calculating the change needed to reach the next tick, instead of adding the same number every time), but eh, negligible, IMO.

Let me know if you're happy with the API, I'll update the wiki.

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -111,7 +111,7 @@
         "format-disconnected": "Disconnected ⚠"
     },
     "pulseaudio": {
-        //"scroll-step": 1,
+        // "scroll-step": 1, // %, can be a float
         "format": "{volume}% {icon}",
         "format-bluetooth": "{volume}% {icon}",
         "format-muted": "",
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -77,7 +77,8 @@ bool waybar::modules::Pulseaudio::handleVolume(GdkEventScroll *e) {
     return false;
   }
   bool       direction_up = false;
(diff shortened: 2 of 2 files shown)


Alexays:
Thanks!
We should print a message to deprecate `scroll-step`.
Also with a touchpad, scrolling is way too fast, we should calculate the next tick..

jeremejevs:
Where should I print the message about the deprecation? Haven't found any existing messages.

About the touchpad scrolling - it's a different and a more complicated issue, and should be solved in a separate PR, IMO. A potential solution that comes to mind would be to keep track of `delta_y` between scroll events and adjust the volume every N pixels scrolled, but I don't know what the N needs to be. Can it be the same for everybody? Or should it be configurable?

Alexays:
After reflection, just leave `scroll-step` and remove the old behavior, people using it will see the change and tweak for new values, what do you think?

jeremejevs:
Up to you :slightly_smiling_face: Since this is pre-1.0, it's technically okay, but is there a timeline for 1.0? I've been annoyed by broken modules multiple times by now, and it would be nice to have some stability at some point.

Alexays:
Configuration will be frozen from 0.8.0, i'll make a milestone for the 1.0 :)

Alexays:
Thanks!

PR #317, opened 2019-05-15
fix(battery): change comparison expr. to assignment

jorgenbele:
It seems like there is a mistaken comparison expression where an assignment statement should be used instead. See the commit for the single line which I am referring to.


Alexays:
Oh my bad! Thx :)

PR #318, opened 2019-05-16
Fix state behavior

unresolvedsymbol:


Alexays:
Thanks, Just pushed to combine the two params.

Issue #225, opened 2019-03-15
Ability to configure an action when scrolling on "blank space"


RX14:
i3bar and swaybar have the behaviour of being able to scroll anywhere on the bar which doesn't have another scroll action to switch workspaces, i'd love to be able to replicate this behaviour in waybar.

PR #322, opened 2019-05-16
Allow scrolling on the entire bar surface

RX14:
It's ugly, but it works.

Closes https://github.com/Alexays/Waybar/issues/225

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -55,6 +55,7 @@ class Bar {
   void setMarginsAndZone(uint32_t height, uint32_t width);
   auto setupWidgets() -> void;
   void getModules(const Factory &, const std::string &);
+  bool handleScroll(GdkEventScroll*);
   void setupAltFormatKeyForModule(const std::string &module_name);
   void setupAltFormatKeyForModuleList(const char *module_list_name);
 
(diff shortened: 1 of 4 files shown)


Alexays:
Thansk!
Oh I didn't understand it like that, I'm going to push on this pr to make the code cleaner :)

RX14:
Thanks! This code is definitely a lot cleaner.

However, the name "disable-workspace-scroll" doesn't make 100% sense with the new location in the config. It should probably be called `"disable-bar-scroll"` now.

Alexays:
Sure you can make a PR and update the wiki if you have time :)

RX14:
Done in https://github.com/Alexays/Waybar/pull/326

RX14:
So, this is supposed to *not* scroll the workspaces when you scroll on a widget with a scroll handler set, for example the pulseaudio widget. 

However, somehow the scroll event gets delivered to the window scroll event handler *first*, *then* the event handler for the widget?? Which, from reading the gtk event handling docs, makes absolutely no sense. And doing some `gdb` debugging, it seems there are two different events?

I'm confused by this, and I'd like to fix this so that widgets with scroll handlers eat the scroll events and do not scroll workspaces, but I don't have time to debug further.

PR #328, opened 2019-05-18
Adding bandwidth support for network module

Organic-Code:
This PR adds support for 4 new tags for the network modules:
* `{bandwidthDownBits}` and `{bandwidthUpBits}`
* `{bandwidthDownOctets}` and `{bandwidthUpOctets}`

Those tags, in combination with the `{interval}` tag, allows users to monitor ongoing network traffic

As of current state, this PR implements those only for linux, but I tried to make it easy to add support for FreeBSD
As I am not running under FreeBSD, I'd rather have some FreeBSD dude add this part. FreeBSD implementation can apparently be done using the [/proc/net/dev](https://www.freebsd.org/cgi/man.cgi?query=proc&apropos=0&sektion=5&manpath=CentOS+7.1&arch=default&format=html) file

If no one is willing to implement this, then I'll go ahead and try to implement it, though I won't be able to test it and would as such like to avoid it

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -302,10 +302,10 @@ auto waybar::modules::Network::update() -> void {
                           fmt::arg("cidr", cidr_),
                           fmt::arg("frequency", frequency_),
                           fmt::arg("icon", getIcon(signal_strength_, connectiontype)),
-                          fmt::arg("bandwidthDownBits", pow_format(bandwidth_down * 8ull, "b")),
-                          fmt::arg("bandwidthUpBits", pow_format(bandwidth_up * 8ull, "b")),
-                          fmt::arg("bandwidthDownOctets", pow_format(bandwidth_down, "o")),
-                          fmt::arg("bandwidthUpOctets", pow_format(bandwidth_up, "o")));
+                          fmt::arg("bandwidthDownBits", pow_format(bandwidth_down * 8ull / interval_.count(), "b/s")),
+                          fmt::arg("bandwidthUpBits", pow_format(bandwidth_up * 8ull / interval_.count(), "b/s")),
+                          fmt::arg("bandwidthDownOctets", pow_format(bandwidth_down / interval_.count(), "o/s")),
+                          fmt::arg("bandwidthUpOctets", pow_format(bandwidth_up / interval_.count(), "o/s")));
   label_.set_markup(text);
   if (tooltipEnabled()) {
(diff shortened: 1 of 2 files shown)


src/modules/network.cpp
```diff
@@ -16,6 +81,21 @@ waybar::modules::Network::Network(const std::string &id, const Json::Value &conf
   if (!id.empty()) {
     label_.get_style_context()->add_class(id);
   }
+
+  auto down_octets = read_netstat(BANDWIDTH_CATEGORY, BANDWIDTH_DOWN_TOTAL_KEY);
```
Alexays: Why this part is needed?
The update method is also called in constructor :)

src/modules/network.cpp
```diff
@@ -1,5 +1,70 @@
 #include "modules/network.hpp"
 #include <sys/eventfd.h>
+#include <fstream>
+
+namespace {
```
Alexays: As the network module starts to get bigger I was thinking about splitting it into two files: netlink.cpp, network.cpp

Alexays:
I don't think all waybar modules are freebsd compatible atm so we can skip it, or if you're really motivated you can try to implement it :smile:

src/modules/network.cpp
Organic-Code: Because InOctets and OutOctets are equal to the total number of octets sent/received since the start of the computer, and as such might not be 0 at the start of Waybar

Otherwise the module may display an artificial network activity spike for the first update

src/modules/network.cpp
Organic-Code: Yeah that could be better to split out this module
Though I admit I don't know what should go in what file, terminology wise D:

Organic-Code:
If it's not a problem then I'd prefer to let it as is, so it's more obvious it does not work

Organic-Code:
By the way, adding logging I saw std::cout was used directly so I used std::cerr aswell, hopefully that's fine

Would you consider a PR adding spdlog or something similar ?

src/modules/network.cpp
Alexays: I'll take care of it :)

Alexays:
Yeah It can be good to use spdlog.
https://wrapdb.mesonbuild.com/spdlog

Alexays:
Thanks!
Could you update the wiki?

Organic-Code:
Updated it

Issue #210, opened 2019-03-09
Show all workspaces?


ReekyMarko:
I'm very new to Sway and Waybar, so sorry if this is a stupid question.

I have my setup almost complete, but I would like to show all workspaces, even empty ones. Is this possible to do somehow? The reason for doing this, is if I open something in workspace 2 before workspace 1, they appear in the wrong order on the bar which is very confusing.

FlorianFranzen:
Yes, just add the following to your config:
```
"sway/workspaces": {
    "all-outputs": true
}
```

ReekyMarko:
I already have that in my config

Alexays:
Sway only gives the workspace that are created, so it is not possible unless the user specifies the workspace in the waybar config file and we can add this feature

AGCaesar:
I would love this feature. AwesomeWM has it as a default behaviour, but sway refuses do implement this. @Alexays what do you mean with specify the workspaces in waybar config?
I tried this, but it still only showed the first workspace:

 "sway/workspaces": {
         "disable-scroll": false,
         "all-outputs": true,
         "format": "{name}: {icon}",
         "format-icons": {
             "1": ".",
             "2": "",
             "3": "",
             "4": ".",

Alexays:
@AGCaesar Yeah it's just a exemple to how we can implement this feature

AGCaesar:
Please try to implement it! I searched for persistent workspaces for sawy/i3 and found many people who would love this feature!
Thanks for your work with this beautiful and easy to use bar!

Alexays:
Done in #330

PR #330, opened 2019-05-18
Adding sway/workspaces:persistant_workspaces

Organic-Code:
Implements https://github.com/Alexays/Waybar/issues/210

Adds the `persistant_workspaces` option to the module `sway/workspaces`.
Within `persistant_workspaces`, you can list workspaces that should always be showed, even when non existant as per sway. It is possible to indicate which outputs they should be displayed on ; if no output was indicated, it will display on all outputs.

example:
```
"sway/workspaces": {
    "persistant_workspaces": {
        "3": [], // Always show a workspace with name '3', on all outputs if it does not exists
        "4": ["eDP-1"], // Always show a workspace with name '4', on output 'eDP-1' if it does not exists
        "5": ["eDP-1", "DP-2"] // Always show a workspace with name '5', on outputs 'eDP-1' and 'DP-2' if it does not exists
    }
}
```

Known issue: upon clicking on a non-existing workspace, the new workspace will be created and opened on the output that is currently focused, instead of on the one where the click occured (except if there are further restriction on sway's config)

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -35,6 +35,7 @@ class Workspaces : public IModule, public sigc::trackable {
   const Bar&                                   bar_;
   const Json::Value&                           config_;
   std::vector<Json::Value>                     workspaces_;
+  std::vector<std::string>                     workspaces_order_;
   waybar::util::SleeperThread                  thread_;
   std::mutex                                   mutex_;
   Gtk::Box                                     box_;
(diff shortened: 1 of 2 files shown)


Alexays:
Just pushed a commit in order to focus the correct workspace, and fix a order pb.
LGTM! Nice work!

Organic-Code:
Fixed another ordering pb :stuck_out_tongue:
If it's good for you, I'll update the wiki then

Thanks!

sethidden:
Sorry for necro, but if anyone comes across this in the future, this is now persist**a**nt_workspaces, not persist**e**nt_workspaces. See https://github.com/Alexays/Waybar/commit/2646390e2817eab0c6f37a0fcea2daebe7aac8c8

ardishco-the-great:
I was about to point that out lmao

ardishco-the-great:
> will this work on hyperland ?

Yes, all wlroots based compositors will work with Waybar's workspace module. 

I am on Hyprland now, and can confirm this is the case

PR #331, opened 2019-05-18
Estimate of time remaining until empty|full on tooltip

mreppen:
Linear estimate like `upower`.

Also removed `current` variables as they are not used since 17291dffdf623f21fe9902d59a0cf90b4f6a594e


Alexays:
Thanks!
Could you also add a format argument for that?

mreppen:
The format argument is there now. Do you want me to squash them?

Alexays:
No that's ok :)
Thanks!
Can you update the wiki?

PR #332, opened 2019-05-19
Adding spdlog

Organic-Code:
Adding spdlog and replacing most text printing

I barely know how to use meson, so hopefully I did it mostly right. Didn't find how to include spdlog while disabling compiler warnings for it though


Alexays:
Thanks!

PR #333, opened 2019-05-19
Fix clock is always a second off

RX14:
I don't understand why these two clocks disagree, but they do, and this way always seems to update the second exactly on time.


Alexays:
Does this if still needed in order to show correct minutes with an interval of 60 seconds?
https://github.com/Alexays/Waybar/blob/5e43b4f587d373f52edf8c3d65e0cac2ff513fc0/src/modules/clock.cpp#L17-L19

Alexays:
Thanks!

PR #334, opened 2019-05-19
Allow formatting memory with used and available memory

RX14:
This adds named format parameters while keeping existing formats using `{}` working.


Alexays:
Thanks!
Can you update the wiki?

RX14:
Done!

PR #336, opened 2019-05-19
backlight: fix for percentage output for amdgpu_bl0 device

isheshukov:
replaced `actual_backlight` with `backlight``


Alexays:
Thanks!

Issue #21, opened 2018-08-15
Libappindicator tray


Alexays:
Tray beta now shipped :)

Bugs I noticed:
- Menu not correctly positioned (also Tooltips) #63 
- ~~Support multi modules~~
- ~~With multiple apps in tray, only the last is interactive, the others cause crash~~

### Electron apps
You need `libappindicator-gtk3` (added to AUR package)
and launch apps with `env XDG_CURRENT_DESKTOP=Unity slack`

Alexays:
Not at the moment, but it's a good feature to have :)

jeremejevs:
I guess I'll report here, instead of spamming issues, since this a beta feature.

When restarting the bar, only Telegram and Keybase icons remain. Slack and Caprine are gone.

Also, Discord never appears at all, for some reason, even with the latest fixes.

jeremejevs:
Upon further usage, it looks like Keybase doesn't re-appear consistently either, only sometimes.

Right-clicking the Keybase tray outright crashes the bar, every time. Other apps crash the bar too when right-clicking, but I can't consistently reproduce it.

Sometimes the bar fails to start. Can't consistently reproduce either, but it eventually happens after launching and killing it repeatedly (segfault, no special error).

Not sure if of any use, but when launching the bar I get these two:

```
An object is already exported for the interface org.kde.StatusNotifierWatcher at /StatusNotifierWatcher
GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: Status Notifier Host with bus name ':1.308' and object path '/StatusNotifierHost' is already registered
```

Alexays:
Do you have a multi-screen configuration?
And the fact that the slack icon doesn't change is normal, it's just that there's no code to manage the tray change 😆 
I can't reproduce the right click :/

jeremejevs:
Yep, I see where this is going :slightly_smiling_face: I have two monitors, both have a bar.

Alexays:
@jeremejevs Can your check if the crash is still present in the last commit? :)

jeremejevs:
Yep, the issue is still there, unfortunately. I have also noticed that in some conditions, if the first icon you right-click on after launching Waybar is Keybase, then right-clicking on other icons still opens Keybase's tray menu. Had that happen with Telegram too. Peculiar.

Alexays:
@jeremejevs I found time to work on it a little bit and it's normally fixed now :)

jeremejevs:
[...]

PR #339, opened 2019-05-20
Tray error handling and logging improvements

alebastr:
Hopefully, that will address or at least explain issues like [this comment](https://github.com/Alexays/Waybar/issues/21#issuecomment-489599739). And make troubleshooting a bit easier for those who are afraid of C++ :smile: 
I had to add log level command line option as there's no way to set it otherwise. Could be reverted if you feel there's a better option, like config file key or whatever else.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -243,10 +243,14 @@ int waybar::Client::main(int argc, char *argv[]) {
   std::string config;
   std::string style;
   std::string bar_id;
+  std::string log_level;
   auto        cli = clara::detail::Help(show_help) |
              clara::detail::Opt(show_version)["-v"]["--version"]("Show version") |
              clara::detail::Opt(config, "config")["-c"]["--config"]("Config path") |
              clara::detail::Opt(style, "style")["-s"]["--style"]("Style path") |
+             clara::detail::Opt(
+                 log_level,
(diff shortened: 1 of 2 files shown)


src/client.cpp
```diff
@@ -243,10 +243,14 @@ int waybar::Client::main(int argc, char *argv[]) {
...
   std::string style;
   std::string bar_id;
+  std::string log_level;
   auto        cli = clara::detail::Help(show_help) |
              clara::detail::Opt(show_version)["-v"]["--version"]("Show version") |
              clara::detail::Opt(config, "config")["-c"]["--config"]("Config path") |
              clara::detail::Opt(style, "style")["-s"]["--style"]("Style path") |
+             clara::detail::Opt(
+                 log_level,
+                 "trace|debug|info|warning|error|critical|off")["-l"]["--log-level"]("Log level") |
```
alebastr: This doesn't look nice, but I haven't found a way to improve formatting without removing list of levels.
```
where options are:
  -?, -h, --help                            display usage information
  -v, --version                             Show version
  -c, --config <config>                     Config path
  -s, --style <style>                       Style path
  -l, --log-level <trace|debug|info         Log level
  |warning|error|critical|off>
  -b, --bar <id>                            Bar id
```

Alexays:
Thanks!
Nice addition :)

Issue #369, opened 2019-06-05
forkExec leaves zombies


toke:
Processes spawned by `forkExec` - via `on-click` or `on-click-right` leaves zombie processes.

To reproduce.

```json
   "custom/demo": {
        "format": "Zombienator{}",
        "exec": "date",
        "return-type": "newline",
        "on-click": "touch /tmp/zombietest",
        "on-click-right": "rm /tmp/zombietest"
    }
```
After clicking on the "Zombienator" defunct processes are visible in the process list.

toke:
Probably a similar issue is happening with the other `open` method.

Alexays:
Fixed in #370

Issue #358, opened 2019-05-30
Child process does not stops


niktob560:
After killing waybar (for refreshing config, etc), child processes keeps working  
I spent 10 minutes on killing custom script with loop for checking capslock state  
Fix that plz :)  
ArchLinux, waybar-git from AUR  
Waybar v0.6.6-28-gfcf2d18 (May 30 2019, branch 'master')

toke:
I had looked into that and I have a "rough" version. But as I'm not fluent in C++ it's a bit harsh and still dies with a segfault some seconds later (which is not as bad a it has to be exited anyway ;))

BUT: As a fix for (at least bash-scripts) it should be possible to add `set -euo pipefail` in them and if one of the outputs fail it should exit.

b-zee:
I have encountered the same problem and can confirm processes do not exit. I was testing styles and 'reloaded' sway multiple times.

Excerpt from `ps --forest aux`:

```
user      17502  0.0  0.0   7172  2804 tty1     S    13:12   0:00 sh -c python $HOME/.config/waybar/mediaplayer.py 2> /dev/null
user      17509  0.0  0.1 177584 19112 tty1     Sl   13:12   0:00  \_ python /home/user/.config/waybar/mediaplayer.py
user      17567  0.0  0.0   7172  2856 tty1     S    13:12   0:00 sh -c python $HOME/.config/waybar/mediaplayer.py 2> /dev/null
user      17574  0.0  0.1 177584 19208 tty1     Sl   13:12   0:00  \_ python /home/user/.config/waybar/mediaplayer.py
user      17657  0.0  0.0   7172  2952 tty1     S    13:14   0:00 sh -c python $HOME/.config/waybar/mediaplayer.py 2> /dev/null
user      17663  0.0  0.1 177584 19188 tty1     Sl   13:14   0:00  \_ python /home/user/.config/waybar/mediaplayer.py
user      17737  0.0  0.0   7172  2724 tty1     S    13:15   0:00 sh -c python $HOME/.config/waybar/mediaplayer.py 2> /dev/null
user      17743  0.0  0.1 177584 19116 tty1     Sl   13:15   0:00  \_ python /home/user/.config/waybar/mediaplayer.py
user      17928  0.0  0.0   7172  2772 tty1     S    13:16   0:00 sh -c python $HOME/.config/waybar/mediaplayer.py 2> /dev/null
[...]

PR #370, opened 2019-06-05
Prevent zombie apocalypse by ignoring SIGCHLD

toke:
Fixes Issue #369


Alexays:
Thanks!
that's a start into fixing #358 cause that doesnt kill children when waybar get killed.

PR #372, opened 2019-06-08
Increase specificity of media stylings

rianadon:
Because of CSS specificity rules, the `#custom-media` style will always override the `custom-spotify` and `custom-vlc` styles, so the background of the media element is always green rather than sometimes orange when VLC is running. I added `#custom-media` in front of each of the class selectors to increase their specificity so they take precedence.

### DIFF
--- a/resources/style.css
+++ b/resources/style.css
@@ -136,11 +136,11 @@ label:focus {
     min-width: 100px;
 }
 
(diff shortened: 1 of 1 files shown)


Alexays:
Nice catch!
Thanks :)

PR #373, opened 2019-06-08
adds the wl output name as a css class

razic:
now you can have a custom styling for each bar. this wasn't possible before, since every bar defined the same set of ids and classes.

also question:

the height/width parameters are useless inside of the css file. i noticed the width and height were being set via a different api. is there any way we can set this to "auto" so the css can take precedence over the width/height of the bar?

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -17,6 +17,7 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
   window.set_title("waybar");
   window.set_name("waybar");
   window.set_decorated(false);
+  window.get_style_context()->add_class(output->name);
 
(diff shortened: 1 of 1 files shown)


razic:
setting the css height of `#waybar` does not work essentially. this may be fine now that i think about it. changing padding on the individual labels increased the height.

Alexays:
Thanks, i'll look for the height, i may have a idea :)

razic:
hey thanks for merging so quick :) let me know what you find! i'm super curious :)

also i opened up https://github.com/swaywm/sway/issues/4231 as i've basically encountered the same problem with not being able to set font size for a menu based on which output it's on.

then i thought about styling the titlebars themselves with css just like waybar. i'd be interested to hear if you have thought about this as well.

PR #381, opened 2019-06-11
Fix twitchy scrolling on touchpads

Synthetica9:
Previously, any and all scroll events were interpreted as reason to switch
workspaces. This resulted in twitchy behaviour, where the scrolling was
practically unusable.

Now, we pool all scroll values, and only scroll if the value is larger than the
new config option "smooth-scrolling-threshold". If this option is not set, the
behaviour is unchanged.

### DIFF
--- a/include/ALabel.hpp
+++ b/include/ALabel.hpp
@@ -36,6 +36,7 @@ class ALabel : public IModule {
 
  private:
   std::vector<int> pid_;
+  gdouble distance_scrolled_;
 };
 
 }  // namespace waybar
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -41,6 +41,7 @@ class Workspaces : public IModule, public sigc::trackable {
   util::JsonParser                             parser_;
   bool                                         scrolling_;
   std::unordered_map<std::string, Gtk::Button> buttons_;
+  gdouble                                      distance_scrolled_;
 
   util::SleeperThread thread_;
   Ipc                 ipc_;
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -97,11 +97,23 @@ bool waybar::ALabel::handleScroll(GdkEventScroll* e) {
   }
   if (e->direction == GDK_SCROLL_SMOOTH) {
     gdouble delta_x, delta_y;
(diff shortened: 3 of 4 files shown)


Synthetica9:
Also, there is inconsistency between the discrete and continuous case. In the discrete case horizontal scroll is also considered, in the continuous case this is ignored. Do we want to change this behaviour? How should this behave on "diagonal" scrolling?

Alexays:
The actual scrolling behaviour is garbage, feel free to redo it completely :)
Also there is also a handleScroll in `ALabel.cpp` if you can also apply your changes here :)

Synthetica9:
@Alexays Should the threshold be configurable for every individual module, or is it possible to add some global option?

Alexays:
@Synthetica9 configurable for each module, soon I will make a kind of global configuration that the modules will inherit

Synthetica9:
@Alexays and the diagonal case?

Alexays:
What do you think is better?
Imo there is no need of the diagonal case.

Synthetica9:
I see a few options:

1. Only use the principle axis,
2. Add the axes, 
3. Ignore "diagonal" scrolls, and only use events that are within a few degrees of either horizontal or vertical. 

1 and 2 seem to work pretty well. I haven't programmed 3 yet.

Synthetica9:
(This implementation is 1. Horizontal scrolling doesn't seem to make sense for ALabel)

Synthetica9:
@Alexays What's currently holding this up?

Alexays:
I've been busy, i'm currently reviewing it :)

Alexays:
Seems quite good!
Just i think it's better to rename `smooth-scrolling-threshold` to `scrolling-threshold`, right?

Synthetica9:
> Just i think it's better to rename smooth-scrolling-threshold to scrolling-threshold, right?

Maybe? It isn't relevant to all scrolling, only to the continuous case

Alexays:
> > Just i think it's better to rename smooth-scrolling-threshold to scrolling-threshold, right?
> 
> Maybe? It isn't relevant to all scrolling, only to the continuous case

Yeah, right, even if I think the smooth case is used in almost all cases, we'll keep `smooth-scrolling-threshold`

Thanks!

PR #385, opened 2019-06-14
Expose total memory to the formatting directive

nenad:
I would like to be able to see the memory usage expressed like `3.3G / 8.0G`, which is not possible at the moment. This PR solves this.

I also noticed that there is no option to display the swap values. If you agree, I'd like to make a PR for that as well.

Note: Wiki page needs to be updated if this gets merged.


Alexays:
Thanks!
For sure, we might need #338 in order to avoid unneeded logic :)
Can you update the wiki?

nenad:
Yeap, on it!

PR #387, opened 2019-06-15
refactor: AModule

Alexays:
@Synthetica9 I took the time to do it :)

PR #389, opened 2019-06-16
AModule: handle X axis scrolling

Synthetica9:
(This is against the `clean` branch, so this should update #387)


Alexays:
Thanks!

PR #391, opened 2019-06-17
Cleanup on clean branch

Synthetica9:
- Implemented review by @panaman67
- General cleanup.

Again, this is against `clean`, affecting #387

(Should probably merge with a rebase, don't see a reason to insert the extra merge commit here)


Alexays:
Thanks!

PR #394, opened 2019-06-28
Conditionally execute modules

jeffa5:
This adds a `show-if` field to the config JSON for some core modules. This allows a user to define a script which is executed to see if the module should be displayed. If this field is not present there should be no change in behaviour.

The scripts are called with relevant arguments. I imagine these would need to be documented somewhere (perhaps the wiki?) with the rest of the details for users.

This is effectively `exec-if` from the custom module but I found `show-if` to be a bit more instructive.

Some things I'm unsure about:
- whether this should be implemented for more or fewer modules
- a good approach to the order of arguments passed to the scripts
-  the choice of arguments passed to the scripts

### DIFF
--- a/include/util/condshow.hpp
+++ b/include/util/condshow.hpp
@@ -1,0 +1,22 @@
+#pragma once
+
+#include "util/command.hpp"
+#include "util/json.hpp"
+
+namespace waybar::util::condshow {
+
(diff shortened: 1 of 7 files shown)


Alexays:
Thanks, but it'll be cleaner to have something like #100,
I'll look on it today :)

Issue #267, opened 2019-04-15
Add support for multiple bar's with the same file


thepenguinthatwants:
When I used polybar I've gotten used to having a bar in bottom and a bar on top. Is it possible with Waybar?

Wondering if its already achievable by now? If so how?

Difrex:
Just use a different config for each bar.
It working.
![2019-04-15_16-49-47](https://user-images.githubusercontent.com/3982213/56138038-b05f0700-5f9e-11e9-81d6-3c043a1c7e18.png)

Alexays:
You can also use a single file, look at #147 :)

thepenguinthatwants:
Thanks I guess I almost understand how it should work.

For some reasons I get only 1 waybar and the bottom one doesnt show. Or does it need another value as layer?

```
[{
    "layer": "top", // Waybar at top layer
 "position": "top", // Waybar position (top|bottom|left|right)
    "modules-left": ["sway/workspaces", "sway/mode"],
    "sway/workspaces": {
        "disable-scroll": true,
        "all-outputs": true,
        "format": " {icon} ",
        "format-icons": {
            "focused": "f",
            "default": "i"
        }
   }
},

{
    "layer": "top", // Waybar at top layer
   "position": "bottom", // Waybar position (top|bottom|left|right)
    "modules-right": [  "pulseaudio"],
    "pulseaudio": {
        "scroll-step": 5,
        "format": "{volume}% {icon}",
        "format-bluetooth": "{volume}% {icon}",
        "format-muted": "",
        "format-icons": {
            "headphones": "",
            "handsfree": "",
            "headset": "",
            "phone": "",
            "portable": "",
            "car": "",
            "default": ["", ""]
        },
        "on-click": "pavucontrol"
    }
}]
```

Alexays:
@thepenguinthatwants Oh yeah if you want multiple bar on same screen without screen restriction, atm you need to launch multiple waybar process, i'll add the feature asap :)

thepenguinthatwants:
Yeah I wanted multiple bar in the same screen. Only using this on my one screen laptop.

Cool!

Alexays:
Done with 79a5e9ecee7fcb537e474071253c383ab82bf2e0

thepenguinthatwants:
Cool!

So would it work just with my earlier example config? Or does it need something special to make it work?

Alexays:
That's it!

Th3Whit3Wolf:
[...]

PR #397, opened 2019-07-04
Allow CSS to select specific waybars when multiple are in use

Jamedjo:
## What

- Adds `name` property for each bar and uses this as a CSS class
- Adds CSS class for the position property

## Why

When using multiple waybars from the same configuration file (https://github.com/Alexays/Waybar/issues/267) it can be preferable to style them individually.

E.g For a bar at the bottom of the screen the `border-bottom` might be better at the top, or a specific named bar might want to have a different background colour.

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -18,6 +18,7 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
   window.set_name("waybar");
   window.set_decorated(false);
   window.get_style_context()->add_class(output->name);
+  window.get_style_context()->add_class(config["position"].asString());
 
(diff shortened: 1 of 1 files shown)


Jamedjo:
If this merges we'll want to add `name` to the wiki under https://github.com/Alexays/Waybar/wiki/Configuration#bar-config. I've also updated the wiki to document GTK's live CSS inspector:

screenshot_2019-07-04-02:55:52

https://github.com/Alexays/Waybar/wiki/Styling/_compare/bbaa58db094d1832231b058f65c3cf8a371c10da

Alexays:
Nice addition, thanks!

Issue #367, opened 2019-06-04
enable-bar-scroll: true option broken


cinocode:
I just tried to add "enable-bar-scroll": true to my config and it seems to be broken. When I scroll (doesn't matter if on the bar or on the workspace switcher) the target workspace flickers for a split second and then the currently active workspace is focussed again.
With this option removed I con scroll on the workspace switcher without any issues.

Here's the relevant part of my waybar config:
    "sway/workspaces": {
        "format": "{name}",
        "disable-scroll-wraparound": true,
        "enable-bar-scroll": true
    },
[Or my entire waybar config - enable bar scroll is not in there](https://github.com/cinocode/dotfiles/tree/master/.config/waybar)

I'm on:
5.1.6-arch1-1-ARCH
sway 1.1.1-1
waybar 0.6.7-1

Thanks for the great work on this project!

junglerobba:
This happens if workspace_auto_back_and_forth is enabled in sway (which it is from looking at your dotfiles). I had the same issue until I disabled that, and can get it again just by turning it back on.

nenad:
Okay, I believe I figured a fix for this without disabling `workspace_auto_back_and_forth` in sway. The fix is to compile waybar (or get it from the `waybar-git` AUR package), and edit the `"smooth-scrolling-threshold"` property of the `sway/workspaces` module to a high value, or at least a value that will suit your needs.

cinocode:
Hi, thanks for the fix. I tested with the waybar-git aur package and your config. Works great.

Just FYI since I look at the PR - my issue was just with mouse scrolling (not touchpad). But the problem is reintroduced, if I delete the smooth-scrolling line from my config.
Maybe it makes sense to have a default value for the scrolling - so other people don't come accross the same issue as me?
Aside from that we could close this since it works for me now.

Synthetica9:
> Maybe it makes sense to have a default value for the scrolling - so other people don't come accross the same issue as me?

I didn't add this to preserve the old behaviour when the config file was unchanged ([xkcd 1172](https://xkcd.com/1172/) springs to mind), but the following would fix that:

```patch
diff --git a/src/AModule.cpp b/src/AModule.cpp
index 38e9785..6b5beb5 100644
--- a/src/AModule.cpp
+++ b/src/AModule.cpp
@@ -64,7 +64,7 @@ AModule::SCROLL_DIR AModule::getScrollDir(GdkEventScroll* e) {
       distance_scrolled_y_ += e->delta_y;
[...]

PR #403, opened 2019-07-06
Fix deadlock on workspace scrolling

### LINKED ISSUE #395: enable-bar-scroll causes waybar to freeze on scroll
after enabling enable-bar-scroll, whenever i scroll on waybar anywhere it changes workspace once, but then the entire bar freezes until I restart it

sheenobu:
Make the mutex guard lifecycle finish before the send ipc command
by adding scope around the code.

Fixes #395


sheenobu:
too fast scrolling was just #367 since i'm on a touchpad.

Alexays:
Thanks!

PR #404, opened 2019-07-06
WIP: dynamic args

Alexays:
I'm not happy with that atm

Modules converted:
- [x] Battery
- [x] CPU
- [x] Memory
- [x] IdleInhibitor
- [x] Custom
- [x] Temperature
- [x] sway/mode
- [x] sway/window
- [x] pulseaudio

TODO:
- [ ] Convert all modules
- [ ] Cleanup
- [ ] #300 
- [ ] Args cache

PR #405, opened 2019-07-06
fix(pulseaudio): connect scroll event handler

crwxrws:
Reconnect the event handler that was removed in commit 527144a.

Fixes Alexays#400


Alexays:
That's not the correct fix, but until i merged #404, that'll fix it.
Thanks :)

crwxrws:
That's fair, my bad. I just wanted to have the feature back for now, as the issue made it into Arch's stable community repo.

PR #421, opened 2019-07-31
 fix(custom): ignore selected_player if not defined. Fixes #419 

### LINKED ISSUE #419: Media player module doesn't appear
The refactored script works well apart from one problem, it doesn't show up if Spotify is started after Waybar. I have to kill Waybar and restart it to get the module to appear. I am using the exact script from resources and the same config block. Running waybar from the terminal I can't see any errors.

SibrenVasse:
When the command line argument `--player <name>` is not defined, custom module is not picking up new players.
Also added example usage of argument to config.


Alexays:
Thanks!

PR #439, opened 2019-08-27
Unbreak with fmtlib 6.0.0

### LINKED ISSUE #436: libfmt 6.0.0 breaks Waybar
Building Waybar master with libfmt 6.0.0 fails with the following error:
```
126 In file included from ../waybar-9999/include/factory.hpp:4,
127                  from ../waybar-9999/src/factory.cpp:1:
128 ../waybar-9999/include/modules/clock.hpp:5:10: fatal error: fmt/time.h: No such file or directory
129     5 | #include "fmt/time.h"
130       |          ^~~~~~~~
```

Full build log attached: [gui-apps:waybar-9999:20190826-191629.log](https://github.com/Alexays/Waybar/files/3542588/gui-apps.waybar-9999.20190826-191629.log)

jbeich:
Fixes #436

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -1,7 +1,7 @@
 #include "bar.hpp"
-#include <spdlog/spdlog.h>
 #include "client.hpp"
 #include "factory.hpp"
+#include <spdlog/spdlog.h>
 
 waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
     : output(w_output),
--- a/include/modules/clock.hpp
+++ b/include/modules/clock.hpp
@@ -1,8 +1,12 @@
 #pragma once
 
 #include <fmt/format.h>
+#if FMT_VERSION < 60000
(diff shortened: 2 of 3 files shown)


src/bar.cpp
```diff
@@ -1,7 +1,7 @@
 #include "bar.hpp"
-#include <spdlog/spdlog.h>
 #include "client.hpp"
 #include "factory.hpp"
+#include <spdlog/spdlog.h>
```
jbeich: Given spdlog itself bundles fmtlib we should probably pass `-DSPDLOG_FMT_EXTERNAL` from meson.build instead to keep fmt API usage consistent.

Alexays: Yeah we should use `SPDLOG_FMT_EXTERNAL`

Alexays: Will do it in another PR

Alexays:
Thanks :)

Issue #63, opened 2018-10-27
Tooltips / Menus not correctly positioned


Alexays:
Need to implement XDG Popup

cjbassi:
The tooltips show up in the middle of the screen for me, but they should appear where my mouse is. Is that what this issue is describing?

Robinhuett:
One interesting find I just made is, that everytime aa tooltip is displayed, waybar gets a new window/container id from sway.
I'm not sure if that helps finding the problem.

matt-snider:
@Alexays I'm also wondering if this refers to the same issue commented above by @cjbassi. Could you please confirm whether this is the case?

Alexays:
@matt-snider Yeah that's the part of the issue!

chzzzh:
[same here](https://github.com/swaywm/sway/issues/3951).

Alexays:
I have debugged to know where it comes from, and it's not a waybar or sway/wlroots fault, but from an update of gtk3

matt-snider:
@Alexays Thanks for the info. Is there any upstream gtk3 issue we can watch?

Alexays:
Surely https://github.com/GNOME/gtk/commit/a73f961e657014dba173ee121f62b53e4a02391c

JadeMaveric:
> One interesting find I just made is, that everytime aa tooltip is displayed, waybar gets a new window/container id from sway.
> I'm not sure if that helps finding the problem.

How do you find the window Class on wayland? One possible workaround I can think of is to use the `for_window [class=window-class] move <left|right|down|up> [<px> px]` to position the windows correctly.

Another variant of the command allows the window to be positioned at the cursor. Here's the i3 manual entry https://i3wm.org/docs/userguide.html#_moving_containers

JadeMaveric:
I tried `for_window [app_id="waybar"] move position mouse`
No, it did not work. :disappointed:  Not the way I imagined anyway

alebastr:
@JadeMaveric, I had this in my sway config. Works slightly better.
```
for_window [app_id="waybar" floating] {
    move position cursor
    move down 120px # adjust if some menus still don't fit
}
```
It's also possible to do correct positioning via ipc:
```javascript
#!/usr/bin/env node
const BAR_HEIGHT = 28;
require("i3wm")
  .Client.connect({ bin: "sway" })
  .then(client => {
    client.subscribe("window");
    client.on("window", ({ change, container }) => {
[...]

PR #441, opened 2019-08-28
Use gtk-layer-shell library for correct positioning of popups

alebastr:
Initial attempt to address #63 - positioning of menus and tooltips with [wmww/gtk-layer-shell](https://github.com/wmww/gtk-layer-shell) library.

While the approach of the library is simple — subscribe to 'realize' for all windows and patch surfaces for anything that looks like popup, there's too much magic happening inside so I'm not even going to try adapting the code directly to waybar.

### How to test:
* install sway 1.2 + wlroots 0.7.0
* install gtk-layer-shell 0.1.0 (optional; meson will fetch it from github if not present)
* checkout git branch from this pull request
* configure waybar with `-Dgtk-layer-shell=enabled` (disabled by default)

### What works:
- menus are opened at mouse pointer position
- tooltips are not grabbing focus and making waybar go insane
- outputs are detected and new bars created for outputs
- menus and submenus are displayed on correct output

### What does not:
- sway < 1.2
  Limitation of wlr-layer-shell protocol: if the compositor has no support for layer-shell popups, wlr-layer-shell.get_popup will make xdg_popup silently disappear, as it does not provide any results or errors.
- keyboard focus on menu. If I understand discussions in wlr-protocols correctly, it's expected to work and the possible reason of the issue is https://github.com/wmww/gtk-layer-shell/issues/9
- closing popup by clicking outside of it: this works only if you click on waybar ¯\\(°_o)/¯﻿. Again, that is caused by missing `xdg_popup_grab` call and should be fixed elsewhere.
- specifying width and height in configuration: the bar will ignore these and resize to screen width/height
- popups are rendered on the same layer as the main window, i.e. underneath other windows if the bar layer is set to bottom. This is a sway bug and it's already fixed in master (sway release 1.3)

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <gdkmm/monitor.h>
 #include <glibmm/refptr.h>
 #include <gtkmm/box.h>
 #include <gtkmm/cssprovider.h>
@@ -15,10 +16,11 @@ namespace waybar {
 
 class Factory;
 struct waybar_output {
-  struct wl_output *     output = nullptr;
-  std::string            name;
-  uint32_t               wl_name;
-  struct zxdg_output_v1 *xdg_output = nullptr;
+  Glib::RefPtr<Gdk::Monitor> monitor;
+  std::string                name;
+
+  std::unique_ptr<struct zxdg_output_v1, decltype(&zxdg_output_v1_destroy)> xdg_output = {
+      nullptr, &zxdg_output_v1_destroy};
 };
 
 class Bar {
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -32,13 +32,12 @@ class Bar {
   auto toggle() -> void;
   void handleSignal(int);
 
-  struct waybar_output *        output;
-  Json::Value                   config;
-  Gtk::Window                   window;
-  struct wl_surface *           surface;
-  struct zwlr_layer_surface_v1 *layer_surface;
-  bool                          visible = true;
-  bool                          vertical = false;
+  struct waybar_output *output;
+  Json::Value           config;
+  Gtk::Window           window;
+  struct wl_surface *   surface;
+  bool                  visible = true;
+  bool                  vertical = false;
 
  private:
   static constexpr const char *MIN_HEIGHT_MSG =
@@ -53,7 +52,9 @@ class Bar {
                                           uint32_t, uint32_t);
   static void layerSurfaceHandleClosed(void *, struct zwlr_layer_surface_v1 *);
 
-  void destroyOutput();
+#ifdef HAVE_GTK_LAYER_SHELL
+  void initGtkLayerShell();
+#endif
   void onConfigure(GdkEventConfigure *ev);
   void onRealize();
   void onMap(GdkEventAny *ev);
@@ -70,6 +71,9 @@ class Bar {
     int bottom = 0;
     int left = 0;
   } margins_;
+  struct zwlr_layer_surface_v1 *layer_surface_;
+  // use gtk-layer-shell instead of handling layer surfaces directly
+  bool                                          use_gls_ = false;
   uint32_t                                      width_ = 0;
   uint32_t                                      height_ = 1;
   uint8_t                                       anchor_;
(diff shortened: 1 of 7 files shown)


Alexays:
How are the outputs destroyed now?

alebastr:
std::remove_if in the end of Client::handleMonitorRemoved erases struct waybar_output from outputs_.
Implicitly defined destructor of struct waybar_output then releases reference to Gdk::Monitor and destroys xdg_output. As wl_output is owned by Gdk::Monitor, nothing here requires writing explicit destruction code.

David96:
What would be the requirements for this to get merged? Is there any way to help?
I tried it out and so far it appears to be working (with the  limitations already mentioned).

Only one thing that did bother me, although I'm not quite sure what the "right" behavior is: The popups use the same layer is the bar. Therefore having the bar layer set to bottom will have the popups show underneath the windows. Setting the layer to top, makes bemenu, a wayland dmenu alternative, (its layer is set to top, too) show underneath waybar. Not sure, whether bemenu should use overlay as layer. https://drewdevault.com/2018/07/29/Wayland-shells.html says that panels should use bottom as a layer though so it might be desireable to use overlay as the popup layer independent of what layer waybar is rendered at? Not sure whether this is possible.

alebastr:
On hold due to me being busy with work and other things. Also because I'm not happy with layer-shell limitations and GTK integration issues.
I can do a cosmetic run over the last commits and set it as ready to merge by end of week.

Regarding the popup layer: I was confused by this behavior as well. You may want to repost the question to [swaywm/wlr-protocols](https://github.com/swaywm/wlr-protocols/issues), a proper place for discussions about layer-shell protocol.

David96:
Alright, thanks for your response :)

DanySpin97:
I am using gtk-layer-shell to test if it fixed #487. Unfortunately, I still get crashes when I disable outputs. Here it is the backtrace:

```
#0  0x00007ffff72a58a9 in send_delete_event () at /usr/x86_64-pc-linux-musl/lib/libgtk-3.so.0
#1  0x00007ffff6e56b48 in gdk_threads_dispatch () at /usr/x86_64-pc-linux-musl/lib/libgdk-3.so.0
#2  0x00007ffff6b0b747 in g_main_dispatch (context=0x5555556c1b40) at ../glib-2.60.0/glib/gmain.c:3189
#3  0x00007ffff6b0b747 in g_main_context_dispatch (context=context@entry=0x5555556c1b40) at ../glib-2.60.0/glib/gmain.c:3854
#4  0x00007ffff6b0bb28 in g_main_context_iterate (context=context@entry=0x5555556c1b40, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at ../glib-2.60.0/glib/gmain.c:3927
#5  0x00007ffff6b0bbac in g_main_context_iteration (context=context@entry=0x5555556c1b40, may_block=may_block@entry=1) at ../glib-2.60.0/glib/gmain.c:3988
#6  0x00007ffff6d2ea6d in g_application_run (application=0x5555556a91a0 [gtkmm__GtkApplication], argc=<optimized out>, argv=0x7fffffffe608) at ../glib-2.60.0/gio/gapplication.c:2516
#7  0x0000555555582ae2 in waybar::Client::main(int, char**) ()
#8  0x0000555555581bbc in main ()
```

It might not be related to Waybar though, I'll try to find the issue in the next days.

eternal-sorrow:
> On hold due to me being busy with work and other things.

So, are you still busy? Maybe someone else could continue the work on this? It's the most important improvement for waybar right now in my opinion. It also would fix a lot of bugs that are already known.

alebastr:
I apologize for the delays. This PR is seriously cursed and various things kept happening whenever I wanted to get back to the code and finish it.

Changed in last revision:
 - use released gtk-layer-shell 0.1.0 instead of git snapshot
 - remove ipc asan fix. This rabbit hole is too deep and deserves a separate PR with larger scope of changes. Like  completely replacing `sigc++` signals from ipc code with something that [actually safe to use across threads](https://developer.gnome.org/gtkmm-tutorial/stable/sec-the-constraints.html.en).
 - minor changes in the bar.cpp

Alexays:
Thanks a lot @alebastr and yeah we need to replace sigc++ that is garbage with threads...

PR #443, opened 2019-08-29
Allow downstream to disable manpages even if scdoc is found

jbeich:
Like sway e.g.,
```
$ pkg-config --modversion scdoc
1.9.6

$ meson -Dman-pages=disabled _build
[...]
Dependency scdoc skipped: feature man-pages disabled
[...]
```


Alexays:
Thanks!

PR #447, opened 2019-09-01
Fix some issues with exclusive zone and margins

alebastr:
Fix some issues with exclusive zones and margins:
- exclusive zone did not include margins after toggling bar
- exclusive zone included incorrect margin when bar is anchored at bottom or right edge
- margins were incorrectly applied over anchored axis (see comment in setSurfaceSize)

There are some remaining issues with specifying size for anchored axis that may require decoupling window size from layer surface size. For now let's just pretend that this works as expected :see_no_evil:


Alexays:
Thanks, I completely missed that ^^

PR #449, opened 2019-09-02
Fix some issues in man pages

MichaelAquilina:
Also enable `man-pages` option in travis so that docs are tested as part of the CI/CD process


MichaelAquilina:
@Alexays seems like you might have forgotten to upload the new docker image for opensuse that contains the scdoc dependency?

Alexays:
Oh just checked and the opensuse image have an issue, i'll check that.
Thanks!

PR #455, opened 2019-09-04
Add configurable time display for battery module

lolzballs:
Adds a `format-time` configuration for the battery module so that users
can configure how they want their remaining time to be displayed.

The default format remains the same as before, i.e. `{H} h {M} min`,
but users can choose something like `{H}:{M:02d}` to give an output
like `4:29` if wanted.


Alexays:
Thanks, can you update the doc and the wiki? 😄

Alexays:
Thanks!

PR #456, opened 2019-09-07
install a systemd user service unit

gdamjan:
add a systemd --user unit/service file, so that one can run waybar as a
--user systemd service.

this feature is automatically enabled if systemd is found, but can be disabled
with -Dsystemd=disabled

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -62,6 +62,13 @@ libnlgen = dependency('libnl-genl-3.0', required: get_option('libnl'))
 libpulse = dependency('libpulse', required: get_option('pulseaudio'))
 libudev = dependency('libudev', required: get_option('libudev'))
 libmpdclient = dependency('libmpdclient', required: get_option('mpd'))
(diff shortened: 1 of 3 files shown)


gdamjan:
I'm not sure about `WantedBy=sway-session.target` since waybar might be used for other wayland compositors too. People have suggested `wayland-session.target` on the internet, so that might be a good option.

(does Waybar run on other wayland compositors?)

Alexays:
Thanks!
Yeah 'wayland-session.target' is better since waybar run on all wlroots compatible compositor

gdamjan:
updated to `wayland-session.target`

Alexays:
Nice, thanks ❤️

gdamjan:
it just dawned on me that I didn't account for the possibility that waybar is not installed in `/usr/bin/waybar`. Not sure if that's expected, but if it is, I'll send a patch to make the service an `.in` file.

PR #458, opened 2019-09-09
service file: allow for waybar installed in different prefix

gdamjan:
ie. don't assume waybar is always gonna be installed in `/usr/bin/waybar`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -66,8 +66,13 @@ systemd = dependency('systemd', required: get_option('systemd'))
 
 if systemd.found()
   user_units_dir = systemd.get_pkgconfig_variable('systemduserunitdir')
-  install_data('./resources/waybar.service',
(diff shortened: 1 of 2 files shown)


Alexays:
👍

meson.build
```diff
@@ -66,8 +66,13 @@ systemd = dependency('systemd', required: get_option('systemd'))
 
 if systemd.found()
   user_units_dir = systemd.get_pkgconfig_variable('systemduserunitdir')
-  install_data('./resources/waybar.service',
-    install_dir: user_units_dir)
+
+  configure_file(
+    configuration: config,
```
Alexays: I reverted this PR, CI says `config` is unknown :?a

gdamjan: it seems I forgot to push that commit :man_facepalming:

PR #460, opened 2019-09-10
service file: allow for waybar installed in different prefix

gdamjan:
ie. don't assume waybar is always gonna be installed in /usr/bin/waybar


gdamjan:
here's the proper fix

gdamjan:
wait, don't merge it still, I'm seeing a warning.

gdamjan:
fixed now, my bad for not checking that there's no configuration_data and prefix in the original meson.build

PR #465, opened 2019-09-17
add padding for mpd module in default style

krystiancha:
In the default style.css many modules (clock, battery, cpu, ...) get horizontal padding, but mpd module does not. This commit adds mpd to the list of modules that get the padding.

Before:
image

After:
image


Alexays:
Thanks!

Issue #281, opened 2019-04-22
SIGABRT in `MPD::checkErrors()`


cole-h:
I'm back again, but no solid STR this time. I'll post a reply if I do find something.

---

After some time (this one actually takes a fairly long while compared to our last two issues), Waybar receives a SIGABRT from libmpdclient, due to a failed assertion: `waybar: ../src/ierror.h:130: mpd_error_get_message: Assertion 'error->message != NULL || error->code == MPD_ERROR_OOM' failed.`

<details>
<summary>Backtrace</summary>

```
#0  0x00007ffff5dddd7f in raise () at /usr/lib/libc.so.6
#1  0x00007ffff5dc8672 in abort () at /usr/lib/libc.so.6
#2  0x00007ffff5dc8548 in _nl_load_domain.cold.0 () at /usr/lib/libc.so.6
#3  0x00007ffff5dd6396 in  () at /usr/lib/libc.so.6
#4  0x00007ffff62c60d1 in  () at /usr/lib/libmpdclient.so.2
#5  0x00005555555f8ee4 in waybar::modules::MPD::checkErrors(mpd_connection*) (this=0x5555557ac000, conn=0x7fffd8000b60) at ../src/modules/mpd.cpp:284
#6  0x00005555555f9053 in waybar::modules::MPD::fetchState() (this=0x5555557ac000) at ../src/modules/mpd.cpp:293
#7  0x00005555555f61f6 in waybar::modules::MPD::update() (this=0x5555557ac000) at ../src/modules/mpd.cpp:49
#8  0x00005555555b08d1 in waybar::Bar::<lambda()>::operator()(void) const (__closure=0x55555585bc30) at ../src/bar.cpp:194
#9  0x00005555555b170e in sigc::adaptor_functor<waybar::Bar::getModules(const waybar::Factory&, const string&)::<lambda()> >::operator()(void) const (this=0x55555585bc30) at /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:256
#10 0x00005555555b15b5 in sigc::internal::slot_call0<waybar::Bar::getModules(const waybar::Factory&, const string&)::<lambda()>, void>::call_it(sigc::internal::slot_rep *) (rep=0x55555585bc00) at /usr/include/sigc++-2.0/sigc++/functors/slot.h:136
[...]

PR #467, opened 2019-09-19
Add more locking in mpd module

jspngh:
This PR aims to solve issue #281 by taking the `connection_lock_` in `waitForEvent()`.

Additionally, I observed that multiple `periodic_updater`s could be running.
This happens when mpd is playing and no events happen.
This causes the mpd connection to drop when `mpd_recv_idle` times out.
If the current `periodic_updater` is sleeping at this moment, `update` will run and start a new one,
while the current one will also continue running (there is a new connection to mpd at this moment, so it sees no reason to stop). Adding a mutex to make sure `update` only continues when the current `periodic_updater` has stopped, fixes this.

It would be nice if more people could try out this code, since I'm not sure #281 is solved by  this, maybe it just happens more infrequently.

### DIFF
--- a/include/modules/mpd.hpp
+++ b/include/modules/mpd.hpp
@@ -65,6 +65,9 @@ class MPD : public ALabel {
   unique_status status_;
   mpd_state     state_;
   unique_song   song_;
+
+  // To make sure the previous periodic_updater stops before creating a new one
+  std::mutex periodic_lock_;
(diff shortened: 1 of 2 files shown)


Alexays:
Thank you ❤️ You can tell here if someone had the bug before and doesn't have it anymore :)

PR #480, opened 2019-10-04
Detect timezone changes

### LINKED ISSUE #479: Timezone changes are not detected
When the timezone is changed via `timedatectl set-timezone XXX` the clock widget does not update accordingly. I tracked down the issue to fmtlib and filed a PR (https://github.com/fmtlib/fmt/pull/1341).

This is a tracking issue to remember to update https://github.com/Alexays/Waybar/blob/048d8d328ec41c6f1a18e2ad695a401a83a4d9b8/meson.build#L50 when a new fmtlib version is released with the patch.

Another option is to patch that directly by adding tzset() before fmt::localtime() invocations, but we might as well just wait for upstream changes to sink in.

layus:
Fixes #479, because upstream does not intend to.

It may be less expensive to do that only once in a while, or to inotify-watch on /etc/timezone, but this is good enough.

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -1,4 +1,5 @@
 #include "modules/clock.hpp"
+#include <time.h>
 
 waybar::modules::Clock::Clock(const std::string& id, const Json::Value& config)
     : ALabel(config, "clock", id, "{:%H:%M}", 60) {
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks! :)

Issue #484, opened 2019-10-16
Failed to run on-click-'s command line including curled braces


Ventto:
I cannot run a command line including curled braces in `on-click-*` options.

Here is an example of configuration file:

```json
{
    "layer": "top",
    "position": "bottom",
    "modules-center": [ "custom/test" ],
    "custom/test": {
        "format": "{}",
        "exec": "${HOME}/.config/waybar/modules/test/exec.sh",
        "on-click": "${HOME}/.config/waybar/modules/test/onclick.sh",
        "on-click-right": "echo command | awk '{print $1}'",
        "interval": 3
    }
}
```

Triggering `on-click` or `on-click-right`, for instance makes *waybar* fail and exist witht the following error message: 

```
(waybar:75012): glibmm-ERROR **: 17:01:49.909: 
unhandled exception (type std::exception) in signal handler:
what: argument not found
```

Located into [`AModule.cpp`](https://github.com/Alexays/Waybar/blob/12b30ca25f8575abf7c3f67c84e6a56ef7330661/src/AModule.cpp#L49), it appears that the `fmt::format()` function looks for the `{arg}` named argument in the given `format` string which does not exist. It will fail if the `format` string contains any `{...}` curled-brace substring.

Alexays:
Fixed with #485

PR #485, opened 2019-10-16
fix: No need to format string with unused 'arg' named argument (#484)

Ventto:
The `fmt::format()` function looks for the `{arg}` named argument in the given "format" string which does not exist. It will fail if the string contains any `{...}` curled-brace substring.

Consequently, any `on-click*` option's command line containing for instance substring like `${var}` or `awk '{...}'` will crash the program.


Alexays:
Oh yeah it was supposed to be used somewhere :/
Thanks :)

PR #495, opened 2019-11-03
Add tooltip formatting to custom module

bonflintstone:
This respects the tooltip-format option, and if tooltip is enabled and this option is a string formats the custom tooltip exactly like it formats the normal text, but using the tooltip_format.

Disclaimer:
I am not a cpp developer and mostly copied the code from the clock module, so it is very likely I did something wrong. I also do not quite understand the code that I replaced. But this seems to work for me quite well.

### DIFF
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -112,10 +112,16 @@ auto waybar::modules::Custom::update() -> void {
     } else {
       label_.set_markup(str);
       if (tooltipEnabled()) {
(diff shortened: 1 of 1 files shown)


bonflintstone:
Clearly this also needs some documentation and perhaps tests, if you like this feature and perhaps give me some pointers to what else needs to be done on this I would be grateful!

bonflintstone:
I noticed this is already possible with the json return type, sorry

PR #510, opened 2019-11-28
Toggle opacity

HardDie:
Now if we toggle bar mode by send SIGUSR1 when bar is still visible
[example](https://imgur.com/a/950sIzg)


Alexays:
Thanks! I hadn't thought about the opacity.

PR #525, opened 2019-12-28
add dependencies of gtk-layer-shell to readme

Akasch:
Add missing dependencies of the new gtk-layer-shell to the README.

I have tested the apt install only on Debian testing, not on Ubuntu 19.10


Alexays:
Thanks

PR #517, opened 2019-12-15
Add info on "on-click-middle" option to custom module man page

fuzxi:
Fixes issue #505

Issue #505, opened 2019-11-22
Document "on-middle-click" option


fuzxi:
Similar to `on-click` and `on-right-click`, I'd like an `on-middle-click` option. This is one of the few features I'm missing from `polybar`, as I used it with the mpd module: left click for previous track, middle click for play/pause, right click for next track.

cprussin:
ha, I was just about to put in this ticket myself earlier today.

While spelunking the code to see how to add it, I found that the option already exists: `on-click-middle`.  It's just not documented.  See https://github.com/Alexays/Waybar/blob/master/src/AModule.cpp#L11

Alexays:
Oh yeah, if someone wants to update wiki and doc to add the missing clicks?

fuzxi:
I made appropriate edits to the wiki and updated the custom module man page in PR #517

Alexays:
Done in #517 by @fuzxi

PR #529, opened 2019-12-29
[Corrected] Add info on "on-click-middle" option to custom module man page

fuzxi:
#505 fix attempt two. #517 had a space on an empty line which caused a build error, this fixes that.

Also changed the spacing in the mpd and temperature modules to use tabs instead of spaces, to be consistent with the other modules.


Alexays:
Thanks 🙏 
Can you resolve conflicts by rebasing on master?

PR #534, opened 2020-01-01
feat: show title if artist isn't available

AdrienLemaire:
Using mps-youtube, most tracks that I'm listening to do not return information on the artist.
Adding an else condition to return the title anyways.


Alexays:
Thanks 🎉

PR #536, opened 2020-01-03
Fix: WindowId recieves 'u'

xPMo:
I noticed a lot of errors of the sort:
```
(waybar:165190): "GLib-GIO-WARNING **: 16:39:02.710: Received property WindowId with type i does not match expected type u in the expected interface"
```

### DIFF
--- a/protocol/dbus-status-notifier-item.xml
+++ b/protocol/dbus-status-notifier-item.xml
@@ -31,7 +31,9 @@
     <property name='Id' type='s' access='read'/>
     <property name='Title' type='s' access='read'/>
     <property name='Status' type='s' access='read'/>
+    <!-- See discussion on pull #536
     <property name='WindowId' type='u' access='read'/>
+    -->
     <property name='IconThemePath' type='s' access='read'/>
     <property name='IconName' type='s' access='read'/>
     <property name='IconPixmap' type='a(iiay)' access='read'/>


Alexays:
That's not the correct fix, we should use unsigned for the windowId

xPMo:
Huh. Where are the messages coming from then?

alebastr:
@Alexays it is a correct fix for described issue though.
My favorite part of SNI specification is that there are several mostly compatible implementations. [Original KDE specification](https://github.com/KDE/plasma-workspace/blob/master/xembed-sni-proxy/org.kde.StatusNotifierItem.xml#L9) defines WindowId as 'i', [Freedesktop spec](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/) has 'u' and [libappindicator](https://bazaar.launchpad.net/~indicator-applet-developers/libappindicator/trunk.16.10/view/head:/src/notification-item.xml) ignores 'WindowId'. And also a couple of independent implementations I've seen on github that are not compatible with any of above.

I don't think we care about `WindowId`, so I'd simply comment the property definition in the xml file. This will disable the validation of property type and suppress the warning message.

Alexays:
@alebastr Thx for the explanation, I'll also prefer to comment the definition :)
Can you make the change @xPMo? :D

xPMo:
force-pushed.

This discussion is mentioned in the comment and commit message.

Alexays:
Thanks 🎉

PR #543, opened 2020-01-06
mpd: add paused format string

cole-h:
I've been using this locally for a while and finally decided to submit this upstream, now that the MPD module doesn't segfault anymore :wink:. Let me know if there are any changes I need to make.

### DIFF
--- a/include/modules/mpd.hpp
+++ b/include/modules/mpd.hpp
@@ -36,6 +36,7 @@ class MPD : public ALabel {
 
   bool stopped();
   bool playing();
+  bool paused();
 
   const std::string module_name_;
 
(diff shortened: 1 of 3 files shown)


Alexays:
Nice!
Can you also update the documentation and wiki 😄

cole-h:
What should I say for its `default`? In its current implementation, it uses the value of `format` if nothing is specified. Does this look right?
```
*format-paused*: ++
    typeof: string ++
    default: value of *format* ++
    Information displayed when a song is paused.
```

I've got docs updated locally and will push if you think this looks good. I'll also update the wiki after we get this merged :wink:

Alexays:
```
*format-paused*: ++
	typeof: string ++
	This format is used when a when a song is paused.
```
Something like this :)

cole-h:
Should be good to go. It breaks away from the standard "Information shown when ___", but I don't really think that's an issue.

Alexays:
Thanks :)
Can you update this wiki page?
https://github.com/Alexays/Waybar/wiki/Module:-MPD

PR #545, opened 2020-01-07
Move bar to background when toggled off

topisani:
This means the bar doesn't have a clickable area that covers other windows when hidden and layer is `top`. Only works with `gtk-layer-shell` so far.

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -368,7 +368,7 @@ auto waybar::Bar::toggle() -> void {
 #endif
   }
   setExclusiveZone(width_, height_);
-  //wl_surface_commit(surface);
+  wl_surface_commit(surface);
 }
 
 void waybar::Bar::getModules(const Factory& factory, const std::string& pos) {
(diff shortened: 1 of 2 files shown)


src/bar.cpp
```diff
@@ -351,13 +351,24 @@ auto waybar::Bar::toggle() -> void {
...
+      auto layer =
+          config["layer"] == "top" ? GTK_LAYER_SHELL_LAYER_TOP : GTK_LAYER_SHELL_LAYER_BOTTOM;
+      gtk_layer_set_layer(window.gobj(), layer);
+      window.show_all();
+    }
+#endif
   }
   setExclusiveZone(width_, height_);
-  wl_surface_commit(surface);
+  //wl_surface_commit(surface);
```
Alexays: Does this is still needed without gtk_layer_shell?

topisani: damn thats a good question, tbh i wrote this code about 6 months ago and just recently rebased on master and realized i should do a PR... so not really sure...

PR #546, opened 2020-01-07
fix(network): add missing include

z3ntu:
Fixes:
```
../src/modules/network.cpp:68:3: error: 'assert' was not declared in this scope
   68 |   assert(starts_with(read, category));
      |   ^~~~~~
../src/modules/network.cpp:6:1: note: 'assert' is defined in header '<cassert>'; did you forget to '#include <cassert>'?
    5 | #include "util/format.hpp"
  +++ |+#include <cassert>
    6 |
```


Alexays:
Thanks ❤️

PR #548, opened 2020-01-09
Keep aspect ratio when scaling tray icons.

Psykar:
Allows things such as https://github.com/udda/multiload-ng to work nicely :)


Alexays:
Thanks :)

PR #550, opened 2020-01-13
Use PA_PROP_DEVICE_FORM_FACTOR for device icon.

PlusMinus0:
It makes finding the icon a little more reliable. [PA_PROP_DEVICE_FORM_FACTOR](https://freedesktop.org/software/pulseaudio/doxygen/proplist_8h.html#a77c49f81eb8426259c13fc53619b9670)

Though this should probably get refactored to use [PA_PROP_DEVICE_ICON_NAME](https://freedesktop.org/software/pulseaudio/doxygen/proplist_8h.html#ad2eee8a3a9bbad4ed65eb7eb094a7518) in the future.

This is how Pulseaudio finds the icon: [sink.c](https://github.com/pulseaudio/pulseaudio/blob/f5d3606fe76302c7dbdb0f6a80400df829a5f846/src/pulsecore/sink.c#L3443)

### DIFF
--- a/include/modules/pulseaudio.hpp
+++ b/include/modules/pulseaudio.hpp
@@ -34,6 +34,7 @@ class Pulseaudio : public ALabel {
   pa_cvolume  pa_volume_;
   bool        muted_;
   std::string port_name_;
+  std::string form_factor_;
   std::string desc_;
   std::string monitor_;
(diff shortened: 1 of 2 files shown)


Alexays:
Nice thanks :)

PR #557, opened 2020-01-16
Download patch files from Github instead of wrapdb

thomaswucher:
Currently wrapdb.mesonbuild.com is offline and its not clear when it
will be up again. Github seems to be the more reliable source for these
files.


Alexays:
Nice,  thanks!

PR #560, opened 2020-01-21
Add timezone support to clock module (closes #223)

### LINKED ISSUE #223: timezone on clock module

mjec:
Please forgive my C++ newbiness!

Notably I had to do the subprojects thing to get around HowardHinnant/date#398, and I don't _love_ that I had to write my own formatter in clock.cpp, but I think this should work.

Feedback, and push back, very welcome. Let me know how to improve the code and I will!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -66,6 +66,8 @@ gtk_layer_shell = dependency('gtk-layer-shell-0',
         required: get_option('gtk-layer-shell'),
         fallback : ['gtk-layer-shell', 'gtk_layer_shell_dep'])
 systemd = dependency('systemd', required: get_option('systemd'))
+date_dep = dependency('date', default_options : [ 'use_system_tzdb=true' ], fallback: [ 'date', 'date_dep' ])
+tz_dep = dependency('date', default_options : [ 'use_system_tzdb=true' ], fallback: [ 'date', 'tz_dep' ])
 
 prefix = get_option('prefix')
 conf_data = configuration_data()
@@ -166,7 +168,9 @@ executable(
         libpulse,
         libudev,
         libmpdclient,
-        gtk_layer_shell
+        gtk_layer_shell,
+        date_dep,
+        tz_dep
     ],
     include_directories: [include_directories('include')],
     install: true,
--- a/meson.build
+++ b/meson.build
@@ -66,7 +66,6 @@ gtk_layer_shell = dependency('gtk-layer-shell-0',
         required: get_option('gtk-layer-shell'),
         fallback : ['gtk-layer-shell', 'gtk_layer_shell_dep'])
 systemd = dependency('systemd', required: get_option('systemd'))
-date_dep = dependency('date', default_options : [ 'use_system_tzdb=true' ], fallback: [ 'date', 'date_dep' ])
 tz_dep = dependency('date', default_options : [ 'use_system_tzdb=true' ], fallback: [ 'date', 'tz_dep' ])
 
 prefix = get_option('prefix')
@@ -169,7 +168,6 @@ executable(
         libudev,
         libmpdclient,
         gtk_layer_shell,
-        date_dep,
         tz_dep
     ],
     include_directories: [include_directories('include')],
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -1,5 +1,5 @@
 #include "modules/clock.hpp"
-#include <time.h>
+#include <date/tz.h>
 
 waybar::modules::Clock::Clock(const std::string& id, const Json::Value& config)
     : ALabel(config, "clock", id, "{:%H:%M}", 60) {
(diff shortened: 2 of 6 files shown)


meson.build
```diff
@@ -66,6 +66,8 @@ gtk_layer_shell = dependency('gtk-layer-shell-0',
         required: get_option('gtk-layer-shell'),
         fallback : ['gtk-layer-shell', 'gtk_layer_shell_dep'])
 systemd = dependency('systemd', required: get_option('systemd'))
+date_dep = dependency('date', default_options : [ 'use_system_tzdb=true' ], fallback: [ 'date', 'date_dep' ])
+tz_dep = dependency('date', default_options : [ 'use_system_tzdb=true' ], fallback: [ 'date', 'tz_dep' ])
```
Alexays: Remove this one :)

meson.build
```diff
@@ -166,7 +168,9 @@ executable(
         libpulse,
         libudev,
         libmpdclient,
-        gtk_layer_shell
+        gtk_layer_shell,
+        date_dep,
+        tz_dep
```
Alexays: Same

layus:
Here are the issues I discovered:

1. date is not trivial to package, at least with nix. It has incomplete cmake support (work is ongoing apparently) and no meson support. mesonbuild wrapper cannot be used as it downloads stuff on the fly. But that is a concern for package maintainers.

2. My TZ dir is /etc/zoneinfo as specified by my $TZ_DIR var. apparetly `date` does not look at TZ_DIR. this sould be upstreamed, but in the meantime this can be a breaking change.

3. The current setup does not follow LC_TIME (fr_BE.UTF-8 here). I do not know if it follows any locale, or just default to a C/en_US locale. This is a breaking change.

4. With several clocks, it may be useful to use a different locale per clock. But this can be left out for later.

layus:
From what I see [here](https://howardhinnant.github.io/date/date.html#format) there is a version of `date::format` that takes the locale as its first argument. We should use that with [`std::locale("")`](https://en.cppreference.com/w/cpp/locale/locale) or find a way to default to the right locale.

meson.build
mjec: It looks like `date_dep` isn't required (I'll remove it) but `tz_dep` is (otherwise we fail to link with missing `date::locate_zone`).

mjec:
> With several clocks, it may be useful to use a different locale per clock. But this can be left out for later.

I like this idea, but after noodling with it for a bit my C++ just isn't up to the task :(.

My thinking was having a new struct containing a localtime and locale that gets passed to the `fmt::format`, and then the locale can be instead of `std::locale` in `formatter::format`. However I wasn't able to figure out a way to deal with `date::zoned_time` being a template type (I think that's the phrase for it?).

layus:
@Alexays I think we have reached the full set of desirable functionalities. Could you have a second look to check C++ style ?

src/modules/clock.cpp
```diff
@@ -11,46 +24,39 @@ waybar::modules::Clock::Clock(const std::string& id, const Json::Value& config)
...
+
 auto waybar::modules::Clock::update() -> void {
-  tzset(); // Update timezone information
-  const date::time_zone* zone;
-  auto now = std::chrono::floor<std::chrono::seconds>(std::chrono::system_clock::now());
-  if (config_["timezone"].isString()) {
-    zone = date::locate_zone(config_["timezone"].asString());
-  } else {
-    zone = date::current_zone();
+  if (!fixed_time_zone_) {
```
Alexays: Why not just use `!config_["timezone"].isString()` and removing `fixed_time_zone_` var?

Alexays:
LGTM Thanks guys ! ❤️

subprojects/date.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+source_url=https://github.com/HowardHinnant/date/archive/v2.4.1.tar.gz
+source_filename=date-2.4.1.tar.gz
+source_hash=98907d243397483bd7ad889bf6c66746db0d7d2a39cc9aacc041834c40b65b98
+directory=date-2.4.1
+
+patch_url = https://wrapdb.mesonbuild.com/v1/projects/hinnant-date/2.4.1/1/get_zip
```
Alexays: Could you also use github as a source?
wrapdb tend to be down sometimes..

src/modules/clock.cpp
layus: To avoid going trough the json parser each time. May not be worth it.

Alexays: The json is already parsed, but yeah, never mind.

mjec:
@Alexays Have updated the patch URL to be github as well, and rewrote history a little bit to use my personal (rather than work) email on the commits.

Let me know if you'd like me to rewrite the history/squash commits at all.

Alexays:
Thanks guys, great work 👍

nerddan:
Will wiki page update?

Alexays:
Man page is updated, but the wiki still need to be updated, @mjec can you give a shot?

mjec:
Just updated the wiki @Alexays

nerddan:
Wait, can I have multiple clock modules with different tz?

mjec:
Yes @nerddan, with a config like:

```json
{
    "modules-right": ["clock#nyc", "clock#sydney"],
    "clock#nyc": {
        "timezone": "America/New_York"
    },
    "clock#sydney": {
        "timezone": "Australia/Sydney"
    }
}
```

nerddan:
When will this be packed into the next release?

Alexays:
@nerddan By the end of the week

Issue #494, opened 2019-11-01
Network module does not detect disconnected state


tobiasjakobi:
Hi,

using git master here. I noticed that the network module incorrectly detects my Ethernet device as 'linked', even though the interface is down and the cable is disconnected.

This is the output of ip:
2: enp4s0f1: <BROADCAST,MULTICAST> mtu 1500 qdisc pfifo_fast state DOWN group default qlen 1000
    link/ether 80:fa:5b:24:a6:69 brd ff:ff:ff:ff:ff:ff

I haven't checked how the network module sources its information, but sysfs provides an operstate and carrier (which can only be read when the interface is UP). I have used this information in my custom Bash script to get the correct information about the if.

--- Tobias

tobiasjakobi:
I had another look at this, and I think we already have the required information available. In handleEvents() we already copy some fields of the ifi object. My suggestion is to also copy ifi_flags and evaluate it in getNetworkState(). ifi_flags seems to provide exactly the information that ip presents in angle brackets (for my example above the BROADCAST and MULTICAST flags are set). For connections of ethernet type UP and LOWER_UP should be present.

Also netlink seems to provide more fine grained carrier state, see section 3.5.9 here:
https://www.infradead.org/~tgr/libnl/doc/route.html

marcplustwo:
This might be similar to what I'm experiencing. In general it seems that the network module does not update properly. When I disable my wifi chip via rfkill, Waybar does not register that change. Only upon clicking on the module twice (once for alt-view and again back to normal), the icon is updated.

I verified that in `network.cpp` update()` is called regularly (interval is set to 10).

This is in version `0.9.0-2` from the arch repos.

Geo25rey:
@marcplustwo Was #568 supposed to fix this issue?

tobiasjakobi:
> @marcplustwo Was #568 supposed to fix this issue?

No, the PR has nothing to do with wired network connections.

niktob560:
I have a little(but not a lot) similar issue
In case of disabling Wi-Fi via hardware button (status accessible via rfkill) network module shows that is already connected to the last Wi-Fi network
``netctl-auto list`` shows that it is not connected to any network in this case

PR #568, opened 2020-01-29
Add rfkill status to network module and new bluetooth module

marcplustwo:
I added an interface to the [linux rfkill utility](https://wireless.wiki.kernel.org/en/users/Documentation/rfkill). This allows modules to update based on changes of the rfkill state. I added this to the existing network module and created a new bluetooth module implementing the feature.

### rfkill
The code is adapted from [here](https://git.kernel.org/pub/scm/linux/kernel/git/jberg/rfkill.git/).

### bluetooth
I was never sure if my bluetooth module was deactivated or not connecting to my devices for another reason.
It's possible to update the module via an interval in addition to rfkill events.

### network module
Two main changes were made:
1. Distinction between no wifi connection present and wifi module turned off. (think: airplane mode)
1. Update status on rfkill event in addition to interval based updates.

The second actually gave me some trouble as there seems to be an unrelated bug. The module doesn't seem to reflect a change in status in the bar, even though the corresponding `update()` function was executed both via the interval as well as based on the rfkill event.
I reproduced this on v0.9 from the arch community repos.  Also see my comment in #494.

I am open to any feedback, especially on style. I don't have much experience with c++.

### DIFF
--- a/include/util/rfkill.hpp
+++ b/include/util/rfkill.hpp
@@ -9,12 +9,11 @@ class Rfkill {
   Rfkill(enum rfkill_type rfkill_type);
   ~Rfkill() = default;
   void waitForEvent();
-  int getState() const;
+  bool getState() const;
 
  private:
   enum rfkill_type rfkill_type_;
   int state_ = 0;
-  int prev_state_ = 0;
 };
 
 }  // namespace waybar::util
--- a/include/util/rfkill.hpp
+++ b/include/util/rfkill.hpp
@@ -5,12 +5,11 @@
 namespace waybar::util {
 
 class Rfkill {
- public:
+ public:;
   Rfkill(enum rfkill_type rfkill_type);
   ~Rfkill() = default;
-  bool isDisabled() const;
   void waitForEvent();
-  int getState();
+  int getState() const;
 
  private:
   enum rfkill_type rfkill_type_;
--- a/include/util/rfkill.hpp
+++ b/include/util/rfkill.hpp
@@ -2,8 +2,20 @@
 
 #include <linux/rfkill.h>
 
-namespace waybar::util::rfkill {
+namespace waybar::util {
 
-bool isDisabled(enum rfkill_type rfkill_type);
+class Rfkill {
+ public:
+  Rfkill(enum rfkill_type rfkill_type);
+  ~Rfkill() = default;
+  bool isDisabled() const;
+  void waitForEvent();
+  int getState();
+
+ private:
+  enum rfkill_type rfkill_type_;
+  int state_ = 0;
+  int prev_state_ = 0;
+};
 
 }  // namespace waybar::util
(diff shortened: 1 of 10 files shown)


man/waybar-bluetooth.5.scd
```diff
@@ -0,0 +1,94 @@
...
+The *bluetooth* module displays information about the status of the device's bluetooth device.
+
+# CONFIGURATION
+
+Addressed by *bluetooth*
+
+*interval*: ++
+	typeof: integer ++
+	default: 60 ++
+	The interval in which the network information gets polled (e.g. signal strength).
```
Alexays: To update :)

src/modules/network.cpp
```diff
@@ -3,8 +3,9 @@
 #include <sys/eventfd.h>
 #include <fstream>
 #include <cassert>
+#include <linux/rfkill.h>
```
Alexays: You can delete this include as it's included in the rfkill util

include/util/rfkill.hpp
```diff
@@ -0,0 +1,20 @@
...
+ public:;
+  Rfkill(enum rfkill_type rfkill_type);
+  ~Rfkill() = default;
+  void waitForEvent();
+  int getState() const;
+
+ private:
+  enum rfkill_type rfkill_type_;
+  int state_ = 0;
+  int prev_state_ = 0;
```
Alexays: Unused

src/util/rfkill.cpp
```diff
@@ -0,0 +1,83 @@
...
+      break;
+    }
+  }
+
+  close(fd);
+  return;
+}
+
+
+int waybar::util::Rfkill::getState() const {
```
Alexays: Why not a bool for the type?

marcplustwo:
Thanks a lot for the feedback! I changed everything you requested.

> For the update that doesn't seem to work, this may be due to the fact that the getInfo stuff needs a delay to be updated and get new information, can you try with a large delay of 1s and reduce it?

As for this, I have tried lots of different combinations of delays at different points, but couldn't figure it out. Maybe you or someone could take another look at that. I don't think my change interfere here - I tested v 0.9 form the repos again and I had the same issue.

marcplustwo:
> Thanks a lot for the feedback! I changed everything you requested.
> 
> > For the update that doesn't seem to work, this may be due to the fact that the getInfo stuff needs a delay to be updated and get new information, can you try with a large delay of 1s and reduce it?
> 
> As for this, I have tried lots of different combinations of delays at different points, but couldn't figure it out. Maybe you or someone could take another look at that. I don't think my changes interfere here - I tested version 0.9 form the repos again and I had the same issue.

The issues disappeared. Now everything is working as expected!

Alexays:
Thanks @marcplustwo, will merge this by the end of the week ❤️

Alexays:
Thanks a lot, can you update the github wiki?

PR #570, opened 2020-01-31
Current month calendar in clock tooltip.

skligys:
A simple single-month calendar in clock tooltip.  Layout follows Unix cal style.  Supports localized weekday names.

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -1,5 +1,4 @@
 #include "modules/clock.hpp"
-#include <mutex>
 #include <sstream>
 
 using zoned_time = date::zoned_time<std::chrono::system_clock::duration>;
@@ -37,14 +36,11 @@ struct CachedCalendar {
   }
 };
 
-std::mutex cached_calendar_mutex;  // protects cached_calendar.
 CachedCalendar cached_calendar;
 
 std::string calendar_text(const waybar_time& wtime, const date::weekday& first_dow) {
   const auto daypoint = date::floor<date::days>(wtime.ztime.get_local_time());
   const auto ymd = date::year_month_day(daypoint);
-
-  const std::lock_guard<std::mutex> lock(cached_calendar_mutex);
   if (cached_calendar.ymd == ymd) {
     return cached_calendar.text;
   }
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -1,101 +1,12 @@
 #include "modules/clock.hpp"
 #include <sstream>
+#include <type_traits>
 #ifdef HAVE_LANGINFO_1STDAY
 #include <langinfo.h>
+#include <locale.h>
 #endif
 
-using zoned_time = date::zoned_time<std::chrono::system_clock::duration>;
-
-struct waybar_time {
-  std::locale locale;
-  zoned_time ztime;
-};
-
-namespace {
-
-#ifdef HAVE_LANGINFO_1STDAY
-// Computations done similarly to Linux cal utility.
-date::weekday first_day_of_week() {
-  const int i = (std::intptr_t) nl_langinfo(_NL_TIME_WEEK_1STDAY);
-  auto ymd = date::year(i / 10000)/(i / 100 % 100)/(i % 100);
-  auto wd = date::weekday(ymd);
-  uint8_t j = *nl_langinfo(_NL_TIME_FIRST_WEEKDAY);
-  return wd + date::days(j - 1);
-}
-#endif
-
-void weekdays_header(const std::locale& locale, const date::weekday& first_dow, std::ostream& os) {
-  auto wd = first_dow;
-  do {
-    if (wd != first_dow) os << ' ';
-    Glib::ustring wd_ustring(date::format(locale, "%a", wd));
-    auto wd_len = wd_ustring.length();
-    if (wd_len > 2) {
-      wd_ustring = wd_ustring.substr(0, 2);
-      wd_len = 2;
-    }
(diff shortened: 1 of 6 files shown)


alebastr:
Have you considered [`Glib::ustring`](https://developer.gnome.org/glibmm/2.58/classGlib_1_1ustring.html#details) for UTF8 string operations? It's already a dependency of waybar.

As a laptop user I'm also slightly concerned about rebuilding calendar on each `Clock::update` call. It would be nice to cache it and rebuild only when necessary.

skligys:
Both good points and should be easy to do.

skligys:
Re: monospace font, yes, it needs it for alignment, will add a CSS class and a style for it.

skligys:
Re: "CSS class and predefined rule", Pango markup doesn't seem to do <div> tags or CSS classes on <span> tags: https://developer.gnome.org/pango/stable/pango-Markup.html

Any suggestions how to do CSS class for the calendar?

alebastr:
> Re: "CSS class and predefined rule", Pango markup doesn't seem to do `<div>` tags or CSS classes on `<span>` tags: https://developer.gnome.org/pango/stable/pango-Markup.html
> Any suggestions how to do the CSS class for the calendar?

Sorry, apparently my knowledge of GTK is lacking. Somehow I expected more from pango markup.
I see that you already wrapped calendar with `<tt>` in a config template. That should be sufficient, please ignore my point about CSS.

Seems like there's no good source for first day of week (well, there's a POSIX `nl_langinfo_l` but it would be a pain to use it with std::locale and I'm not so sure about BSD compatibility). It might be worth adding a config option.

Also I believe that `AModule::update` is always running in the main thread (invoked by `Glib::Dispatcher  dp`) and does not require mutex.
Otherwise looks good, thanks!

Alexays:
Thanks @skligys,
Great stuff, can you move most of stuct sutff to the header file?

skligys:
Updates:
- removed the mutex
- if `nl_langinfo` with `_NL_TIME_WEEK_1STDAY`, `_NL_TIME_FIRST_WEEKDAY` is available, use it for getting the first day of week

Not sure what you mean by "can you move most of stuct sutff to the header file" though.  Get rid of anonymous namespace and make all those new functions into private methods?

Alexays:
@skligys Both if you can :D

src/modules/clock.cpp
```diff
@@ -1,4 +1,101 @@
...
+  date::year_month_day ymd;
+  std::string text;
+
+  void set(const date::year_month_day& ymd_, std::string text_) {
+    ymd = ymd_;
+    text = text_;
+  }
+};
+
+CachedCalendar cached_calendar;
```
alebastr: Didn't notice that on the first pass. There could be multiple `waybar::Clock` instances with different locale/timezone values, thus it's better to cache rendered calendar per `Clock` instance as a class field.

src/modules/clock.cpp
```diff
@@ -1,4 +1,101 @@
...
+struct waybar_time {
+  std::locale locale;
+  zoned_time ztime;
+};
+
+namespace {
+
+#ifdef HAVE_LANGINFO_1STDAY
+// Computations done similarly to Linux cal utility.
+date::weekday first_day_of_week() {
```
alebastr: Something like this would allow correctly rendering calendar with non-default locale. 
```cpp
date::weekday first_day_of_week(const std::locale& locale) {
#ifdef HAVE_LANGINFO_1STDAY
  using locale_ptr_t = std::unique_ptr<std::remove_pointer<locale_t>::type, decltype(&freelocale)>;
  locale_ptr_t loc{newlocale(LC_ALL, locale.name().c_str(), nullptr), &freelocale};
  if (loc) {
    // Computations done similarly to Linux cal utility.
    const int i = (std::intptr_t)nl_langinfo_l(_NL_TIME_WEEK_1STDAY, loc.get());
    auto      ymd = date::year(i / 10000) / (i / 100 % 100) / (i % 100);
    auto      wd = date::weekday(ymd);
    uint8_t   j = *nl_langinfo_l(_NL_TIME_FIRST_WEEKDAY, loc.get());
    return wd + date::days(j - 1);
  }
#endif
  return date::Sunday;
}
```

src/modules/clock.cpp
skligys: Done.

src/modules/clock.cpp
skligys: Thanks, missed this, done.

skligys:
> @skligys Both if you can :D

PTAL:
- made anonymous namespace functions into provate methods
- calendar now is cached per clock instance
- first day of week properly handles locales in clock configuration

src/modules/clock.cpp
skligys: Hm, will this work if a user doesn't set a locale in Waybar's clock configuration but sets an environment variable LANG=... instead?  From testing it looks that it works fine but I don't understand how.

alebastr: https://github.com/Alexays/Waybar/blob/83b12fc8a7053d455f37326826120a6f64cf5da6/src/modules/clock.cpp#L12-L16
At least in libstdc++ and libc++ implementations std::locale is built on top of locale_t. And for [newlocale(3)](http://man7.org/linux/man-pages/man3/newlocale.3.html) "" means taking value from environment. Resulting std::locale object will have correct locale name.

skligys:
Thanks for review!

Alexays:
LGTM Good work!
Thanks

Issue #492, opened 2019-10-27
Option to specify mpd password


maxigaz:
My mpd server is protected with a password. While I can access it from [Polybar](https://gitlab.com/polybar/polybar/wikis/Module:-mpd) by giving the `password` the correct value in its configuration file, I can’t find such an option in the man page of Waybar v0.8.0 or in the [master branch](https://github.com/Alexays/Waybar/blob/master/man/waybar-mpd.5.scd).

Alexays:
@minijackson Can you give a shot? :D

minijackson:
I can try, but since I don't have much off time, I'd prefer to mentor someone, if someone wants to give this issue a shot ^^

Alexays:
Done in #888 888

PR #576, opened 2020-02-04
[WIP do not merge] MPD: try to support password

torstehu:
This is a crude and untested implementation, but am I on the right track?

Tries to deal with #492


Alexays:
I don't know so much mpd, but seems good, anyone can test?

torstehu:
Fine by me!

Issue #565, opened 2020-01-28
clock module shows thousand seperator in year


Maclanky:
I recently updated to the latest commit and now the clock module shows the year as 2,020 using the default tooltip-format.

My config for the clock module:

```
"clock": {
        // "timezone": "America/New_York",
        "tooltip-format": "{:%Y-%m-%d | %H:%M}",
        "format": "{:%Y-%m-%d | %H:%M}",
        "format-alt": "{:%Y-%m-%d}"
    },
```

jomority:
I have the same problem. It was introduced in #560.

As a workaround, you can add `"locale": "C"` to your clock config.

However, this does not fix the displaying of fractional seconds. For example with `"format": "{:%Y-%m-%d %H:%M:%S}"` I get `2020-01-28 19:09:12.000123432`.

honza:
> As a workaround, you can add "locale": "C" to your clock config.

Thank you!

Xartos:
> However, this does not fix the displaying of fractional seconds. For example with `"format": "{:%Y-%m-%d %H:%M:%S}"` I get `2020-01-28 19:09:12.000123432`.

A workaround for this is to change `%S` to `%OS` in the format string

skligys:
I think it's this bug in date library:

https://github.com/HowardHinnant/date/issues/525
https://github.com/HowardHinnant/date/commit/a1843097862aacb752a81a7e98231e9d9464a645

I patched in that change in my build, and it fixed it.

Alexays:
Fixed with #715 unless timezone is specified.
While waiting for a [HowardHinnant/date](https://github.com/HowardHinnant/date) update.

PR #579, opened 2020-02-05
fix(clock): lower precision of zoned_time to avoid fractional seconds in output

alebastr:
`date` includes fractional seconds into `%S` format string output (see https://github.com/HowardHinnant/date/issues/543 for details).
This change should address the issue with seconds formatting reported in https://github.com/Alexays/Waybar/issues/565#issuecomment-579382135


Alexays:
Can you rebase? :D

alebastr:
> Can you rebase? :D
Done.

PR #583, opened 2020-02-11
fix(meson): Support libc++ >=9.0.0

DanySpin97:
From LLVM libc++ documentation:
"Prior to LLVM 9.0, libc++ provides the implementation of the
filesystem library in a separate static library."

Now the filesystem library (not the experimental one) is shipped
inside the libc++.so library.

Check if '-lc++fs' link flag is needed and supported before adding
it.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -9,6 +9,8 @@ project(
     ],
 )
 
+compiler = meson.get_compiler('cpp')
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

Issue #588, opened 2020-02-15
Network label text not updated properly when formats contain Unicode characters


JordanL2:
EDIT: Original title "Network format-disconnected doesn't allow Unicode characters"

This is reproducible in both the latest stable version and the latest git master version.

If I have this in my config (if the chars don't show, I'm using Font Awesome chars):

```
    "network": {
        "tooltip": true,
        "tooltip-format-ethernet": "{ifname} ({ipaddr})",
        "tooltip-format-wifi": "{essid} ({ipaddr}) {signalStrength}",
        "tooltip-format-disconnected": "DISCONNECTED",
        "format-ethernet": "  ",
        "format-wifi": "  ",
        "format-disconnected": "  ",
        "on-click": "nm-connection-editor"
    },
```

when the ethernet is disconnected (and there is no wifi), the ethernet connected icon shows. However, the field is red as determined by its disconnected status and my style.css, so Waybar does know it's disconnected, it's just showing the wrong icon.

If I change the format-disconnected to eg:

```
        "format-disconnected": " -- ",
```

then it correctly displays "--" when it is disconnected.

JordanL2:
Ok this is weird. If I change the disconnected format to:
"format-disconnected": " H ",

then it shows it correctly when disconnected.

JordanL2:
Yep, it seems it's literally only a problem when the icon is just surrounded with spaces. Which just happens to be exactly what I want.

JordanL2:
If it just has one space to the left of it, or one to the right, it works fine. If there is one space on either side (and nothing else, "  ") then that's the only situation in which it doesn't work.

JordanL2:
If I change the ethernet and disconnected formats to:

```
"format-ethernet": "  l",
"format-disconnected": "  ",
```

then they work fine. Is there a map somewhere, which uses the format as a key, and it's getting them mixed up (possibly because the unicode chars gets turned into the same char)?

JordanL2:
I think I *might* have found the problem.
https://github.com/Alexays/Waybar/blob/master/src/modules/network.cpp
line 282:

```
if (text != label_.get_label()) {
```

I bet that text comparison doesn't cope with the unicode characters. So it doesn't change the label text when the ethernet is disconnected, because it thinks the new text is identical.

JordanL2:
Created pull request https://github.com/Alexays/Waybar/pull/589

PR #589, opened 2020-02-15
Fix for 'Network label text not updated properly when formats contain…

JordanL2:
… Unicode characters' (#588)

A fix for this issue: https://github.com/Alexays/Waybar/issues/588


Alexays:
Thanks for the fix!

PR #596, opened 2020-02-18
Add a way to exclude specific output

vesim987:
This adds ability to exclude specific output for a bar with `!` character, for example:
```
[
  {
    "layer": "bottom",
    "output": "eDP-1",
    # config for main display
  },
  {
    "layer": "bottom",
    "output": "!eDP-1",
    # config for extra displays
  }
]
```

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -64,28 +64,22 @@ void waybar::Client::handleOutput(struct waybar_output &output) {
 }
 
 bool waybar::Client::isValidOutput(const Json::Value &config, struct waybar_output &output) {
+  if (config["output"].isArray()) {
+    for (auto const &output_conf : config["output"]) {
+      if (output_conf.isString() && output_conf.asString() == output.name) {
+        return true;
+      }
+    }
+  }
 
-  auto check_output = [](std::string_view config_output_name, struct waybar_output &output) {
+  if (config["output"].isString()) {
+    auto config_output_name = config["output"].asString();
     if (!config_output_name.empty()) {
       if (config_output_name.substr(0, 1) == "!") {
           return config_output_name.substr(1) != output.name;
       }
       return config_output_name == output.name;
     }
-    return false;
-  };
-
-  if (config["output"].isArray()) {
-    for (auto const &output_conf : config["output"]) {
-      if (output_conf.isString() && check_output(output_conf.asString(), output)) {
-        if(check_output(output_conf.asString(), output)) {
-          return true;
-        }
-      }
-    }
-  }
-  if (config["output"].isString()) {
-    return check_output(config["output"].asString(), output);
   }
 
   return false;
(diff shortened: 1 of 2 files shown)


src/client.cpp
```diff
@@ -65,21 +65,25 @@ void waybar::Client::handleOutput(std::unique_ptr<struct waybar_output> &output)
 
 bool waybar::Client::isValidOutput(const Json::Value &                    config,
                                    std::unique_ptr<struct waybar_output> &output) {
-  bool found = true;
+  auto check_output = [](std::string_view                       config_output_name,
+                         std::unique_ptr<struct waybar_output> &output) {
+    if (config_output_name.length() > 0) {
```
Alexays:
```suggestion
    if (!config_output_name.empty()) {
```

Alexays: Can you avoid using another function and don't return when the result is false as it'll skip others screen.

Alexays:
Otherise, LGTM, pretty simple with a !
Could you update the man and the wiki? 😃

src/client.cpp
vesim987: WDYM by another function? I've used lambda to reduce duplication of that code.

vesim987:
Rebased against latest master, and also added info to man and [wiki](https://github.com/Alexays/Waybar/wiki/Configuration#configuration-of-multiple-outputs).

src/client.cpp
```diff
@@ -64,21 +64,31 @@ void waybar::Client::handleOutput(struct waybar_output &output) {
...
+    return false;
+  };
+
   if (config["output"].isArray()) {
-    bool in_array = false;
     for (auto const &output_conf : config["output"]) {
-      if (output_conf.isString() && output_conf.asString() == output.name) {
-        in_array = true;
-        break;
+      if (output_conf.isString()) {
```
Alexays:
```suggestion
      if (output_conf.isString() && check_output(output_conf.asString(), output)) {
```

Alexays: Mb we should not do this check here, but only when only one screen name is passed, otherwise when there is
["!edp1", "edp2"] all screens will enter the config

Alexays:
> Mb we should not do this check here, but only when only one screen name is passed, otherwise when there is
> ["!edp1", "edp2"] all screens will enter the config

Did you see this comment :?

vesim987:
Ouch, somehow I've missed that comment, sorry.
Yep, that's good point.

src/client.cpp
```diff
@@ -64,21 +64,25 @@ void waybar::Client::handleOutput(struct waybar_output &output) {
...
+    if (!config_output_name.empty()) {
+      if (config_output_name.substr(0, 1) == "!") {
+          return config_output_name.substr(1) != output.name;
+      }
+      return config_output_name == output.name;
+    }
   }
-  return found;
+
+  return false;
```
Alexays: Should be true by default, when no output is defined in the config file.

PR #598, opened 2020-02-19
Use the same StatusNotifierWatcher for all trays

### LINKED ISSUE #359: tray lost after dpms toggle
Config: https://github.com/swalladge/dotfiles/tree/master/host-archdesk/config/waybar

To reproduce: `swaymsg "output * dpms off"; sleep 30; swaymsg "output * dpms on"`, then observe that tray is not shown in waybar. No empty gap where tray was shown, just nothing as if the tray module was disabled.

Applications displaying tray icons are not affected, and the icons can be seen back in the tray if waybar is killed and restarted.

This is waybar running on git master at the time of writing.

layus:
Fixes #359

This uses a singleton so we only ever use one StatusNotifierWatcher for
all of our bars (and all of their trays).

Before we had one per tray module. The first one to get the DBus name
would prevent the others from getting it. When that tray disappears (due
to dpms toggles for example), the other watchers do not react to it
correctly. Plus, I am not even sure that it is correct for the same
process to try to acquire the same name multiple times.

This does not solve the fact that we should be able to lose the
name and get it back on the bus, but at least we are not deceiving
ourselves :).

### DIFF
--- a/src/modules/sni/watcher.cpp
+++ b/src/modules/sni/watcher.cpp
@@ -3,14 +3,13 @@
 
 using namespace waybar::modules::SNI;
 
-Watcher::Watcher(std::size_t id)
+Watcher::Watcher()
     : bus_name_id_(Gio::DBus::own_name(Gio::DBus::BusType::BUS_TYPE_SESSION,
                                        "org.kde.StatusNotifierWatcher",
                                        sigc::mem_fun(*this, &Watcher::busAcquired),
                                        Gio::DBus::SlotNameAcquired(), Gio::DBus::SlotNameLost(),
                                        Gio::DBus::BUS_NAME_OWNER_FLAGS_ALLOW_REPLACEMENT |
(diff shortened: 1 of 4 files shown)


src/modules/sni/watcher.cpp
```diff
@@ -23,6 +22,7 @@ Watcher::~Watcher() {
     g_slist_free_full(items_, gfWatchFree);
     items_ = nullptr;
   }
+  Gio::DBus::unown_name(bus_name_id_);
```
Alexays: It's not already the case when it's destroyed?

layus: Where is it destroyed ? I worked on the assumption that (bus_name_id_) is only an int, and therefore RAII does not apply.

Alexays: Ok you're right, I assumed that Gio::DBus::own_name will do it, but no.

PR #612, opened 2020-03-03
systemd service: fix start up ordering

gdamjan:
the service needs to have After=wayland-session.target otherwise it'll
be started in parallel to the compositor which might not be fully
configured.

ps. sorry I overlooked this setting when I was sending the previous pull request.


Alexays:
Thanks!

freed00m:
Where is the wayland-session.target coming from, who provides it?

Only what I have is graphical-session.target

PR #625, opened 2020-03-14
Switch default Makefile rule from run to build

Xyene:
This commit makes it so that running `make` without arguments builds rather than builds and runs, which is unconventional and surprising behaviour.


Alexays:
Thanks!

PR #629, opened 2020-03-20
fix(memory): provide better free memory approximation on old kernels

Xyene:
The approximation should include `SReclaimable`, and subtract `Shmem`. To prevent the parsing code from ballooning in size, this commit also refactors the parsing into a map.


Alexays:
Thanks, that's a lot cleaner!

Issue #600, opened 2020-02-21
Terminator doesn't startup


masm11:
I have a config including:
```
    "custom/terminal": {
        "format": " Terminator",
        "on-click": "terminator",
        "tooltip": false
    },
```

I start waybar, and click the Terminator button, then terminator window opens.
However I click the icon the second time, then one doesn't open.

I traced in the terminator. It uses `os.uname()` of python, and it forks and executes
`uname -p`, `read()` from the process, and `wait()`s finishing of the process.
But the process already disappears, so can't `wait()`.

I think who is `wait()`ing, and it may be waybar, because waybar is anscestor of the `uname -p`.

Is there any posibility?

I'm using waybar 0.9.0 with wayfire 2d49da0 on archlinux.

kepet19:
I just tried it on my machine, and there does not seem to be a problem, I used the same config as you and I uses sway. 1.4
waybar 0.8.0

maybe you can fixed it with is instead `"terminator & disown"`
```
    "custom/terminal": {
        "format": " Terminator",
        "on-click": "terminator & disown",
        "tooltip": false
    },
```

FreeFull:
Alternatively, you could do something like `"on-click": "swaymsg exec terminator",`

masm11:
Thank you.

Terminator runs successfully on the first click, but it doesn't on the second click.

Here is the error message:
```
Traceback (most recent call last):
  File "/usr/bin/terminator", line 114, in <module>
    TERMINATOR = Terminator()
  File "/usr/lib/python2.7/site-packages/terminatorlib/terminator.py", line 72, in __init__
    self.prepare_attributes()
  File "/usr/lib/python2.7/site-packages/terminatorlib/terminator.py", line 97, in prepare_attributes
    self.pid_cwd = get_pid_cwd()
  File "/usr/lib/python2.7/site-packages/terminatorlib/cwd.py", line 42, in get_pid_cwd
    system = platform.system()
  File "/usr/lib/python2.7/platform.py", line 1303, in system
    return uname()[0]
  File "/usr/lib/python2.7/platform.py", line 1270, in uname
    processor = _syscmd_uname('-p','')
  File "/usr/lib/python2.7/platform.py", line 1005, in _syscmd_uname
    rc = f.close()
IOError: [Errno 10] 子プロセスがありません
```
The last line is in Japanese. It means "No child processes".

`terminator & disown` doesn't help for me.

[...]

PR #636, opened 2020-03-27
restore SIGCHLD settings to SIG_DFL. #600

masm11:
I startup "Terminator", a terminal emulator, from Waybar button.

Waybar forks on button click, and the child process executes Terminator,
and the Waybar process itself setups SIGCHLD to SIG_IGN to discard defunct child processes.
But the SIGCHLD setting is inherited to a new child process on fork when I click the button next time.

That is why Terminator startups with SIGCHLD setting of SIG_IGN on the second click.

Terminator forks, executes `uname -p`, and waits finishing of the `uname -p` process.
But because Terminator's SIGCHLD setting is SIG_IGN, `uname -p` process disappears just after it exits,
and Terminator can't wait it, raise an exeception, and exits.

The solution is to restore SIGCHLD setting of child processes to SIG_DFL after forks.
Then, Terminator's SIGCHLD setting becomes SIG_DFL, and `uname -p` process doesn't disappears until
Terminator waits.

### DIFF
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -88,6 +88,7 @@ inline int32_t forkExec(std::string cmd) {
   // Child executes the command
   if (!pid) {
     setpgid(pid, pid);
+    signal(SIGCHLD, SIG_DFL);
     execl("/bin/sh", "sh", "-c", cmd.c_str(), (char*)0);
     exit(0);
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #653, opened 2020-04-07
fix(client): suppress intempestive error message

### LINKED ISSUE #646: Unknown option -b
Hi,

I noticed the following message in sway's output: `Unknown option -b`, while trying to figure out why waybar had disappeared (I only found #642 later on).

It seems this message is printed by waybar (or more precisely Gtk). Here is a reproducer:
```console
user@host:~$ waybar -b bar-0
[2020-04-04 16:55:17.427] [info] Resources files: /home/user/.config/waybar/config, /home/user/.config/waybar/style.css
Unknown option -b
```

It does not seem to matter which option is passed:
```console
user@host:~$ waybar -l off
Unknown option -l
```

My guess is that this is an issue in how you call [`Gtk::Application::create()`](https://developer.gnome.org/gtkmm/stable/classGtk_1_1Application.html#a6432c724238bf3bf6dd24dd16ed56795). I applied the following patch, and it seems to resolve the issue:
```diff
diff --git a/src/client.cpp b/src/client.cpp
index f7c70e0..7b5e078 100644
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -249,7 +249,7 @@ int waybar::Client::main(int argc, char *argv[]) {
   if (!log_level.empty()) {
     spdlog::set_level(spdlog::level::from_str(log_level));
   }
-  gtk_app = Gtk::Application::create(argc, argv, "fr.arouillard.waybar");
+  gtk_app = Gtk::Application::create("fr.arouillard.waybar");
   gdk_display = Gdk::Display::get_default();
   if (!gdk_display) {
     throw std::runtime_error("Can't find display");
```

Cheers!

qb-cea:
Before this commit, launching waybar with any option would produce the
the following message on stderr: "Unknown option <option>".

Fixes #646

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -249,7 +249,7 @@ int waybar::Client::main(int argc, char *argv[]) {
   if (!log_level.empty()) {
     spdlog::set_level(spdlog::level::from_str(log_level));
   }
-  gtk_app = Gtk::Application::create(argc, argv, "fr.arouillard.waybar");
+  gtk_app = Gtk::Application::create("fr.arouillard.waybar");
   gdk_display = Gdk::Display::get_default();
   if (!gdk_display) {
     throw std::runtime_error("Can't find display");


Alexays:
That's not the correct this, we should use something like `G_APPLICATION_HANDLES_COMMAND_LINE `
c.f. https://developer.gnome.org/CommandLine/

Alexays:
Fixed by e72b153f87de2f5a33dc314a951b755d99cf86f2

PR #659, opened 2020-04-10
WIP: Keyboard layout module

BlueGone:


meson.build
```diff
@@ -75,6 +75,7 @@ gtkmm = dependency('gtkmm-3.0', version : ['>=3.22.0'])
 dbusmenu_gtk = dependency('dbusmenu-gtk3-0.4', required: get_option('dbusmenu-gtk'))
 giounix = dependency('gio-unix-2.0', required: get_option('dbusmenu-gtk'))
 jsoncpp = dependency('jsoncpp')
+pugixml = dependency('pugixml')
```
BlueGone, line 78: I suppose a wrap file could be appreciated. But on the other hand, pugixml is packaged in most distros, so I don't know if this is really necessary.

Alexays: If a wrap file is available, i'll go for it 😄 
Otherwise think about adding the dependency in the readme.

BlueGone: The wrap file doesn't exist in the wrapdb. I updated the README.

BlueGone:
:wave: @Alexays Is there any way to access the travis status ? I am doing changes on the Dockerfiles, but can't have any feedback on it.

Alexays:
https://travis-ci.org/github/Alexays/Waybar/jobs/673431112
I need to update images on Docker hub 😄 
I'll take care of it.

BlueGone:
Should I drop the commit affecting the docker images ?

man/waybar-sway-layout-5.scd
```diff
@@ -0,0 +1,82 @@
...
+
+# NAME
+
+waybar - sway layout module
+
+# DESCRIPTION
+
+The *layout* module displays the current keyboard layout of Sway
+
+# CONFIGURATION
```
BlueGone, line 11: I am not sure about the different fields here, I picked them from waybar-disk(5). Are they herited from ALabel ?

Alexays:
> Should I drop the commit affecting the docker images ?

I can't update docker image based on a PR, could you do another PR with only docker file changs?
And once merged, pipeline should be ok.

BlueGone:
> > Should I drop the commit affecting the docker images ?
> 
> I can't update docker image based on a PR, could you do another PR with only docker file changs?
> And once merged, pipeline should be ok.

#661

Alexays:
> #661
https://travis-ci.org/github/Alexays/Waybar
It's obviously not the right packages for alpine and fedora, maybe the devel one?

include/modules/sway/layout.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+ private:
+
+  using ShortNames = std::tuple<std::string, std::string>;
+  using MemoizedShortNames = std::unordered_map<
+    std::string,
+    std::tuple<std::string, std::string>
+  >;
+
+  static inline const fs::path xbk_file_ = "/usr/share/X11/xkb/rules/evdev.xml";
```
eternal-sorrow, line 43: Should this be `xkb_file_`?

Also, kind of hardcoded, isn't it? May this file be at some different path? Like under another prefix?

BlueGone:
> https://travis-ci.org/github/Alexays/Waybar
> It's obviously not the right packages for alpine and fedora, maybe the devel one?

#665

include/modules/sway/layout.hpp
BlueGone: This path is defined and standart by https://linux.die.net/man/7/xkeyboard-config . I fairly assume there will be an issue only with nixos. Are you using nixos, and if so, have you some suggestions for a workaround ?

eternal-sorrow: You can get the path to the `/usr/share/X11/xkb` dir with `pkg-config xkeyboard-config --variable xkb_base`.

eternal-sorrow: Also, is it intentional or typo `xkb_file_` -> `xbk_file_`?

maximbaz:
@BlueGone would it be possible to add CSS classes to distinguish layouts and variants? For example, I would definitely want to use different colors for different layouts. I guess the simplest way would be to add CSS classes `short-<value>` and `variant-<value>`, such as `short-en` and `variant-dvorak`.

include/modules/sway/layout.hpp
BlueGone: The type is not intended. I will fix it, thanks for the feedback.
I suppose it is possible for meson.build to generate a macro which contains the path `XKB_BASE`. I don't know how to use meson.build properly, I'll explore a bit, I'll keep you informed.

BlueGone:
> @BlueGone would it be possible to add CSS classes to distinguish layouts and variants? For example, I would definitely want to use different colors for different layouts. I guess the simplest way would be to add CSS classes `short-<value>` and `variant-<value>`, such as `short-en` and `variant-dvorak`.



> @BlueGone would it be possible to add CSS classes to distinguish layouts and variants? For example, I would definitely want to use different colors for different layouts. I guess the simplest way would be to add CSS classes `short-<value>` and `variant-<value>`, such as `short-en` and `variant-dvorak`.

Hum, nice idea. I'll take a look at it, get back to you as soon as I see it if I can.

jbeich:
@BlueGone, to unbreak [FreeBSD on CI](https://travis-ci.org/github/Alexays/Waybar/jobs/689981563) apply the following:
```diff
diff --git a/.travis.yml b/.travis.yml
index 3085d7b..962065c 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
       install:
         - sudo sed -i '' 's/quarterly/latest/' /etc/pkg/FreeBSD.conf
         - sudo pkg install -y date gtk-layer-shell gtkmm30 jsoncpp libdbusmenu
-               libfmt libmpdclient libudev-devd meson pulseaudio scdoc spdlog
+               libfmt libinput libmpdclient libudev-devd meson pugixml pulseaudio scdoc spdlog
       script:
         - meson build -Dman-pages=enabled
         - ninja -C build
```

meson.build
```diff
@@ -83,6 +83,7 @@ dbusmenu_gtk = dependency('dbusmenu-gtk3-0.4', required: get_option('dbusmenu-gt
 giounix = dependency('gio-unix-2.0', required: get_option('dbusmenu-gtk'))
 jsoncpp = dependency('jsoncpp')
 pugixml = dependency('pugixml')
+libinput = dependency('libinput')
```
jbeich, line 86: Why? I don't see libinput references in the code.

jbeich: See also 74db69dcb7cc

jbeich: FWIW, this PR builds fine if libinput is removed from meson.build and libinput package deinstalled.

FjorgeTheHuman:
Would it be possible to add support for showing the keyboard modes (num lock, scroll lock, caps lock, etc.) in this module?

PR #665, opened 2020-04-13
WIP : Layout dockerfiles

BlueGone:
This PR fixed alpine and opensuse Dockerfiles for building #659 .
This time I was able to test on my own computer, so I am more confident about the validity of this.

I have a problem with fedora : [pugixml-devel (fedora)](https://apps.fedoraproject.org/packages/pugixml-devel) is a terrible package. It doesn't include `/usr/lib/pkgconfig/pugixml.pc` so pkg-config is confused about this :
```
Run-time dependency pugixml found: NO (tried pkgconfig)

meson.build:78:0: ERROR: Dependency "pugixml" not found, tried pkgconfig
```
I don't know what is the expected way to fix this : should I manually download the file during the Dockerfile build, add a fedora specifig meson.build.patch, or add a wrapfile ?
ping @Alexays

### DIFF
--- a/Dockerfiles/alpine
+++ b/Dockerfiles/alpine
@@ -2,4 +2,4 @@
 
 FROM alpine:latest
 
-RUN apk add --no-cache git meson alpine-sdk libinput-dev wayland-dev wayland-protocols mesa-dev libxkbcommon-dev eudev-dev pixman-dev gtkmm3-dev jsoncpp-dev pugixml libnl3-dev pulseaudio-dev libmpdclient-dev scdoc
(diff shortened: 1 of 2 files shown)


Alexays:
As some fedora user will build from source, maybe go for a wrap file?

Alexays:
Can you also add missing deps like ` libdbusmenu-gtk3 libsigc++ libnl libmpdclient libappindicator-gtk3` 😄

alebastr:
> I have a problem with fedora : pugixml-devel (fedora) is a terrible package.

It's not terrible, it's just a bit outdated. Pugixml started shipping usable pkgconfig file in 1.10 but Fedora 31 has only 1.9.

Anyways, there are 2 options:
1. add cmake to the list of packages for fedora and meson will use cmake to resolve dependency
2. update base image to `fedora:32` which already has pugixml 1.10

Alexays:
Thanks @alebastr, our fedora image was still on the 30, just updated to 32

PR #684, opened 2020-04-28
Build bluetooth module man page

DiegoGuidaF:
The man page for bluetooth was present but not being built by meson.


Alexays:
Thanks!

PR #692, opened 2020-05-04
foreign-toplevel-manager based taskbar module

l3nkz:
Although waybar was originally designed to work with sway, some people use it as a bar with other compositors now. People such as me ;) One feature I was missing from waybar, was a simple taskbar with all the windows that I have open (similar to tint2 for X). After some searching I found wf-panel and its taskbar implementation based on the foreign-toplevel-manager protocol defined by wlroots. I took a lot of inspiration from this implementation and developed a module with similar features for waybar.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -7,6 +7,9 @@
 #include "modules/sway/window.hpp"
 #include "modules/sway/workspaces.hpp"
 #endif
+#ifdef HAVE_WLR
+#include "modules/wlr/taskbar.hpp"
+#endif
 #ifndef NO_FILESYSTEM
 #include "modules/battery.hpp"
 #endif
--- a/include/modules/wlr/taskbar.hpp
+++ b/include/modules/wlr/taskbar.hpp
@@ -11,10 +11,13 @@
 
 #include <gdk/gdk.h>
 
+#include <glibmm/refptr.h>
+
 #include <gtkmm/box.h>
 #include <gtkmm/button.h>
 #include <gtkmm/image.h>
 #include <gtkmm/label.h>
+#include <gtkmm/icontheme.h>
 
 #include <wayland-client.h>
 #include "wlr-foreign-toplevel-management-unstable-v1-client-protocol.h"
@@ -128,6 +131,8 @@ class Taskbar : public waybar::AModule
     Gtk::Box box_;
     std::vector<TaskPtr> tasks_;
 
+    Glib::RefPtr<Gtk::IconTheme> icon_theme_;
+
     struct zwlr_foreign_toplevel_manager_v1 *manager_;
     struct wl_seat *seat_;
 
@@ -148,6 +153,8 @@ class Taskbar : public waybar::AModule
 
     bool show_output(struct wl_output *) const;
     bool all_outputs() const;
+
+    Glib::RefPtr<Gtk::IconTheme> icon_theme() const;
 };
 
 } /* namespace waybar::modules::wlr */
--- a/include/modules/wlr/taskbar.hpp
+++ b/include/modules/wlr/taskbar.hpp
@@ -25,7 +25,8 @@ class Taskbar;
 class Task
 {
    public:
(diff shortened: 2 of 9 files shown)


Alexays:
Oh nice stuff, i'll review it when i'll get some time 😃

Alexays:
You forgot to add man file to build https://github.com/Alexays/Waybar/blob/master/meson.build#L227
Code LGTM apart the code style, but i'll take care of it in an upcoming PR.

Nice work!
Great addition so far!

When you added the man file to build, can you update the github wiki and i'll merge this PR ❤️

l3nkz:
I pushed the change that adds the man page generation for the taskbar.

The PR now has conflicts with master. Do you want me to rebase my branch on the current master and force push? Or do you want to resolve the conflicts yourself.

Furthermore, I had some crashes recently and I am still trying to figure out what happens there. The crash occurs when I unplug my external monitor. First the bar gets unresponsive (doesn't update texts, …) and after some time crashes completely with the following message:
```
waybar: ../src/ierror.h:130: mpd_error_get_message: Assertion `error->message != NULL || error->code == MPD_ERROR_OOM' failed.
```
I still have to look in the core dump.

Alexays:
Conflicts fixed :)
For the crash, It's probably from a module destrutor, obviously the mpd's destrutor here.

l3nkz:
Ok, so it is not related to my changes … I will double check with the original master branch without my module when I have time.

Alexays:
Just merged!
Your issue should be fixed with latest master changes

Alexays:
Thanks!

l3nkz:
> Just merged!
> Can you update the github wiki?
> Your issue should be fixed with latest master changes

I updated the wiki now. Sorry for the long delay. I was very busy the last weeks.

Anakael:
Hello!
I try to launch taskbar and get error:
`[error] Failed to register as toplevel manager`
In config i have:
`"modules-center": ["wlr/taskbar"],`
Maybe I need latest sway version? (Now I have the one from community arch repo).
Great work! Interesting, how you solve problem with icon. Need to study your solution :)

l3nkz:
Hm, it seems as if your compositor doesn't support the foreign-toplevel-manager protocol. This is a limitation of sway (See issue #728). But according to the issue using the git version of sway from the AUR might solve the problem.

Which problem with the icon do you mean?

Anakael:
Problem that wayland doesn't hava xdg-open analog and icon path for application is not fixating.

Anakael:
I just rebuild sway with wlroots from master branches and there is still the same issue.
With which compositor did you test it?

Anakael:
It works now!
But is there way for separate task bar for tag? 
And there is no icon on tasks...

Anakael:
And close on middle-click doesn't work.

Anakael:
About middle-click:
Found out that is mistake in example. Work fine. I will commit that fix if manage to solve another tasks in taskbar.

Anakael:
-> And there is no icon on tasks...
Oh, that bug still not fixed.
You set icon in app_id subscription. But not all nodes in sway have that property. 
For example: Alacritty has, but firefox, visual-studio-code, gnome-terminal - hasn't.
I will try to search sway about that.

l3nkz:
> About middle-click:
> Found out that is mistake in example. Work fine. I will commit that fix if manage to solve another tasks in taskbar.
Good catch. I fixed the wiki page. Thank you.

> I just rebuild sway with wlroots from master branches and there is still the same issue.
> With which compositor did you test it?
I use wayfire.

> -> And there is no icon on tasks...
> Oh, that bug still not fixed.
> You set icon in app_id subscription. But not all nodes in sway have that property.
> For example: Alacritty has, but firefox, visual-studio-code, gnome-terminal - hasn't.
> I will try to search sway about that.
Hm, that is interesting. I will also have a look into this. But this might be a bug of sway. Because if I don't get an app_id, I don't really know how to identify the app. The title is usually not very helpful.

Anakael:
It's not really bug, but nuance of Wayland and sway. app_id used only for pure Wayland application and missing in xwayland. 
For xwayland apps .class property can be useful.

PR #693, opened 2020-05-04
sway/workspaces added current_output CSS class to buttons.

f0rki:
All workspace buttons that are visible on the same output as the current waybar can be styled with the `current_output` css class.

This is really only useful in combination with the `"all-outputs":
true`. Then the workspaces that are on the current output can be styled
differently than the workspace on other outputs, while all workspaces are visible in every waybar.


Alexays:
Thanks! can you update the wiki?

f0rki:
Sure. I updated the wiki.

Issue #482, opened 2019-10-09
workspace 10 order is not correct


guanghwang:
## Platform
```
arch Linux, sway 1.2-3, waybar 0.8.0-3
```

## Problem

The order of workspace 10 is not correct. It is not next to workspace 9. Instead, it is between workspace 1 and workspace 2. If I use sway-bar, workspace 10 is right next to workspace 9.

## How to reproduce
```
1. add workspace 10 in sway config
2. switch to workspace 10
3. It is between workspace 1 and workspace 2
```

I expect that workspace 10 is right next to workspace 9.

Alexays:
It is because the comparison is made on a string so the 1 of 10 is taken into account, i'll look into that.

Ventto:
I have tried with 10 workspaces (even with 11) and it seems to works  (c.f the screenshot below). Here is the sway config:

```
set $ws1 "1: A"
set $ws2 "2: B"
set $ws3 "3: C"
set $ws4 "4: D"
set $ws5 "5: E"
set $ws6 "6: F"
set $ws7 "7: G"
set $ws8 "8: H"
set $ws9 "9: I"
set $ws10 "10: J"

workspace $ws1
workspace $ws2
workspace $ws3
workspace $ws4
workspace $ws5
workspace $ws6
workspace $ws7
workspace $ws8
workspace $ws9
workspace $ws10
```
![waybar example](https://i.ibb.co/Q608nrS/region-2019-10-14-202723.png)

Can you share your configuration ?

guanghwang:
I use the default config in sway. 

```
    bindsym $sup+1 workspace 1
    bindsym $sup+2 workspace 2
    bindsym $sup+3 workspace 3
    bindsym $sup+4 workspace 4
    bindsym $sup+5 workspace 5
    bindsym $sup+6 workspace 6
    bindsym $sup+7 workspace 7
    bindsym $sup+8 workspace 8
    bindsym $sup+9 workspace 9
    bindsym $sup+0 workspace 10
```
waybar config
```
    "sway/workspaces": {
        "disable-scroll": true,
        "format": "{name}:{icon}",
        "format-icons": {
            "1": "😎",
            "2": "🌐",
            "3": "📁",
            "4": "🤖",
            "5": "📝",
            "6": "📒",
            "7": "📖",
            "8": "🎸",
            "9": "📧",
            "10": "🌞"
        }
    },
```

Alexays:
[...]

PR #696, opened 2020-05-05
sway/workspaces: sort by the "num" property provided by sway

f0rki:
Sway provides the workspace "num" property which is an integer number of the workspace, i.e., workspace `"3"` -> `3` and also `"3dev"` -> `3`. This PR uses this property to sort the workspaces, which makes sense when persistent workspaces or all-output is specified. This commit also adds a new configuration option, whether the numeric workspaces come in front or after workspaces that have non-numeric name.

* I think this should fix issue #482.
* Tested only with sway version 1.4
* Always sorts workspaces, regardless of any settings.

### DIFF
--- a/man/waybar-sway-workspaces.5.scd
+++ b/man/waybar-sway-workspaces.5.scd
@@ -64,6 +64,10 @@ Addressed by *sway/workspaces*
 	typeof: string ++
 	Command to execute when the module is updated.
 
+*numeric-first*: ++
+	typeof: bool ++
+	Whether to put workspaces starting with numbers before workspaces that do not start with a number.
+
 # FORMAT REPLACEMENTS
 
 *{value}*: Name of the workspace, as defined by sway.
@@ -107,6 +111,7 @@ n.b.: the list of outputs can be obtained from command line using *swaymsg -t ge
 "sway/workspaces": {
     "disable-scroll": true,
     "all-outputs": true,
+    "numeric-first": false,
     "format": "{name}: {icon}",
     "format-icons": {
         "1": "",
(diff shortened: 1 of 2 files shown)


Alexays:
Nice thanks!
Should be good for #482 
Can you update the wiki?

f0rki:
@Alexays sure I'll update the wiki

imjasonmiller:
Given that this is fixed, I think I might have misconfigured something on my end? 

```shell
sway version 1.4-5c32a484 (May 20 2020, branch 'master')
Waybar v0.9.2-140-g759602a (branch 'master')
```
**sway/config**
```bash
# Workspaces
set $ws1 1:Web
set $ws2 2:Code
set $ws3 3:Other
set $ws4 4:Media
set $ws5 5:Chat
```

**waybar/config**
```json
"sway/workspaces": {
    "all-outputs": true,
    "format": " {name} ",
    "persistent_workspaces": {
        "1:Web": [],
        "2:Code": [],
        "3:Other": [],
        "4:Media": [],
        "5:Chat": []
    }
},
```

This gives me the following workspace order, where `Web` (1) is placed at the very end:

Waybar taskbar

Thank you for your time.

f0rki:
@imjasonmiller Hmm I think you have stumbled upon a rather strange bug. I tried to reproduce this in addition to my usual numeric workspaces. That did not work, but in a nested sway session I managed to reproduce, but strangely enough it only shows up if the "1:Web" (or "1:X" in my case) workspace is the first workspace that is focused...

When I launch sway and switch to X
20200722_15h17m18s_grim
20200722_15h17m38s_grim

However, when I launch sway and switch to C the ordering stays intact.
20200722_15h17m18s_grim
20200722_15h17m25s_grim

f0rki:
@imjasonmiller I think this might be that same issue as #777

PR #701, opened 2020-05-07
Confirm build-instructions for Ubuntu 20.04 too.

josteink:


Alexays:
Thanks for the update!

PR #702, opened 2020-05-08
Add a criticial threshold for both CPU and Memory usage

becausewat:
This will allow users to configure percentage based thresholds where the CSS for a module can be changed once the threshold is reached.

It's pretty much just copied from the way that the temperature module implements it.


becausewat:
That was not what I meant to do. Will fix whatever just happened and make another request.

PR #703, opened 2020-05-08
 Add a criticial threshold for both CPU and Memory usage

becausewat:
This will allow users to configure percentage based thresholds where the CSS for a module can be changed once the threshold is reached.

It's pretty much just copied from the way that the temperature module implements it.

(this PR should be good, and not messed up like my previous one)

### DIFF
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -22,6 +22,7 @@ class Cpu : public ALabel {
  private:
   static inline const std::string         data_dir_ = "/proc/stat";
   uint16_t                                getCpuLoad();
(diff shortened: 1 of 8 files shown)


maximbaz:
@becausewat what you want to achieve is already possible:

    "cpu": {
        "format": "{usage}%",
        "states": {
            "warning": 70,
            "critical": 90
        }
    },

```
#cpu.warning {
  border-bottom: 3px solid yellow;
}

#cpu.critical {
  border-bottom: 3px solid red;
}
```

Alexays:
As mentioned by maximbaz it's already possible.

PR #719, opened 2020-05-23
Simplify .travis.yml

jbeich:
I'm still new to TravisCI thus have misunderstood how `jobs.include` work. It seems topmost phases are not appended but overwritten. So, let's go back to the previous matrix for (implicit) `os: linux`.

While here
- Drop `sudo` because it has [no effect](https://config.travis-ci.com/ref/sudo)
- Move `language` to the top, so `os: linux` builds aren't tagged as Ruby

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,34 @@
+language: cpp
+
+services:
+    - docker
+
 git:
     submodules: false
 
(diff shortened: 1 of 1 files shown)


Alexays:
Even better, thank you 🙏

PR #736, opened 2020-06-07
sway/workspaces: make clicking on workspaces idempotent

Xyene:
Previously, clicking on the same workspace you were on would throw you
to another workspace if `workspace_auto_back_and_forth yes` was
specified in your sway config. This also fixes workspace output moving
misbehaving and doing the same.

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -20,6 +20,8 @@ class Workspaces : public AModule, public sigc::trackable {
   auto update() -> void;
 
  private:
+  static inline const std::string workspace_switch_cmd_ = "workspace --no-auto-back-and-forth \"{}\"";
+
   void              onCmd(const struct Ipc::ipc_response&);
   void              onEvent(const struct Ipc::ipc_response&);
   bool              filterButtons();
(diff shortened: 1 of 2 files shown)


src/bar.cpp
```diff
@@ -305,6 +305,7 @@ void waybar::Bar::setupAltFormatKeyForModuleList(const char* module_list_name) {
 void waybar::Bar::handleSignal(int signal) {
   for (auto& module : modules_left_) {
     auto* custom = dynamic_cast<waybar::modules::Custom*>(module.get());
+    module->update();
```
Alexays, line 308: Why did you added that?

Alexays:
LGTM except the comment 😃

src/bar.cpp
Xyene: Whoops, that snuck in from a patch I maintain locally! Removed it from the patch :+1:

Alexays:
Thanks!

mischw:
> Previously, clicking on the same workspace you were on would throw you
> to another workspace if `workspace_auto_back_and_forth yes` was
> specified in your sway config.

Isn't that what this config flag was supposed to do? I used it to toggle between two workspaces. Any way to re-enable this behavior?

Moonlight-Angel:
@Xyene Hi, just to let you know #1014 was just merged into `master`.
In order to keep the behavior you introduced with this PR, there is a new config flag `disable-auto-back-and-forth` you have to set to `true` in the Sway/Workspaces config.

PR #747, opened 2020-06-16
Only real batteries are counted as batteries

Jannusch:
I have a Wacom pen which is loaded via my laptop. Currently its battery has been added to the laptop battery, which has doubled the total capacity, so I have never been able to charge my laptop over 50% according to the display. But pens are from the type USB, so it is not recognized anymore by my fix.


Alexays:
Thanks!

PR #751, opened 2020-06-24
Fixes for wlr/taskbar

### LINKED ISSUE #728: Taskbar module questions
I just tried the taskbar module.
When I set this to icon, "format": "{icon}", this work
I tried "format": "{title}" or "{app_id}" but the taskbar doesn't appeared.

Also how to define the style.css?
I added this but it has no effect on waybar.
#taskbar
#taskbar button
#taskbar button.maximized
#taskbar button.minimized
#taskbar button.active
#taskbar button.fullscreen

Ok I will wait until the wiki has been updated you close this If you want, by the way I been using the taskbar module since 4 days now and it working well! here my config:
 "wlr/taskbar": {
    "format": "{icon}",
	"tooltip-format": "{title}",
	"on-click": "activate",
	"all-outputs": false,
	"active-first": true,
	"icon-size": 32,
	"icon-theme" : "Obsidian",	
	"on-click-right": "minimize"

### LINKED ISSUE #735: Taskbar: sh: minimize: command not found
I think I don't fully understand the logic behind the taskbar.
My understanding is that "activate" should bring me to this window/viewport, but it does nothing. (I'm using wayfire). "minimize" should put the window into the tray, but it calls shell instead. So far I just can see what windows are open, without possibility of any sencible action.

l3nkz:
These two commits fix the problems with the 'on-click-right' click events and the 'format' option.

Resolves #728, Resolves #735

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -214,7 +214,7 @@ Task::Task(const waybar::Bar &bar, const Json::Value &config, Taskbar *tbar,
 
     /* Handle click events if configured */
     if (config_["on-click"].isString() || config_["on-click-middle"].isString()
-            || config_["on-click-left"].isString()) {
+            || config_["on-click-right"].isString()) {
         button_.add_events(Gdk::BUTTON_PRESS_MASK);
         button_.signal_button_press_event().connect(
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #761, opened 2020-06-30
Change find icon priority and to_lower for app_id

Anakael:

Issue #754, opened 2020-06-25
Taskbar: Missing application icons when the .desktop is in a folder not being searched or hicolor theme is being used as fallback


ghost:
Some applications are not displaying their taskbar icons.

https://github.com/Alexays/Waybar/blob/8f684f703eb9c8a4e8b8ad02890622c486e960ea/src/modules/wlr/taskbar.cpp#L30

Does not include values like  `~/.local/share/applications` and Flatpak's `~/.local/share/flatpak/exports/share/applications` (--user), `/var/lib/flatpak/exports/share/applications`. Maybe the `$XDG_DATA_DIRS` environment variable, with the additional /applications sub-folder directory, could be used too?

The second search is from the icon theme.
https://github.com/Alexays/Waybar/blob/8f684f703eb9c8a4e8b8ad02890622c486e960ea/src/modules/wlr/taskbar.cpp#L67

`hicolor` is [the default fallback theme](https://developer.gnome.org/icon-theme-spec/) and it's why the icons were showing up for me on GNOME and Plasma. It could be hardcoded as the fallback theme after everything else fails.

Or multiple choices for `icon-theme`.

```json
"wlr/taskbar": {
    "icon-theme": "my-main-theme, my-custom-theme, hicolor"
}
```

l3nkz:
> Some applications are not displaying their taskbar icons.
> 
> https://github.com/Alexays/Waybar/blob/8f684f703eb9c8a4e8b8ad02890622c486e960ea/src/modules/wlr/taskbar.cpp#L30
> 
> Does not include values like `~/.local/share/applications` and Flatpak's `~/.local/share/flatpak/exports/share/applications` (--user), `/var/lib/flatpak/exports/share/applications`. Maybe the `$XDG_DATA_DIRS` environment variable, with the additional /applications sub-folder directory, could be used too?

I just looked in the definition of `$XDG_DATA_DIRS` and it seems as if this is a good idea. Because if this environment variable is not set one should use `/usr/share` and  `/usr/local/share` which are the folders that I already use. Changing the code so that I use `$XDG_DATA_DIRS` should be an easy change.

> The second search is from the icon theme.
> https://github.com/Alexays/Waybar/blob/8f684f703eb9c8a4e8b8ad02890622c486e960ea/src/modules/wlr/taskbar.cpp#L67
> 
[...]

PR #766, opened 2020-07-07
Support for multiple icon themes and XDG_DATA_DIRS

l3nkz:
Add support for defining multiple icon themes in the config which are used in the given order to lookup icons. With this approach it is possible to define one (or multiple) fallback icon themes if the app cannot be found in the current main icon theme. In addition the module will always lookup icons in the system default icon theme (for GTK this is hicolor).

Also respect XDG_DATA_DIRS now if set to get correct lookup paths for desktop files in the system. Especially for flatpack this is of importance. See #754

In addition I also refactored a little bit the lower_app_id changes done in PR #761

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -20,6 +20,27 @@
 
 namespace waybar::modules::wlr {
 
+/* String manipulation methods */
+const std::string WHITESPACE = " \n\r\t\f\v";
+
+static std::string ltrim(const std::string& s)
+{
+	size_t start = s.find_first_not_of(WHITESPACE);
+	return (start == std::string::npos) ? "" : s.substr(start);
+}
+
+static std::string rtrim(const std::string& s)
+{
+	size_t end = s.find_last_not_of(WHITESPACE);
+	return (end == std::string::npos) ? "" : s.substr(0, end + 1);
+}
+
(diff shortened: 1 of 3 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -539,13 +580,23 @@ Taskbar::Taskbar(const std::string &id, const waybar::Bar &bar, const Json::Valu
...
-        icon_theme_ = Gtk::IconTheme::create();
-        icon_theme_->set_custom_theme(config_["icon-theme"].asString());
-        spdlog::debug("Use custom icon theme: {}.", config_["icon-theme"].asString());
-    } else {
-        spdlog::debug("Use system default icon theme");
-        icon_theme_ = Gtk::IconTheme::get_default();
+        auto icon_theme_config = config_["icon-theme"].asString();
+        size_t start = 0, end = 0;
+
+        do {
```
Alexays, line 586: Switch icon-theme to array, and keep `if (config_["icon-theme"].isString()) {` if we pass just one icon theme.
Otherwise, add an `if (config_["icon-theme"].isArray()) {` it'll much simpler than splitting by `,`

l3nkz: Thank you, will fix this.

Alexays:
Thanks!

PR #772, opened 2020-07-16
CI: unbreak FreeBSD

jbeich:
`/quarterly` package set was recently updated to **2020Q3**, so let's use it instead of bumpy `/latest`.


Alexays:
Thanks!

Issue #739, opened 2020-06-08
Custom modules disappear on-click


luispabon:
Here's my config: https://github.com/luispabon/sway-dotfiles/tree/master/configs/waybar

I have two custom modules, one for cpu speed and a power button one. As soon as click (right or left) on either, they disappear. Sometimes from all displays, sometimes from the one I clicked on. Sometimes both modules disappear.

 https://youtu.be/ig2MkdemrGg 

The click-action does work - clicking on the cpu speed one does bring a terminal with htop etc.

I can't say when this started happening, but relatively recently.

wlroots 0.10
sway 1.4
Ubuntu 20.04
waybar 4b2e6b54a7fcfa5aae2bbc06f8683bbcf3855849 (1 Jun, latest in master as of writing this)

X0rg:
This issue appeared for me since I build Waybar from the master branch. It may be a regression introduced by fixes for #572.

alexarice:
I'm also getting this problem with a custom spotify module, after running `systemctl --user status waybar` there is the following message:
```
Couldn't map window 0x134f860 as subsurface because its parent is not mapped.
```

alexarice:
I think the message I posted was unrelated, I bisected the issue down to one of

6b32aca094ad108fa4b4a40084bb8fbd6fb118db
6ca4e14b296fa6c1b18e22f28160ba9965ff4bb4
9b9d13ab0d90f16f6d64e7e9dc7394bf80121785

The module didn't show up at all in two of these commits and before the module works fine and after it disappears on click

Alexays:
Fixed with #778

luispabon:
Thank you, I can config that fixed the problem.

PR #778, opened 2020-07-21
Handle SIGCHLD for exec/forkExec

excellentname:
When forkExec is called it begins to ignore all SIGCHLD signals for the rest of the progam's execution so that they are automatically reaped. However, this means that subsequent waitpid calls in the exec function will always fail. So instead handle SIGCHLD by reaping any processes created by forkExec and ignoring all others so that they can be handled directly by the exec function.

I had a custom module disappearing on scroll-up which is fixed by this and I'm 99% sure this is what causes #739 while still fixing #369. I didn't know where to put the signal handler function, so I just put it in main.cpp. It adds a couple of ugly global variables, but I couldn't think of any other way to do it. If waybar was relying on signal(SIGCHLD, SIG_IGN) to reap other child processes not created by forkExec then this won't work, but I couldn't find any other instances of fork(), so I think it's ok.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,30 +1,70 @@
 #include <csignal>
 #include <list>
+#include <mutex>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <spdlog/spdlog.h>
 #include "client.hpp"
 
-sig_atomic_t is_inserting_pid = false;
+std::mutex reap_mtx;
 std::list<pid_t> reap;
 
-static void handler(int sig) {
-  int saved_errno = errno;
-  if (!is_inserting_pid) {
-    for (auto it = reap.begin(); it != reap.end(); ++it) {
-      if (waitpid(*it, nullptr, WNOHANG) == *it) {
-        it = reap.erase(it);
-      }
+void* signalThread(void* args) {
+  int err, signum;
+  sigset_t mask;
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -1,7 +1,32 @@
 #include <csignal>
+#include <list>
+#include <sys/types.h>
+#include <sys/wait.h>
 #include <spdlog/spdlog.h>
 #include "client.hpp"
 
+sig_atomic_t is_inserting_pid = false;
```
Alexays, line 8: Should be replaced with a mutex, no?

excellentname: I forgot signals can be handled by any thread, so yeah it's not safe atm. But I'm pretty sure using a mutex in a signal handler is unsafe so we can't do that. Apparently a good way handle signals in multithreaded programs is to have a dedicated thread that handles signals like [here](https://www.linuxjournal.com/article/2121), [here](https://devarea.com/linux-handling-signals-in-a-multithreaded-application/), and [here](http://poincare.matf.bg.ac.rs/~ivana/courses/ps/sistemi_knjige/pomocno/apue/APUE/0201433079/ch12lev1sec8.html). That way we can properly use mutexes. So I will try that

Alexays:
Thanks!

Issue #777, opened 2020-07-20
[sway/workspaces] Persistent workspaces always appear first if not created by sway


mjec:
Given a persistent workspaces configuration, if any workspace has not been created (i.e. does not appear in `swaymsg -t get_workspaces`), that workspace button will always appear first in sway.

I believe this is because since 6fe764540c431657b37b575cf3ac103cf369d315 the workspace `num` property is used for sort order, however that is [not set for persistent workspaces](https://github.com/Alexays/Waybar/blob/759602af64e10948af9665813af05237e36ad845/src/modules/sway/workspaces.cpp#L88-L91). I believe it's therefore cast to zero, appearing first.

One consequence of this is that switching to such a workspace causes the workspace button to jump to the correct position.

The fix for this isn't obvious to me. We could set the `num` property to -1, which would result in these being grouped with non-numeric workspaces, and their position dependent on the `numeric-first` config parameter. However this behavior is a little inconsistent, as persistent workspaces may be numeric. This will still, I assume, make the workspace button jump around when changing to that workspace.

My best idea so far is to set `num` to the `strtoi` of the workspace name, or -1 if strtoi fails. This should mirror the behavior of sway, which should at least prevent the button jumping around. The downside is that if sway's behavior ever changes, this will break.

I'm open to ideas here, and happy to code up whatever is decided as the best approach.

f0rki:
For reference, there are some comments with screenshots in the pull request that lead to this problem: #696 

> My best idea so far is to set num to the strtoi of the workspace name, or -1 if strtoi fails. This should mirror the behavior of sway, which should at least prevent the button jumping around. 

This is something I thought of doing before discovering sway's "num" property, which was simply easier to reuse in my PR.

> The downside is that if sway's behavior ever changes, this will break.

I would suggest to simply not use the "num" property anymore. If the "num" property is ignored completely by waybar and the logic of assigning a number to a workspace is duplicated in waybar, then waybar will be independent of sway's "num" assignments. Since this is only used for ordering and displaying, this should not cause too much trouble even if waybar and sway disagree on the workspace "num" in the future.

Alexays:
Fixed by #783 🎉

PR #783, opened 2020-07-22
make waybar itself assign numbers to workspaces like sway

f0rki:
sway assigns a "num" property to every workspace and this PR copies the logic of sway into waybar. While this is duplicated effort, it seems to be needed to properly sort persistent workspace, which might not exist in the sway IPC response, but needs something like the "num" property anyway.

Not sure this is really the best way but it seems to fix the issue #777 and the comment in the PR #696 .

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -1,8 +1,27 @@
 #include "modules/sway/workspaces.hpp"
 
+#include <string>
+#include <cctype>
 #include <spdlog/spdlog.h>
 
+
 namespace waybar::modules::sway {
+
+// this is the code that sway uses to assign a number to a workspace. This is
+// taken quite verbatim from `sway/ipc-json.c`.
+int sway_wsname_to_num(std::string name) {
+  if (isdigit(name[0])) {
+    errno = 0;
+    char *    endptr = NULL;
+    long long parsed_num = strtoll(name.c_str(), &endptr, 10);
+    if (errno != 0 || parsed_num > INT32_MAX || parsed_num < 0 || endptr == name.c_str()) {
+      return -1;
+    } else {
+      return (int)parsed_num;
+    }
+  }
+  return -1;
+}
 
 Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value &config)
     : AModule(config, "workspaces", id, false, !config["disable-scroll"].asBool()),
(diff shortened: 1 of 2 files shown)


src/modules/sway/workspaces.cpp
```diff
@@ -102,13 +121,27 @@ void Workspaces::onCmd(const struct Ipc::ipc_response &res) {
                     // the "num" property (integer type):
                     // The workspace number or -1 for workspaces that do
                     // not start with a number.
-                    auto l = lhs["num"].asInt();
-                    auto r = rhs["num"].asInt();
+                    //auto l = lhs["num"].asInt();
+                    //auto r = rhs["num"].asInt();
```
Alexays, line 125: Shouldn't we keep it as a fallback?

f0rki: Maybe, but this way waybar is completely independent from how sway assigns the `num` property resulting in less unexpected behaviour in waybar, in case sway ever changes the way the `num` property is set.

f0rki:
Btw @imjasonmiller and @mjec could you please confirm this fixes the issues you noticed on your setups?

mjec:
> @mjec could you please confirm this fixes the issues you noticed on your setups?

Yup this is a complete fix for what I reported as #777

imjasonmiller:
> @imjasonmiller ... could you please confirm this fixes the issues you noticed on your setups?

Lovely! I just built it and it works. Here's two instances of Waybar running, with the above being the fixed version:

image

src/modules/sway/workspaces.cpp
```diff
@@ -1,9 +1,28 @@
...
+#include <string>
+#include <cctype>
 #include <spdlog/spdlog.h>
 
+
 namespace waybar::modules::sway {
 
+// this is the code that sway uses to assign a number to a workspace. This is
+// taken quite verbatim from `sway/ipc-json.c`.
+int sway_wsname_to_num(std::string name) {
```
Alexays, line 12: Can you just move this func as a method of Workspaces class, and i'll merge this PR :)

f0rki: Is it really needed? The the lambda used in the `std::sort` call would then need to capture `this` as well.

Alexays: At least a static method, that's just in order to avoid anonymous func outside the class.

f0rki: @Alexays done :)

Alexays:
Thanks!

PR #787, opened 2020-07-28
Clang format

carlsmedstad:
Added a formatting job that fails if code is not formatted with clang-format. This will fail now since the code currently is not.

Edit: Never mind, it's actually formatted, my bad. Using the default Travis environment provides clang-format 3.8 which is from 2016. Optionally, one could fetch the latest stable version of clang-format or use Travis's Ubuntu 18.04 environment which should provide clang-format 6.0.0. This would probably require a reformatting of the code in order for that job to be green.

Edit2: In order to do as I've done clang-format version 10.0.0 is needed due that's when `--dry-run` and `--Werror` was introduced. Using an older version of clang-format will require some more scripting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,27 @@
 language: cpp
 
 services:
-    - docker
+  - docker
 
 git:
-    submodules: false
+  submodules: false
 
 env:
-    - distro: debian
-    - distro: archlinux
-    - distro: fedora
-    - distro: alpine
-    - distro: opensuse
+  - distro: debian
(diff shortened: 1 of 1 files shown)


carlsmedstad:
Should the choice to stay on a clang-format version less than 10, one could achieve the same goal of failing non-formatted code by formatting in-place and ask git for changes:
```sh
clang-format -i $FILES
git diff-index HEAD --quiet
```
The second line will produce a non-zero exit code should any files have changed.

PR #788, opened 2020-07-28
Fix 'date' dependency in meson

JonasToth:
The 'date' library dependency was incompletly imported with meson.
Only the target 'date::date' seemed to be caught by meson, but
'date::date-tz' not, which lead to build errors in gentoos sandbox
environment.

See this question, too:
https://stackoverflow.com/questions/62951569/meson-doesnt-link-library

Adding the modules in meson imports them all and the library builds
nice.
Note, that this did not happen with a standard checkout and local build
on my system, but only when creating an ebuild for the package.


Alexays:
Nice, good to know, thx!

PR #793, opened 2020-07-31
network: make missing nl80211 non-fatal

uggedal:
This will enable the networking module to be used for ethernet
interfaces on kernels without nl80211 support.

It should be reasonable to allow desktop systems without
wireless interfaces to run custom kenrel configs
without nl80211 compiled in.

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -158,7 +158,7 @@ void waybar::modules::Network::createInfoSocket() {
   }
   nl80211_id_ = genl_ctrl_resolve(sock_, "nl80211");
   if (nl80211_id_ < 0) {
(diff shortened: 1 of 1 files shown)


Alexays:
Should we add a check here
https://github.com/Alexays/Waybar/blob/fb8cda9d903d5a10153f8ece720a0fb7ad3efbd9/src/modules/network.cpp#L734-L738
Before the genlmsg_put call?

uggedal:
I'm not experienced enough with this code to say, but I've not noticed any crashes while running with my simple patch for the last month: https://github.com/uggedal/overlay/commit/69642772cac571ebea6884fb829e401a88e1235a#diff-bb5ca82bc940d3f35b9d30373ebd2a6a

Alexays:
Thanks!

PR #800, opened 2020-08-06
switch workspace on mouse-down to match swaybar

### LINKED ISSUE #686: sway/workspaces switches workspace on mouse-up, not mouse-down
This makes it feel notably less responsive. The normal bar switches workspace on mouse-down.

NotAFile:
closes #686

Previously, the workspace only switched when you release the mouse, which feels slow and clunky compared to i3bar and swaybar.


Alexays:
LGTM, thanks!

Issue #810, opened 2020-08-10
Network module crashes waybar with bandwidthDownBits format specifier


rgouicem:
Hi!

Since the latest update (version 0.9.3), my network module crashes waybar on startup. More specifically, the `bandwidthDownBits` format does. Using the `-l trace` option gives no information on the error, I just get a segmentation fault.

Here is my network module configuration:
```
"network": {
        "format-ethernet": " {ifname} ({bandwidthDownBits})",
        "format-linked": " (No IP) ⚠",
        "format-wifi": " {essid} ({signalStrength}%)",
        "format-disconnected": " 睊 ⚠",
	"on-click": "gnome-terminal -- nmtui",
	"on-click-right": "~/.config/rofi/scripts/network.sh",
        "tooltip": false
}
```

If I remove the `{bandwidthDownBits}` format, it works correctly. Note that all four formats that check the speed of the connection cause the crash.

Since I use Arch Linux with a recent 5.7.12 kernel, I thought it might be due to a change in the `proc/net/netstat`  file, but it does not seem to be the case as far as I can tell.

pbartyik:
similar behaviour with `bandwidthUpBits` too.

maximbaz:
I confirm, this seem to happen since we upgraded `fmt` to v7. I just downgraded `fmt` to `6.2.1`, recompiled and network module works as expected, no segfaults.

With `fmt v7.0.3` this is a dump of a debug build:

```
       Message: Process 95054 (waybar) of user 1000 dumped core.
                
                Stack trace of thread 95054:
                #0  0x0000555ec303749a n/a (/home/maximbaz/Waybar/build/waybar + 0x4349a)
                #1  0x0000555ec303736d n/a (/home/maximbaz/Waybar/build/waybar + 0x4336d)
                #2  0x00007fcc72461e7a _ZN3fmt2v76detail23parse_replacement_fieldIcRNS1_14format_handlerINS1_13arg_formatterISt20back_insert_iteratorINS1_6bufferIcEEEcEEcNS0_20basic_format_contextIS8_cEEEEEEPKT_SG_SG_OT0_ (libfmt.so.7 + 0x17e7a)
[...]

PR #813, opened 2020-08-10
Fix crash with fmt

xenrox:
For https://github.com/Alexays/Waybar/issues/810


Alexays:
Even if it's not the right fix, it fixes the crash.

Alexays:
Thanks!

PR #814, opened 2020-08-10
Allow enabing pango markup in the taskbar string

danieldg:
The fix for taskbar tooltips in 6a2d214b55 was incomplete: it causes the label
to contain escaped titles.  Allow choosing whether to enable markup or not.

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -306,7 +306,7 @@ std::string Task::state_string(bool shortened) const
 
 void Task::handle_title(const char *title)
 {
-    title_ = Glib::Markup::escape_text(title);
+    title_ = title;
 }
 
 void Task::handle_app_id(const char *app_id)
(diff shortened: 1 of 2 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -460,38 +460,78 @@ bool Task::operator!=(const Task &o) const
...
-                    fmt::arg("title", title_),
-                    fmt::arg("app_id", app_id_),
-                    fmt::arg("state", state_string()),
-                    fmt::arg("short_state", state_string(true))
-                )
-        );
-        text_after_.show();
-    }
+    bool markup = config_["markup"].isBool() ? config_["markup"].asBool() : false;
+    if (markup) {
```
Alexays, line 464: Rather than this big if/else, could you save the fmt::format and then doing an if/else with just the set_markup/set_label?

danieldg: Since the arguments to the format (title, app_id) also need to change, I did it this way to avoid making a copy of those strings unless they needed escaping. I could switch to always making local strings, using them for the format, and then choosing set_{markup,text} if you think that's clearer.

Alexays (approved):

Alexays:
Thanks!

Alexays:
Can you update the wiki?

PR #817, opened 2020-08-12
Add option for no workspace switch on click

pedrocr:
In sway/workspaces, just like `disable-scroll` turns on/off the ability to change workspaces by scrolling the mouse add `disable-click` that turns on/off the ability to change workspaces by clicking.


Alexays:
Thanls, can you update the wiki?

pedrocr:
Sure, done.

PR #819, opened 2020-08-13
Add IDs to sway workspace buttons for CSS styling

pedrocr:
In case you want to style a specific workspace add IDs to the workspace buttons. Styling is done by matching button#sway-workspace-${name}.


Alexays:
Thanks, good idea!
Can you update the wiki?

pedrocr:
Sure, done. Sent a PR for a missing `*` in the manpage. It was building fine but it was still wrong.

PR #821, opened 2020-08-13
New clock features

danieldg:
Timezone switching inspired by KDE's clock module; calendar formatting fix because the bold/underline is easy to miss on my font/monitor (I use "<b><span color='green'>{}</span></b>").


Alexays:
Thanks!
Can you update the wiki with `today-format` and `timezones`?

Issue #822, opened 2020-08-14
Waybar crashes on startup because of rfkill


W-joe:
I don't know what rfkill is and I don't want to know.
I went by just fine without it and I'm living a happy life.
However since waybar 0.9.3, more specifically from the change log Network: Added Rfkill status #568 @marcplustwo, waybar refuses to start which makes me sad.

W-joe:
P.S. this is my first Issue/Pull request on github, so please bear my mistakes and I'd appreciate it if those were pointed out to me. Thanks.

Alexays:
Perfect!

theduke:
@Alexays I'm getting the same crash:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Can't open RFKILL control device
```

A compile time config option isn't exactly a great solution...

I disabled the network module for now, which does make it work.

But that's obviously suboptimal.

W-joe:
@theduke not wanting to pay for what you don't use isn't odd at all. Just think any custom Linux kernel that doesn't include unwanted/unneeded functionality.
Or source based Linux distributions where you can just disable unneeded features via use flags or some such.

So from my point of view this 'rfkill optional' works as advertised and, at least for me, perfectly fine.

If you come from a binary distribution's point of view one could argue it's incomplete and, indeed, it does still crash for the very same reasons as before.
But the problem is with the incomplete implementation of #568 by @marcplustwo who doesn't check whether or not rfkill is available prior to using it.

Anyways if you need a quick fix maybe make the author of the feature aware of the issue as he is much more familiar with #568 and the problem domain or look into it yourself and make a pull request because I won't be able to allocate a time slot for this any time soon.

mcepl:
I guess a duplicate of this bug is https://bugzilla.suse.com/show_bug.cgi?id=1183499 (with the backtrace attached).

mcepl:
And it is probably a duplicate of #994.

PR #823, opened 2020-08-14
Rfkill optional

W-joe:
Add a feature to meson config which allows for optionally building rfkill stuffs.

See #822


Alexays:
Nice, thanks for the fix!

W-joe:
No, thank you :)

Issue #799, opened 2020-08-06
If bar height gets very large at some point it never liberates the reserved space


vberger:
While testing some stuff, I started a demo Wayland app which set its title to some very large string, containing several newlines. As a result, waybar grew a lot in height to display the full title, including the newlines, and the space reserved for it by sway grew as well.

Once that demo app was closed, waybar returned to its normal size, but the space reserved for it by sway did not, resulting in a huge empty gap on my screen between the bar and the windows.

Looks like the reserved space in layer shell is only updated when the bar grows, but not when it shrinks?

alebastr:
> Looks like the reserved space in layer shell is only updated when the bar grows, but not when it shrinks?

Looking at the code, that's exactly how it works when gtk-layer-shell is used. The old layer-shell code was relying on `layer_surface.configure` event for that, gtk-layer-shell code path is not using it and `Bar::onConfigure` code doesn't look right.

Need to look how to fix that without affecting fallback layer-shell code.

PR #824, opened 2020-08-15
Fix resizing of the bar with gtk-layer-shell

alebastr:
In the hindsight, it wasn't a good idea to have one set of GTK event handlers for both code paths. Took me a while to remember which code is needed for gtk-layer-shell and which one is for bare layer-surface. Apparently `Bar::onConfigure` wasn't passing the right values to `setExclusiveZone` because it was written with assumption that `layerSurfaceHandleConfigure` already done everything required.
***
At some point GTK stopped listening to `resize`/`set_size_request` if the requested size is insufficient. This is now causing occasional bar size changes and warnings in the log for both code paths.
This would also cause an interesting behavior if `gtk-layer-shell` is disabled and the bar height is a fixed value from the config file; the bar would expand according to a content size, but the exclusive zone size would stay as configured.

Fixes: #799 (only for gtk-layer-shell).
Also fixes xdg_wm_base.version mismatch warning via gtk-layer-shell 0.3.0 update.

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -53,13 +53,18 @@ class Bar {
   static void layerSurfaceHandleClosed(void *, struct zwlr_layer_surface_v1 *);
 
 #ifdef HAVE_GTK_LAYER_SHELL
+  /* gtk-layer-shell code */
   void initGtkLayerShell();
+  void onConfigureGLS(GdkEventConfigure *ev);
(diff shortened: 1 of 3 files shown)


Alexays:
Perfect as always!
Definitely cleaner by splitting the event handlers
Thanks!

PR #828, opened 2020-08-19
Prevent line breaks in ellipsized labels

### LINKED ISSUE #827: Waybar becomes three lines high with a specific title
When visiting this website in Firefox:

https://spectrum.ieee.org/aerospace/aviation/with-ultralight-lithiumsulfur-batteries-electric-airplanes-could-finally-take-off

Waybar becomes three lines high instead of cropping the title to fit the space. The full content is:

```
With Ultralight Lithium-Sulfur Batteries, Electric Airplanes Could Finally Take Off   - IEEE Spectrum
```

But after "Take Off" there are two sequences of 0xE280A9. Looking that up in UTF8 encoding those are:

```
U+2029 | e2 80 a9 | PARAGRAPH SEPARATOR
```
https://www.utf8-chartable.de/unicode-utf8-table.pl

So they're just like `\n` apparently but encoded as unicode.

pedrocr:
If a label is being ellipsized it doesnt make sense to allow it to use line breaks to have multiple lines.

Fixes #827

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -22,8 +22,10 @@ ALabel::ALabel(const Json::Value& config, const std::string& name, const std::st
   if (config_["max-length"].isUInt()) {
     label_.set_max_width_chars(config_["max-length"].asUInt());
     label_.set_ellipsize(Pango::EllipsizeMode::ELLIPSIZE_END);
+    label_.set_single_line_mode(true);
   } else if (ellipsize && label_.get_max_width_chars() == -1) {
     label_.set_ellipsize(Pango::EllipsizeMode::ELLIPSIZE_END);
+    label_.set_single_line_mode(true);
   }
 
   if (config_["rotate"].isUInt()) {


src/ALabel.cpp
```diff
@@ -22,8 +22,10 @@ ALabel::ALabel(const Json::Value& config, const std::string& name, const std::st
   if (config_["max-length"].isUInt()) {
     label_.set_max_width_chars(config_["max-length"].asUInt());
     label_.set_ellipsize(Pango::EllipsizeMode::ELLIPSIZE_END);
+    label_.set_property("single-line-mode", true);
```
alebastr, line 25: gtkmm provides a wrapper for this property, `label_.set_single_line_mode(true)`.

pedrocr: Is there any advantage to doing that? The GtkLabel docs list the properties and not that wrapper, so using `set_property()` seems appropriate. The wrapper doesn't really do anything here.

alebastr: Mostly cosmetic. It's just gtkmm C++ style of doing things vs gtk pure C way.
By calling `set_x` instead of `set_property(x)` you are allowing compiler to check that the property exists (it's exposed by gtkmm bindings generator) and the argument type is correct.

pedrocr: Makes sense. Changed it.

Alexays (approved):

Alexays:
Thanks!

PR #829, opened 2020-08-19
Defensively ignore markup in window titles

pedrocr:
Display the text of window titles without HTML markup interpretation. This is not used by applications and might be a vector for problems in the future.


pedrocr:
Seems like at least firefox sets titles with `&` escape sequences. So this isn't usable as is. Maybe input sanitization is required instead.

PR #830, opened 2020-08-19
Allow customization of battery module tooltip

Markaos:
This commit allows users to change tooltip format using `tooltip-format[-status][-state]` parameters. It also adds `autoTooltip` (I couldn't come up with anything better) format argument containing the "default" tooltip text.

This shouldn't affect existing configs


src/modules/battery.cpp
```diff
@@ -169,22 +169,37 @@ auto waybar::modules::Battery::update() -> void {
...
   // Transform to lowercase  and replace space with dash
   std::transform(status.begin(), status.end(), status.begin(), [](char ch) {
     return ch == ' ' ? '-' : std::tolower(ch);
   });
   auto format = format_;
   auto state = getState(capacity, true);
+  auto time_remaining_formatted = formatTimeRemaining(time_remaining);
+  if (tooltipEnabled()) {
+    std::string tooltip_text_default;
+    std::string tooltip_format = "{autoTooltip}";
```
Alexays, line 182: Not a big fan of the `autoTooltip` naming.
Maybe something like `timeTo`

Alexays (changes requested):
Sorry for the delay, here's a review :)

Markaos:
I agree, `timeTo` sounds like a reasonable name

Alexays:
Thanks!

Alexays:
Could you update github wiki? :)

Markaos:
Sure, I'm on it

PR #832, opened 2020-08-20
Don't roundtrip to escaped text

### LINKED ISSUE #833: sway/window tooltips show HTML escape codes
sway/window tooltips will include things like `&apos;` when there's an apostrophe in the title. For an example see this webpage:

https://www.reddit.com/r/formula1/comments/icnmra/hamilton_hopes_something_positive_comes_up_for/

pedrocr:
GtkLabel can just display text with no escaping, so use that directly instead of first escaping text and then interpreting the escapes.

Fixes #833

### DIFF
--- a/src/modules/sway/window.cpp
+++ b/src/modules/sway/window.cpp
@@ -56,7 +56,7 @@ auto Window::update() -> void {
     bar_.window.get_style_context()->remove_class("solo");
     bar_.window.get_style_context()->remove_class("empty");
   }
-  label_.set_markup(fmt::format(format_, window_));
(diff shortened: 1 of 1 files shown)


Alexays:
With that, we don't allow anymore to pass markup in the format, That's what the set_markup was for, to let the possibility of setting markup in the config

pedrocr:
Replaced by #834 then. The format for sway/window is strange though. As the only thing that can be used in the format is the window title itself, even though the code grabs other stuff. Maybe something that was left midway?

PR #834, opened 2020-08-20
Don't escape sway/window tooltip text

### LINKED ISSUE #833: sway/window tooltips show HTML escape codes

pedrocr:
Tooltip text doesn't accept markup so don't escape the text for that and end up with `&apos;` and other HTML escape codes literally in the tooltip.

Fixes #833

### DIFF
--- a/src/modules/sway/window.cpp
+++ b/src/modules/sway/window.cpp
@@ -56,7 +56,8 @@ auto Window::update() -> void {
     bar_.window.get_style_context()->remove_class("solo");
     bar_.window.get_style_context()->remove_class("empty");
   }
-  label_.set_markup(fmt::format(format_, window_));
+  std::string escaped_window = Glib::Markup::escape_text(window_);
+  label_.set_markup(fmt::format(format_, escaped_window));
   if (tooltipEnabled()) {
     label_.set_tooltip_text(window_);
   }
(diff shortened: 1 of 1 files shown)


Alexays (approved):

src/modules/sway/window.cpp
```diff
@@ -56,7 +56,8 @@ auto Window::update() -> void {
     bar_.window.get_style_context()->remove_class("solo");
     bar_.window.get_style_context()->remove_class("empty");
   }
-  label_.set_markup(fmt::format(format_, window_));
+  std::string escaped_window = Glib::Markup::escape_text(window_);
+  label_.set_markup(fmt::format(format_, escaped_window));
   if (tooltipEnabled()) {
     label_.set_tooltip_text(window_);
```
asas1asas200, line 62: Consider to use `label_.set_tooltip_markup()` instead,  like #1669 ?

Alexays:
Should use something like in https://github.com/Alexays/Waybar/pull/1710

flexibeast:
@Alexays: i'm willing to create a separate PR to fix this issue, if the code involved is straightforward; could you explain what you mean by "something like in #1710"?

PR #840, opened 2020-08-27
Fix "restart-interval" in the custom module

### LINKED ISSUE #621: Process all output lines from modules with `interval` property
Suppose my custom module is based on a long-running script. I would like it to show "loading" state, like so:

```sh
#!/bin/sh

printf '{"text":"loading..."}\n'

sleep 5s # simulates some long-running job...

printf '{"text":"answer: 42"}\n'
```

I want this module to refresh on interval, plus if I tell it so manually.

```
"custom/task": {
  "exec": "waybar-task",
  "return-type": "json",
  "signal": 1,
  "on-click-right": "pkill -RTMIN+1 -x waybar",
  "interval": 1200
}
```

If I set this to refresh on `interval` like above, only "loading..." will show-up in waybar, I will never see "answer: 42".

If I remove the `interval` property, then waybar correctly shows "loading..." and then changes to "answer: 42", but it expects my script to loop in itself. 

Even if I wrap my script in `while true; do .....; sleep 1200; done`, then I would not be able to refresh it manually via `pkill`.

I would expect that waybar processes all lines containing JSON output when I have the `interval` property, not only the first line.

tamirzb:
This pull request fixes the issue where the process would restart immediately and the thread would sleep after the process has restarted, and not before.

Fixes #621

### DIFF
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -50,7 +50,6 @@ void waybar::modules::Custom::continuousWorker() {
   thread_ = [this, cmd] {
     char*  buff = nullptr;
     size_t len = 0;
-    bool   restart = false;
     if (getline(&buff, &len, fp_) == -1) {
       int exit_code = 1;
       if (fp_) {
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
Thanks for the fix!

PR #842, opened 2020-08-28
Implement support for reloading of config files.

### LINKED ISSUE #759: Would it be possible to reload just the config file?
I don't want my screen or bar to do a full reload everytime I change the config. And I keep an executable file to change the config and rerun the Waybar. But it shouldn't flicker like this. I would like just the config file to be updated, which would be faster and would be far less noticeable.
Is it possible?

rdnetto:
Closes #759.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -269,10 +269,9 @@ int waybar::Client::main(int argc, char *argv[]) {
   gtk_app->hold();
   gtk_app->run();
   bars.clear();
-  zxdg_output_manager_v1_destroy(xdg_output_manager);
-  zwlr_layer_shell_v1_destroy(layer_shell);
-  zwp_idle_inhibit_manager_v1_destroy(idle_inhibit_manager);
-  wl_registry_destroy(registry);
-  wl_display_disconnect(wl_display);
   return 0;
+}
+
+void waybar::Client::reset() {
+  gtk_app->quit();
 }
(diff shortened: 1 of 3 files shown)


src/client.cpp
```diff
@@ -269,10 +269,9 @@ int waybar::Client::main(int argc, char *argv[]) {
   gtk_app->hold();
   gtk_app->run();
   bars.clear();
-  zxdg_output_manager_v1_destroy(xdg_output_manager);
-  zwlr_layer_shell_v1_destroy(layer_shell);
-  zwp_idle_inhibit_manager_v1_destroy(idle_inhibit_manager);
-  wl_registry_destroy(registry);
-  wl_display_disconnect(wl_display);
```
rdnetto, line 276: `Gdk::Display::get_default()` is a singleton, and since `wl_display` is retrieved from it, is (and the related components here) are also singletons. This means that if we destroy them we can't recreate them, so in order to execute `Client::main()` multiple times the simplest solution is to just not destroy them.

Since the functions return the same objects every time, this does not result in a memory leak.

Alexays (approved):

Alexays (approved):

Alexays:
Thanks, could you update the wiki?

rdnetto:
> Thanks, could you update the wiki?

Sure thing - I added it to the FAQ.

rieje:
Can you clarify how reloading a configuration is different from restarting waybar? I simply modified an `on-click` setting in the config and reloading the config did not apply the change in setting, only restarting waybar did.

PR #843, opened 2020-08-30
style(workspaces): align text with other modules

tchebb:
Currently, the bottom border on workspace buttons eats into the box size and causes the text to sit higher than in other modules. This is ugly when there are other modules (like the window title) right next to the workspace module. To fix the issue, create the bottom border using an inset box-shadow, which doesn't affect the box's content sizing.

Before:
Before

After:
After


Alexays (approved):

Alexays:
Thanks for updating the default config!

PR #844, opened 2020-08-30
style: add 4px margins to window and workspaces modules

tchebb:
These modules, unlike others, have no horizontal margins by default. This means that they'll appear uncomfortably close together in any config that puts them side-by-side. In general, the default style should make configs with any module ordering look good. Add the same 4px horizontal margins that other module have to these.

This affects rendering of the default config by moving the workspace buttons 4px to the right.

Before:
Before

After:
After

### DIFF
--- a/resources/style.css
+++ b/resources/style.css
@@ -83,6 +83,21 @@ window#waybar.chromium {
     color: #ffffff;
 }
 
+#window,
+#workspaces {
+    margin: 0 4px;
+}
+
+/* If workspaces is the leftmost module, omit left margin */
+.modules-left > widget:first-child > #workspaces {
+    margin-left: 0;
+}
+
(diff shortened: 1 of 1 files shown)


Alexays:
Because for the default config, workspaces is fixed on the left side, so no need, but for the window module it may be necessary. 
Can you update the PR with just applying margin to the window module?

tchebb:
I updated the commit so that if the workspaces module is the leftmost or rightmost module, it won't have margin on that side. This should preserve the existing default appearance but still behave well for people who move the module elsewhere in their config.

tchebb:
@Alexays just making sure you didn't miss my update here, since you responded to all my other PRs yesterday but not this one.

Alexays (approved):

Alexays:
I missed that one 😄 
Thanks for the update!

Issue #845, opened 2020-08-30
Memory module tooltip is wrong and hardcoded


HarHarLinks:
https://github.com/Alexays/Waybar/blob/44119db436da9c06d98d200f1b69411f6bd52359/src/modules/memory/common.cpp#L39

The number is in Gigabytes (or actually more likely, Gibibytes GiB), however it is formatted as Gigabits (Gb).

Further this is hardcoded here and would be nice to be able to manipulate via the usual "tooltip-format" in the config.

Alexays:
Fixed with #587

PR #847, opened 2020-09-02
Add format for memory tooltip

MusiKid:
This commit adds support of format string for memory module tooltip (#845).


Alexays (approved):

Alexays:
Thanks for the fix and the improvement!

PR #848, opened 2020-09-05
feat: optional rfkill

Alexays:


Alexays:
@theduke can you try with these changes?

PR #849, opened 2020-09-06
Add sndio module.

ericonr:
- can control sndio: change volume, toggle mute
- appearance is somewhat dynamic: takes muted status into account
- uses polling inside sleeper thread to update values

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,9 @@ jobs:
       compiler: clang
       env:
       before_install:
-        - sudo pkg install -y gtk-layer-shell gtkmm30 jsoncpp libdbusmenu
+        - export CPPFLAGS+=-isystem/usr/local/include LDFLAGS+=-L/usr/local/lib # sndio
+        - sudo sed -i '' 's/quarterly/latest/' /etc/pkg/FreeBSD.conf
+        - sudo pkg install -y gtk-layer-shell gtkmm30 jsoncpp libdbusmenu sndio
                libfmt libmpdclient libudev-devd meson pulseaudio scdoc spdlog
       script:
         - meson build -Dman-pages=enabled
--- a/src/modules/sndio.cpp
+++ b/src/modules/sndio.cpp
@@ -1,0 +1,167 @@
+#include "modules/sndio.hpp"
+#include <algorithm>
+#include <cstdlib>
+#include <poll.h>
+#include <fmt/format.h>
+
+namespace waybar::modules {
+
+void ondesc(void *arg, struct sioctl_desc *d, int curval) {
+  auto self = static_cast<Sndio*>(arg);
+  if (d == NULL) {
+    // d is NULL when the list is done
+    return;
+  }
+  self->set_desc(d, curval);
+}
+
+void onval(void *arg, unsigned int addr, unsigned int val) {
+  auto self = static_cast<Sndio*>(arg);
+  self->put_val(addr, val);
+}
+
+Sndio::Sndio(const std::string &id, const Json::Value &config)
+    : ALabel(config, "sndio", id, "{volume}%"),
+      hdl_(nullptr),
+      pfds_(0),
+      addr_(0),
+      volume_(0),
+      old_volume_(0),
+      maxval_(0),
+      muted_(false) {
+  hdl_ = sioctl_open(SIO_DEVANY, SIOCTL_READ | SIOCTL_WRITE, 0);
+  if (hdl_ == nullptr) {
+    throw std::runtime_error("sioctl_open() failed.");
+  }
+
+  if(sioctl_ondesc(hdl_, ondesc, this) == 0) {
+    throw std::runtime_error("sioctl_ondesc() failed.");
+  }
+
+  sioctl_onval(hdl_, onval, this);
+
+  pfds_.reserve(sioctl_nfds(hdl_));
+
+  event_box_.show();
+
+  event_box_.add_events(Gdk::SCROLL_MASK | Gdk::SMOOTH_SCROLL_MASK | Gdk::BUTTON_PRESS_MASK);
+  event_box_.signal_scroll_event().connect(
+    sigc::mem_fun(*this, &Sndio::handleScroll));
+  event_box_.signal_button_press_event().connect(
+    sigc::mem_fun(*this, &Sndio::handleToggle));
+
+  thread_ = [this] {
+    dp.emit();
+
+    int nfds = sioctl_pollfd(hdl_, pfds_.data(), POLLIN);
+    if (nfds == 0) {
+      throw std::runtime_error("sioctl_pollfd() failed.");
+    }
+    while (poll(pfds_.data(), nfds, -1) < 0) {
+      if (errno != EINTR) {
+        throw std::runtime_error("poll() failed.");
+      }
+    }
+
+    int revents = sioctl_revents(hdl_, pfds_.data());
+    if (revents & POLLHUP) {
+      throw std::runtime_error("disconnected!");
+    }
+  };
+}
+
+Sndio::~Sndio() {
+  sioctl_close(hdl_);
+}
+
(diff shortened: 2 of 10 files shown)


ericonr:
Hmm I believe this will also require a symbol check, since I'm using functions introduced in version 1.7.0 of sndio.

src/modules/sndio.cpp
```diff
@@ -0,0 +1,167 @@
...
+    int revents = sioctl_revents(hdl_, pfds_.data());
+    if (revents & POLLHUP) {
+      throw std::runtime_error("disconnected!");
+    }
+  };
+}
+
+Sndio::~Sndio() {
+  sioctl_close(hdl_);
+}
```
ericonr, line 75: Does this destructor also kill the thread, or would I need to do it explicitly?

Alexays: Nope should be good :)

ericonr:
@Alexays did this compile for FreeBSD with `sndio` support? I can't see CI results. I have now added `sndio-dev` to the alpine dockerfile as well, to increase the coverage.

src/modules/sndio.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+    while (poll(pfds_.data(), nfds, -1) < 0) {
+      if (errno != EINTR) {
+        throw std::runtime_error("poll() failed.");
+      }
+    }
+
+    int revents = sioctl_revents(hdl_, pfds_.data());
+    if (revents & POLLHUP) {
+      throw std::runtime_error("disconnected!");
```
ericonr, line 68: Due to this, restarting `sndiod` (which should be rare) makes Waybar abort. How would I throw an error that indicates to `waybar` that this module should be turned off?

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  disconnected!
fish: 'waybar' terminated by signal SIGABRT (Abort)
```

ericonr: It could also try reconnecting...

.travis.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       compiler: clang
       env:
       before_install:
-        - sudo pkg install -y gtk-layer-shell gtkmm30 jsoncpp libdbusmenu
+        - sudo pkg install -y gtk-layer-shell gtkmm30 jsoncpp libdbusmenu sndio
```
jbeich, line 32:
```suggestion
        - export CPPFLAGS=-isystem/usr/local/include LDFLAGS=-L/usr/local/lib # sndio
        - sudo pkg install -y gtk-layer-shell gtkmm30 jsoncpp libdbusmenu sndio
```

jbeich (commented):
> @Alexays did this compile for FreeBSD with sndio support? I can't see CI results.

Looking at https://travis-ci.org/github/Alexays/Waybar/pull_requests

> Library sndio found: NO

Adding `CPPFLAGS=-isystem/usr/local/include` + `LDFLAGS=-L/usr/local/lib` leads to

> Library sndio found: YES
> Checking for function "sioctl_open" with dependency -lsndio: YES

Seems to build fine after that.

.travis.yml
ericonr: Why not use `-I`?

Thanks for testing and for the link to CI results! Do you use `sndio`? Would be nice to get opinions on usability and such.

jbeich: > Why not use `-I`?

`-isystem` puts include path after `-I` (avoiding conflicts with local headers) and silences warnings (that cannot be fixed by this project).

See also https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html

> Do you use `sndio`? Would be nice to get opinions on usability and such.

No. Maybe @t6 can help.

ericonr: I see, thanks!

jbeich:
From https://travis-ci.org/github/Alexays/Waybar/jobs/727586703
> [88/97] Installing sndio-1.6.0...
[...]
> Library sndio found: YES
> Checking for function "sioctl_open" with dependency -lsndio: NO 

Until 2020-10-10 you may need to revert #772 (e.g., [commit](https://github.com/Alexays/Waybar/commit/81ae5ba3a85939cd62aec4e76b47cee456e51b8d), [build](https://travis-ci.com/github/jbeich/Waybar/jobs/385412638)). https://github.com/freebsd/freebsd-ports/commit/68eb5c3a71f59793b00a22f9390d84e4375c8c7f is missing on 2020Q3 but 2020Q4 will be branched in ~2 weeks + ~5 days to build due to many updates.

ericonr:
I see, thanks. I guess I will revert to get it to build at least once in CI, then I can revert it.

ericonr:
@jbeich thanks a lot, it did build with `sndio` support now: https://travis-ci.org/github/Alexays/Waybar/jobs/727695954

I think now that a job has run I should move back to `/quarterly`, right?

ericonr:
Rebased on latest.

src/modules/sndio.cpp
Alexays: Go for implementing reconnecting if it's not too complex.

Alexays:
Beside the reconnection part, it looks perfect!
Waiting for the first point.
Thanks!

src/modules/sndio.cpp
ericonr: I will have to look into how to do that without blocking. Will let you know how it goes.

ericonr:
@Alexays thanks for the incentive, reconnection now works! It should be good to go, then.

src/modules/sndio.cpp
```diff
@@ -20,27 +21,28 @@ void onval(void *arg, unsigned int addr, unsigned int val) {
...
   hdl_ = sioctl_open(SIO_DEVANY, SIOCTL_READ | SIOCTL_WRITE, 0);
   if (hdl_ == nullptr) {
     throw std::runtime_error("sioctl_open() failed.");
   }
 
-  if(sioctl_ondesc(hdl_, ondesc, this) == 0) {
-    throw std::runtime_error("sioctl_ondesc() failed.");
-  }
-
+  sioctl_ondesc(hdl_, ondesc, this);
```
Alexays, line 30: Shouldn't we keep the check here?

src/modules/sndio.cpp
```diff
@@ -65,7 +67,23 @@ Sndio::Sndio(const std::string &id, const Json::Value &config)
 
     int revents = sioctl_revents(hdl_, pfds_.data());
     if (revents & POLLHUP) {
-      throw std::runtime_error("disconnected!");
+      spdlog::warn("sndio disconnected!");
+      sioctl_close(hdl_);
+      hdl_ = nullptr;
+
+      // reconnection loop
+      while (1) {
```
Alexays, line 75: Could you replace with `while (thread_.isRunning())`?

Alexays:
Two small comments and then it's good.

src/modules/sndio.cpp
ericonr: I was a bit unclear on this function's failure modes, but a better look at the source confirmed that we should indeed check for errors. I added a check for `sioctl_onval` too.

src/modules/sndio.cpp
ericonr: Just did.

src/modules/sndio.cpp
```diff
@@ -0,0 +1,198 @@
...
+      sioctl_close(hdl_);
+      hdl_ = nullptr;
+
+      // reconnection loop
+      while (thread_.isRunning()) {
+        try {
+          connect_to_sndio();
+        } catch(std::runtime_error const& e) {
+          // avoid leaking hdl_
+          if (hdl_) sioctl_close(hdl_);
```
ericonr, line 85: Added this.

Alexays:
Thanks!

ericonr:
Sure! Thanks :D

PR #850, opened 2020-09-06
Fix non-standard usage of <sys/poll.h>

z3ntu:
Fixes the following build warning with musl libc:
```
In file included from ../src/util/rfkill.cpp:24:
/usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
    1 | #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
      |  ^~~~~~~
```


Alexays (approved):

Alexays:
Thanks!

PR #851, opened 2020-09-06
Add an "exec-on-event" config for the custom module

### LINKED ISSUE #841: Able to disable execution on click/scroll for custom module
Currently, a custom module that has click and/or scroll events configured, would be re-executed on a click/scroll accordingly. At first I thought this was a bug, but looking at commit ff28de048289c21aa54ff88529f21823947f6ac6, it seems that this was intentional.

Now, I wonder why? I think that in various configs, this isn't necessary (for instance, the github notifications and pacman configs from the wiki). Also given how the signal config works, this can be done by adding a command that sends a signal to an event.

Is it possible to change this behavior? Perhaps if you really don't want to change the current behavior, maybe add a config that disables this?

tamirzb:
This pull request adds a new config for the custom module: `exec-on-event`.
This config allows disabling the default behavior of re-executing the script whenever an event that has a command set is triggered.

Fixes #841

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -22,6 +22,7 @@ class Custom : public ALabel {
   void continuousWorker();
   void parseOutputRaw();
   void parseOutputJson();
+  void handleEvent();
   bool handleScroll(GdkEventScroll* e);
   bool handleToggle(GdkEventButton* const& e);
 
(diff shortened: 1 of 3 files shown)


Alexays:
Thanks, can you update the wiki? ❤️

tamirzb:
Sure, done.

PR #860, opened 2020-09-21
Add module "sway/hide" for swaybar integration

### LINKED ISSUE #255: Only show waybar when modifier key is pressed
Is there any way to configure sway/waybar so that it behaves similar to i3's [mode hide](https://i3wm.org/docs/userguide.html#_display_mode)?

Sway already supports `mode hide` for their `swaybar`. I was wondering if it would be possible to implement the same for `waybar`?

Basically what I'm looking for is having this sway configuration work as expected:
```
bar {
  mode hide
  swaybar_command waybar
}
```

somini:
Rebases and continues work by @xPMo in https://github.com/xPMo/Waybar/tree/sway-hide

This allows auto-showing the bar when the modifier is pressed
    
Closes #255
    
Setup instructions:
- Set the `mode` of the bar to "hide" in sway configuration file
 - Add "sway/hide" into `modules-left` in Waybar configuration file

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -260,7 +260,7 @@ void waybar::Bar::onMap(GdkEventAny* ev) {
 
 void waybar::Bar::setExclusiveZone(uint32_t width, uint32_t height) {
   auto zone = 0;
-  if (visible && config["layer"] != "overlay") {
+  if (visible) {
     // exclusive zone already includes margin for anchored edge,
     // only opposite margin should be added
     if (vertical) {
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -389,8 +389,8 @@ void waybar::Bar::layerSurfaceHandleClosed(void* data, struct zwlr_layer_surface
   o->modules_right_.clear();
 }
 
-auto waybar::Bar::toggle() -> void {
-  visible = !visible;
+auto waybar::Bar::setVisible(bool nvis) -> void {
+  visible = nvis;
   if (!visible) {
     window.get_style_context()->add_class("hidden");
     window.set_opacity(0);
@@ -402,6 +402,10 @@ auto waybar::Bar::toggle() -> void {
   if (!use_gls_) {
     wl_surface_commit(surface);
   }
+}
+
+auto waybar::Bar::toggle() -> void {
+  return setVisible(!visible);
 }
 
 void waybar::Bar::getModules(const Factory& factory, const std::string& pos) {
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -260,7 +260,7 @@ void waybar::Bar::onMap(GdkEventAny* ev) {
 
 void waybar::Bar::setExclusiveZone(uint32_t width, uint32_t height) {
   auto zone = 0;
-  if (visible) {
+  if (visible && config["layer"] != "overlay") {
     // exclusive zone already includes margin for anchored edge,
     // only opposite margin should be added
     if (vertical) {
--- a/src/factory.cpp
+++ b/src/factory.cpp
@@ -22,6 +22,9 @@ waybar::AModule* waybar::Factory::makeModule(const std::string& name) const {
     if (ref == "sway/window") {
       return new waybar::modules::sway::Window(id, bar_, config_[name]);
     }
+    if (ref == "sway/hide") {
+      return new waybar::modules::sway::Hide(id, bar_, config_[name]);
+    }
 #endif
 #ifdef HAVE_WLR
     if (ref == "wlr/taskbar") {
--- a/src/modules/sway/hide.cpp
+++ b/src/modules/sway/hide.cpp
@@ -5,9 +5,12 @@
 
 namespace waybar::modules::sway {
 
+std::string current_mode = "hide";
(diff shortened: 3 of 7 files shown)


src/modules/sway/hide.cpp
```diff
@@ -0,0 +1,45 @@
...
+    : ALabel(config, "hide", id, "{}", 0, true), bar_(bar), windowId_(-1) {
+  ipc_.subscribe(R"(["bar_state_update","barconfig_update"])");
+  ipc_.signal_event.connect(sigc::mem_fun(*this, &Hide::onEvent));
+  bar_.window.get_style_context()->add_class("hidden");
+  // Launch worker
+  worker();
+}
+
+void Hide::onEvent(const struct Ipc::ipc_response& res) {
+  auto payload = parser_.parse(res.payload);
```
Alexays, line 19: Add a mutex here, might fix your crash

src/factory.cpp
```diff
@@ -22,6 +22,9 @@ waybar::AModule* waybar::Factory::makeModule(const std::string& name) const {
     if (ref == "sway/window") {
       return new waybar::modules::sway::Window(id, bar_, config_[name]);
     }
+    if (ref == "sway/hide") {
+      return new waybar::modules::sway::Hide(id, bar_, config_[name]);
```
alebastr, line 26: Should that be a module?
Maybe we can implement it as a generic SwayBarConfigHandler class that is created by the `waybar::Client` instance when the SWAYSOCK is defined and controls all the bars.
The benefit is that on multi-monitor configurations bar events would be handled once.

src/bar.cpp
```diff
@@ -260,7 +260,7 @@ void waybar::Bar::onMap(GdkEventAny* ev) {
 
 void waybar::Bar::setExclusiveZone(uint32_t width, uint32_t height) {
   auto zone = 0;
-  if (visible) {
+  if (visible && config["layer"] != "overlay") {
```
alebastr: `sway-bar(5)`:
> (in overlay mode the bar is transparent to input events)

I wonder if it makes sense to support overlay without that part...

Also, if we're speaking about `bar mode overlay`, I assume that the layer should be 'top'. Need to check swaybar code to confirm though.
Otherwise we will see waybar on top of the lockscreen or fullscreen windows and that's not nice

src/modules/sway/hide.cpp
```diff
@@ -0,0 +1,45 @@
...
+  if (payload.isMember("mode")) {
```
alebastr, line 21: Note that each waybar instance supposed to have bar id (`waybar -b bar-0`) and events from sway could be addressed to the specific bar or to all bars.
Somewhere around here you should check if the current bar needs to handle the event.

src/modules/sway/hide.cpp
somini: My C++ knowledge is almost 0, I just copy pasted some bits from other places that looked correct. :disappointed: 

2253f3e0a27be97a2b9d2e1a4e987dba530846e5 crashes less, but it still fails:

```
**
Gtk:ERROR:../gtk/gtk/gtkcssinheritvalue.c:33:gtk_css_value_inherit_free: code should not be reached
**
Gtk:ERROR:../gtk/gtk/gtkcssinheritvalue.c:33:gtk_css_value_inherit_free: code should not be reached
Bail out! Gtk:ERROR:../gtk/gtk/gtkcssinheritvalue.c:33:gtk_css_value_inherit_free: code should not be reached
Bail out! Gtk:ERROR:../gtk/gtk/gtkcssinheritvalue.c:33:gtk_css_value_inherit_free: code should not be reached
Aborted (core dumped)
```
Sounds like hiding a painting bar is a no-no.

I think the mutex should be on the `setVisible` module for the bar itself, but waiting for the first paint. Is there some kind of event queue where we can add the hide/show events, instead of being done ASAP?

somini: Rebased this to the lastest master and force-pushed

alebastr (commented):
I can try to rework `waybar::Bar` in order to provide methods to apply certain parts of the bar_config object, but that'll have to wait until the weekend.

Switching to [`Gio::Socket`](https://developer.gnome.org/glibmm/stable/classGio_1_1Socket.html) in the main loop might resolve all threading issues. But that's a lot of refactoring for sway ipc code...

src/bar.cpp
somini: Yes, I removed this on the latest code. One thing at a time.

We can support only `mode "hide"` for now.

src/modules/sway/hide.cpp
somini: I was going to use just call `setVisible` on the  `bar_` object, but I got an error about mismatching `const` which is above my meagre C++-fu.

somini:
@alebastr Thanks for the review. From the GTK logs, a mutex on the `setVisible` method on the bar itself might work, as long as we can ignore event that appear during the transitions (i.e. when the mutex is locked). Sort of like a rate-limiter. It's better to ignore events than crashing.

I don't know enough C++ to do much more than this, but I'm available for testing.

somini:
Continued in #1241

PR #862, opened 2020-09-21
Add state to disk module

xenrox:
I prefer to hide my disk information until only a certain percentage is left. This would implement that functionality in a way like the CPU/memory modules use `state`.


src/modules/disk.cpp
```diff
@@ -73,6 +73,7 @@ auto waybar::modules::Disk::update() -> void {
       ));
   }
   event_box_.show();
+  getState((stats.f_blocks - stats.f_bavail) * 100 / stats.f_blocks);
```
Alexays, line 76: Could save the operation into a variable in order to avoid computing it twice 😄

Alexays (changes requested):

src/modules/disk.cpp
xenrox: Done 😃

Alexays (approved):

Alexays:
Thanks!
Can you update the GitHub wiki? ❤️

xenrox:
> Can you update the GitHub wiki? ❤️

Done as well. Thanks for your work!

PR #868, opened 2020-09-26
Fix "on-click-backward" when "on-click-forward" is not present

danieldg:


Alexays (approved):

Alexays:
Thanks for the fix Daniel!

PR #869, opened 2020-09-26
Add format-icons for workspace's name entry in sway/workspaces module

lrhel:
Hello,
Little hack to have workspace name like `2:Firefox` (to get the default sway workspace bindsym working) with an associated icon without the need to repeat 10 times the line in the configuration file for each case.
~~thanks for the bar~~


Alexays (approved):

Alexays:
Thanks!

Issue #714, opened 2020-05-22
XDG_CONFIG_DIRS not respected


betaboon:
i think this #425 introduced a regression on handling config-file-paths.

if i understand this correctly:

before #425 `/etc/xdg/waybar/config` was checked (which seems to have been a wrong implementation of XDG_CONFIG_DIRS in the first place)

according to [the specification](https://specifications.freedesktop.org/basedir-spec/0.6/ar01s03.html) the order of precendence should be:
- $XDG_CONFIG_HOME
- $XDG_CONFIG_DIRS (maintaining its order, and defaulting to `/etc/xdg` if empty

jbeich:
> before #425 `/etc/xdg/waybar/config` was checked

Should still be true for most Linux distributions:
```
$ meson --prefix=/usr build
$ ninja -C build
$ strings _build/waybar | fgrep etc/xdg
//etc/xdg/waybar/style.css
//etc/xdg/waybar/config
```

jbeich:
> `man 5 waybar` doesnt reflect the change introduced in #425 either. it still states that `/etc/xdg/waybar/config` is checked

This is similar to sway which hardcodes `/etc/sway/config` in manpages. Less preprocessing makes it easier to edit documentation.

betaboon:
@jbeich yes it should still be true for every distribution that builds waybar (and sway) with the default prefix of `/`. in nixos we don't, thus forcing us to maintain patches that check `/etc/xdg/waybar` instead of `$prefix/etc/xdg/waybar` (as we currently do for sway).
to avoid this i am proposing to fully support the base-dir-spec linked to above, and thus supporting $XDG_CONFIG_DIRS in addition to $XDG_CONFIG_HOME.

jbeich:
> in nixos we don't, thus forcing us to maintain patches that check `/etc/xdg/waybar`

Why do you need to patch what is supported Meson out of the box?
```
$ meson --prefix=/whatever --sysconfdir=/etc _build
$ ninja -C _build
$ strings _build/waybar | fgrep etc/xdg
//etc/xdg/waybar/style.css
//etc/xdg/waybar/config
```

betaboon:
as far as i understand the `meson.build` does
```
add_project_arguments('-DSYSCONFDIR="/@0@"'.format(join_paths(prefix, sysconfdir)), language : 'cpp')
```
[...]

PR #870, opened 2020-09-29
Add back unprefixed global config directory

minijackson:
This still allows loading the installed default config, but also add back the path `/etc/xdg/waybar` unconditionally for the people who were using it as their configuration directory and have a "prefixed" waybar installation. Should fix https://github.com/Alexays/Waybar/issues/714 and https://github.com/NixOS/nixpkgs/issues/99057

Also added the sysconfdir-prefixed path in the man page to prevent some confusion. The meson code is kinda messy, sorry about that. If there's a better way to do this, please do tell.


Alexays (approved):

Alexays:
Thanks!

PR #873, opened 2020-10-02
Use 5 Moon emoji icons for the backlight module

chaitunya:
This is a simple commit which adds the backlight icons that [polybar uses](https://github.com/polybar/polybar/wiki/Module:-backlight#additional-formatting) to the default config. These emojis have a scale from 0-4 (4 being full), so it provides greater accuracy than the simple 0-1 scale that was there before.


Synthetica9:
Shouldn't this array be inversed?

syntezoid:
@Synthetica9 Yup, you're right.

PR #874, opened 2020-10-04
[RFC] Put the generic configuration options in a separate page, to allow for shorter module specific man pages.

ericonr:


Alexays:
This should also be done for the GitHub wiki 😃

man/waybar-sway-workspaces.5.scd
```diff
@@ -65,10 +65,6 @@ Addressed by *sway/workspaces*
     default: empty ++
     Lists workspaces that should always be shown, even when non existent
 
-*on-update*: ++
```
Alexays, line 68: Since sway/workspaces is not a Label module,  it has no common config, should we keep it here?

ericonr: Updated to keep this config.

ericonr:
Sorry for the delay! This is now updated with the new pages.

PR #877, opened 2020-10-07
mpd: revamped to event-driven, single-threaded

jbenden:
Fix MPD connection issues by converting/rewriting module into a
state-machine driven system. It is fully single-threaded and uses
events for transitioning between states. It supports all features
and functionality of the previous MPD module.

Has additionally been running without problems, locally, for about a week.


Alexays:
Ohh, thanks!
Give me some days to review it 😃

include/modules/mpd/mpd.hpp
```diff
@@ -0,0 +1,66 @@
...
+  void checkErrors(mpd_connection* conn);
+  void fetchState();
+  void queryMPD();
+
+  inline bool stopped() const { return connection_ && state_ == MPD_STATE_STOP; }
+  inline bool playing() const { return connection_ && state_ == MPD_STATE_PLAY; }
+  inline bool paused() const { return connection_ && state_ == MPD_STATE_PAUSE; }
+};
+
+#if !defined(MPD_NOINLINE)
```
Alexays, line 62: Can you explain the purpose of this define?

jbenden: This is defined is here and in a cpp file.

The purpose is it serves as a way to have functions inline or to have them be all in a single translation unit.

So when `-Os` is used, one would probably want `-DMPD_NOINLINE` to help keep size decreased.

Alexays:
LGTM, big thanks!

Alexays:
Reverted as CI not green :/
https://travis-ci.com/github/Alexays/Waybar/jobs/401473971

PR #881, opened 2020-10-12
Support format-{state} for cpu/disk/memory

olemartinorg:
According to the [wiki page on states](https://github.com/Alexays/Waybar/wiki/States), the following is true:

- Some modules support 'states'
- [E]ach state can have its own `format`. Those can be configured via `format-<name>`.

I sadly found this not to be true when I wanted to hide the CPU, Disk and Memory modules whenever my states are 'good'. The text here seems to have been copied directly from the Battery module wiki page, even though no other modules supported it. I'll update the wiki to reflect these changes if this gets merged.


Alexays (approved):

Alexays:
Thanks!

PR #884, opened 2020-10-14
enable-bar-scroll: only switch workspaces with the mouse wheel if over the background or workspaces module

emmanueltouzery:
waybar is great! One little nitpick that I have is that I like to enable the scroll wheel to change workspace AND ALSO to change the volume. Both options are available for waybar!
  "enable-bar-scroll": true is available for the workspace switching, and the pulseaudio module does it out of the box.

 However that means that if we enable the bar-scroll feature then when we're over the pulseaudio module, we'll do both: switch workspace AND change the volume. In effect we can't anymore change the volume using the mouse wheel, because of the side-effect of switching the workspace.

What this PR does is modify enable-bar-scroll to only have an effect on the "background" of waybar: so enable-bar-scroll doesn't have any effect on top of waybar modules (which obviously includes the pulseaudio module). The workspaces module is handled specially so that the effect works there too.

I tried a little to make the scroll wheel work everywhere except on modules that actually use the scroll wheel (like pulseaudio), but I didn't manage to make it work. I think this is good enough and pretty intuitive.

Thank you again for waybar!


Alexays (approved):

Alexays:
Sorry for the delay 😢 
Can you try to fix conflict?
Then i'll merge the PR 🎉

emmanueltouzery:
great, i'll fix the conflicts tonight!

emmanueltouzery:
It should be OK now!

alebastr:
I thought scroll bubbling was finally fixed in GTK 3.24.13 a year ago. The described issue is not reproducible for me with the latest GTK3 and the changes here would make 'enable-bar-scroll' less usable for other people who don't have this bug. Would it make sense to make the changed behavior conditional?

On the code itself: since I don't have this bug I can't really test, but is it possible to use [`Gtk::Widget::intersect`](https://developer.gnome.org/gtkmm/stable/classGtk_1_1Widget.html#acb010d9cbff38078be998d617e61c0fb) with rectangle { x, y, 1, 1 } to check if the pointer coordinates belong to the module area?

emmanueltouzery:
the issue is still reproducable for me with the waybar shipped by fedora 33, meaning 3.24.23-1.
I'm surprised things work OK for you.

I don't know if intersects brings much. There are three boxes, left, center and right. We check if the cursor if within any of these. Seems to be as simple as it's going to be with the current code.

alebastr:
> the issue is still reproducible for me with the waybar shipped by fedora 33, meaning GTK 3.24.23-1 and waybar 0.9.4.

Then it's really surprising. I'm also on Fedora 33 and I just checked that scrolling over pulseaudio module does not trigger workspace change. I wonder what's the difference.

> I don't know if intersects brings much. There are three boxes, left, center and right. We check if the cursor if within any of these. Seems to be as simple as it's going to be with the current code.

Center box requires more careful handling, because it's not really in the center. It's constrained between left and right which could have different widths. Intersect could let you avoid calculating the real position of center box.

emmanueltouzery:
my config is there: https://github.com/emmanueltouzery/dotfiles/tree/master/.config/waybar
nothing much special i guess.

emmanueltouzery:
> Center box requires more careful handling, because it's not really in the center. It's constrained between left and right which could have different widths. Intersect could let you avoid calculating the real position of center box.

good point, I didn't realize that. Let's see what you find out maybe, how come I have the problem and not you, and if we decide this change still makes sense, then I can change that.

alebastr:
Oh... TIL that GTK handles mouse and touchpad scroll differently :scream:

Mouse wheel generates both smooth scroll and regular scroll events. EventBox (pulseaudio module) receives only events with direction `GDK_SCROLL_SMOOTH` and windows receives both (or only regular if smooth is handled by module).

Ignoring events with `e->direction != GDK_SCROLL_SMOOTH` in `Workspace::handleScroll` seems to solve the issue, but I'm not sure what are downsides of this. Need to look deeper into Gdk wayland backend to check if smooth events are always generated for mouse wheel.

emmanueltouzery:
I believe #935 is a superior fix for this issue, closing this PR.

PR #887, opened 2020-10-18
mpd: revamped to event-driven, single-threaded v2

jbenden:
Fix MPD connection issues by converting/rewriting module into a
state-machine driven system. It is fully single-threaded and uses
events for transitioning between states. It supports all features
and functionality of the previous MPD module.

v2 change:
* fixes fmt header includes.


jbenden:
Sorry about the last submission; but an unfortunate sequence of events occurred. The CI builds weren't working. I have to locally make fmt changes. The two didn't mix well. :)

jbenden:
Also, not really sure why the CI says 1/2 complete, but there is only 1 over on Travis.

Alexays:
Ok thanks, we have a legacy Travis CI that i have to remove, don't worry!

PR #888, opened 2020-10-19
mpd: support password protected MPD

jbenden:
I saw an unfinished draft pull-request which added MPD password support; so, I went ahead and completed it. It works locally, with and without a password (but not without password when MPD requires one.)

- Add MPD module option `password`, and document it.
- Add logic to send the password, directly after connecting to
  MPD.

Fixes: #576

### DIFF
--- a/include/modules/mpd/mpd.hpp
+++ b/include/modules/mpd/mpd.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <mpd/client.h>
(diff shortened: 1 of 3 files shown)


Alexays:
Perfect, you rock!
Can you update the Github Wiki?

jbenden:
Thanks!

PS: I've updated the wiki!

PR #892, opened 2020-10-22
Refactor layer surface implementations (again)

alebastr:
An attempt to update the bar code for #860 that went overboard. I'm pretty sure that something is broken in this branch, just haven't figured out what. I'll be keeping this as a draft until I'm convinced that the only collateral damage is logs.

`GLSSurfaceImpl` works fine for me so far, which is not surprising with how small it is after letting `gtk-layer-shell` manage exclusive zone. `RawSurfaceImpl` is not so extensively tested, but I'm going to switch to it for next few days and try all the configuration changes possible via IPC.

***
The end goal is a live configuration reload and updates from sway IPC events, but the path is long (https://github.com/alebastr/Waybar/compare/master...swaybar-ipc is a somewhat ugly proof of concept for the IPC part).
It might also require some feedback and breaking changes as I don't know how close we want to be to swaybar behavior and how good is the idea to apply all supported configuration from sway's `barconfig_update` event.

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -36,11 +36,11 @@ class BarSurface {
   BarSurface() = default;
 
  public:
-  virtual void set_exclusive_zone(bool enable) = 0;
-  virtual void set_layer(const std::string_view &layer) = 0;
-  virtual void set_margins(const struct bar_margins &margins) = 0;
-  virtual void set_position(const std::string_view &position) = 0;
-  virtual void set_size(uint32_t width, uint32_t height) = 0;
+  virtual void setExclusiveZone(bool enable) = 0;
+  virtual void setLayer(const std::string_view &layer) = 0;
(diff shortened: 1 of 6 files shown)


alebastr:
Okay, I haven't found any *new* issues with both implementations. Guess it's ready for review and wider testing...
I could split surface implementations into new files or move them after the Bar code if you think it'll make things cleaner.

This PR also supersedes #545

Alexays:
Awesome stuff @alebastr!
The only thing that bothers me is the snake case on the name of the functions instead of the camelca.
Otherwise, I am very happy with this refacto! 🎉

alebastr:
> The only thing that bothers me is the snake case on the name of the functions instead of the camelca.

Sorry, that wasn't conscious. Been working with different code style too much :smile: 
Fixed and rebased.

Alexays:
Tested and LGTM, thanks for this refacto so far.
❤️

PR #895, opened 2020-10-23
Fix various mpd bugs

Flakebi:
- Add elapsedTime and totalTime to tooltip format arguments
- Catch format exceptions and print error
- Copy mpd connection error message before it gets freed
- Update display after connection to mpd was lost


Alexays (approved):

Alexays:
Thanks!

PR #898, opened 2020-10-25
fix(wlr/taskbar): do not bind to unsupported protocol versions

### LINKED ISSUE #897: Waybar doesn't start anymore
Gnome 3.38
Wayfire version 0.5.0-e92f43a
Wlroots-git 0.11.0.r69.g79c7591d-1

> waybar
[2020-10-24 07:43:32.979] [info] Resources files: /home/bluebyt/.config/waybar/config, /home/bluebyt/.config/waybar/style.css
[2020-10-24 07:43:58.012] [warning] Using different foreign toplevel manager protocol version: 3
[2020-10-24 07:43:58.018] [warning] For a functional tray you must have libappindicator-* installed and export XDG_CURRENT_DESKTOP=Unity
Gdk-Message: 07:43:58.036: Error 2 (No such file or directory) dispatching to Wayland display.
wc: write error

alebastr:
It's not allowed to bind to a higher version of a wayland protocol than supported by the client. Binding wlr-foreign-toplevel-manager-v1 v3 to a generated code for v2 causes errors in libwayland due to a missing handler for `zwlr_foreign_toplevel_handle_v1.parent` event.

This PR just allows to run waybar on the compositors with newer protocols. Someone else will have to update the protocol file, read https://github.com/swaywm/wlr-protocols/pull/52 and add `zwlr_foreign_toplevel_handle_v1.parent` event support for detecting window stacks and allowing to ignore child windows in the taskbar. Wayfire git master already supports sending the event and could be used for testing.

Note: I'm using interface version instead of the constant `2` because I'm assuming that whoever updates the protocol file would  add a handler for the `parent` event in the same PR. Doing only the former would make waybar crash again.

Fixes #897

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -32,6 +32,8 @@ void waybar::Client::handleGlobal(void *data, struct wl_registry *registry, uint
                                   const char *interface, uint32_t version) {
   auto client = static_cast<Client *>(data);
   if (strcmp(interface, zwlr_layer_shell_v1_interface.name) == 0) {
+    // limit version to a highest supported by the client protocol file
+    version = std::min<uint32_t>(version, zwlr_layer_shell_v1_interface.version);
     client->layer_shell = static_cast<struct zwlr_layer_shell_v1 *>(
         wl_registry_bind(registry, name, &zwlr_layer_shell_v1_interface, version));
   } else if (strcmp(interface, zxdg_output_manager_v1_interface.name) == 0 &&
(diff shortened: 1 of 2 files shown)


Alexays (approved):

Alexays:
Thanks for the fix Aleksei!

bluebyt65:
Everything solved! Thanks that was fast!

PR #901, opened 2020-10-28
Replace lowercase "k" with uppercase "K" to make it look more consistent

1sixth:
Other units are all uppercased, so using an uppercased "K" makes it look more consistent (especially when `{bandwidthUpBits}` or something like that is used).

### DIFF
--- a/include/util/format.hpp
+++ b/include/util/format.hpp
@@ -42,7 +42,7 @@ namespace fmt {
 
       template<class FormatContext>
         auto format(const pow_format& s, FormatContext &ctx) -> decltype (ctx.out()) {
-          const char* units[] = { "", "k",  "M",  "G",  "T",  "P",  nullptr};
+          const char* units[] = { "", "K",  "M",  "G",  "T",  "P",  nullptr};
 
           auto base = s.binary_ ? 1024ull : 1000ll;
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
Thanks!

HarHarLinks:
@Alexays please revert this commit, it is wrong as explained by berbiche.

1sixth:
Sorry about it, I mixed up [kilobit](https://en.wikipedia.org/wiki/Kilobit) and [kibibit](https://en.wikipedia.org/wiki/Kibibit).

HarHarLinks:
> Sorry about it, I mixed up [kilobit](https://en.wikipedia.org/wiki/Kilobit) and [kibibit](https://en.wikipedia.org/wiki/Kibibit).

While the capital K (and the other respective capital letters) are [sometimes used to denote the power of 2 variants](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units), this is in my opinion bad since it is easily confused with the SI-prefixes (or wrongly used SI-prefixes in case of K) and thus better to use the KiB notation for these.

Note @Alexays that I have not verified what these units actually correspond to.
Thanks for the swift reaction and your work, I enjoy waybar very much. :)

Issue #668, opened 2020-04-13
Drop date dependency and use C++20 (once compilers support it)


bisco2:
Hi,

according to https://github.com/HowardHinnant/date/issues/565 C++20 seems to implement the functionality of the `date` library, that Waybar uses. Most compilers don't fully implement C++20 yet, but when they do, it would probably make sense to drop the dependency.

(For transparency: I've uploaded releases of `waybar` to Debian in the past, but at the moment I'm blocked by the missing `date` dependency. I looked into packaging `date`, but given the fact that its functionality will be provided by C++20 I'm not sure it makes much sense.)

Alexays:
I'll go 100% on C++20 once possible!

nschloe:
I've worked on waybar for debian and definitely can say that the dependency on hhdate hinders waybars adoption. It'd help waybar's distribution if the dependency on hhdate was optional, even if that costs timezone support.

b1rger:
Just for the record, I've uploaded `waybar` to Debian unstable, but for now it is compiled without the `clock` module. I have added a README to the Debian package, pointing out that the `clock` module is missing and that some of its functionality can be replaced by the `custom` module.

nschloe:
I've packages hhdate _and_ waybar to get the full functionality. [Here's the PPA](https://launchpad.net/~nschloe/+archive/ubuntu/waybar). FWIW it might be ready for inclusion into debian.

diraol:
Hi @b1rger can you give me a hand setting up the clock using the `custom` module, please?

I've done it this way:

```
{
    "layer": "top",
    "position": "top",
    "height": 26,
    "modules-left": [
    ],
    "modules-center": [
        "custom/clock"
    ],
    "modules-right": [],
    "custom/clock": {
        "exec": "date '+%Y-%m-%d %H:%M'",
        "format": "{}",
        "interval": "45"
    }
}
```

It almost work, but once the clock is initialized it does not update anymore, even though having the `interval` set. I am not sure what I'm doing wrong here. :disappointed: Any tips?

b1rger:
> Hi @b1rger can you give me a hand setting up the clock using the custom module, please?

I'm not sure, maybe drop the `format`?
This works for me:
```
    "custom/clock": {
        "exec": "date +'%H:%M %S'",
        "interval": 5
    },
```

jacksgt:
[...]

PR #903, opened 2020-10-29
Add simpleclock as fallback when hhdate is not available

spk:
ref https://github.com/Alexays/Waybar/issues/668

This will allow to have the clock module (before hhdate was introduce) when its not available (Debian disable subprojects download)

wdyt?


Alexays (approved):

Alexays:
I'm not a big fan of that, but it's a needed workaround.

Issue #890, opened 2020-10-20
Custom module updating frequently breaks all tooltips


JordanL2:
Using latest Waybar / Sway on Arch.

With GTK Layer Shell enabled.

I added a custom module that runs an endlessly looping script. I've noticed that if the script outputs an update frequently, say every 0.5s, all Waybar tooltips stop working and won't appear when I hover the mouse over anything. Changing the update frequency to every 0.6s fixes it. This happens even if the script doesn't output a tooltip and the module is customed with "tooltip": "false".

Example config:
```
    "custom/cpu": {
        "exec": "/home/jordan/.config/waybar/cpu_graphical 0.5",
        "format": " {} ",
        "tooltip": true,
        "return-type": "json"
    },
```

Script: https://github.com/JordanL2/LinuxSetup/blob/master/Sway/config/waybar/cpu_graphical

JordanL2:
I'm guessing this is because tooltips have a delay before they appear, and the module updating resets the timer.

JordanL2:
I submitted a patch that partially fixes this issue.

Alexays:
Closed by #905

PR #905, opened 2020-10-31
Tooltips no longer completely break if a custom module updates too frequently

JordanL2:
This is a fix for https://github.com/Alexays/Waybar/issues/890

The issue was caused by ```label_.set_tooltip_markup``` being called too frequently (more than once every 0.6s), which somehow prevents any tooltips from appearing at all. Now, ```label_.set_tooltip_markup``` is only called if the tooltip markup has changed. This means modules can now output text updates more frequently without breaking tooltips.

It would be good if it was possible to also update tooltip markup frequently without breaking tooltips, but that seems to be a GTK problem, and this is a good start.


Alexays (approved):

Alexays:
Thanks!

Issue #296, opened 2019-05-01
The Idle Inhibitor icon isn`t changed on another monitor


gorsheninmv:
I\`ve got an external monitor for my laptop and when I activate/deactivate the Idle Inhibitor by clicking on the icon on one monitor it won\`t switch on another one. It doesn\`t depend which monitor I click on.

Alexays:
Idle inhibitor is linked to a surface which is different on each bar, it might not be possible to fix that.

Synthetica9:
@Alexays Maybe you could create a file somewhere if idle is being inhibited (and watch it with something like libinotify/[GFileMonitor](https://developer.gnome.org/gio/stable/GFileMonitor.html))? 

This would have the additional advantage of allowing easy external control over the inhibit with a simple `rm`/`touch`.

JordanL2:
I raised a draft PR for this, it works but the part that causes the idle inhibitor modules on other bars needs to be worked on.

Alexays:
Closed by #906

PR #906, opened 2020-10-31
Fix for syncing idle inhibitor across outputs

JordanL2:
Fix for https://github.com/Alexays/Waybar/issues/296

Changes:
1. The idle_inhibitor surface is no longer part of the idle inhibitor module, it has been moved to Client, so there can only be one shared between all idle inhibitor modules.
2. After toggling the idle_inhibitor, the module needs to ensure all other idle inhibitor modules update. I'm unsure what the best way to do this is, so currently it just forces ALL modules to update.

### DIFF
--- a/src/modules/idle_inhibitor.cpp
+++ b/src/modules/idle_inhibitor.cpp
@@ -5,6 +5,7 @@ waybar::modules::IdleInhibitor::IdleInhibitor(const std::string& id, const Bar& 
                                               const Json::Value& config)
     : ALabel(config, "idle_inhibitor", id, "{status}"),
       bar_(bar),
+      idle_inhibitor_(nullptr),
       pid_(-1) {
   event_box_.add_events(Gdk::BUTTON_PRESS_MASK);
   event_box_.signal_button_press_event().connect(
--- a/src/modules/idle_inhibitor.cpp
+++ b/src/modules/idle_inhibitor.cpp
@@ -2,7 +2,7 @@
 #include "util/command.hpp"
 
 std::list<waybar::AModule*> waybar::modules::IdleInhibitor::modules;
-std::string                 waybar::modules::IdleInhibitor::status = "deactivated";
+bool                        waybar::modules::IdleInhibitor::status = false;
 
 waybar::modules::IdleInhibitor::IdleInhibitor(const std::string& id, const Bar& bar,
                                               const Json::Value& config)
@@ -37,7 +37,7 @@ waybar::modules::IdleInhibitor::~IdleInhibitor() {
 
 auto waybar::modules::IdleInhibitor::update() -> void {
   // Check status
-  if (status == "activated") {
+  if (status) {
     label_.get_style_context()->remove_class("deactivated");
     if (idle_inhibitor_ == nullptr) {
       idle_inhibitor_ = zwp_idle_inhibit_manager_v1_create_inhibitor(
@@ -51,11 +51,12 @@ auto waybar::modules::IdleInhibitor::update() -> void {
     }
   }
 
+  std::string status_text = status ? "activated" : "deactivated";
   label_.set_markup(
-      fmt::format(format_, fmt::arg("status", status), fmt::arg("icon", getIcon(0, status))));
-  label_.get_style_context()->add_class(status);
+      fmt::format(format_, fmt::arg("status", status_text), fmt::arg("icon", getIcon(0, status_text))));
+  label_.get_style_context()->add_class(status_text);
   if (tooltipEnabled()) {
-    label_.set_tooltip_text(status);
+    label_.set_tooltip_text(status_text);
   }
   // Call parent update
   ALabel::update();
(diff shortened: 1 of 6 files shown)


JordanL2:
I found a better way of updating the idle inhibitors. The updateAll method has been removed from Bar, and a list of idle inhibitor modules is kept on the Client. Now only idle inhibitor modules are being updated.

alebastr:
Note that the `idle_inhibitor` object is created for surface and becomes invalid once the surface is destroyed or no longer visible.

From the protocol standpoint, the current implementation with independent indicators makes the most sense as it actually conveys what would happen if you disconnect the output with active idle inhibitor.
***
If you want to address this, you could redesign the module in a way where a shared part of the module tracks all the bars displayed and manages idle inhibitor objects for all of them (including cleanup when the instance of bar assigned to specific output is removed and the inhibitor object no longer has a corresponding surface).
And it might be possible to contain all of that within the idle inhibitor module and avoid any changes to the Client class.

JordanL2:
I forgot about bars being created and destroyed when outputs are connected / disconnected. This means when the idle inhibitor modules are destroyed they need to be removed from the vector.

Ok, so I think I should:
1. move the ```idle_inhibitor``` object back to being part of the module
2. ```Client``` simply has the ```status``` property keeping track of whether or not idle inhibiting is enabled
3. when handleToggle is run:
- get the current state from Client
- invert it, then set it back on the Client
- call update on all inhibitor modules
4. when update is run:
- get the current state from Client
- if enabled, and module has no ```idle_inhibitor``` object, create it, otherwise delete it if it exists
5. when idle inhibitor module is destroyed, ensure it is removed from Client's ```idle_inhibitor_modules``` vector

I think this is all that's needed, when an output is connected the new module will automatically get its ```idle_inhibitor``` object if needed when update is called.

JordanL2:
Change committed, implementing the above.

JordanL2:
@Alexays @alebastr

On a slightly unrelated note, I noticed idle_inhibitor module sets this property called "click_param" to the value of status. This is a property that is part of ALabel, but nothing else seems to reference it. Is this used, or can it be removed?

Alexays:
@JordanL2, that's an old code, you can remove it. 😃

JordanL2:
Done :)

JordanL2:
Ok, so basically I haven't really used C++ since my days of university so I'm picking it up as I go along, mostly over the course of this weekend. I've now discovered class static members, so I've moved ```idle_inhibitor_status``` and ```idle_inhibitor_modules``` out of ```Client``` and into ```idle_inhibitor``` as static members so they're shared among the idle inhibitor module instances. This means ```Client``` is now untouched by this PR and it's a lot cleaner.

src/modules/idle_inhibitor.cpp
```diff
@@ -19,37 +25,60 @@ waybar::modules::IdleInhibitor::~IdleInhibitor() {
...
 }
 
 bool waybar::modules::IdleInhibitor::handleToggle(GdkEventButton* const& e) {
   if (e->button == 1) {
-    label_.get_style_context()->remove_class(status_);
-    if (idle_inhibitor_ != nullptr) {
-      zwp_idle_inhibitor_v1_destroy(idle_inhibitor_);
-      idle_inhibitor_ = nullptr;
-      status_ = "deactivated";
+    std::string status = waybar::modules::IdleInhibitor::status;
```
alebastr, line 65: You don't need to use fully qualified name to refer to static class member.
It might be also worth to store status as a `static bool active` since string representations are only used for UI update. I.e. the construction here shortens to `active = !active;`

src/modules/idle_inhibitor.cpp
```diff
@@ -19,37 +25,60 @@ waybar::modules::IdleInhibitor::~IdleInhibitor() {
...
+    label_.get_style_context()->remove_class(status);
```
alebastr, line 66: Doesn't seem to be invoked for other modules. Probably should go to `update` so that all UI changes would be in the same place.

src/modules/idle_inhibitor.cpp
```diff
@@ -19,37 +25,60 @@ waybar::modules::IdleInhibitor::~IdleInhibitor() {
...
+      status = "activated";
     }
-    click_param = status_;
+    waybar::modules::IdleInhibitor::status = status;
   }
+
+  // Make all other idle inhibitor modules update
+  for (auto const& module : waybar::modules::IdleInhibitor::modules) {
+    if (module != this) {
+      module->update();
```
alebastr, line 78: Current module also needs to be updated.

JordanL2: Current module is updated after its handleToggle is run, I'm not sure how/where but it is definitely triggered.

src/modules/idle_inhibitor.cpp
JordanL2: I've pushed a fix, I now just use the static class member directly.

src/modules/idle_inhibitor.cpp
JordanL2: I've moved this to update.

src/modules/idle_inhibitor.cpp
JordanL2: Also, I agree about making status a bool, pushed another commit.

src/modules/idle_inhibitor.cpp
alebastr: Right, `AModule::handleToggle` already triggers update by calling `dp.emit`. Forgot about that.

alebastr (approved):
LGTM. Thanks for making all these changes! :)

JordanL2:
Cheers! No problem, it's been a good learning experience :)

Alexays (approved):

Alexays:
LGTM, Thanks for the fix!

Issue #490, opened 2019-10-25
Dual battery bugs


SairesArt:
Removing one of my thinkpad's two batteries makes waybar lose it's sanity.
It usually shows a random amount above 200%. After a couple of minutes it shows either 0%, 5% whilst flashing white/red or copies the second batteries amount. Clicking the battery module to show remaining time and back to percentage again resets the amount to a random value above 200%. Although I'd really love to have my batteries charge past 100% :]
Also for some reason it breaks the wifi module the first time I remove one of the batteries, making it shows disconnected for about a minute, even though I'm clearly still connected to a hotspot...
![bugs](https://user-images.githubusercontent.com/10562480/67602744-984f1800-f777-11e9-9a8e-87bf25cd81f0.png)

pedrocr:
@SairesArt this is possibly fixed in current master. Could you please test and see?

PR #923, opened 2020-11-26
Simplify and improve battery state calculations

pedrocr:
I've come across a few inconsistencies in the battery state calculations so I had a go at the code to try and make it simpler and fix some of the issues. This PR includes:

- Calculating the battery state from just the energy values. Using other values for the total capacity was resulting in broken results in other cases. Since the energy information is already enough and was already present this simplifies the code and avoids the inconsistency.
- Clamp battery states above 100%. This can happen when the battery recalibrates and charges more than the capacity it thinks it has. Just show 100% in those cases.
- Round the battery state so that 99.9% shows as 100%. This avoids some cases where the last bit of battery doesn't charge but the battery is effectively full already.
- Whenever the capacity has been calculated as 100% show the state as full, so that the formatting rules are applied accordingly

I don't have a way to check as I don't have a removable second battery but this may help with #490. To fully fix that it's probably best to enumerate all the batteries on update instead of generating the list of batteries on startup.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -31,16 +31,16 @@ class Battery : public ALabel {
  private:
   static inline const fs::path data_dir_ = "/sys/class/power_supply/";
 
-  void                                          getBatteries();
+  void                                          refreshBatteries();
   void                                          worker();
   const std::string                             getAdapterStatus(uint8_t capacity) const;
   const std::tuple<uint8_t, float, std::string> getInfos() const;
   const std::string                             formatTimeRemaining(float hoursRemaining);
 
-  std::vector<fs::path> batteries_;
+  int                   global_watch;
+  std::map<fs::path,int> batteries_;
   fs::path              adapter_;
   int                   fd_;
-  std::vector<int>      wds_;
   std::string           old_status_;
 
   util::SleeperThread   thread_;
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM, thanks!
Can you add your idea about enumerate all the batteries on update?

pedrocr:
I've now also added making `full-at` show the battery at 100% when the state is at `full-at`. Before it was capping at the `full-at` value which doesn't really make sense. This way the range is back to 0-100% with `full-at` being the new 100%.

pedrocr:
Enumerating batteries on update doesn't sound too hard but it will do two things that may not be wanted:

- If the config has a specific battery and that has disappeared it will throw an error. That can be worked around to just show that battery as unknown but I don't have a way to test this.
- There will be a lot of `inotify_add_watch` and `inotify_rm_watch` calls on every update. No idea what that does for performance.

pedrocr:
I've implemented the hotplug stuff without doing any storm of `inotify_watch` calls. The only thing I'm missing is a way to have `refreshBatteries()` be called only when the `CREATE` or `DELETE` inotify event happened instead of every time. That would make the performance better. But maybe this is enough.

Since my laptop doesn't have a removable battery I have no way to test if this has solved the hotplug stuff though.

Alexays:
Could you add a TODO about calling `refreshBatteries()` only on `CREATE` and `DELETE` event?

pedrocr:
The TODO is already there.

By the way that reminds me that this and other modules should have a refresh every time the clock advances quickly or similar. When returning from suspend the battery updates very slowly.

Alexays:
@pedrocr We should create a separate issue for that.
I'll merge this PR, thanks!

PR #929, opened 2020-12-01
Update fmt subproject to 7.1.3

tchebb:
There is no particular change in this update that we require. However, our previous version, 5.3.0, is nearly two years old, so it seems prudent to pull in all the upstream fixes that have been made since then.

New wrap file taken from https://wrapdb.mesonbuild.com/fmt and modified to download from GitHub as per commit 99dde1aff8db ("Download patch files from Github instead of wrapdb").


Alexays:
Thanks!

PR #930, opened 2020-12-01
Fix a few compiler warnings

tchebb:
There was one uninitialized value warning and two mismatched-sign compare warnings. They both appear valid, the first occurring when MPD's `format-stopped` contains `{songPosition}` or `{queueLength}` and the second occurring when the clock's `timezones` array is more than 2 billion items long (not likely, I admit). Fix both issues.


Alexays:
Thanks!

PR #932, opened 2020-12-03
Don't update battery list on every update

pedrocr:
Speedup battery state update by only updating the battery list when we get a CREATE/DELETE event in the directory or whenever we do a full refresh on the interval.

This doesn't seem to make much difference in CPU performance either way that I've been able to measure. But the code doesn't look too bad, although it does need a mutex. It's a judgment call weather to do this or to just get rid of the TODO and live with the refresh on every update. There may be situations where the battery state is updating very fast where this is a relevant gain.


Alexays (approved):

Alexays:
LGTM, thanks for the quick update!

PR #934, opened 2020-12-03
Update to the latest version of the foreign toplevel manager protocol

l3nkz:
There was an update the of the toplevel manager protocol. Unfortunately,
there are no new interesting updates with regard to the taskbar
implementation. Nonetheless, update the protocol xml files to the latest
version so that the implementation is up-to-date.

While being there, also change the debug warning that is shown when
there is a version mismatch between the server and client version of the
protocol.

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -187,6 +187,12 @@ static void tl_handle_done(void *data, struct zwlr_foreign_toplevel_handle_v1 *h
     return static_cast<Task*>(data)->handle_done();
 }
 
+static void tl_handle_parent(void *data, struct zwlr_foreign_toplevel_handle_v1 *handle,
+        struct zwlr_foreign_toplevel_handle_v1 *parent)
+{
+    /* This is explicitly left blank */
+}
+
 static void tl_handle_closed(void *data, struct zwlr_foreign_toplevel_handle_v1 *handle)
 {
     return static_cast<Task*>(data)->handle_closed();
@@ -200,6 +206,7 @@ static const struct zwlr_foreign_toplevel_handle_v1_listener toplevel_handle_imp
     .state = tl_handle_state,
     .done = tl_handle_done,
     .closed = tl_handle_closed,
+    .parent = tl_handle_parent,
 };
 
 Task::Task(const waybar::Bar &bar, const Json::Value &config, Taskbar *tbar,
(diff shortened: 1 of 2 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -661,9 +661,10 @@ void Taskbar::register_manager(struct wl_registry *registry, uint32_t name, uint
         spdlog::warn("Register foreign toplevel manager again although already existing!");
         return;
     }
-    if (version < ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_FULLSCREEN_SINCE_VERSION) {
-        spdlog::warn("Using different foreign toplevel manager protocol version: {}", version);
+    if (version <= ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_FULLSCREEN_SINCE_VERSION) {
```
alebastr, line 664: `version <= 2`? Is there a reason to require version 3?
Also, strictly speaking, the module should work just fine with version 1 and 2 is not required. wayland protocols are backward compatible and supporting version X means that everything  < X should work as well.

alebastr (commented):
Please, define `parent` event handler. Otherwise libwayland-client could attempt to dereference `toplevel_handle_impl.parent` and crash.

src/modules/wlr/taskbar.cpp
l3nkz: I require at least version 2, because the taskbar uses the fullscreen functionality of the foreign-toplevel-manager. Hence, that's why I require at least version 2 ... Does this makes sense?

l3nkz:
@alebastr True, totally forgot about adding the event handler for this. Will update the pull request.

src/modules/wlr/taskbar.cpp
alebastr: Your change in condition now requires 3 and prints warning for 2.
That does make sense, but as I said, taskbar would work without fullscreen functionality. Although we don't have compositors with version 1 to test that :smile:

l3nkz: Ah sorry. You are right. I got the condition totally wrong in my head. I will update the warning as well.

alebastr: Oh, right, there are methods with since="2"...
Passive usage (like checking flags) does not require anything. Calling `set_fullscreen` may cause a protocol error and it would be better to have a check in `Task::fullscreen`.  Something like
```c++
if (zwlr_foreign_toplevel_handle_v1_get_version(handle_) < ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_FULLSCREEN_SINCE_VERSION) {
    fmt::warn(...);
    return;
}
```

l3nkz: Fixed in the latest version. Thank you very much for the feedback.

alebastr (approved):
Looks good. Thanks!

Alexays (approved):

Alexays:
LGTM, thx!

PR #935, opened 2020-12-04
fix(sway/workspaces): ignore emulated scroll events

### LINKED ISSUE #386: enable-bar-scroll shouldn't trigger on modules that handle scroll events
@RX14 mentioned in [a comment](https://github.com/Alexays/Waybar/pull/322#issuecomment-494037425) in #322 that the scrolling event should be consumed by the modules first, if they handle it. Personally, I use the scroll wheel to control the volume on the pulseaudio module, but with `enable-bar-scroll`, my workspaces are changed as well.

**Expected behavior**

My volume is changed, but not my workspaces when I scroll over a module that handles mouse wheel events.

alebastr:
GDK Wayland backend can emit two events for mouse scroll: one is a
GDK_SCROLL_SMOOTH and the other one is an emulated scroll event with
direction. We only receive emulated events on a window, thus it is not
possible to handle these in a module and stop propagation.

Ignoring emulated events should be safe since those are duplicates of
smooth scroll events anyways.

Fixes #386
CC @emmanueltouzery, @somini: please, confirm if that works and fixes the issue for you.

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -291,6 +291,12 @@ std::string Workspaces::getIcon(const std::string &name, const Json::Value &node
 }
 
 bool Workspaces::handleScroll(GdkEventScroll *e) {
+  if (gdk_event_get_pointer_emulated((GdkEvent *)e)) {
+    /**
+     * Ignore emulated scroll events on window
+     */
+    return false;
+  }
   auto dir = AModule::getScrollDir(e);
   if (dir == SCROLL_DIR::NONE) {
     return true;


emmanueltouzery:
YES, I tried it, and it works :)
it's better than #884, because the wheel events do still change the workspace on most of waybar (for me, everywhere but the volume icon and the calendar).

great work!

Alexays:
Thanks a lot @alebastr!

somini:
Finally got around to test this and can confirm what @emmanueltouzery mentioned.

It's merged already, so yay!

Issue #933, opened 2020-12-03
Feature-Request: pipewire module


ilpianista:
May you please add [PipeWire](https://pipewire.org/) support?

pkulak:
Or maybe if anyone has gotten anything to work they could post here? Even with pipewire-pulse installed, the pulse module doesn't work, which surprised me.

ilpianista:
@pkulak may you please give #937 a try?

krtko1:
> Or maybe if anyone has gotten anything to work they could post here? Even with pipewire-pulse installed, the pulse module doesn't work, which surprised me.

I find out it is working if you restart waybar

mastertinner:
For me, waybar-pulseaudio works fine with `pipewire-pulse`. But somehow, it doesn't recognize if I connect a bluetooth device (e.g. my Apple Air Pods Pro) and doesn't switch to the respective Icon specified in `format-bluetooth` in the waybar config.

nloewen:
If pipewire-pulse isn't exposing the bluetooth info in the same way as native PulseAudio, I think that's PipeWire's bug (or missing feature).

That said, a native PipeWire module is probably still a good idea to get access to things like latency data that could be useful in a status bar, but wouldn't be available in the PulseAudio compatibility layer.

pkulak:
> For me, waybar-pulseaudio works fine with `pipewire-pulse`. But somehow, it doesn't recognize if I connect a bluetooth device (e.g. my Apple Air Pods Pro) and doesn't switch to the respective Icon specified in `format-bluetooth` in the waybar config.

Hmm... not sure what's different with my setup. For me, the module has been empty since I switched to pipewire-pulse.

https://github.com/pkulak/dotfiles/blob/master/.config/waybar/config

nloewen:
Make sure your PipeWire version is up to date I guess? PipeWire is under heavy development, and regularly gets new features and fixes. Latest is 0.3.20, and the only issue I'm seeing is Waybar isn't showing Bluetooth info.

tokyovigilante:
I had to rebuild Waybar-git after installing pipewire-pulseaudio on FC33, but works fine for volume display and control after doing so.

niktob560:
Anyway we have ``pipewire-pulse``
Is Waybar working with it?

Narrat:
> Anyway we have `pipewire-pulse`
> Is Waybar working with it?

works fine for me

eternal-sorrow:
[...]

PR #937, opened 2020-12-04
Fix waybar-pulseaudio with pipewire-pulse

ilpianista:
Yesterday I opened #933, but actually this change was enough to make `waybar-pulseaudio` work with `pipewire-pulse`. What's your opinion about this change?


Alexays:
LGTM, thanks!

Issue #434, opened 2019-08-25
Define configuration format


lovesegfault:
Right now the configuration file's format is a weird JSON variation with `//` comments, at least as I understand it.

It's be nice if we, instead, used a more config-friendly format (such as TOML or even YAML) that allowed for things like comments. Moreover, it's be useful if the file was called `config.json` or `config.toml` so that editors can pick-up the filetype automatically.

I'll leave a link to an article I enjoy on why JSON is a poor choice for a configuration language: https://www.lucidchart.com/techblog/2018/07/16/why-json-isnt-a-good-configuration-language/

cafehaine:
That would mean breaking the compatibility with all installs of waybar. The user-base might be small, but this still would cause a lot of troubles.

lovesegfault:
@cafehaine Not necessarily, the old config could remain, but be removed from wiki/readme and leave alongside the TOML one for some time. If you load the old config format it works with a deprecation warning.

A breaking change nonetheless, but with a nice adaption period.

asmigala:
The great thing about yaml is that you can actually use a yaml parser to parse json. So that would kind of ensure backwards compatibility with older configs, the only problem being the `//` comments.

A dirty way to convert old config to yaml:
```
cat ~/.config/waybar/config | sed 's|//.*$||' | yq -y .
```

(this won't work if you have `//` in a string somewhere, but you get the idea)

eoli3n:
+10 i'm really confused about that choice...

eoli3n:
Toml is a bad choice too : [https://hitchdev.com/strictyaml/why-not/toml/](https://github.com/crdoconnor/strictyaml#why-strictyaml)
Yaml is still the easier and cleaner human readable configuration language and would be a great choice for that kind of project.

lovesegfault:
@eoli3n Agree to disagree and I'll avoid bikeshedding this :laughing:

WhyNotHugo:
FWIW, current format is just "a JSON variation", it's called `jsonc`, and actually used by plenty of tools. So I wouldn't discredit is as non-standard.

eoli3n:
Any news ?

billop:
> FWIW, current format is just "a JSON variation", it's called `jsonc`, and actually used by plenty of tools. So I wouldn't discredit is as non-standard.

I was only able to find that VS Code uses this format for configuration. What other tools were you talking about here?

WhyNotHugo:
[...]

PR #941, opened 2020-12-08
Add filetype hint for vim.

ferdinandschober:
This will enable syntax highlighting in vim / nvim without the
need of setting it manually via :set syntax=json by providing a vim
filetype hint.


ferdinandschober:
this would provide a solution for #434 at least for vim.

WhyNotHugo:
Correct type is `jsonc` (JSON with inline comments).

Alexays:
Closing in favor of #1112

PR #946, opened 2020-12-18
Fix Sway #waybar.solo CSS rule applying on split

narilth:
This error occurs because of an incorrect assumption that the size of the list of nodes that contains the focused window is the number of windows in a workspace.

The windows in a workspace are stored as a tree by Sway, rather than a list, so the number of windows has to be found by counting the leaves of a workspace tree.


Alexays (approved):

Alexays:
Thanks!

PR #949, opened 2020-12-23
CI: add FreeBSD to Actions

jbeich:
The last build on TravisCI was 20 days ago, so let's try an alternative.
https://travis-ci.com/github/Alexays/Waybar/builds
https://news.ycombinator.com/item?id=25338983

Example build: https://github.com/jbeich/Waybar/runs/1602295701


Alexays:
Yeah, Travis is credit limited...
We should move slowly to Github Action.
LGTM

PR #956, opened 2020-12-25
Added waybar_output.identifier support. #602

### LINKED ISSUE #602: Identify output by model and serial
Currently it seems only possible to match against "connector names" like `DP-3` and such.
It would be a nice addition when one could match with the full name and serial number, like for example in Sway: `Goldstar Company Ltd IPS235 Serial Number` (and yes, that model as "Serial Number" for a serial number :man_facepalming:).

That would be helpful in case of unreliable "connector names" (they might change on every boot for example), or when the same connector is sometimes used for different purposes.

AndreasBackx:
This resolves #602 allowing you to specify the output identifier exactly like Sway allows you to do.

It does this by getting the identifier from the description that is reported. This is how swaybar does it as well: https://github.com/swaywm/sway/pull/3415/files#diff-f1816f4a24dbbcf1a1dc419a0daeee3ac8bfcb410a44bd9d9e75900d1210bbbcR264-R279

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -144,10 +144,6 @@ void waybar::Client::handleOutputName(void *      data, struct zxdg_output_v1 * 
   auto client = waybar::Client::inst();
   try {
     auto &output = client->getOutput(data);
-    spdlog::debug("Output detected with name: {} ({} {})",
-                  name,
-                  output.monitor->get_manufacturer(),
-                  output.monitor->get_model());
     output.name = name;
   } catch (const std::exception &e) {
     std::cerr << e.what() << std::endl;
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -1,12 +1,14 @@
 #include "client.hpp"
+
 #include <fmt/ostream.h>
 #include <spdlog/spdlog.h>
+
 #include <fstream>
 #include <iostream>
-#include "util/clara.hpp"
-#include "util/json.hpp"
 
 #include "idle-inhibit-unstable-v1-client-protocol.h"
+#include "util/clara.hpp"
+#include "util/json.hpp"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 
 waybar::Client *waybar::Client::inst() {
@@ -59,8 +61,8 @@ void waybar::Client::handleOutput(struct waybar_output &output) {
       .logical_position = [](void *, struct zxdg_output_v1 *, int32_t, int32_t) {},
       .logical_size = [](void *, struct zxdg_output_v1 *, int32_t, int32_t) {},
       .done = [](void *, struct zxdg_output_v1 *) {},
-      .name = &handleOutputName,
-      .description = [](void *, struct zxdg_output_v1 *, const char *) {},
+      .name = [](void *, struct zxdg_output_v1 *, const char *) {},
+      .description = &handleOutputDescription,
   };
   // owned by output->monitor; no need to destroy
   auto wl_output = gdk_wayland_monitor_get_wl_output(output.monitor->gobj());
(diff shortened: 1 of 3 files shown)


src/client.cpp
```diff
@@ -59,8 +61,8 @@ void waybar::Client::handleOutput(struct waybar_output &output) {
       .logical_position = [](void *, struct zxdg_output_v1 *, int32_t, int32_t) {},
       .logical_size = [](void *, struct zxdg_output_v1 *, int32_t, int32_t) {},
       .done = [](void *, struct zxdg_output_v1 *) {},
-      .name = &handleOutputName,
-      .description = [](void *, struct zxdg_output_v1 *, const char *) {},
+      .name = [](void *, struct zxdg_output_v1 *, const char *) {},
+      .description = &handleOutputDescription,
```
AndreasBackx, line 65: I wanted to keep the old name callback as well. Unfortunately there _seems_ to be a conflict in how Waybar handles them. I don't know the details and didn't want to fall in a rabbit hole.

I have 2 Dell U2515H monitors and it would only callback the description once instead of twice. So DP-4 and DP-5 would be reported twice via the name callback, but their description (and thus identifiers) would only be reported for 1 instead of 2. Only relying on the description solves it.

src/client.cpp
```diff
@@ -112,16 +117,20 @@ std::vector<Json::Value> waybar::Client::getOutputConfigs(struct waybar_output &
...
-                  output.monitor->get_manufacturer(),
-                  output.monitor->get_model());
+    auto &      output = client->getOutput(data);
+    const char *open_paren = strrchr(description, '(');
+    const char *close_paren = strrchr(description, ')');
+
+    // Description format: "identifier (name)"
+    size_t identifier_length = open_paren - description;
+    output.identifier = std::string(description, identifier_length - 1);
+    output.name = std::string(description + identifier_length + 1, close_paren - open_paren - 1);
```
alebastr, line 131: Quote from xdg-output-unstable-v1:
```
Many compositors can produce human-readable descriptions of their
outputs.  The client may wish to know this description as well, to
communicate the user for various purposes.

The description is a UTF-8 string with no convention defined for its
contents. Examples might include 'Foocorp 11" Display' or 'Virtual X11
output via :1'.
```
It's not correct to rely on any specific content or format of description string.

If you need to handle both `name` and `description` you'll have to define `name` and `description` callbacks that only set corresponding properties, subscribe to `zxdg_output_v1.done` or  `wl_output.done` and wait for the `done` callback before taking any actions. The callback should be selected based on zxdg_output_v1 version; anything >= 3 should use `wl_output.done`, previous versions should use `zxdg_output_v1.done`.

alebastr: > The callback should be selected based on zxdg_output_v1 version;

I checked the code and we are binding to version 2 explicitly. You can ignore that part and use only `zxdg_output_v1.done`.

AndreasBackx:
Addressed comments and the following scenarios were tested for the following monitors:

* HDMI-A-1: Acer _Serial Number_
* DP-4: Dell _Serial Number_
* DP:-5 Dell _Serial Number_

1. String `"HDMI-A-1"`: only shows on that one monitor. 
2. String `"!HDMI-A-1"`: only shows on other 2 monitors.
3. String `"Acer Technologies XF270HU SERIAL"`: only shows on that one monitor. 
4. String `"!Acer Technologies XF270HU SERIAL"`: only shows on other 2 monitors.
5. Array `["HDMI-A-1"]`: only show on that one.
6. Array `["Acer Technologies XF270HU SERIAL"]`: only show on that one.

src/client.cpp
AndreasBackx: I addressed your comments. The worst that can happen now is that the identifier is some garbage in case the description is not in the format we want. Though I feel like this is acceptable as swaybar does this? :)

src/client.cpp
```diff
@@ -139,6 +139,36 @@ void waybar::Client::handleOutputName(void *      data, struct zxdg_output_v1 *
   }
 }
 
+void waybar::Client::handleOutputName(void *      data, struct zxdg_output_v1 * /*xdg_output*/,
+                                      const char *name) {
+  auto client = waybar::Client::inst();
+  try {
+    auto &output = client->getOutput(data);
+    spdlog::debug("Output detected with name: {} ({} {})",
```
alebastr, line 147: This log duplicates the one in `done` handler and can be safely removed.

alebastr (approved):
Thanks for the changes!
I think it's fine to match swaybar behavior. In fact this would be useful when I finish work on swaybar ipc compatibility layer.

AndreasBackx:
Maybe the project should move to GitHub actions btw. Travis CI's free model seems to be gutted which is probably why CI takes so long.

Koffeinfriedhof:
Hi!
Just was searching for this feature and implemented the changes. Works for Dell U2715H (same here) and Philips PHL 272B8Q too.
Thanks for implementing!

Alexays (approved):

Alexays:
LGTM, thanks!
Maybe we need to update the wiki to add some lines about the change.

PR #880, opened 2020-10-11
added module group selectors for styling

ChilloManiac:
Added css selectors to the module groups. This allows for individual styling of each group seperately. Suggested in #652

PR #959, opened 2020-12-28
Fix for group module selectors when bar is vertical

dorgnarg:
I noticed that #880 didn't seem to work when the bar was in vertical mode, and it looks like that was because the boxes were being re-assigned after adding the selectors, so I moved the selectors to just after the re-assignment, so now they should work in either mode :)


Alexays (approved):

Alexays:
Thanks!

PR #964, opened 2021-01-03
Add an option to use battery design capacity

Ocisra:
This little PR implements a new option `design-capacity` to the battery module with default value `false`. If set to `true`, Waybar will use the design capacity of the battery instead of the actual full capacity which may be lower due to deterioration of the battery. With this option, ti should report the same battery percentage than i3status' default behavior.
I personally like to see it this way because I know how worn off my battery is.

I will edit the wiki as soon as you merge this PR.


src/modules/battery.cpp
```diff
@@ -156,17 +158,20 @@ const std::tuple<uint8_t, float, std::string> waybar::modules::Battery::getInfos
...
       total_power += power_now;
       total_energy += energy_now;
       total_energy_full += energy_full;
+      total_energy_full_design += energy_full_design;
     }
     if (!adapter_.empty() && status == "Discharging") {
       bool online;
       std::ifstream(adapter_ / "online") >> online;
-      if (online) {
+      if (online) {
```
Alexays, line 174: Remove extra space here

Alexays (changes requested):

Alexays:
Hi, thanks for the change.
Just a typo change and i'll merge this PR!

Ocisra:
Sorry this change wasn't intended.
I changed it, the extra commit is because I forgot to exclude my `.cache` folder

Alexays:
LGTM, thx, could you update github wiki?

Ocisra:
Done,
thanks for merging

PR #965, opened 2021-01-03
Fixing logic in getIcon

### LINKED ISSUE #896: Battery module doesn't work with format-icons object
Using the following configuration for the battery module, nothing is displayed during either charging or discharging. I had a good look through `ALabel.cpp` and I'm pretty sure this config should work just fine, but nothing shows up. If I change `"default"` to `"discharging"`, then the capacity percentage at least shows while charging or discharging, but no icons are shown. If I replace `format-icons` with a plain array of icons, then everything works fine, but I'd like to have a different array depending on the battery status (charging/discharging).

```jsonc
"battery": {
        "bat": "BAT0",
        "adapter": "AC",
        "states": {
            // "good": 95,
            "warning": 25,
            "critical": 15
        },
        "format-icons": {
            "default": ["󰂎", "󰁺", "󰁻", "󰁼", "󰁽", "󰁾", "󰁿", "󰂀", "󰂁", "󰂂", "󰁹"],
            "charging": ["󰢟", "󰢜", "󰂆", "󰂇", "󰂈", "󰢝", "󰂉", "󰢞", "󰂊", "󰂋", "󰂅"]
        }
    }
```

ofwinterpassed:
fixes #896

This fixes a logical error in the loop checking several alts to get the right icon. The old logic assigned "default" on each failure to find the right one. This made the loop fail on the next iteration unless default also was an object that contained the other alts looked for.

My fix changes the logic to use a default value for the name that should be used and assigns to that name if it finds the value. Finally it assigns to `format_icons` once it either found the one we're looking for or nothing was found.

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -60,14 +60,14 @@ std::string ALabel::getIcon(uint16_t percentage, const std::string& alt, uint16_
 std::string ALabel::getIcon(uint16_t percentage, std::vector<std::string>& alts, uint16_t max) {
   auto format_icons = config_["format-icons"];
   if (format_icons.isObject()) {
+    std::string _alt = "default";
     for (const auto& alt : alts) {
       if (!alt.empty() && (format_icons[alt].isString() || format_icons[alt].isArray())) {
-        format_icons = format_icons[alt];
+        _alt = alt;
         break;
-      } else {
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
Thanks!

PR #969, opened 2021-01-09
Fix issues in wlr/taskbar

alebastr:
See commit descriptions for individual issues.

A better fix for the protocol error would be to separate the foreign toplevel manager code from Gtk UI and keep a single instance shared between bars, but that requires more time and effort than I can spare right now :frowning_face:

CC: @danieldg, @l3nkz

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -367,16 +367,16 @@ void Task::handle_output_leave(struct wl_output *output)
 void Task::handle_state(struct wl_array *state)
 {
     state_ = 0;
-    for (auto* entry = static_cast<uint32_t*>(state->data);
-         entry < static_cast<uint32_t*>(state->data) + state->size;
-         entry++) {
-        if (*entry == ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_MAXIMIZED)
+    size_t size = state->size / sizeof(uint32_t);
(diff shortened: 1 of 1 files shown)


l3nkz (approved):

l3nkz (approved):

l3nkz:
Thank you very much for your fixes. All looks good to me.
I will think about your idea to separate the foreign toplevel manager handling code from the ui stuff. And might submit a PR for that. Thanks for the input.

Alexays (approved):

Alexays:
LGTM

l3nkz:
Hm, I get now some very wired behavior when unplugging from an external output. Sometimes my compositor (wayfire) crashes now which is even worse than waybar crashing. But I first need to look into this more closely.

alebastr:
@l3nkz do you have a wayfire backtrace?
The only change from the compositor standpoint is that the wlr-foreign-toplevel-manager serverside object is now deleted on output unplug. That doesn't seem to be causing any issues on sway.

l3nkz:
@alebastr I am working on it. However, I couldn't yet reproduce it when I start wayfire directly from console instead of starting it via SDDM. I will continue digging into the problem. Maybe this is unrelated with this change. I will keep you updated.

l3nkz:
After some more digging, I think this is actually unrelated to your change. Thank you again for your work.

Issue #925, opened 2020-11-27
[Feature Request] - Split pulseaudio module into sink and source modules


chron-isch:
Hi, most of all, thanks for all the work on waybar.

For most mobile system the way the pulseaudio module currently works is perfectly practical, but as soon as you start using multiple sinks and sources you run into configuration problems. That's why I wanted to ask if it was feasible to split it into separate modules for sink and source and make the actual device it's referring to configurable and just use default/fallback if no device is specified by the user.
That way you could also have separate left-, middle-, right-click and on-scroll actions for sinks and sources.

subnut:
Any progress on this? Currently, there is no way to control the the source (microphone) volume using waybar. Scrolling on _any_ pulseaudio module results in changing the sink (speaker) volume, which is very unintuitive IMO.

mknapik:
@subnut Controlling source volume works for me when repeating the module twice. `volume.sh` is a custom script which is a wrapper for `pamixer`.
```
...
"pulseaudio#input": {
        "format-source": "",
        "format-source-muted": "",
        "format": "{format_source}",
        "on-click": "volume.sh input-mute",
        "on-scroll-up": "volume.sh input-up",
        "on-scroll-down": "volume.sh input-down",
        "tooltip": false
    },
    "pulseaudio#output": {
        "format": "{volume} {icon}",
        "format-icons": {
            "default": [
                "",
                "",
                "",
                ""
            ]
        },
        "on-click": "volume.sh mute",
        "on-scroll-up": "volume.sh up",
        "on-scroll-down": "volume.sh down",
        "scroll-step": 1,
        "tooltip": false
    },
...
```

OJFord:
Thanks for that - it's not a complete solution though since both will be styled 'muted' whenever the _sink_ is; ignoring the state of the source.

subnut:
@OJFord You can use
```css
#pulseaudio.output.source-muted {
  background: red;
}
```

OJFord:
Thanks @subnut, (though that should be `.input.source-muted` if using the names above) ok as a workaround then but I still agree this would be more convenient if they were separate.

abhinandanudupa:
> Thanks @subnut, (though that should be `.input.source-muted` if using the names above) ok as a workaround then but I still agree this would be more convenient if they were separate.

Does this work now? In my testing it does not.

ret2src:
Thanks to the comments above, I was able to control the icons for input and output separately. Here's my `config`:

``` json
[...]

Issue #853, opened 2020-09-08
[Feature Request] - tooltip-format for pulseaudio module


returntrip:
I use two `pulseaudio` modules: one for input and one for output. It would be great if the tooltip for the input could show the source instead of the description of the currently selected output.

kamushadenes:
Same here, also different classes for sink mute and source mute would be awesome!

OJFord:
`tooltip-format` seems to have been added in #974 linked above.

returntrip:
Closed by #974 . Thanks @kamushadenes!

PR #974, opened 2021-01-12
Improve Pulseaudio sink/source separation

kamushadenes:
Hello,

I'm not really experience with C++, so apologies for any mistakes.

I opened this PR due to https://github.com/Alexays/Waybar/issues/853.

This should be temporary, as https://github.com/Alexays/Waybar/issues/925 exists.

## Changes

- Add additional fields, namely `source_volume` and `source_desc`
- Add `tooltip-format`, reverting to default behavior if not specified
- Add additional CSS classes, namely `sink-muted` and `source-muted`

### DIFF
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -207,6 +207,7 @@ const std::string waybar::modules::Pulseaudio::getPortIcon() const {
 
 auto waybar::modules::Pulseaudio::update() -> void {
   auto format = format_;
+  std::string                 tooltip_format;
   if (!alt_) {
     std::string format_name = "format";
     if (monitor_.find("a2dp_sink") != std::string::npos) {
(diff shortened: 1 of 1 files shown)


src/modules/pulseaudio.cpp
```diff
@@ -222,28 +223,53 @@ auto waybar::modules::Pulseaudio::update() -> void {
       }
       format_name = format_name + "-muted";
       label_.get_style_context()->add_class("muted");
+      label_.get_style_context()->add_class("sink-muted");
```
Alexays, line 226: This class is really needed?
Seems a dups of `muted` one

Alexays (changes requested):

src/modules/pulseaudio.cpp
kamushadenes: This is to separately handle sink and source mutes, while preserving default behavior.

Alexays (approved):

PR #980, opened 2021-01-15
add option 'tag-labels' to river/tags

sjtio:
add option (array) to river/tags so that you can define the labels to display rather than 1,2,3...


Alexays (approved):

Alexays:
Thx, could you also update the Github wiki?

sjtio:
just updated the wiki

thanks for merging!

PR #984, opened 2021-01-16
[WIP] Add ability to use icons for language module

3ap:


alebastr (changes requested):
Looking at the other modules, the common style is to define getIcon method and use it for the `icon` argument. Something like
```c++
std::string Language::getIcon(const std::string& lang) {
  auto icons = config_["format-icons"];
  if (icons.isObject() && icons[lang].isString()) {
    return icons[lang].asString();        
  }
  // default to name if no icon is configured
  return lang;
}
// ...
fmt::format(format_, fmt::arg("language", lang_), fmt::arg("icon", getIcon(lang_)));
// ...

File: src/ALabel.cpp [hunk 1]
```diff
@@ -80,6 +80,14 @@ std::string ALabel::getIcon(uint16_t percentage, std::vector<std::string>& alts,
   return "";
 }
 
+std::string ALabel::getIcon(const std::string& key) {
```
Alexays, line 83: You should define getIcon in the language class :)

Alexays (changes requested):

src/modules/sway/language.cpp
```diff
@@ -53,11 +53,21 @@ void Language::onEvent(const struct Ipc::ipc_response& res) {
...
+  }
+  return "";
+}
+
 auto Language::update() -> void {
   if (lang_.empty()) {
     event_box_.hide();
   } else {
-    label_.set_markup(fmt::format(format_, lang_));
+    label_.set_markup(fmt::format(format_,
```
Alexays, line 68: You should also keep default arg for config compability purpose

Alexays (changes requested):

PR #1002, opened 2021-01-31
add length limits for MPD module tags

nullobsi:
adds the following configuration options to the MPD module:

 - artist-len
 - album-len
 - album-artist-len
 - title-len

to control the maximum length of the tags when they are displayed on the bar.

### DIFF
--- a/src/modules/mpd/mpd.cpp
+++ b/src/modules/mpd/mpd.cpp
@@ -2,7 +2,7 @@
 
 #include <fmt/chrono.h>
 #include <spdlog/spdlog.h>
-
+#include <glibmm/ustring.h>
 #include "modules/mpd/state.hpp"
 #if defined(MPD_NOINLINE)
 namespace waybar::modules {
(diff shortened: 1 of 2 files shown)


YerinAlexey (approved):
Looks good

YerinAlexey (approved):

src/modules/mpd/mpd.cpp
```diff
@@ -144,13 +144,25 @@ void waybar::modules::MPD::setLabel() {
   bool        singleActivated = mpd_status_get_single(status_.get());
   std::string singleIcon = getOptionIcon("single", singleActivated);
 
+  auto artistLen = config_["artist-len"].isInt() ?
```
Alexays, line 147: I would prefer overriding artist with something like:
```
if (config_["artist-len"].isInt()) {
    artist = artist.substr(0, config_["artist-len"].asInt());
}
```
Rather than substr with the string length if length not provided, same for other variables.

Alexays (changes requested):
Nice addition, but what do you think of my proposal?

src/modules/mpd/mpd.cpp
YerinAlexey: Yeah, that makes sense

nullobsi: that makes more sense, I'll fix it up in a bit

nullobsi:
alright, i've fixed it up :smile:

Alexays (approved):

Alexays:
LGTM, thx

Alexays:
Could you also update the github wiki?

Issue #685, opened 2020-04-28
Bluetooth module error


DiegoGuidaF:
When using the bluetooth example in the Man page I get the following error:
`[error] bluetooth: argument not found`

No extra info in debug or trace log levels.

Alexays:
What's your config file?

DiegoGuidaF:
Even with a barebone config I get the error:
`// vim:ft=json
{
    "layer": "top",
    "height": 22,
    "position": "top",

    "modules-center": ["bluetooth"],

    "bluetooth": {
    	"interval": 30,
    	"format": "{icon}",
    	"format-alt": "bluetooth: {status}",
    	"format-icons": {
    		"enabled": "",
    		"disabled": ""
    	},
    	"tooltip-format": "{status}"
    }
}
`

Alexays:
Your problem seems to be the tooltip-format, the status arg doest exist, you must use `{}`
It'll be fixed soon otherwise

PR #1003, opened 2021-01-31
Update waybar-bluetooth.5.scd

joshuachp:
Remove the `status` from the `tooltip-format` example since it will
throw error. Related to #685


Alexays:
Thanks!

PR #1006, opened 2021-01-31
Improved wlr/taskbar icon search.

jgmdev:
Fixes application icons not found like Quassel Client where the icon is not named org.kde.quassel but quassel or Wayfire config manager which .desktop file is not named org.gtk.wayfire-config-manager.desktop but instead wayfire-config-manager.desktop.


Alexays (approved):

Alexays:
Thanks!

Issue #883, opened 2020-10-12
Properly handling multiple Waybars


nyamcoder:
Hi all, hope you're fine!

I lately started using Sway along with Waybar, and I like it so far very much – especially as it looks like Waybar would easily deal with multiple instances (at least it's mentioned in the wiki). Since I have a small 1024x768px screen and would like to organize my panel stuff by a couple of stacked Waybars; preferably with Waybar tools only.

And if I am not wrong, as for now it seems not possible to optionally explicitely limit the width of the modules (see ([my other issue](https://github.com/Alexays/Waybar/issues/882)) to avoid covering each other, I would rather need even up to six individual Waybars: 1 upper left for its `"modules-left"` only, 1 upper centered for `"modules-center"`, 1 upper right for `"modules-right"`, 1 lower left for its `"modules-left"` only, 1 lower centered for its own `"modules-center"`, 1 lower right for its own `"modules-right"`, respectively.

While on the one hand I got them all named and started – what does not work as intended is

1. to display them in their own shortened widths (per px), so that all upper 3 waybar instances would arrange themselves automatically in a single line next to each other (imagine `"display: inline-block"` in HTML-CSS), and the lower waybars in another line below the upper one accordingly. So if I set 2 x 3 (say, 450px+124px+450px) = 6 individual bars and widths such as:
```
[{
    "name": "waybar#testbar1", 
    "position": "top", // Waybar position (top|bottom|left|right) -- i.e. screen edge
    "height": 30, // Waybar height (to be removed for auto height)
    "width": 450, // Waybar width -- no effect?
    "modules-left": ["sway/window"] // -- kind of multi-module alignment or gravity
},
...
etc. 
...
]
```
– all bars appear stacked in full width though (and also now cover a `bemenu` app-starter):
```
[2020-10-11 12:08:55.641] [info] Bar configured (width: 1024, height: 30) for output: LVDS-1
[2020-10-11 12:08:55.812] [info] Bar configured (width: 1024, height: 30) for output: LVDS-1
[2020-10-11 12:08:55.814] [info] Bar configured (width: 1024, height: 30) for output: LVDS-1
[2020-10-11 12:08:55.815] [info] Bar configured (width: 1024, height: 30) for output: LVDS-1
[2020-10-11 12:08:55.820] [info] Bar configured (width: 1024, height: 30) for output: LVDS-1
[2020-10-11 12:08:55.826] [info] Bar configured (width: 1024, height: 30) for output: LVDS-1
```
[...]

PR #1008, opened 2021-01-31
improve module sizing options

nullobsi:
Add `min-length` config property, can be used in combination with `max-length` to fix the length of a module to a specific size
Also add `align` property to align text in a label

this should help with #883

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -20,7 +20,7 @@ ALabel::ALabel(const Json::Value& config, const std::string& name, const std::st
   }
   event_box_.add(label_);
   if (config_["max-length"].isUInt()) {
-    label_.set_max_width_chars(config_["max-length"].asUInt());
+    label_.set_max_width_chars(config_["max-length"].asInt());
     label_.set_ellipsize(Pango::EllipsizeMode::ELLIPSIZE_END);
     label_.set_single_line_mode(true);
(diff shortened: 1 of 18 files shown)


src/ALabel.cpp
```diff
@@ -20,14 +20,23 @@ ALabel::ALabel(const Json::Value& config, const std::string& name, const std::st
...
     label_.set_single_line_mode(true);
   }
 
+  if (config_["min-length"].isUInt()) {
+    label_.set_width_chars(config_["min-length"].asUInt());
+  }
+
+  if (config_["align"].isDouble()) {
+    auto align = config_["align"].asFloat();
+    label_.set_xalign(align);
```
Alexays, line 37: Does we also want xalign if the bar is in vertical mode?

nullobsi: I think it should be based on the mode of the bar. Let me try and add a check.

nullobsi: Okay, I've made it based off the rotate property. This should make align match with how the max and min lengths affect the label.

Alexays:
Nice, could you add `min-length` to man, then i'll merge this PR :)

nullobsi:
I've updated the man pages that have `max-length`. Thank you!

Alexays:
Thanks!

PR #1011, opened 2021-02-02
[wlr/taskbar] More icon search improvements.

jgmdev:
While using waybar noticed that other apps where also missing icon and added these extra changes:

* Added ~/.local/share prefix to search in current user defined apps.
* Add support for apps that don't properly set an id like pamac-manager which desktop info file is org.manjaro.pamac.manager.desktop but wayfire reports its id as pamac-manager.
* Lastly, If no icon is found use "unknown" as default icon instead of displaying an empty square button.

For icon issues like with pamac-manager I had to use g_desktop_app_info_search directly since I didn't see any bind to it on the giomm c++ binding.

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -15,6 +15,7 @@
 #include <gtkmm/icontheme.h>
 
 #include <giomm/desktopappinfo.h>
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
Thanks, taskbar should be good now :)

PR #1014, opened 2021-02-02
Add a way to configure auto_back_and_forth on Sway workspaces

### LINKED ISSUE #985: [Sway/Workspaces] Enable auto back-and-forth
Hi Alexays,
I would like to have the option to go back to the previous workspace when clicking on the actual one. If you are OK with it, I'm going to implement it but I have one question about it : do you prefer that this setting follows `workspace_auto_back_and_forth` in the sway config (really easy to implement) or that I create a new setting to control this behavior (I will need a tester to be sure that it does work on a multi-monitor setup) ?

Moonlight-Angel:
#736 introduced a behavior that some may like, but some may not ; I'm part of the latter.

This PR adds a configuration entry to Sway workspaces module in order to enable or disable this behavior.
By default, the behavior is to follow what's configured in the Sway config file (seems more logical that way IMO), and you're able to explicitely tell Waybar not to follow what's configured in the Sway config file.
However, the `--no-auto-back-and-forth` flag is conserved when moving workspaces to other outputs or when changing workspaces by scrolling on the bar, since these behaviors were rightfully fixed by #736 and that's fine by me.

The config entry is named `disable-auto-back-and-forth` and is a boolean.

Closes #985

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -257,11 +257,19 @@ Gtk::Button &Workspaces::addButton(const Json::Value &node) {
           ipc_.sendCmd(
               IPC_COMMAND,
               fmt::format(workspace_switch_cmd_ + "; move workspace to output \"{}\"; " + workspace_switch_cmd_,
+                          "--no-auto-back-and-forth",
                           node["name"].asString(),
                           node["target_output"].asString(),
+                          "--no-auto-back-and-forth",
                           node["name"].asString()));
         } else {
-          ipc_.sendCmd(IPC_COMMAND, fmt::format(workspace_switch_cmd_, node["name"].asString()));
+          ipc_.sendCmd(
+              IPC_COMMAND,
+              fmt::format("workspace {} \"{}\"",
+                          config_["disable-auto-back-and-forth"].asBool()
+                            ? "--no-auto-back-and-forth"
+                            : "",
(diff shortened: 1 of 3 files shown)


src/modules/sway/workspaces.cpp
```diff
@@ -322,7 +330,9 @@ bool Workspaces::handleScroll(GdkEventScroll *e) {
     }
   }
   try {
-    ipc_.sendCmd(IPC_COMMAND, fmt::format(workspace_switch_cmd_, name));
+    ipc_.sendCmd(
+        IPC_COMMAND,
+        fmt::format(workspace_switch_cmd_, "--no-auto-back-and-forth", name));
```
Alexays, line 335: Missing `config_["disable-auto-back-and-forth"].asBool()` no?

Moonlight-Angel: Nope, disabling it when scrolling through workspaces on the bar is a behavior we want.
Technically, you can't "re-click" on the same workspaces when scrolling through them, but I chose to keep this behavior introduced by #736 since it seemed to fix some glitch.

Alexays:
LGTM, thanks!

Alexays:
Could you update github wiki?

Moonlight-Angel:
> Could you update github wiki?

Done ! Thanks for merging this !

PR #1015, opened 2021-02-03
rfkill code refactoring

### LINKED ISSUE #994: Crash on start - 'std::runtime_error' - Wrong size of RFKILL event
I run Fedora Rawhide, with wlroots and sway from git master.
Building and running git master results in the following:

waybar -l debug
[2021-01-23 15:04:20.085] [info] Resources files: /home/dturner/.config/waybar/config, /home/dturner/.config/waybar/style.css
[2021-01-23 15:04:45.167] [debug] Output detection done: DVI-D-1 (Ancor Communications Inc ASUS VW224 A4LMTF074092)
[2021-01-23 15:04:45.250] [debug] Cmd exited with code 127
[2021-01-23 15:04:45.250] [error] media stopped unexpectedly, is it endless?
[2021-01-23 15:04:45.250] [debug] Received SIGCHLD in signalThread
[2021-01-23 15:04:45.274] [warning] For a functional tray you must have libappindicator-* installed and export XDG_CURRENT_DESKTOP=Unity
terminate called after throwing an instance of 'std::runtime_error'
  what():  Wrong size of RFKILL event
Aborted (core dumped)

Any thoughts?

alebastr:
I only wanted to fix that one event size check in `rfkill.cpp`. Should have more willpower to stop and ignore the rest :sob:

Summary:
 * Reopening "/dev/rfkill" for each event isn't optimal and could miss some events. Especially since we were ignoring ADD event with the initial state of the device.
 * Rfkill doesn't even need a thread. Glib::MainLoop is fast enough to handle events on a couple of file descriptors.
 * `Bluetooth::interval_thread_` was not doing anything useful because it had no way to get changed status value. Now we have 3 less threads :smile:
 * I'm actually ashamed of the last commit, but untangling the locks and threads in `network.cpp` was beyond an evening coding exercise. Although I'm quite curious why we have to sleep for 5 seconds while holding the mutex and will check that on another evening.

Fixes #994

### DIFF
--- a/src/util/rfkill.cpp
+++ b/src/util/rfkill.cpp
@@ -19,60 +19,63 @@
 #include "util/rfkill.hpp"
 
 #include <fcntl.h>
+#include <glibmm/main.h>
 #include <linux/rfkill.h>
-#include <poll.h>
-#include <stdlib.h>
+#include <spdlog/spdlog.h>
 #include <unistd.h>
 
 #include <cerrno>
-#include <cstring>
-#include <stdexcept>
 
-waybar::util::Rfkill::Rfkill(const enum rfkill_type rfkill_type) : rfkill_type_(rfkill_type) {}
-
-void waybar::util::Rfkill::waitForEvent() {
-  struct rfkill_event event;
-  struct pollfd       p;
-  ssize_t             len;
-  int                 fd, n;
-
-  fd = open("/dev/rfkill", O_RDONLY);
-  if (fd < 0) {
-    throw std::runtime_error("Can't open RFKILL control device");
+waybar::util::Rfkill::Rfkill(const enum rfkill_type rfkill_type) : rfkill_type_(rfkill_type) {
+  fd_ = open("/dev/rfkill", O_RDONLY);
+  if (fd_ < 0) {
+    spdlog::error("Can't open RFKILL control device");
     return;
   }
+  int rc = fcntl(fd_, F_SETFL, O_NONBLOCK);
+  if (rc < 0) {
+    spdlog::error("Can't set RFKILL control device to non-blocking: {}", errno);
+    close(fd_);
+    fd_ = -1;
+    return;
+  }
+  Glib::signal_io().connect(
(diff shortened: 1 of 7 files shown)


alebastr:
I haven't found time for a better look at the `network.cpp`, but let's not make that block the review. Suggestions on dealing with rfkill wakeup in network module would be appreciated.
***
@rxguy, @pvalena, do you mind trying this PR? Only the first commit is required to resolve #994 and that's the only commit I'm planning to cherry-pick for Fedora package, but it'll be nice to test the rest of the changes here on another system.

pvalena:
@alebastr Sure I'd try... can you build the package in COPR for rawhide?

alebastr:
@pvalena sure, https://copr.fedorainfracloud.org/coprs/build/1956328

src/util/rfkill.cpp
```diff
@@ -19,60 +19,63 @@
...
+      spdlog::error("Reading of RFKILL events failed: {}", errno);
+      return false;
     }
 
-    if (len != RFKILL_EVENT_SIZE_V1) {
-      throw std::runtime_error("Wrong size of RFKILL event");
-      continue;
+    if (len < RFKILL_EVENT_SIZE_V1) {
+      if (errno != EAGAIN) {
+        spdlog::error("Wrong size of RFKILL event: {}", len);
```
Alexays, line 65: Do we want to return false here?

Alexays:
Just a small comment, otherwise LGTM, no suggestion for the network part, relying on the timer thread is a good idea.

src/util/rfkill.cpp
alebastr: No, that's a recoverable error that doesn't affect future reads. The logic here should follow [util-linux/sys-utils/rfkill.c](https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/sys-utils/rfkill.c#n206).
But thanks for highlighting this block; I noticed that I'm checking for `EAGAIN` in a wrong condition :smile:

Alexays:
LGTM thanks Aleksei!

PR #1018, opened 2021-02-04
[wlr/taskbar] Fix unhandled exception crash when icon name is a path.

jgmdev:
Fixes a crash and handles the edge case where the application.desktop 'Icon' property points to the absolute path of the icon file instead of an icon name, for example:

```
[Desktop Entry]
Name=Cendric
Comment=RPG and Platformer
Exec=/usr/bin/Cendric
Terminal=false
Icon=/usr/share/Cendric/res/icon.ico
Type=Application
Categories=Game;
```

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -81,8 +81,6 @@ static Glib::RefPtr<Gdk::Pixbuf> load_icon_from_file(std::string icon_path, int 
     try {
         auto pb = Gdk::Pixbuf::create_from_file(icon_path, size, size);
         return pb;
-    } catch(Glib::Error&) {
-        return {};
     } catch(...) {
         return {};
     }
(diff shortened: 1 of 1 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -74,6 +76,18 @@ static std::vector<std::string> search_prefix()
     return prefixes;
 }
 
+static Glib::RefPtr<Gdk::Pixbuf> load_icon_from_file(std::string icon_path, int size)
+{
+    try {
+        auto pb = Gdk::Pixbuf::create_from_file(icon_path, size, size);
+        return pb;
+    } catch(Glib::Error&) {
```
Alexays, line 84: No need of this catch as the next one catch all errors

Alexays (approved):

Alexays:
Thanks!

PR #1021, opened 2021-02-07
[wlr/taskbar] Check StartupWMClass

jgmdev:
Checks the StartupWMClass property of all .desktop files returned by g_desktop_app_info_search() in order to select the correct .desktop file. 

This fixes issues when applications with similar names are installed, eg: codeblocks, codelite, visual studio code, in this scenario visual studio code which appid is Code and .desktop file is visual-studio-code.desktop, without the check for StartupWMClass would end with either codeblocks or codelite icon.

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -141,8 +141,10 @@ static std::string get_from_desktop_app_info_search(const std::string &app_id)
                 auto tmp_info = Gio::DesktopAppInfo::create(desktop_list[0][i]);
                 auto startup_class = tmp_info->get_startup_wm_class();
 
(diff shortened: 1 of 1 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -137,8 +137,15 @@ static std::string get_from_desktop_app_info_search(const std::string &app_id)
         for (size_t i=0; desktop_list[0][i]; i++) {
             if(desktop_file == "") {
```
Alexays, line 138: Add missing space after if

Alexays (changes requested):
Just a small change, LGTM

Alexays:
Thanks!

PR #992, opened 2021-01-21
Fix duplicate bars on xdg_output property changes

alebastr:
Since both `name` and `description` are not mutable in the protocol version we bind to, it's fine to drop the xdg_output object after receiving an initial set of properties.

Also cleanup the code for attaching styles. There's always only one screen in Gdk so there's no reason to do that for every window.

Fixes #990

PR #1027, opened 2021-02-09
Fix more issues with duplicated bars

### LINKED ISSUE #990: Duplicated waybar on output config updates
Whenever my outputs update for whatever reason (eg changing the scale or adding or removing displays) waybar duplicates on one of the displays, I believe the one that has focus.

Here's an example:
https://user-images.githubusercontent.com/6388823/105233140-5753d800-5b61-11eb-8bff-6bc6b88d99a9.mp4

Sway & wlroots master
Waybar 66e5fda418271a077d850c2ab86e1339b83c30bb

alebastr:
In #992 I missed a certain case: If `kanshi` or something similar automatically aplies configuration when monitors are plugged, resulting xdg_output events could be received and dispatched in a single batch: initial configuration, done, configuration modified by kanshi, done [, ...repeat for each output configured by kanshi].

Also, as #1019 shows outputs may unexpectedly disappear from under our code due to async callback dispatch magic. I addressed a couple of obvious crashes, but there are more, much more... :scream:

Fixes #990 (again), makes #1019 slightly less likely.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -120,16 +120,26 @@ void waybar::Client::handleOutputDone(void *data, struct zxdg_output_v1 * /*xdg_
   auto client = waybar::Client::inst();
   try {
     auto &output = client->getOutput(data);
-    spdlog::debug("Output detection done: {} ({})", output.name, output.identifier);
+    /**
+     * Multiple .done events may arrive in batch. In this case libwayland would queue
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
LGTM, thanks again Aleksei!

PR #1032, opened 2021-02-12
Terminate custom module scripts on exit

### LINKED ISSUE #358: Child process does not stops

matteodelabre:
(Fixes #358.)

Subprocesses created for custom module scripts were previously left running when the parent Waybar process exited. This patch sets the parent-death signal of child processes (PR_SET_PDEATHSIG on Linux, PROC_PDEATHSIG_CTL on FreeBSD) to SIGTERM.

Caveats:

* This uses Linux-specific or FreeBSD-specific calls. I don’t know if this project targets other systems?
* There is a possibility that Waybar exits after calling `fork()`, but before calling `prctl` to set the parent-death signal. In this case, the child will not receive the SIGTERM signal and will continue to run. I did not handle this case as I consider it quite unlikely, since module scripts are usually launched only when Waybar starts. Please let me know if you think it needs to be handled.

Testing:

* With `htop` open, run Waybar v0.9.5 with a custom module that has an `exec` script. Terminate the Waybar process and notice that the script’s subprocess stays alive and is now a child of the init process.
* Run Waybar with this patch and follow the same steps as above. Notice that this time the script’s subprocess terminates when the parent exits.

### DIFF
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -5,6 +5,13 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#ifdef __linux__
+#include <sys/prctl.h>
+#endif
+#ifdef __FreeBSD__
+#include <sys/procctl.h>
+#endif
+
 #include <array>
 
 extern std::mutex reap_mtx;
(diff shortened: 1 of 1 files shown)


matteodelabre:
I guess the CI just answered my question about portability. :-) Apparently FreeBSD has an equivalent mechanism called PROC_PDEATHSIG_CTL, I’ll look into using that.

Alexays:
Thanks @matteodelabre!

PR #1036, opened 2021-02-19
Configure systemd.service file to allow reloading

WhyNotHugo:
This allows `systemctl --user reload waybar` to reload waybar's config as expected.


Alexays (approved):

Alexays:
Thanks!

PR #1037, opened 2021-02-20
Add cpu min/max/avg frequencies

### LINKED ISSUE #764: feature request: cpu: show frequency
For laptop users it would be great to have access to `{frequency}` in the format string.

Moonlight-Angel:
Adds format replacements for the CPU module to get the current min, max and average frequencies.

The implementation for *BSD platforms has been stubbed out for the time being, until someone on these platforms implements it.

Here's a quick example of the output it produces :
image

Closes #764

### DIFF
--- a/src/modules/cpu/bsd.cpp
+++ b/src/modules/cpu/bsd.cpp
@@ -2,8 +2,10 @@
 
 #include <sys/types.h>
 #include <sys/sysctl.h>
+#include <spdlog/spdlog.h>
 #include <cstdlib>            // malloc
 #include <unistd.h>           // sysconf
+#include <cmath>              // NAN
 
 #if defined(__NetBSD__) || defined(__OpenBSD__)
 #  include <sys/sched.h>
@@ -97,5 +99,10 @@ std::vector<std::tuple<size_t, size_t>> waybar::modules::Cpu::parseCpuinfo() {
 }
 
 std::vector<float> waybar::modules::Cpu::parseCpuFrequencies() {
-  throw std::runtime_error("Cpu frequency is not implemented on BSD.");
+  static std::vector<float> frequencies;
+  if (frequencies.empty()) {
+    spdlog::warn("cpu/bsd: parseCpuFrequencies is not implemented, expect garbage in {*_frequency}");
+    frequencies.push_back(NAN);
+  }
+  return frequencies;
 }
--- a/src/modules/cpu/bsd.cpp
+++ b/src/modules/cpu/bsd.cpp
@@ -95,3 +95,7 @@ std::vector<std::tuple<size_t, size_t>> waybar::modules::Cpu::parseCpuinfo() {
   free(cp_time);
   return cpuinfo;
 }
+
+std::vector<float> waybar::modules::Cpu::parseCpuFrequencies() {
+  throw std::runtime_error("Cpu frequency is not implemented on BSD.");
+}
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -22,7 +22,9 @@ class Cpu : public ALabel {
  private:
   uint16_t                                getCpuLoad();
   std::tuple<uint16_t, std::string>       getCpuUsage();
+  std::tuple<float, float, float>         getCpuFrequency();
   std::vector<std::tuple<size_t, size_t>> parseCpuinfo();
+  std::vector<float>                      parseCpuFrequencies();
 
   std::vector<std::tuple<size_t, size_t>> prev_times_;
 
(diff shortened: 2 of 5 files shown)


David96:
Just throwing ideas out there but maybe it'd be good to use the min/max of the frequencies and provide an option for which to use? My usecase would be to see whether my laptop is throttling the CPU which could easily be seen if the max frequency is constantly under some threshold.
Imo this provides more information than just picking one core at random.

Moonlight-Angel:
@David96 You're totally right, so I started again from the beginning :joy: 
I'm now getting the frequency for all CPU cores (using `/proc/cpuinfo`), and providing the minimum, maximum and average frequency to the formatter. :+1:

alebastr:
For FreeBSD `dev.cpu.0.freq` and `dev.cpu.0.freq_levels` sysctl variables should provide the necessary information. See [`cpufreq(4)`](https://www.freebsd.org/cgi/man.cgi?query=cpufreq&sektion=4&apropos=0&manpath=FreeBSD+12.2-stable) for the details. Defaulting these to 0 if unavailable sounds reasonable.

I can't help with testing though (the only FreeBSD I have right now is a VM), so maybe @jbeich can offer more help.

Moonlight-Angel:
@alebastr Thanks for the hint. I tried to check how it's implemented in Conky, since their source code has helped me for the Linux implementation.
However, it looks like there is a different way to get the frequency on [FreeBSD](https://github.com/brndnmtthws/conky/blob/e4fad17dc0bb69745236f1425ccf88c998a95c6c/src/freebsd.cc#L501), [OpenBSD](https://github.com/brndnmtthws/conky/blob/e4fad17dc0bb69745236f1425ccf88c998a95c6c/src/openbsd.cc#L555), [Dragonfly](https://github.com/brndnmtthws/conky/blob/e4fad17dc0bb69745236f1425ccf88c998a95c6c/src/dragonfly.cc#L507) and NetBSD (NetBSD doesn't even have it implemented in Conky).
Moreover, their code only queries one CPU core (`dev.cpu.0.freq`, as you said) and I would like to query all CPU cores to avoid having different behaviors on platforms. I don't know if I can query `dev.cpu` and filter out the lines containing the frequency, just like I'm doing it on Linux :shrug: 
All in all, it all seems too blurry for me to be able to implement this properly on all *BSD platforms, without being able to test it live, and honestly I don't want to invest more time into researching this and installing *BSD distros to test and check things out :confused:

src/modules/cpu/bsd.cpp
```diff
@@ -95,3 +95,7 @@ std::vector<std::tuple<size_t, size_t>> waybar::modules::Cpu::parseCpuinfo() {
   free(cp_time);
   return cpuinfo;
 }
+
+std::vector<float> waybar::modules::Cpu::parseCpuFrequencies() {
+  throw std::runtime_error("Cpu frequency is not implemented on BSD.");
```
jbeich, line 100: Try turning error into warning and maybe replace zero with something easy to notice e.g.,
```c++
#include <spdlog/spdlog.h>
#include <cmath>              // INFINITY
```
```suggestion
  static std::vector<float> frequencies;
  if (frequencies.empty()) {
    spdlog::error("cpu/bsd: parseCpuFrequencies is not implemented");
    frequencies.push_back(INFINITY);
  }
  return frequencies;
```

jbeich (commented):
> NetBSD doesn't even have it implemented in Conky

See [PkgSrc patch](https://github.com/NetBSD/pkgsrc/blob/ab0a39d05e0f6ba09365448ca3a09f6208e415e2/sysutils/conky/patches/patch-src_netbsd.c#L554). Less popular BSDs (or community driven OSes) don't upstream as much partially due to lack of manpower (turning layers of hacks into into PRs is hard work) and partially due to users having lower expectations for software outside of system packages.

> I don't want to invest more time into researching this and installing *BSD distros to test and check things out 😕

Stubbing is fine. I or someone else may try implementing it later without time pressure.

Moonlight-Angel:
Thanks @jbeich, I applied your suggestion, and thanks for the link to the NetBSD patch :+1:

Alexays (approved):

Alexays:
Thanks @Moonlight-Angel, Do you plan to implement the BSD part? If not, I will probably merge the PR as it 😃

Moonlight-Angel:
> Thanks @Moonlight-Angel, Do you plan to implement the BSD part? If not, I will probably merge the PR as it smiley

I have no plans on implementing it on *BSD platforms, so you can merge it if you're ok with it :+1:

Alexays:
Ok, thanks @Moonlight-Angel!

PR #1050, opened 2021-03-12
Fix: CPU load values

### LINKED ISSUE #769: Incorrect values for CPU load format replacement?
The CPU `{load}` format replacement doesn't show the same values as `uptime` and `top`, for example.

The latter two are what is mostly presented as the load metric. However, in Waybar, the value obtained with `getloadavg(load, 1)` is scaled with the number of CPUs to show it as a percentage of CPU usage.

This seems to be a very rare way to present the load value, and I also think that it does not make so much sense as there is already the `{usage}` format.

Is there any reason for it?

cmovcc:
Remove the division of the load by the number of online (or available) processors.

The current displayed value is quite counter-intuitive and hard to get a grasp on as it is different from the values of `uptime` and `top`; moreover there already is the other `{usage}` format, as argued in #769 by @gvccandido.
Furthermore, this (very small) fix would lead Waybar CPU load output to be the same as the one displayed by `i3status`.
Another possibility would be to keep the current format and add a new format for this "more standard" load.

This would fix #769 .
Thank you for Waybar, it is a very useful software I use daily.

### DIFF
--- a/include/modules/cpu.hpp
+++ b/include/modules/cpu.hpp
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <fmt/format.h>
-#include <unistd.h>
 #include <cstdint>
 #include <fstream>
 #include <numeric>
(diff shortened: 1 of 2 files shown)


Alexays (approved):

Alexays:
Agree, let's forget the old format, thanks!

PR #1051, opened 2021-03-13
Add default styling for disk module

lunik1:
This patch adds some default styling for the disk module, which previously did not have any styling specified. It uses a brown background since brown isn't used by any other modules.


Alexays:
Thanks!

PR #1052, opened 2021-03-17
meson.build: add missing waybar-sway-language manpage

Logarithmus:
There's *.scd file for sway/language module's manpage, but it's not mentioned in meson.build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -315,6 +315,7 @@ if scdoc.found()
         'waybar-network.5.scd',
         'waybar-pulseaudio.5.scd',
         'waybar-river-tags.5.scd',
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -315,6 +315,7 @@ if scdoc.found()
         'waybar-network.5.scd',
         'waybar-pulseaudio.5.scd',
         'waybar-river-tags.5.scd',
+		'waybar-sway-language.5.scd',
```
Alexays, line 318: There's an indent problem

Logarithmus: Sorry, fixed.

Alexays:
Thanks!

PR #1055, opened 2021-03-21
Add option to rewrite sway/window title

vrld:
Rewrites window title according to config option "rewrite".

"rewrite" is an object where keys are regular expressions and values are rewrite rules if the expression matches. Rules may contain references to captures of the expression. Regex and replacement follow ECMA-script rules. If no regex matches, the title is left unchanged.

Useful to add icons depending on the window title, e.g.:
```json
"sway/window": {
  "rewrite": {
    "(.*) - Mozilla Firefox": " $1",
    "(.*) - zsh": " [$1]",
  }
}
```

Examples:
- Firefox:
  -  before before patch
  -  patched after patch
- Terminal / zsh: 
  - before before patch
  - patched after patch

### DIFF
--- a/include/modules/sway/window.hpp
+++ b/include/modules/sway/window.hpp
@@ -22,6 +22,7 @@ class Window : public ALabel, public sigc::trackable {
   std::tuple<std::size_t, int, std::string, std::string> getFocusedNode(const Json::Value& nodes,
                                                                         std::string&       output);
   void                                                   getTree();
+  std::string                                            rewriteTitle(const std::string& title);
 
   const Bar&       bar_;
   std::string      window_;
--- a/man/waybar-sway-window.5.scd
+++ b/man/waybar-sway-window.5.scd
@@ -66,12 +66,32 @@ Addressed by *sway/window*
 	default: true ++
 	Option to disable tooltip on hover.
 
+*rewrite*: ++
+	typeof: object ++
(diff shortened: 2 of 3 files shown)


vrld:
Failed test seems to be caused by the runner, not by the changed code:

> ERROR: Clock skew detected. File /Users/runner/work/Waybar/Waybar/build/../subprojects/date-3.0.0/meson_options.txt has a time stamp 10.6111s in the future.

vrld:
@Alexays is anything missing?

Alexays:
LGTM, can you update the man in order to add the change?
Thanks!

Alexays (approved):

Synthetica9:
Very cool, using this locally and it's working like a charm! +1 for this feature.

Synthetica9 (approved):

Synthetica9 (changes requested):
Actually, I'd like to request 1 change: if the regex is malformed in some way, it's probably better to keep the window title unchanged rather than crapping out completely and not setting the window title at all as is currently the case.

vrld:
@Synthetica9 Good catch. Added try/catch around the regex stuff to fix this issue

@Alexays Added documentation in manpage and rebased on current master.

Alexays:
Thx! ❤️ 

vrld:
@Alexays done :)

PR #1056, opened 2021-03-22
Add style for battery state "plugged"

martin2250:
Hi Alex,

the default style does not apply the green background to the battery module if the status is reported as "plugged" (eg. when the charge limiter is active). I think re-using the style currently used for "charging" is a better default.

Cheers,
Martin


Alexays (approved):

Alexays:
Thanks!

PR #1061, opened 2021-03-28
Use the correct battery status when multiple batteries are present

akheron:
Use the most relevant battery status, instead of the status of the last battery. The most relevant is the first of charging, discharging, not charging, full or unknown, in this order. So, e.g. when one of the batteries is charging and others are not charging or full, report use charging as the status. Or, when one of the batteries is discharging and others are full, use discharging as the status.


Alexays (approved):

Alexays:
Nice, thanks @akheron

PR #1070, opened 2021-04-14
[Module CPU] fix crash due to empty frequencies.

jgmdev:
On some systems (eg: ARM) the supported frequencies of the CPU are not properly reported by /proc/cpuinfo so use lscpu instead which is more reliable.

### DIFF
--- a/src/modules/cpu/linux.cpp
+++ b/src/modules/cpu/linux.cpp
@@ -1,3 +1,4 @@
+#include <filesystem>
 #include "modules/cpu.hpp"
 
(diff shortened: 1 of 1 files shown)


Alexays:
I'm not a big fan of having to execute a command

jgmdev:
I thought the same... will have to check lscpu source code to see how it is retrieving the min/max cpu frequencies :)

jgmdev:
Ok, rewrote the commit to use /sys/devices/system/cpu/cpufreq/policy[0-9]/cpuinfo_[max|min]_freq so no command execution now.

Alexays (approved):

Alexays:
LGTM, will release a new version with this fix, thx!

PR #1081, opened 2021-04-25
[modules/pulseaudio] fix bluetooth class for PipeWire

David96:
Apparently, pipewire-pulse slightly changed the naming of the sink.

To me it looks like this name parsing is in general quite prone to this kind of bugs but I don't know pulseaudio/pipewire and what additional ways it might provide for this kind of information so for now I just sticked to the way it was already handled anyway.


Alexays (approved):

Alexays:
Thx!

PR #1084, opened 2021-04-26
[modules/battery] allow format-discharging-full

### LINKED ISSUE #1031: Battery has no discharging status
On my system (Slimbook pro X 15), `/sys/class/power_supply/BAT0/status` has 3 possible states: 
* Discharging : power adapter is not plugged in
* Charging : power adapter is plugged in and battery is not full
* Full : power adapter is plugged in and battery is full 

In waybar (version v0.9.5-92-g08e1960), `format-discharging` is never active, and `format-full` is active when battery is at 100% even if the state in  `/sys/class/power_supply/BAT0/status` says "Discharging". I think this might be a bug, since the documentation states that the `status` should correspond to the one in  `/sys/class/power_supply/BAT0/status`. The corresponding CSS styles are not triggered as well.

gabegorelick:
`format-discharging-full` has been impossible since #923 made it impossible to be full and discharging at the same time. This should fix that by only making `format-charging-full` impossible. Whether or not that should be allowed is a good question, but beyond the scope of this change.

Fixes #1031

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -216,7 +216,7 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
       capacity = 100.f;
     }
     uint8_t cap = round(capacity);
-    if (cap == 100) {
+    if (cap == 100 && status == "Charging") {
       // If we've reached 100% just mark as full as some batteries can stay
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
LGTM, thx !

PR #1087, opened 2021-04-26
rewriteTitle: allow multiple sequential rewrites

Synthetica9:
For example, before if I had

```json
"rewrite": {
  "a": "x",
  "b": "y"
}
```

and my window title was `ab` it would get rewritten to `xb`, with this PR it's `xy` (for a specific use-case, I use it to rewrite my home dir to ~ and add icons without having really weird regex logic)

Open question: do we want to have this match in the middle of titles? You can easily opt out of that behaviour with `^$` but it's a lot harder to opt in (you'd need something like `(.*)pattern(.*) => $1...$n`)

### DIFF
--- a/src/modules/sway/window.cpp
+++ b/src/modules/sway/window.cpp
@@ -132,12 +132,13 @@ void Window::getTree() {
   }
 }
 
-std::string Window::rewriteTitle(const std::string& title)
-{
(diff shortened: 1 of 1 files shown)


Alexays (approved):

Alexays:
LGTM, thx !

Issue #1098, opened 2021-05-09
Not showing right battery level, fix proposed


Max1Truc:
On the Pinebook Pro, running Manjaro ARM and Sway, the battery module shows 100% or 0% randomly, and never shows the right amount of battery.

I tracked the problem to the absence of both charge_now and energy_now in `/sys/class/power_supply/cw2015-battery/` ( See [battery.cpp](https://github.com/Alexays/Waybar/blob/cdce3e03ea9c22ef8825d53001b84520e3e32c98/src/modules/battery.cpp#L167) )

I fixed it by reading capacity instead of energy_now and charge_now if they are not available.

If capacity was read, I then multiply the variable `energy_now` by `energy_full` and divide the result by 100.

```
$ ls -l /sys/class/power_supply/cw2015-battery/
total 0
-r--r--r-- 1 root root 4096  9 mai   10:18 capacity
-r--r--r-- 1 root root 4096  9 mai   10:51 charge_counter
-r--r--r-- 1 root root 4096  9 mai   10:18 charge_full
-r--r--r-- 1 root root 4096  9 mai   10:18 charge_full_design
-r--r--r-- 1 root root 4096  9 mai   10:18 current_now
lrwxrwxrwx 1 root root    0  9 mai   10:37 device -> ../../../4-0062
drwxr-xr-x 3 root root    0  4 févr.  2013 hwmon4
drwxr-xr-x 2 root root    0  9 mai   10:51 power
-r--r--r-- 1 root root 4096  9 mai   10:18 present
-r--r--r-- 1 root root 4096  4 févr.  2013 status
lrwxrwxrwx 1 root root    0  4 févr.  2013 subsystem -> ../../../../../../../class/power_supply
-r--r--r-- 1 root root 4096  9 mai   10:18 technology
-r--r--r-- 1 root root 4096  9 mai   10:51 time_to_empty_now
-r--r--r-- 1 root root 4096  4 févr.  2013 type
-rw-r--r-- 1 root root 4096  4 févr.  2013 uevent
-r--r--r-- 1 root root 4096  9 mai   10:18 voltage_now
drwxr-xr-x 2 root root    0  4 févr.  2013 wakeup12
```

Patch I created:
```patch
diff --git a/src/modules/battery.cpp b/src/modules/battery.cpp
index e598858..75b033e 100644
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
[...]

PR #1101, opened 2021-05-10
fix: incorrect battery percentage on Pinebook Pro

Max1Truc:
As discussed on #1098 , little fix for pinebook pro laptops.


Alexays (approved):

danielrparks:
This breaks the battery display on my system. Now it's showing half the percentage it should be. That's weird, because I looked at the code changes and there's no way it should have affected the battery calculation with my battery.

PR #1105, opened 2021-05-15
Fix power calculation when battery units are in μA instead of μW

Amanieu:
`current_now` needs to be multiplied by `voltage_now` to get the current power usage in `μW`.


Alexays:
Thanks!

PR #1113, opened 2021-05-22
Add config options for exclusive zone and input event passthrough

### LINKED ISSUE #1091: Non exclusive bar.
Is it possible to make the bar's surface non exclusive?

alebastr:
So I sneakily added the `overlay` layer value in one of the previous PRs, but it wasn't behaving as a real `overlay` mode of `swaybar`. Here's the two missing bits: input passthrough flag that makes the bar ignore any pointer events and allows the window underneath to handle those, and a flag to disable the exclusive zone.

Possible uses:
 - A bar on the `bottom` layer with `"exclusive": false` which is only visible on an empty workspace.
 - A transparent bar on the `overlay` or `top` layer which shows some data over the application window, but cannot be interacted with.
 - If you look closely at the swaybar `hide` mode (the one where you only see a bar when the binding is pressed), it actually does not set exclusive zone but still accepts input.
   Now we can get _almost_ similar behavior with creative use of `bindsym --no-repeat`, `bindsym --no-repeat --release` and `SIGUSR1`. With possibility of races, bar states out of sync between different screens and other wonderful bugs.

Fixes #1091

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -65,6 +65,7 @@ class Bar {
   struct waybar_output *output;
   Json::Value           config;
   struct wl_surface *   surface;
+  bool                  exclusive = true;
   bool                  visible = true;
   bool                  vertical = false;
   Gtk::Window           window;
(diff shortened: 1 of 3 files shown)


Alexays (approved):

Alexays:
Thanks again @alebastr ❤️ 
Can you update the Github wiki with these man changes?

alebastr:
> Can you update the Github wiki with these man changes?

Updated.

PR #1116, opened 2021-05-27
Fix network module rework

tperard:
- one fix for wifi network appearing as disconnected after some time.
- one improvement, to fetch wifi essid as soon as available instead of after some time.


Alexays (approved):

Alexays:
Thanks again @tperard, you rock!

PR #1125, opened 2021-06-05
sway-language: ignore events with empty layout

maximbaz:
When using e.g. [wtype](https://github.com/atx/wtype), which also sends events to emulate keyboard presses, this confuses Waybar and incorrectly hides the module. This fix will ensure compatibility with `wtype`, simply ignoring events where layout is not set, and only reacting on events which actually change the keyboard layout.


Alexays:
Thanks Maxim!

PR #1126, opened 2021-06-05
Fix network interface auto detection

tperard:
That should fix the interface auto detection when NetworkManager is used.


Alexays (approved):

Alexays:
Thanks for all your changes!

Issue #1065, opened 2021-04-09
Waybar Crash (Fedora 34)


sclel016:
Hi,

The problem signature is reproducible with the Waybar 0.9.5-4 available through DNF and using HEAD on master (97e4b53).

I'm running latest SwayWM on Fedora 34. In a dual monitor setup, Waybar crashes when one of the displays in disabled. I've attached the debug logs from a crash where I disable my laptops internal display while connected to an external monitor.

I'm happy to help in any way possible, love the project!

Stew
[waybar.log](https://github.com/Alexays/Waybar/files/6282473/waybar.log)

martinpitt:
I get that crash on either changing displays (like docking/undocking Thunderbolt) or suspend/resume. When starting wayland in a terminal:
```
(waybar:97366): dbind-WARNING **: 10:28:56.027: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.NameHasNoOwner: Could not activate remote peer.
[2021-04-11 10:28:56.047] [info] Resources files: /home/martin/.config/waybar/config, /etc/xdg/waybar/style.css
[2021-04-11 10:28:56.055] [error] media stopped unexpectedly, is it endless?
[2021-04-11 10:28:56.055] [warning] module battery#bat2: Disabling module "battery#bat2", No battery named BAT2
[2021-04-11 10:28:56.055] [warning] For a functional tray you must have libappindicator-* installed and export XDG_CURRENT_DESKTOP=Unity

** (waybar:97366): WARNING **: 10:28:56.060: gtk-layer-shell v0.6.0 may not work on GTK v3.24.28. If you experience crashes, check https://github.com/wmww/gtk-layer-shell/blob/master/compatibility.md
[2021-04-11 10:28:56.204] [info] Bar configured (width: 1920, height: 17) for output: eDP-1
```

then close / reopen lid:
```
(waybar:97366): GLib-GIO-WARNING **: 10:29:12.512: Invalid id 2 passed to g_bus_unwatch_name()
terminate called without an active exception
Aborted (core dumped)
```

martinpitt:
Ignore me please -- "my" crash is fixed in current master (97e4b53cf3b as well), and but suspending and changing displays work well. Time for another release? There are dozens of commits since 0.9.5

Thank you!

uaftab:
Hi

I am also facing the same crash as reported by the original poster. Waybar 0.9.5-4 available through DNF in SwayWM on Fedora 34. 

[...]

PR #1132, opened 2021-06-12
fix(util): protect std::condition_variable methods from pthread_cancel

alebastr:
The changes in GCC 11.x made `std::condition_variable` implementation
internals `noexcept`. `noexcept` is known to interact particularly bad
with `pthread_cancel`, i.e. `__cxxabiv1::__force_unwind` passing through
the `noexcept` call stack frame causes a `std::terminate` call and
immediate termination of the program

Digging through the GCC ML archives[1] lead me to the idea of patching
this with a few pthread_setcancelstate's. As bad as the solution is, it
seems to be the best we can do within C++17 limits and without major
rework.

[1]: https://gcc.gnu.org/legacy-ml/gcc/2017-08/msg00156.html

Fixes: #1065
***
I'm running the patch for last few days and the crashes are gone. Let's ask more people to test before merging though.
Fedora build: https://copr.fedorainfracloud.org/coprs/build/2251643

### DIFF
--- a/include/util/sleeper_thread.hpp
+++ b/include/util/sleeper_thread.hpp
@@ -8,6 +8,20 @@
 
 namespace waybar::util {
 
+/**
+ * Defer pthread_cancel until the end of a current scope.
+ *
+ * Required to protect a scope where it's unsafe to raise `__forced_unwind` exception.
+ * An example of these is a call of a method marked as `noexcept`; an attempt to cancel within such
+ * a method may result in a `std::terminate` call.
+ */
+class CancellationGuard {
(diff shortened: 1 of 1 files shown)


danielrparks:
Works for me.

danielrparks:
I wonder if you can use pthread's thread cancellation handlers to call stop() before the unwind exception is generated.

Alexays (approved):

Alexays:
LGTM as users reporting that it fixes their problems!
Thanks again @alebastr

Issue #1143, opened 2021-06-26
Waybar compile problem with libfmt8


yibaba:
Hello, I'm leaving this here for the sake of visibility, but some problems have arised in gentoo with waybar. Here is the bugpost.

https://bugs.gentoo.org/797649

Sorry for not being that helpful.

PR #1144, opened 2021-06-30
libfmt >=8.0.0 compatibility

ajakk:
Attempting to fix compatibility with libfmt-8.0.0 mentioned in #1143 and https://bugs.gentoo.org/797649. I don't really have any idea what I'm doing in C++, so criticism welcome (and probably necessary).

The issue with parse_nonnegative_int comes from the libfmt change [here](https://github.com/fmtlib/fmt/commit/e421d527), where the third argument becomes an integer. Previously on error the function returned -1, now it returns the third argument, which I've changed to -1 here. I'm unsure about side effects of the ctx object no longer being passed into `parse_nonnegative_int`. This function was also moved from `fmt/format.h` to `fmt/core.h` and namespaces seem to have changed along the way, so specifying `fmt::detail::` makes it compile (for me) but not sure if there's a better way.

`tm_format` was also apparently moved out of libfmt's public API. Declare it similarly here. This seems to work at runtime for me.

### DIFF
--- a/include/util/format.hpp
+++ b/include/util/format.hpp
@@ -35,7 +35,11 @@ namespace fmt {
             // The rationale for ignoring it is that the only reason to specify
             // an alignment and a with is to get a fixed width bar, and ">" is
             // sufficient in this implementation.
+#if FMT_VERSION < 80000
             width = parse_nonnegative_int(it, end, ctx);
+#else
+            width = detail::parse_nonnegative_int(it, end, -1);
+#endif
           }
           return it;
         }
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -196,6 +196,9 @@ template <>
 struct fmt::formatter<waybar_time> : fmt::formatter<std::tm> {
   template <typename FormatContext>
   auto format(const waybar_time& t, FormatContext& ctx) {
+#if FMT_VERSION >= 80000
+	auto& tm_format = specs;
+#endif
     return format_to(ctx.out(), "{}", date::format(t.locale, fmt::to_string(tm_format), t.ztime));
   }
 };


src/modules/clock.cpp
```diff
@@ -196,6 +196,7 @@ template <>
 struct fmt::formatter<waybar_time> : fmt::formatter<std::tm> {
   template <typename FormatContext>
   auto format(const waybar_time& t, FormatContext& ctx) {
+    basic_memory_buffer<char> tm_format;
```
alebastr, line 199: This breaks clock formatting when a timezone is specified.

I'd try following, but let me think more on that...
```suggestion
#if FMT_VERSION >= 80000
    auto& tm_format = specs;
#endif
```

meson.build
```diff
@@ -79,7 +79,7 @@ is_netbsd = host_machine.system() == 'netbsd'
 is_openbsd = host_machine.system() == 'openbsd'
 
 thread_dep = dependency('threads')
-fmt = dependency('fmt', version : ['>=5.3.0'], fallback : ['fmt', 'fmt_dep'])
+fmt = dependency('fmt', version : ['>=8.0.0'], fallback : ['fmt', 'fmt_dep'])
```
alebastr, line 82: This breaks compilation with subprojects as fmt 8.0.0 is not yet available from wrapdb.
Also, I believe it's possible to fix compilation in a way that does not require latest fmt on a rolling release.

include/util/format.hpp
```diff
@@ -35,7 +35,7 @@ namespace fmt {
             // The rationale for ignoring it is that the only reason to specify
             // an alignment and a with is to get a fixed width bar, and ">" is
             // sufficient in this implementation.
-            width = parse_nonnegative_int(it, end, ctx);
+            width = fmt::detail::parse_nonnegative_int(it, end, -1);
```
alebastr, line 38:
```suggestion
#if FMT_VERSION < 80000
            width = parse_nonnegative_int(it, end, ctx);
#else
            width = detail::parse_nonnegative_int(it, end, -1);
#endif
```

alebastr (changes requested):
Overall, I'd avoid touching anything related to fmt before the pending 8.0.1 release with regression fixes. I doubt it's going to fix anything affected here, but it might change even more things.

meson.build
ajakk: I agree, my bad!

alebastr (approved):
LGTM, thanks for the patch!

Since CI is taking a break, I'll do its job and confirm that the change compiles and works on Fedora both with old and new fmt.

ajakk:
Thank you for reviewing! Seems like CI needs to be approved for first time contributors to prevent abuse. Hopefully this helps certain rolling release downstreams in any case

Alexays:
Thanks for the patch!

PR #1145, opened 2021-06-30
tray module enhancements

### LINKED ISSUE #1138: High CPU usage in tray module
old

https://github.com/Alexays/Waybar/blob/d08fbb2ef2b2a2cada05fd64e51ffe041aabb47c/src/modules/sni/item.cpp#L169-L176

This is caused by `value.equal(old_value)`.Maybe an upstream issue, the implement of `g_variant_equal` is compare them by their pretty-printing string, which use many CPU cycles.(This is the source code: https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gvariant.c#L2736)

I remove this compare and the CPU usage reduce nearly to zero:

```c++
    for (const auto& [name, value] : properties) {
        proxy_->set_cached_property(name, value);
        setProperty(name, const_cast<Glib::VariantBase&>(value));
    }
```
( But this give me a lot warning : `[2021-06-22 23:58:50.837] [warning] Failed to set tray item property: Fcitx.WindowId, value = 0, err = std::bad_cast` )

new

alebastr:
Assorted enhancements for tray module.

I had to drop the tooltip icon changes with intent to finish those later, once I have a complete grasp of a possible implementation.
Need to take another look at `update_pending_` for possibility of races. Shouldn't be a blocker for the merge though, as I haven't found anything harmful so far. Only a minuscule chance of missing a property update.

### Changes:

 - Try to optimize updates by assuming that signals don't lie and could be using to infer all modified properties. A somewhat risky assumption, given how broken is an average sni client implementation, but I've seen other trays working like that.
 - Support tooltips for tray items (text only).
 - Support `Scroll`  SNI method.
 - Support `Status` property: hide `Passive` items by default; attach corresponding CSS classes to the icon.

Fixes #1138, #1150

### DIFF
--- a/src/modules/sni/item.cpp
+++ b/src/modules/sni/item.cpp
@@ -105,8 +105,6 @@ void Item::setProperty(const Glib::ustring& name, Glib::VariantBase& value) {
       title = get_variant<std::string>(value);
     } else if (name == "Status") {
       status = get_variant<std::string>(value);
-    } else if (name == "WindowId") {
-      window_id = get_variant<int32_t>(value);
     } else if (name == "IconName") {
       icon_name = get_variant<std::string>(value);
     } else if (name == "IconPixmap") {
--- a/src/modules/sni/item.cpp
+++ b/src/modules/sni/item.cpp
@@ -51,10 +51,15 @@ Item::Item(const std::string& bn, const std::string& op, const Json::Value& conf
   if (config["smooth-scrolling-threshold"].isNumeric()) {
     scroll_threshold_ = config["smooth-scrolling-threshold"].asDouble();
   }
+  if (config["show-passive-items"].isBool()) {
+    show_passive_ = config["show-passive-items"].asBool();
+  }
   event_box.add(image);
   event_box.add_events(Gdk::BUTTON_PRESS_MASK | Gdk::SCROLL_MASK | Gdk::SMOOTH_SCROLL_MASK);
   event_box.signal_button_press_event().connect(sigc::mem_fun(*this, &Item::handleClick));
   event_box.signal_scroll_event().connect(sigc::mem_fun(*this, &Item::handleScroll));
+  // initial visibility
+  event_box.set_visible(show_passive_);
 
   cancellable_ = Gio::Cancellable::create();
 
@@ -81,7 +86,7 @@ void Item::proxyReady(Glib::RefPtr<Gio::AsyncResult>& result) {
 
     this->proxy_->signal_signal().connect(sigc::mem_fun(*this, &Item::onSignal));
 
-    if (this->id.empty() || this->category.empty() || this->status.empty()) {
+    if (this->id.empty() || this->category.empty()) {
       spdlog::error("Invalid Status Notifier Item: {}, {}", bus_name, object_path);
       return;
     }
@@ -127,7 +132,7 @@ void Item::setProperty(const Glib::ustring& name, Glib::VariantBase& value) {
         event_box.set_tooltip_markup(title);
       }
     } else if (name == "Status") {
-      status = get_variant<std::string>(value);
+      setStatus(get_variant<Glib::ustring>(value));
     } else if (name == "IconName") {
       icon_name = get_variant<std::string>(value);
     } else if (name == "IconPixmap") {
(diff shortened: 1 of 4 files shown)


alebastr:
@12101111, do you mind checking if the CPU usage is acceptable with this PR?
I haven't managed to get a good data with `perf`, likely because most of my profiling experience is with other tools and I don't have access to any of those at home :frowning_face:

src/modules/sni/item.cpp
```diff
@@ -121,7 +147,10 @@ void Item::setProperty(const Glib::ustring& name, Glib::VariantBase& value) {
     } else if (name == "AttentionMovieName") {
       attention_movie_name = get_variant<std::string>(value);
     } else if (name == "ToolTip") {
-      // TODO: tooltip
+      tooltip = get_variant<ToolTip>(value);
+      if (!tooltip.text.empty()) {
+        event_box.set_tooltip_markup(tooltip.text);
```
Alexays, line 152: ❤️

src/modules/sni/item.cpp
```diff
@@ -167,33 +209,48 @@ void Item::processUpdatedProperties(Glib::RefPtr<Gio::AsyncResult>& _result) {
...
 }
 
+/**
+ * Mapping from a signal name to a set of possibly changed properties.
+ * Commented signals are not handled by the tray module at the moment.
+ */
+static const std::map<std::string_view, std::set<std::string_view>> signal2props = {
+    {"NewTitle", {"Title"}},
+    {"NewIcon", {"IconName", "IconPixmap"}},
+    // {"NewAttentionIcon", {"AttentionIconName", "AttentionIconPixmap", "AttentionMovieName"}},
```
Alexays, line 233: No need of theses mapping?

Alexays (approved):
Apart from some questions, LGTM
Thanks!

src/modules/sni/item.cpp
alebastr: Commented mappings are things we don't support at the moment. I.e. no reason to process AttentionIconPixmap updates if we aren't capable of displaying it.

Alexays: Ok got it

12101111:
> @12101111, do you mind checking if the CPU usage is acceptable with this PR?
> I haven't managed to get a good data with `perf`, likely because most of my profiling experience is with other tools and I don't have access to any of those at home frowning_face

Sorry for the late, the CPU usage is nearly to 0 and perf result looks great.
图片

rieje:
The issue that prompted #1150 was `udiskie`'s `--smart-tray` option which only shows the icon if when actions are available. Prior to this fix, it will always show the icon. With this fix, the icon is appropriately hidden *except* on the first instance of waybar, where it will always show the icon. Killing and starting another instance correctly shows the icon as hidden. 

I have waybar autostarted but even if disable autostart and start waybar for the first time at a later time, the tray icon is incorrectly showing until I kill the instance and start another. I am on latest version 0.9.8.

alebastr:
@rieje can you please check `waybar -l trace` output when running the first time? Does it get the right `Status` from udiskie?

rieje:
Ok, just did some quick testing and it does not happen only on first time. It happens randomly. I run multiple waybars to see what happens (the status is `passive` on all these instances which is correct), and on each instance that starts, it may trigger show/hide of the udiskie icon on all other waybars independently. [-l trace output](full log](https://pastebin.com/EGv1XEL4) of the typical instance.

alebastr:
I believe it's not so random. I managed to reproduce the issue, and everything is right from our side: we're calling `event_box.set_visible(false)` for the udiskie tray item and subsequent `event_box.get_visible()` returns `false` as well.

Do you happen to notice if the issue is reproducible when udiskie is not the first icon in the tray? My suspicion is that the first item in a GtkBox cannot be hidden, but I'll need to check GTK sources to confirm that.

rieje:
Yes, still happens if not the first (leftmost) icon. I see it as the second icon, and less frequently as the third (last) icon.

alebastr:
So I got some time to debug the issue and found that it's our fault. I just never looked at the code dealing with updates on the whole tray module level and missed the right (or wrong in this case) line.

@rieje do you mind giving https://github.com/alebastr/Waybar/commit/dc67bcc3bd5d8e61920e52fef1a81c7bef7bccfe a try? Worked for me, but I only did limited testing.

rieje:
This works, I have not experienced any unexpected results at all. Looks fixed!

PR #1157, opened 2021-07-10
Do not fail to parse a multi-bar config

### LINKED ISSUE #1152: Configuration of multiple outputs fails
The configuration syntax for multiple outputs causes waybar to fail on start with a JSON related error.

The syntax is explained in [the wiki here][syntax-explanation]. Essentially, it says to put your config JS object into an array and use the `output` property as an output selector.

### Version
* My setup is: Linux kernel 5.12.9, sway 1.6-31b4b96e, and alacritty 0.8.0.
* I use the waybar-git AUR package. I get this error at the latest waybar commit (7c1303f), but not at a commit from last month (2016074). I hope that helps track down the culprit LOC.

### Error message
```bash
$ waybar --log-level trace
[2021-07-03 17:12:40.746] [info] Resources files: /home/ian/.config/waybar/config, /home/ian/.config/waybar/style.css
[2021-07-03 17:12:40.746] [error] in Json::Value::resolveReference(key, end): requires objectValue
```

### Steps to reproduce
1. Build waybar from source using the latest commit (7c1303f)
2. Save the following minimal config file. (I took it from [this example][minimal-config-example] in the wiki).
    ```js
    {
        "layer": "top",
        "modules-left": ["sway/workspaces", "sway/mode"],
        "modules-center": ["sway/window"],
        "modules-right": ["battery", "clock"],
        "sway/window": {
            "max-length": 50
        },
        "battery": {
            "format-icons": ["", "", "", "", ""]
        },
        "clock": {
            "format-alt": "{:%a, %d. %b  %H:%M}"
        }
    }
    ```
2. Apply the syntax for multiple outputs, ie. put the JS object into an array. Realistically, a one element config array isn't an interesting example, but the next example shows that multiple config objects fail as well.
    ```js
    // ~/.config/waybar/config
    // Example-1
    [{
        "layer": "top",
        "modules-left": ["sway/workspaces", "sway/mode"],
        "modules-center": ["sway/window"],
        "modules-right": ["battery", "clock"],
        "sway/window": {
            "max-length": 50
        },
        "battery": {
            "format-icons": ["", "", "", "", ""]
        },
        "clock": {
            "format-alt": "{:%a, %d. %b  %H:%M}"
        }
    }]
    ```
3. Restart waybar and startup will fail
4. Apply a more interesting multi-output config.
    ```js
    // ~/.config/waybar/config
    // Example-2
    [{
        "layer": "top",
        "output": "YOUR-OUTPUT-1",
        "modules-left": ["sway/workspaces", "sway/mode"],
        "modules-center": ["sway/window"],
        "modules-right": ["battery", "clock"],
        "sway/window": {
            "max-length": 50
        },
        "battery": {
            "format-icons": ["", "", "", "", ""]
        },
        "clock": {
            "format-alt": "{:%a, %d. %b  %H:%M}"
        }
    },
    {
        "layer": "top",
        "output": "YOUR-OUTPUT-2",
        "modules-left": ["sway/workspaces", "sway/mode"],
        "modules-center": ["sway/window"],
        // NOTE: clock isn't shown on this output
        "modules-right": ["battery"],
        "sway/window": {
            "max-length": 50
        },
        "battery": {
            "format-icons": ["", "", "", "", ""]
        }
    }]
    ```
5. Restart waybar and startup will fail

[syntax-explanation]: https://github.com/Alexays/Waybar/wiki/Configuration#configuration-of-multiple-outputs
[minimal-config-example]: https://github.com/Alexays/Waybar/wiki/Configuration#minimal-config

tiosgz:
Fixes #1152.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -245,22 +245,50 @@ auto waybar::Client::setupConfig(const std::string &config_file, int depth) -> v
   std::string      str((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
   util::JsonParser parser;
   Json::Value      tmp_config_ = parser.parse(str);
-  if (tmp_config_["include"].isArray()) {
-    for (const auto &include : tmp_config_["include"]) {
-      spdlog::info("Including resource file: {}", include.asString());
-      setupConfig(getValidPath({include.asString()}), ++depth);
+  if (tmp_config_.isArray()) {
+    for (auto &config_part : tmp_config_) {
+      resolveConfigIncludes(config_part, depth);
     }
+  } else {
+    resolveConfigIncludes(tmp_config_, depth);
   }
   mergeConfig(config_, tmp_config_);
 }
 
-auto waybar::Client::mergeConfig(Json::Value &a_config_, Json::Value &b_config_) -> void {
-  for (const auto &key : b_config_.getMemberNames()) {
-    if (a_config_[key].type() == Json::objectValue && b_config_[key].type() == Json::objectValue) {
-      mergeConfig(a_config_[key], b_config_[key]);
-    } else {
-      a_config_[key] = b_config_[key];
+auto waybar::Client::resolveConfigIncludes(Json::Value &config, int depth) -> void {
+  Json::Value includes = config["include"];
+  if (includes.isArray()) {
+    for (const auto &include : includes) {
+      spdlog::info("Including resource file: {}", include.asString());
+      setupConfig(getValidPath({include.asString()}), ++depth);
     }
+  } else if (includes.isString()) {
+    spdlog::info("Including resource file: {}", includes.asString());
+    setupConfig(getValidPath({includes.asString()}), ++depth);
+  }
+}
+
+auto waybar::Client::mergeConfig(Json::Value &a_config_, Json::Value &b_config_) -> void {
+  if (!a_config_) {
+    // For the first config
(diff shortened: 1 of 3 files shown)


colemickens:
I don't want to hold up this fix being merged but it'd be great to have a test here to prevent regressions in the future.

tiosgz:
I wonder if there is a test suite? Looking everywhere I can find only that Waybar is built, but not tested.

src/client.cpp
```diff
@@ -245,22 +245,50 @@ auto waybar::Client::setupConfig(const std::string &config_file, int depth) -> v
...
+    }
+  } else if (a_config_.isArray() && b_config_.isArray()) {
+    // This can happen only on the top-level array of a multi-bar config
+    for (Json::Value::ArrayIndex i = 0; i < b_config_.size(); i++) {
+      if (a_config_[i].isObject() && b_config_[i].isObject()) {
+        mergeConfig(a_config_[i], b_config_[i]);
+      }
     }
+  } else {
+    spdlog::error("Cannot merge config, conflicting or invalid JSON types");
```
tiosgz, line 291: Maybe I should overwrite A with B here, because B is the config which called the include and A the configs included so far?

Alexays: Keep the error, and if it's happened let's overwrite A with B

Alexays:
Thanks for the fix!

PR #1169, opened 2021-07-20
pulseaudio: Control currently running sink

roosemberth:
In a system with multiple sinks, the default sink may not always be
the once currently being used. It is more useful to control the
currently active sink rather than an unused one.

This patch does not make any difference if the system only uses the
default sink.

Signed-off-by: Roosembert Palacios <roosemberth@posteo.ch>


Alexays:
Thanks!

Issue #1077, opened 2021-04-20
Blurry tray icons


BrightYC:
I have using a Hi-DPI screen, and at x2 scaling in sway i have this:

![20210420_16h25m33s_grim](https://user-images.githubusercontent.com/28766913/115432517-fa1a0200-a1f5-11eb-895f-2e5eaaa36086.png)

Waybar v0.9.7-5-g1c9b62d (branch 'master')

Tray icons are perfectly scaled on swaybar, but not in waybar :(

PR #1174, opened 2021-07-21
Fix blurry tray icons for HiDPI displays

mswiger:
Fixes https://github.com/Alexays/Waybar/issues/1077. The approach here is to scale the icon size using the Gtk scale factor, create a cairo surface from the loaded pixbuf, and use that cairo surface for the image. The cairo surface will use the scale factor from the given window, which makes everything render correctly.

Screenshots:
before
after


Alexays:
Thanks!

Issue #1175, opened 2021-07-24
Tray icons with incorrect size on fractionally scaled output


minhduc0711:
The tray icons on the scaled monitor are weirdly sized on my scaled output
![image](https://user-images.githubusercontent.com/34294427/126855153-7863dbe9-ed3a-4764-8958-df417b6a6640.png)
On my other unscaled output, they appear normally
![image](https://user-images.githubusercontent.com/34294427/126855166-c84d709c-e9d1-4e25-a3db-10b911b38d0d.png)
Could it be related to https://github.com/Alexays/Waybar/pull/1174?

Version: `Waybar v0.9.7-107-g1b4ddbc (branch 'master')`

luispabon:
Having the same problem:

![image](https://user-images.githubusercontent.com/6388823/127145633-065241b6-bcc6-410c-8f18-9687e425893a.png)

the wifi signal icon (nm-applet) is 1/2 the size on a scale 2 display. This is from a recent update.

Mathew-D:
If you go back to before commit https://github.com/Alexays/Waybar/commit/811f0896c9965de2672f03a1fcf673b0781a270d that fixes the error and the icons are the correct size.  Something in that commit breaks the size on some icons.

Mathew-D:
Also it gives this error if you use the master: 
gtk_widget_set_accel_path: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed

luispabon:
Reverting to 67d482d fixes it for me as well.

Alexays:
@mswiger can you take a look?

minhduc0711:
There's a small issue left: upon first launching the bar, the icons are way bigger than it should be. They would go back to their normal size if I do something that change the icons' states (e.g. change my ibus input method).
![image](https://user-images.githubusercontent.com/34294427/127799621-53bf1456-bf75-4f03-91e5-cf0810f9ec55.png)

mswiger:
I am able to reproduce the issue that @minhduc0711 is seeing, but only with a multiple-display setup, and it does not reproduce 100% of the time (maybe like 25% of the time for me). It is relatively easy to reproduce by repeatedly reloading the Sway config file until the issue appears ($mod+Shift+c for me).

After spending a bunch of time debugging, this bug seems to behave like some kind of race condition. For tray items that take longer to populate in the tray, this issue never seems to happen. For tray items that populate quickly after waybar is loaded, this problem seems to happen pretty consistently.

[...]

PR #1190, opened 2021-08-01
Fix incorrect tray icon scaling

mswiger:
This PR fixes https://github.com/Alexays/Waybar/issues/1175 and refactors the code around updating the image to separate fetching the correct pixmap and scaling the pixmap.

The root cause of https://github.com/Alexays/Waybar/issues/1175 is that for certain icons, such as the network manager wireless signal icons, the requested icon size may not be available (the network manager applet only supplies a 22x22 pixel version of the wireless signal icon). There is some logic in `updateImage` for proportionally scaling the icon pixbuf if it did not match the requested icon size. However, the code branch in `updateImage` where the icon was loaded using `getIconByName` did not apply this resize logic, resulting in situations where, e.g., the pixbuf is too small.

In this PR, `updateImage` has been updated to apply the logic for proportionally resizing the icon pixbuf to all pixbufs. Additionally, the logic for fetching the correct pixbuf for a given item has been factored out into `getIconPixbuf`.

Side note, it does appear that the network manager applet [supplies "symbolic" SVG icons](https://gitlab.gnome.org/GNOME/network-manager-applet/-/merge_requests/84), which are scalable monochrome icons, but it's not clear to me how to use these symbolic icons in the Waybar tray module.

### Screenshot (2.0 scale factor)
screenshot

### DIFF
--- a/include/modules/sni/item.hpp
+++ b/include/modules/sni/item.hpp
@@ -66,7 +66,9 @@ class Item : public sigc::trackable {
 
   void                      updateImage();
   Glib::RefPtr<Gdk::Pixbuf> extractPixBuf(GVariant* variant);
+  Glib::RefPtr<Gdk::Pixbuf> getIconPixbuf();
   Glib::RefPtr<Gdk::Pixbuf> getIconByName(const std::string& name, int size);
+  double                    getScaledIconSize();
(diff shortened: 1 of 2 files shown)


Alexays:
LGTM

PR #1194, opened 2021-08-06
Fix sway/language to handle xkb_layout command

gavinbeatty:
This change fixes a few issues:

- Fix to handle users changing layouts via "absolute" commands like:
    `swaymsg input type:keyboard xkb_layout es`
  which previously evaluated {long}, {short}, {variant} as empty.
  Only "relative" commands like:
    `swaymsg input type:keyboard xkb_switch_layout next`
  were handled.
  Previously, layout information was only stored internally if it was
  one of the available layouts on the device with the most such layouts.
  Even though memory was allocated (and leaked, see below) for every possible
  configurable layout, it wasn't stored internally in a way we could
  look up later. Now, all are stored internally for later lookup, and
  leaks plugged.

- Simplify how layouts are stored internally, and how events are handled.
  I admit I didn't understand the modifications to short_name to append
  an integer and the lookup by short name logic, but I have removed it.
  To simplify and reduce (but not eliminate) the incidence of the observed
  layout flickering when a stream of events are coming in,
  the following heuristic is applied:
  - initially, pick the "identifier" that has the most layouts, if any.
  - keep track of the `event_count`, i.e., how many times `full_name`
    has changed for each identifier.
  - when handling events, if the new `event_count` is greater than that
    of the previously chosen identifier, switch to the new identifier.

- Fix the above memory leaks of all Layout instances, one for every
  layout configured on the system.

- Fix potential data race due to lack of lock in `update()`.
  It is called from a different thread than `onCmd` and `onEvent`.
  Also take the lock in the constructor, to guarantee config occurs before IPC.

- Handle errors in parsing the default xkb ruleset config.
  The result is that we can't enumerate any layouts though.

- Remove unused layout_ member in XKBContext.

### DIFF
--- a/include/modules/sway/language.hpp
+++ b/include/modules/sway/language.hpp
@@ -5,6 +5,7 @@
 
 #include <map>
 #include <string>
+#include <unordered_map>
 
 #include "ALabel.hpp"
 #include "bar.hpp"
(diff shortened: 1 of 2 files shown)


gavinbeatty:
A trivial demonstration is up at https://youtu.be/gDNsym6f7O4

Before this change, the xkb_layout "es" and "hu" changes would result in no layout information being displayed, instead of "Spanish" and "Hungarian" respectively.

gavinbeatty:
My error to not check for existing sway/language PRs!

- Re: #1192, my changes here match the existing format, wild or not, as the case may be. I can adapt this PR to rebase on it and fix new lines' formatting, no problem.
- Re: #984, I actually have a branch containing such a change (not yet published), with the following differences from that PR:
  - added documentation (example config and manpage/scdoc)
  - format-icons mapping is keyed on {long} not the first two characters of {long}
  - mapping of icons is built in constructor -- in general, I see a lot of `operator[]` calls on mutable `Json::Value` instances, and that call can allocate, so I prefer to use `const auto` for such values instead of just `auto`, and create a simple `map<string, string>` in the constructor and just use `find` on it.

PR #1211, opened 2021-08-17
Fix tray icon scaling on multi-display setups

mswiger:
Fixes another trailing issue from https://github.com/Alexays/Waybar/issues/1175#issuecomment-890679048 where the incorrect tray icon size was sometimes used on multi-display setups. This issue seems to occur only on multi-display setups, and it seems to happen when tray items are added before the bar has been properly configured for both displays. This fix adds an event handler to the `signal_configure_event` in the tray `Item` class that calls `updateImage`, which results in the image being correctly resized when the bar's window configuration changes.

Before this change, I was able to repro the issue 1/3 of the time (out of 30 tries). Afterwards, I am unable to repro.

### DIFF
--- a/include/modules/sni/host.hpp
+++ b/include/modules/sni/host.hpp
@@ -5,13 +5,15 @@
 #include <glibmm/refptr.h>
 #include <json/json.h>
 #include <tuple>
+#include "bar.hpp"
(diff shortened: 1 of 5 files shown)


Alexays:
Thanks!

PR #1216, opened 2021-08-19
fix: Add number to shortDescripton

Anakael:
Add number to the end of shortDescription if several variants for layouts and they aren't displayed.


Alexays:
You're on fire, thanks!

PR #1217, opened 2021-08-20
Fix memory leak and data race

gavinbeatty:
- Delete previous Layout before creating next one, and in destructor
- Use stack XKBContext instead of local new+delete
- Lock mutex in update() as it is called from a different thread than onEvent(res)

### DIFF
--- a/src/modules/sway/language.cpp
+++ b/src/modules/sway/language.cpp
@@ -94,6 +94,7 @@ void Language::onEvent(const struct Ipc::ipc_response& res) {
 }
 
 auto Language::update() -> void {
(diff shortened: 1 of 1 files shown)


gavinbeatty:
I see sway/language is being actively developed, so just suggesting a minimal fix for the leak and data race instead of #1194.

gavinbeatty:
Based on looking at some other open PRs, it looks like continuous-integration/travis-ci doesn't finish, so presumably this has passed CI.

Alexays:
lgtm, thanks!

PR #1222, opened 2021-08-22
Use g_memdup2 instead of g_memdup

mswiger:
This fixes the following compiler warning:

```
[89/91] Compiling C++ object ....p/src_modules_sni_item.cpp.
../src/modules/sni/item.cpp: In member function ‘Glib::RefPtr<Gdk::Pixbuf> waybar::modules::SNI::Item::extractPixBuf(GVariant*)’:
../src/modules/sni/item.cpp:290:48: warning: ‘void* g_memdup(gconstpointer, guint)’ is deprecated: Use 'g_memdup2' instead [-Wdeprecated-declarations]
  290 |           array = static_cast<guchar*>(g_memdup(data, size));
      |                                        ~~~~~~~~^~~~~~~~~~~~
In file included from /usr/include/glib-2.0/glib.h:82,
                 from /usr/include/glib-2.0/gobject/gbinding.h:28,
                 from /usr/include/glib-2.0/glib-object.h:22,
                 from /usr/include/glib-2.0/gio/gioenums.h:2,
                 from /usr/include/glib-2.0/gio/giotypes.h:2,
                 from /usr/include/glib-2.0/gio/gio.h:26,
                 from protocol/dbus-status-notifier-item.h:1,
                 from ../include/modules/sni/item.hpp:3,
                 from ../src/modules/sni/item.cpp:1:
/usr/include/glib-2.0/glib/gstrfuncs.h:257:23: note: declared here
  257 | gpointer              g_memdup         (gconstpointer mem,
      |                       ^~~~~~~~
```

It seems that the code in `item.cpp` is indeed passing a `gsize` to `g_memdup` rather than the expected `guint`. The new function takes a `gsize` instead.

See: https://discourse.gnome.org/t/port-your-module-from-g-memdup-to-g-memdup2-now/5538

### DIFF
--- a/src/modules/sni/item.cpp
+++ b/src/modules/sni/item.cpp
@@ -287,7 +287,11 @@ Glib::RefPtr<Gdk::Pixbuf> Item::extractPixBuf(GVariant* variant) {
           if (array != nullptr) {
             g_free(array);
           }
+#if GLIB_MAJOR_VERSION >= 2 && GLIB_MINOR_VERSION >= 68
           array = static_cast<guchar*>(g_memdup2(data, size));
+#else
+          array = static_cast<guchar*>(g_memdup(data, size));
+#endif
           lwidth = width;
           lheight = height;
         }
--- a/src/modules/sni/item.cpp
+++ b/src/modules/sni/item.cpp
@@ -287,7 +287,7 @@ Glib::RefPtr<Gdk::Pixbuf> Item::extractPixBuf(GVariant* variant) {
           if (array != nullptr) {
             g_free(array);
           }
-          array = static_cast<guchar*>(g_memdup(data, size));
+          array = static_cast<guchar*>(g_memdup2(data, size));
           lwidth = width;
           lheight = height;
         }


alebastr:
Note that g_memdup2 is a new API introduced in glib 2.68 earlier this year. My concerns with this PR:
1. The change will break compilation on a distributions with older versions of glib (Ubuntu 20.10 or below, Debian testing, Fedora 33, etc)
2. The updated requirement should be in meson.build, so that the build could fail early and with better diagnostics.

Alternatively, we can just use `-DGLIB_VERSION_MIN_REQUIRED=...` to silence the warning. gsize -> gint conversion overflow is highly unlikely in this code as we're reading an image from GVariant container received through DBus.
Seems to be working with lowest available value `GLIB_VERSION_2_26`.

mswiger:
Agreed with your concerns, as does the build pipeline! Rather than restricting which deprecation warnings are shown by setting GLIB_VERSION_MIN_REQUIRED, what do you think about adding a compile-time check to use the correct/available version of `g_memdup` based on the linked GLib version? It is a bit ugly, to be sure, but I think it communicates clearly that the code calling `g_memdup` is deprecated and should hopefully be deleted eventually.

I've updated the PR with this approach, let me know what you think!

Alexays:
Thanks!

PR #1229, opened 2021-08-25
`sway/language` country flag

kraftwerk28:


Alexays:
Sorry for the delay, can you update the PR in order to fix conflict? :)

include/modules/sway/language.hpp
```diff
@@ -32,6 +32,7 @@ class Language : public ALabel, public sigc::trackable {
     std::string short_name;
     std::string variant;
     std::string short_description;
+    std::string country_flag() const;
```
Anakael, line 35: I think we can use country name from `short_name` in case multi-language layouts.
It also implement this https://github.com/Alexays/Waybar/issues/1245.

kraftwerk28:
Done. Added some additional checking to prevent rendering flags for non-language-code layouts (such as `mao`)

Issue #610, opened 2020-03-02
network: bandwidth speed is wrong, shows as sum over all interfaces


maximbaz:
Currently placeholders like `{bandwidthDownBits}` show the sum across all interfaces, this is a problem if you use a VPN and so all your traffic goes through tunnel interface and then a wireless interface, the numbers shown by waybar are doubled comparing to real internet speed.

I would like to ask that all these placeholders show information about the currently shown interface name.

I tried to enforce the interface name via config `"interface": "wlan0"` but this doesn't affect these placeholders.

captainju:
Relevant code is in src/modules/network.cpp
It currently uses `/proc/net/netstat` which gives only global bandwidth speed.
It should use `/proc/net/dev` for having info per interface.

I'll try to look into it if I can (I'm not familiar with C++).

kushraj:
yeah, It's even adding the speed of localhost ( playing local videos on browser using jellyfin )

b10n:
Not sure if it's related but for me the reported speed is too low (~6% with VPN, ~3% without).
I'm using Waybar v0.9.7-35-g94a882b on Linux 5.12.3-arch1-1.

ciarand:
I spent some time trying to make the changes to read from /proc/net/dev, but I'm not super comfortable with c++ and struggled to incorporate these changes into the project as it is.

In case it's helpful for the next person that tries to tackle this, here's a standalone program that prints the total bandwidth used for the specified interface https://gist.github.com/ciarand/568e6e8d99aa2f1ba0d489ecba9b0d68

Feel free to steal/copy this code if it's helpful (or ignore it if it's not)

captainju:
Looks fixed to me, thanks @ciarand !

PR #1230, opened 2021-08-25
Switch network module to read /proc/net/dev

ciarand:
This fixes issue #610 by reading bandwidth usage per-interface from
/proc/net/dev instead of globally via /proc/net/netstat. It supports the
same matching logic as elsewhere, so setting interface to '*' should
display the same sum-total bandwidth usage as the previous
implementation.


droc12345:
Added this to my copy of git on Nov 19, seems to be working fine. 
More realistic values for up/download, I use a tun adapter and it now show proper with what looks like valid values.

Alexays:
Thanks!

PR #1236, opened 2021-09-07
Add environment condition to systemd service file

jtheoof:
Sometimes the waybar service fails to start because the wayland environment is not fully setup. I would get errors such as:

```
Sep 06 21:21:40 ansan waybar[4847]: Unable to init server: Could not connect: Connection refused
Sep 06 21:21:40 ansan waybar[4847]: cannot open display:
Sep 06 21:21:40 ansan systemd[4753]: waybar.service: Main process exited, code=exited, status=1/FAILURE
```

With this `ConditionEnvironment` we make sure systemd waits until the `WAYLAND_DISPLAY` env variable becomes available.

See the equivalent in [mako](https://github.com/emersion/mako/blob/master/contrib/systemd/mako.service#L7) which works fine all the time.

### DIFF
--- a/resources/waybar.service.in
+++ b/resources/waybar.service.in
@@ -4,6 +4,8 @@ Documentation=https://github.com/Alexays/Waybar/wiki/
 PartOf=graphical-session.target
 After=graphical-session.target
 Requisite=graphical-session.target
+# ConditionEnvironment requires systemd v247 to work correctly
+ConditionEnvironment=WAYLAND_DISPLAY
 
 [Service]
 ExecStart=@prefix@/bin/waybar


jtheoof:
Sadly, this doesn't work as well as mako. Not sure why. 

I get:

```
Sep 06 21:40:50 ansan systemd[1320]: Condition check resulted in Highly customizable Wayland bar for Sway and Wlroots based compositors. being skipped.
```
And systemd never retries it. Not sure how this works for mako.

jtheoof:
@emersion sorry to bother you. Any idea why this patch on `waybar.service` doesn't behave like `mako.service`? With mako, initially the conditions are not met (see logs) but a few seconds after they are met (when sway finished setup). However with `waybar.service` the conditions are not "retried" and the waybar never shows. With my patch both `waybar.service` and `mako.service` are nearly identical so I wonder what's happening. I also tried to remove the `Requisite` but no luck either. A little help would be appreciated.

```
❯ journalctl -b --user -eu mako.service
Sep 06 21:56:04 ansan systemd[1265]: Condition check resulted in Lightweight Wayland notification daemon being skipped.
Sep 06 21:56:40 ansan systemd[1265]: Starting Lightweight Wayland notification daemon...
Sep 06 21:56:40 ansan systemd[1265]: Started Lightweight Wayland notification daemon.
```

```
❯ journalctl -b --user -eu waybar.service
Sep 06 21:40:50 ansan systemd[1320]: Condition check resulted in Highly customizable Wayland bar for Sway and Wlroots based compositors. being skipped.
```

emersion:
Sorry, I don't know.

PR #1238, opened 2021-09-08
round brightness

Matan1x:
round the brightness percentage instead of flooring it and displaying inaccurate results


Alexays:
Thanks!

PR #1239, opened 2021-09-09
Added support for {gwaddr} as a variable

mazunki:
I didn't find any way to include the gateway without altering the code. So I did. I greped all the files for `ipaddr`, hoping I catched all the conventions.

I also added a new makefile target (`debug-run`), just because I was annoyed it didn't exist already.

Having `"format-ethernet": "{ipaddr} / {cidr}", "tooltip-format": "{ifname} via {gwaddr}"` feels good and natural. Check the `resources/config` I modified.

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -662,7 +662,7 @@ int waybar::modules::Network::handleEvents(struct nl_msg *msg, void *data) {
         net->ifid_ = temp_idx;
         net->route_priority = priority;
         net->gwaddr_ = temp_gw_addr;
-        spdlog::debug("network: new default route via if{} metric {}", temp_idx, priority);
+        spdlog::debug("network: new default route via {} on if{} metric {}", temp_gw_addr, temp_idx, priority);
 
         /* Ask ifname associated with temp_idx as well as carrier status */
         struct ifinfomsg ifinfo_hdr = {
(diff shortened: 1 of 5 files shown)


src/modules/network.cpp
```diff
@@ -655,6 +661,8 @@ int waybar::modules::Network::handleEvents(struct nl_msg *msg, void *data) {
         net->clearIface();
         net->ifid_ = temp_idx;
         net->route_priority = priority;
+	net->gwaddr_ = temp_gw_addr;
+	spdlog::debug("netwok: gateway {}", net->gwaddr_);
```
Alexays, line 665: Can you remove this debug and indent `net->gwaddr_ = temp_gw_addr;`?

Alexays (changes requested):

mazunki:
Thanks for checking it out, @Alexays . Commit 5f08319 updates the changes requested. 

I also fixed the Makefile dependency from default build to debug build. The 6142dfb is a suggestion, to include the gateway IP in the debug log, feel free to take it or leave it.

As a note, I recently updated my kernel, and there seems to be an issue with reaching the case block for the `RTA_GATEWAY`, which worked when I pushed the commit.

While it still works on the standalone script [I initially tested on](https://gist.github.com/mazunki/e44181731294baad3bb1c8fc7cb7667f), it doesn't seem to work through the Waybar script.

I haven't yet looked much into what's going on here.

Alexays:
Thanks!

mazunki:
Perfect! I updated the wiki too. I'm surprised there weren't any permission limitations for editing it. Is that a default behaviour of GitHub?

PR #1221, opened 2021-08-22
[Taskbar] feat: Add name format replacementf feat: Add app ids mapping (ready for review)

Anakael:
Implements:
https://github.com/Alexays/Waybar/issues/982
https://github.com/Alexays/Waybar/issues/1028
https://github.com/Alexays/Waybar/issues/1135

Issue #1135, opened 2021-06-14
[wlr/Taskbar] Class name to icon mapping


Anakael:
In some case application's class not equal to icon name for such application.
For example: firefox-developer-edition has `firefoxdeveloperedition` class in sway VM and use `firefox-developer-edition` icon.

So, I suggest add option to `wrl/Taskbar` module to specify icon name for such classes.
Example:
```
...
class_icon: {
  "firefoxdeveloperedition": "firefox-developer-edition"
}
...
```

It can also be useful to overwrite icon for some application.

sjoblomj:
@Anakael : Looks like this issue can be closed as it was implemented in #1221

Alexays:
Closing — this is fixed/implemented on current master and will be in the next release. Please reopen if it still occurs after updating. 🙏

Issue #1028, opened 2021-02-09
[wlr/taskbar] icon name


kushraj:
Will it be okay to replace `app_id` with `Name` from desktop entry ? Lot of apps use weird naming convention like Nautilus(org.gnome.Nautilus) which  is not something that should be displayed on taskbar .

Its also fine if `Name` could be added to wlr-taskbar as a separete element , but I don't see the point of using app_id as it don't see any naming convention except for no spaces.

WillPower3309:
Duplicate of #982 , and similar functionality already exists. Use `title`
```
    "wlr/taskbar": {
        "format": "{icon} {title}",
        "icon-size": 13,
        "tooltip": false,
        "on-click": "activate",
        "on-click-right": "close"
    },
```

kushraj:
using `title` isn't really intuiting, we already have window module for it. Even opening two browser/editor windows with `title` messes up entire taskbar because of big names, and its really a hassle because waybar display all windows across workspaces.

kushraj:
As somebody else said earlier, `app_info->get_display_name()` should be made default option as name which could fallback upon app_id if not available

Alexays:
Closing as a duplicate of #982 (also already achievable via {title}). 🙏

Issue #982, opened 2021-01-15
[Taskbar Feature Request] Add a {name} field


qoheniac:
A `{name}` field that works like this: if a `.desktop` file is found it returns the value of `Name` and `app_id` else.

savchenko:
Sounds like it can be covered as the part of https://github.com/Alexays/Waybar/issues/1000

WillPower3309:
similar functionality already exists. Try `title`. Its not the .desktop filename but better than `app_id` if you'd like a solution that works now
```
    "wlr/taskbar": {
        "format": "{icon} {title}",
        "icon-size": 13,
        "tooltip": false,
        "on-click": "activate",
        "on-click-right": "close"
    },
```

qoheniac:
Thank you, but I knew about {title} and I prefer {app_id}. On the one hand {title} is often very long and for some programs on the other hand it's empty.

savchenko:
@WillPower3309 , neat tweak, thanks for sharing!

@qoheniac , maybe having `max-length` defined on per-module basis would help? It seems like `wlr/taskbar` already does 90% of it.

qoheniac:
(A maybe capitalized version of) `{app_id}` (or a shortened version of `{title}`) would be OK for many applications but for many others this produces misleading and/or ugly labels. The expected label would be the value of Name from the `.desktop` file as that's what launchers use and I think this should be possible as `wlr/taskbar` is already looking for the desktop files searching for icons. So if it found a `.desktop` file it could also read the `Name` field and provide that through `{name}`.

qoheniac:
Here is what I mean:

https://github.com/Alexays/Waybar/blob/08e19602f75a2e13087024db6c65bf7177a59b41/src/modules/wlr/taskbar.cpp#L112

This `app_info` should already contain the value of `Name` if I understand it correctly.

qoheniac:
`app_info->get_display_name()->to_string()` should return what I would like to be provided through a `{name}` placeholder.

WillPower3309:
If this isn't implemented by end of April I can look into it then (exams will be done then)

WillPower3309:
Did some investigation because needed a study break:p
[...]

PR #1240, opened 2021-09-11
[Taskbar] Add name field

WillPower3309:
Implements:
#982
#1028
#1135

Realize as I was finishing this that this is basically an alternative solution to #1221 , however decided to post this as a PR anyway, as it seems less intrusive as an alternate solution.

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -45,7 +45,6 @@ static std::string trim(const std::string& s)
 }
 
 
-/* Icon loading functions */
 static std::vector<std::string> search_prefix()
 {
     std::vector<std::string> prefixes = {""};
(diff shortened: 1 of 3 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -76,18 +75,7 @@ static std::vector<std::string> search_prefix()
...
-        auto pb = Gdk::Pixbuf::create_from_file(icon_path, size, size);
-        return pb;
-    } catch(...) {
-        return {};
-    }
-}
-
-/* Method 1 - get the correct icon name from the desktop file */
-static std::string get_from_desktop_app_info(const std::string &app_id)
+Glib::RefPtr<Gio::DesktopAppInfo> gen_app_info(const std::string app_id)
```
Anakael, line 78: Why `gen*` instead of `get*`?

src/modules/wlr/taskbar.cpp
```diff
@@ -383,6 +390,15 @@ void Task::handle_app_id(const char *app_id)
 {
     app_id_ = app_id;
 
+    Glib::RefPtr<Gio::DesktopAppInfo> app_info = gen_app_info(app_id);
```
Anakael, line 393: In case missing icon and name options it will cost unnecessary overhead. And it will be searched twice, as app_info was't saved.

WillPower3309:
Closed due to #1221 being merged

PR #1241, opened 2021-09-12
Updated sway-hide work to the latest master

### LINKED ISSUE #255: Only show waybar when modifier key is pressed

emirror-de:
Updates the previous work by @somini and @xPMo to the latest master branch.

Fixes #255

Changes:
* set_opacity has been removed, as it seems to raise an address boundary sigsegv
* The bar does not reserve any space on the screen when the hide mode is enabled (value is being set on initialization)
* When shown, the bar is on the `TOP` layer, when hidden on the `BOTTOM`


src/modules/sway/hide.cpp
```diff
@@ -0,0 +1,52 @@
...
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (payload.isMember("mode")) {
+    // barconfig_update: get mode
+    auto mode = payload["mode"].asString();
+    if (mode == "hide") {
+      // Hide the bars when configuring the "hide" bar
+      for (auto& bar : waybar::Client::inst()->bars) {
+        bar->setVisible(false);
+      }
+    }
```
xPMo, line 29: With swaybar, users can easily toggle whether their bar is in `hide` mode or `dock` (normal) mode.
```
bindsym $binding bar mode toggle
```
This alternates between `dock` and `hide` mode. When in `hide` mode, we should:

- Set the layer to `bottom`
- Remove the exclusive zone

When in `dock` mode, we should:

- Set the layer to the user-configured layer
- Re-enable the exclusive zone

xPMo:
Excellent, it works! I'm fine with merging this as is and improving it later, but I can see one swaybar feature that we should definitely address sooner than later. [I left a comment on the file](https://github.com/Alexays/Waybar/pull/1241/files#r707698671).

src/modules/sway/hide.cpp
```diff
@@ -0,0 +1,52 @@
...
+  } else if (payload.isMember("visible_by_modifier")) {
+    // bar_state_update: get visible_by_modifier
+    visible_by_modifier_ = payload["visible_by_modifier"].asBool();
+    spdlog::info("WayBar Shown: {}", visible_by_modifier_);
```
xPMo, line 33: This gets pretty noisy by default, and `"WayBar"` is redundant. My suggestion:

```
spdlog::debug("sway/hide: visible: {}", visible_by_modifier_);
```

A `spdlog::info` could be useful above when we get a `mode` change though.

src/bar.cpp
```diff
@@ -594,15 +594,10 @@ auto waybar::Bar::setVisible(bool nvis) -> void {
   visible = nvis;
   if (!visible) {
     window.get_style_context()->add_class("hidden");
-    window.set_opacity(0);
   } else {
     window.get_style_context()->remove_class("hidden");
-    window.set_opacity(1);
```
alebastr: You are fixing the symptom and also breaking existing features (show/hide by signal).

`set_opacity` should not fail on a correct visible Gtk window. The only cases of crash I can imagine without looking at the backtrace are when the window is not yet initialized or already destroyed.
Can you try checking `window.get_realized()` or `window.get_mapped()` right after setting the `visible` property?

src/bar.cpp
```diff
@@ -607,6 +590,26 @@ void waybar::Bar::handleSignal(int signal) {
...
+    window.get_style_context()->add_class("hidden");
+    surface_impl_->setLayer(bar_layer::BOTTOM);
+  } else {
+    window.get_style_context()->remove_class("hidden");
+    surface_impl_->setLayer(bar_layer::TOP);
+  }
+  wl_surface_commit(surface);
+}
+
+auto waybar::Bar::removeExclusiveZone() const -> void {
```
alebastr, line 605: I'd suggest to do it like this:
```cpp
void waybar::Bar::setExclusive(bool value) {
  exclusive = value;
  surface_impl_->setExclusiveZone(exclusive && visible);
} 
```

src/bar.cpp
```diff
@@ -607,6 +590,26 @@ void waybar::Bar::handleSignal(int signal) {
...
 }
 
+auto waybar::Bar::setVisible(bool nvis) -> void {
+  visible = nvis;
+  if (!visible) {
+    window.get_style_context()->add_class("hidden");
+    surface_impl_->setLayer(bar_layer::BOTTOM);
+  } else {
+    window.get_style_context()->remove_class("hidden");
+    surface_impl_->setLayer(bar_layer::TOP);
```
alebastr, line 600: This overrides the layer set in the config file.
You can implement that differently:
```cpp
void waybar::Bar::setLayer(bar_layer layer) {
  layer_ = layer;
  surface_impl_->setLayer(layer_);
}
```
and call the method from the same location where you are disabling exclusive zone.

src/bar.cpp
```diff
@@ -607,6 +590,26 @@ void waybar::Bar::handleSignal(int signal) {
...
+  }
+  wl_surface_commit(surface);
```
alebastr, line 602: Extraneous `wl_surface_commit` is known to be able to crash Gtk in certain cases (as found and fixed in gtk-layer-shell). Also, surface can be unset you call that on an unmapped window.
That's why the code in master branch is calling `surface_impl_->commit()` which is noop for the most cases.

include/bar.hpp
```diff
@@ -58,8 +58,9 @@ class Bar {
   Bar(const Bar &) = delete;
   ~Bar() = default;
 
-  void setVisible(bool visible);
-  void toggle();
+  auto setVisible(bool nvis) -> void;
+  auto toggle() -> void;
```
alebastr, line 62: I know I'm nitpicking, but why was that change necessary?

alebastr (commented):
Certain changes in bar.cpp are breaking the existing functionality.
I also don't feel comfortable with reverting `setVisible` to an implementation before all my changes.

include/bar.hpp
emirror-de: No that's totally justified. I took the wrong definition while resolving the merge conflicts. I will correct this.

src/modules/sway/hide.cpp
emirror-de: Yes, that's right. Will correct this.

src/bar.cpp
emirror-de: Sorry by introducing breaking changes, that was not my intention. I will reset it to the current master branch implementation.

src/modules/sway/hide.cpp
```diff
@@ -0,0 +1,63 @@
...
+      spdlog::debug("sway/hide: hiding bar(s)");
+      for (auto& bar : waybar::Client::inst()->bars) {
+        bar->setVisible(false);
+        bar->setExclusive(false);
+      }
+    } else if (mode == "dock") {
+      spdlog::info("sway/hide: showing bar(s)");
+      for (auto& bar : waybar::Client::inst()->bars) {
+        bar->setVisible(true);
+        bar->setExclusive(true);
```
emirror-de, line 32: Would it be better to apply this once on initialization for every bar?

src/modules/sway/hide.cpp
emirror-de: I merged your fork but replaced the functions in favor of @alebastr suggestion.

src/bar.cpp
emirror-de: I reset the changes to the setVisible function. Calling `get_realized` or `get_mapped` did not make a difference.

The following error occurs when `set_opacity` is in the function and is used when the modifier key is pressed.
``` 
fish: “./build/waybar” terminated by signal SIGSEGV (Address boundary error)
```

alebastr:
@emirror-de I just recovered and updated my old attempt on this feature: #1244.
Do you mind checking if you see the same crash on that branch? If you do, please share your config and I'll try to reproduce.

> in my opinion this should be an "either, or" decision for the user to use the signal or the module

I agree with that, because otherwise the state gets inconsistent with sway internal state. And as I did in my approach, `mode` value from IPC should be able to override user config for `layer`, `exclusive` and `passthrough`. I.e. waybar is either controlled via config and signals or ipc `barconfig_update`/`bar_state_update`.

> This use case is not possible with the current state, as the toggle gets applied to all. Maybe this can be extended later.

This is an interesting use case. You need set bar `id` for each entry in the multi-bar config and then filter IPC messages by `id`.  And convince sway that you have multiple bars running via placeholder `swaybar_command` if you need to control more that one bar with IPC.

One thing to be aware of: if you connect new output, a new `waybar::Bar` instance will be created with the initial configuration and state and desired mode won't apply until the next IPC update. That could be addressed by reading and applying initial barconfig before spawning worker thread (see my branch for example).

emirror-de:
@alebastr thanks for your response.

> I just recovered and updated my old attempt on this feature: #1244.

I merged your attempt to this PR (locally) and pushed it to [this branch](https://github.com/emirror-de/Waybar/tree/sway-hide-ipc-pr-1244). Hope that is what you meant?
I tested it and still get the `Address boundary error`, however it occurs only if I spam the modifier key as fast as I can. That is indeed not the way it should be used, but it is still a bug/thing to consider I think.

> `mode` value from IPC should be able to override user config for `layer`, `exclusive` and `passthrough`

As I am new to the Waybar source I do not have knowledge this deep. But if I did not get you wrong, we should merge #1244 in this PR to get the `"either, or"` strategy?

> This is an interesting use case. You need set bar id for each entry in the multi-bar config and then filter IPC messages by id. And convince sway that you have multiple bars running via placeholder swaybar_command if you need to control more that one bar with IPC.
> 
> One thing to be aware of: if you connect new output, a new waybar::Bar instance will be created with the initial configuration and state and desired mode won't apply until the next IPC update. That could be addressed by reading and applying initial barconfig before spawning worker thread (see my branch for example).

Thanks for this idea, I think we should keep that in mind. In my opinion this use case should be addressed later on, after the basic functionality has been stabilized. What do you think?

emirror-de:
I got the following message in the console before the Address boundary error occurs. Maybe this helps a bit more:

```
** (waybar:56488): CRITICAL **: 16:25:26.014: Custom wayland shell surface is not a layer surface, your Wayland compositor may not support Layer Shell
```

alebastr:
> I merged your attempt to this PR (locally) and pushed it to [this branch](https://github.com/emirror-de/Waybar/tree/sway-hide-ipc-pr-1244). Hope that is what you meant?
> I tested it and still get the `Address boundary error`, however it occurs only if I spam the modifier key as fast as I can. That is indeed not the way it should be used, but it is still a bug/thing to consider I think.

I meant to test my branch alone, but you got all the relevant changes to bar.cpp anyways.
So I tried to repro with your config and the code in this PR, but I guess I'm not spamming the modifier key fast enough to see the crash. And the error about the custom wayland shell surface does not make sense. We initialize the window with the right type of custom wayland surface and it's not expected to change. Not something I see happening without an odd GTK bug.

Also, interactions of `set_opacity` and css transitions sure are weird. IIUC, `set_opacity` is there as a fallback for missing or incorrect CSS styles, so I believe it should be kept there. Unless that could be replaced with some GTK magic and builtin CSS style definition, but I haven't even tried if that works...

> 
> > `mode` value from IPC should be able to override user config for `layer`, `exclusive` and `passthrough` ...
> 
> As I am new to the Waybar source I do not have knowledge this deep. But if I did not get you wrong, we should merge #1244 in this PR to get the `"either, or"` strategy?

Not quite. #1244 attempts to implement the same feature in a different and incompatible way. It's unlikely that merging these branches will result in anything useful. You may benefit from https://github.com/alebastr/Waybar/commit/6eec6aec3a6c650185a3377c7f604ca6c9938ae7 though, as it's an API that could be used by both approaches.

> > This is an interesting use case. You need set bar id for each entry in the multi-bar config and then filter IPC messages by id. And convince sway that you have multiple bars running via placeholder swaybar_command if you need to control more that one bar with IPC.
> > One thing to be aware of: if you connect new output, a new waybar::Bar instance will be created with the initial configuration and state and desired mode won't apply until the next IPC update. That could be addressed by reading and applying initial barconfig before spawning worker thread (see my branch for example).
> 
> Thanks for this idea, I think we should keep that in mind. In my opinion this use case should be addressed later on, after the basic functionality has been stabilized. What do you think?

Fair enough, that could be addressed later once the basic functionality stops crashing.

Random thought: were you aware that each bar (either on different screens or on the same one with multi-bar configuration) is a separate instance of `waybar::Bar` that holds separate instances of each configured and enabled module?
I.e. if you have two screens with two bars, you also have two `sway/hide` objects. And both of them are trying to control all the bars, calling Bar methods twice. Multiply as necessary according to the number of screens and configured bars :)
That may cause some instabilities, potentially including the crash you observed. As a solution, you can change `Hide` module instance to control only it's own `Bar`, the one passed to the module constructor. That would also restrict IPC visibility processing to the bars that have `sway/hide` in the list of modules.

emirror-de:
> And the error about the custom wayland shell surface does not make sense. 

It occured only one time, and I did not get it again yet, so I think we can skip that.

> Not quite. #1244 attempts to implement the same feature in a different and incompatible way. It's unlikely that merging these branches will result in anything useful. You may benefit from alebastr@6eec6ae though, as it's an API that could be used by both approaches.

Ah okay. Thanks I will try that tomorrow, especially the setMode function looks nice.

> Random thought: were you aware that each bar (either on different screens or on the same one with multi-bar configuration) is a separate instance of waybar::Bar that holds separate instances of each configured and enabled module?

I think your "random" thought found the bug :smile:. No I was not aware of this, don't know how I could oversee this. My development setup is indeed two monitors. Looping over all bars available also does not make any sense!

> As a solution, you can change Hide module instance to control only it's own Bar, the one passed to the module constructor. That would also restrict IPC visibility processing to the bars that have sway/hide in the list of modules.

This is how I updated the implementation and for now it did not crash anymore. I also found that I forgot to unlock the mutex in the module.
I will test it tomorrow as my daily driver to see if issues occur. In addition to that, I tested a multibar configuration, one with the sway/hide module and one without. And it works! No crash and the one without stays visible with exclusive zone. :partying_face:

emirror-de:
After working one day with this setup, no crash happend :partying_face: 

However, I found the following behavior for which I do not have ideas howto fix them yet:

### Behavior No1
* *Monitor setup:* Two monitors, one bar per monitor
* *Behavior*: Resuming from swayidle, one bar is visible and with exclusive set to true, the other one is hidden with exclusive set to false. After pressing the modifier key, both bars are hidden with exclusive set to false.
* *Occurrence*: Everytime

### Behavior No2
* *Monitor setup:* Two monitors, one bar per monitor
* *Behavior*: Resuming from swayidle, one bar is visible and with exclusive set to true, the other one does not show up. Even after pressing the modifier key once, the second bar stays away, not showing anymore.
* *Occurrence*: Once in ~8hours

Does someone that has detailed knowledge about the internals of Waybar an idea what the reason could be for those behaviors?

alebastr:
> ### Behavior No1
> 
>     * _Monitor setup:_ Two monitors, one bar per monitor
> 
>     * _Behavior_: Resuming from swayidle, one bar is visible and with exclusive set to true, the other one is hidden with exclusive set to false. After pressing the modifier key, both bars are hidden with exclusive set to false.
> 
>     * _Occurrence_: Everytime

Most likely, one of the outputs was disabled while idle and enabled again. I.e. you've got a new bar instance with clean initial state.
That's the case where you'd want to request bar configuration in the constructor of sway/hide and set mode immediately (before the window is visible).

The possible code for fetching an initial mode is in my branch, https://github.com/Alexays/Waybar/pull/1244/files#diff-0d4cf783ddd7af6ea0c6abcfa5a8b50f29b433a0397b9e8f4c24634a452ca092R14-R20. Feel free to use that and modify as necessary for this PR.
You'll need bar_id though, which should be set in `waybar::Client` (see `waybar::Client::main`). And maybe overridden via `sway/hide` module config if we want to support waybar being launched without `swaybar_command` or have multiple bars with different sway bar ids.

> ### Behavior No2
> 
>     * _Monitor setup:_ Two monitors, one bar per monitor
> 
>     * _Behavior_: Resuming from swayidle, one bar is visible and with exclusive set to true, the other one does not show up. Even after pressing the modifier key once, the second bar stays away, not showing anymore.
> 
>     * _Occurrence_: Once

Ugh... Sounds like waybar haven't received the GTK event that the output in question is available again. Or somehow failed to process one.
I'd love to see `waybar -l trace` output for that scenario, even if it's not verbose enough to see all the details we want. But I guess it would be hard to reproduce.

> ### Behavior No3
> 
>     * _Monitor setup:_ Two monitors, one bar per monitor
> 
>     * _Behavior_: While holding down the modifier key, no window reacts to a click, except the bar.
> 
>     * _Occurrence_: Everytime

If you are using sway and have the same `floating_modifier` and bar modifier, sway would handle the clicks outside of the bar and will ignore events on tiled windows.

somini:
I have another use case, which is a single monitor with two bars. One stays always on, the other one is hidden, shown with the modifier.

With the latest code here, the bar defaults to being show, and when the modifier toggles it, it changes layer (it overlaps the visible window), but it's not hidden. Sounds like another case of multiple bars @alebastr mentioned in the last comment.

xPMo:
> I have another use case, which is a single monitor with two bars. One stays always on, the other one is hidden, shown with the modifier.

When `Hide::onEvent` receives an event object, the `"id"` field contains the bar in question which should be modified.  So instead of applying the change to all bars, the hide/show should be restricted to the bar in question.

I don't know why resuming from `swayidle` should change anything. Is Sway inconsistent with the events it sends, or is it a /gtk(-layer-shell)?/ thing?  Just curious whether Waybar's behavior agrees or disagrees with the events received from sway:

```sh
swaymsg -t subscribe -m '["barconfig_update", "bar_state_update"]' | jq
```

emirror-de:
@alebastr 

> Most likely, one of the outputs was disabled while idle and enabled again. I.e. you've got a new bar instance with clean initial state.

Ok that sounds reasonable. My monitors are in energy saving mode, so this could be the case here.

> That's the case where you'd want to request bar configuration in the constructor of sway/hide and set mode immediately (before the window is visible).

> The possible code for fetching an initial mode is in my branch, https://github.com/Alexays/Waybar/pull/1244/files#diff-0d4cf783ddd7af6ea0c6abcfa5a8b50f29b433a0397b9e8f4c24634a452ca092R14-R20. Feel free to use that and modify as necessary for this PR.

I did not know how to change your suggestion for this case. So I added two lines to the constructor [in this commit](https://github.com/Alexays/Waybar/pull/1241/commits/09da1f03e59837107bdeb151c39e02741ee4c36d). I do not know if this update is correct?

> Ugh... Sounds like waybar haven't received the GTK event that the output in question is available again. Or somehow failed to process one.
> I'd love to see waybar -l trace output for that scenario, even if it's not verbose enough to see all the details we want. But I guess it would be hard to reproduce.

I will run waybar with the trace option and hope for it to occur :)

> I have another use case, which is a single monitor with two bars. One stays always on, the other one is hidden, shown with the modifier.
> 
> With the latest code here, the bar defaults to being show, and when the modifier toggles it, it changes layer (it overlaps the visible window), but it's not hidden. Sounds like another case of multiple bars @alebastr mentioned in the last comment.

@somini, @xPMo, this use case should be handled already [from this commit on](https://github.com/emirror-de/Waybar/commit/73de008a6a67a4e3c41001130d47756fcdece151). The iteration through all bars has been replaced by the bar passed in the constructor.

@somini Could you share your config for this use case? So it is easier for me to have a look at the behavior.

> I don't know why resuming from swayidle should change anything. Is Sway inconsistent with the events it sends, or is it a /gtk(-layer-shell)?/ thing? Just curious whether Waybar's behavior agrees or disagrees with the events received from sway:

I will leave that command running during testing.

emirror-de:
Hm ok. So I just found another behavior that is close to No2.

Resuming from swaylock (monitors are on, did not switch off while swaylock is active).
One bar is hidden and shows when modifier is pressed, the other bar does not show up anymore. Both do not have exclusive set to true.

somini:
> @somini Could you share your config for this use case? So it is easier for me to have a look at the behavior.

Waybar Config:

```json
{
    "layer": "top",
    "position": "top",
    "name": "main",
    "modules-left": [
	"sway/hide",
	"sway/workspaces",
	"sway/mode"
    ],
    "modules-center": [
	"tray"
    ],
    "modules-right": [
	"idle_inhibitor",
	"backlight",
	"network#wifi",
	"network#ethernet",
	"pulseaudio",
	"battery",
	"clock#localtime"
    ],
    "sway/workspaces": {
	"disable-scroll-wraparound": true
    },
    "sway/mode": {},
    "idle_inhibitor": {
        "format": "{icon}",
        "format-icons": {
            "activated": "",
            "deactivated": ""
        }
    },
    "backlight": {
        "format": "{percent}",
	"on-click": "brightnessctl -q set 100%",
	"on-scroll-up": "brightnessctl -q set 1%+",
	"on-scroll-down": "brightnessctl -q set 1%-"
    },
    "network#wifi": {
	"interface": "wifi",
	"interval": 5,
        "format-wifi": "",
        "format-linked": "⚠",
        "format-disconnected": "",
        "tooltip-format": "{ifname}: {ipaddr}/{cidr}",
	"tooltip-format-linked": "{ifname}: No IP",
	"tooltip-format-disconnected": "{ifname}: OFF",
        "format-alt": "{ifname}[{essid}]: {ipaddr}/{cidr}"
    },
    "network#ethernet": {
	"interface": "ethernet",
	"interval": 5,
        "format-ethernet": "",
        "format-linked": "⚠",
        "format-disconnected": "",
        "tooltip-format": "{ifname}: {ipaddr}/{cidr}",
	"tooltip-format-linked": "{ifname}: No IP",
	"tooltip-format-disconnected": "{ifname}: OFF",
        "format-alt": "{ifname}: {ipaddr}/{cidr}"
    },
    "pulseaudio": {
        "format": "{icon}{volume}%|{format_source}",
        "format-muted": "|{format_source}",
        "format-bluetooth": "{icon}{volume}%|{format_source}",
        "format-bluetooth-muted": "|{format_source}",
        "format-source": "{volume}%",
        "format-source-muted": "",
        "format-icons": {
            "headphone": "",
            "hands-free": "",
            "headset": "",
            "phone": "",
            "portable": "",
            "car": "",
            "default": ["", "", ""]
        },
        "on-click": "pulseaudio-ctl mute toggle",
        "on-click-right": "pavucontrol",
	"on-clock-middle": "pulseaudio-ctl set 20"
    },
    "battery": {
	"format": "{icon}",
        "format-charging": "{capacity}%",
        "format-plugged": "{capacity}%",
        "format-alt": "{icon}{capacity}%",
	"format-time": "{H}h{M}",
	"format-icons": [
	    "",
	    "",
	    "",
	    "",
	    ""
	],
	"tooltip": true
    },
    "tray": {
	"spacing": 10
    },
    "clock#localtime": {
	"format": "{:%d%b %H:%M:%S}",
	"tooltip": true,
	"tooltip-format": "<big>{:%Y %B}</big>\n<tt>{calendar}</tt>"
    }
}
```

Sway Config:

```
# Bars
bar {
    id main
    swaybar_command waybar-configured
    mode dock
    hidden_state show
    modifier none
}

bar {
    id info
    swaybar_command waybar-configured
    mode hide
    hidden_state hide
    modifier $mod
    position top
    tray_output none
}
```

`waybar-configured` is just a wrapper script to allow multiple configurations:

```bash
#!/bin/bash
bar_id=
while getopts ":b:" opt; do
	case "$opt" in
		b)
			bar_id="$OPTARG"
			;;
		*)
			echo "Usage: $0 [-b bar] args"
			exit 2
			;;
	esac
done
shift $((OPTIND - 1))

if [ -z "$bar_id" ]; then
	echo "Missing bar ID"
	exit 3
fi

bar_folder="$HOME/.config/waybar/$bar_id"
if [ ! -d "$bar_folder" ]; then
	echo "Missing bar configuration folder: $bar_folder"
	exit 4
fi

declare -a ARGS
folder_config="$bar_folder/config.json"
if [ -r "$folder_config" ]; then
	ARGS+=(--config "$folder_config")
fi
folder_style="$bar_folder/style.css"
if [ -r "$folder_style" ]; then
	ARGS+=(--style "$folder_style")
fi

exec waybar -b "$bar_id" "${ARGS[@]}" "$@"
```

woutdp:
This PR is great, it works for me. It wasn't clear to me how to configure this but here's how I did it:

Include the following in:

`~./config/sway/config`
```
bar {
    mode hide
    swaybar_command waybar
}
```

`~./config/waybar/config`
```
    ...
    "layer": "bottom",  // <-- Set this to bottom
    "modules-left": [
        "sway/hide", // <--- Include this module
        "sway/mode",
        "sway/workspaces",
        ...
    ],
    ...
```

emirror-de:
Thanks @woutdp! Nice that it works for you.

> While holding down the Mod key, it is not possible to move a window down to be tiled at the area of the screen (bottom in my case).

I do not know why, but this issue does not occur any longer here as well. At least on my notebook it is possible to tile at the side of the screen where the bar is. Next week I am going to test multi monitor setup and will have a look at your configuration @somini.

In addition, I think it would be nice to add an "alert" that fires when the battery hits critical level. This could prevent surprising shutdowns when the bar is hidden. With "alert" I mean, that the bar automatically switches to be shown when the critical battery level has been reached. From then on, the bar behaves the same as before, if the user presses the MOD key, the bar hides again. Maybe this could be combined with the battery module, so that it automatically will be enabled when the battery module is enabled.

somini:
Makes sense. Many thanks for that configuration option.

I guess this is ready now, you can remove the Draft option now.

Great work, I'll be using this more widely as soon as it goes in a release.

emirror-de:
I did some refactoring on the code to make it more readable. Also the behavior has been changed a bit:

* Exclusive Zone has been removed for the bar. This means the bar draws over other windows on startup without having exclusive space on the screen. So no resizing any longer after the first MOD press. Toggling by USR1 signal has been untouched.
* A new boolean configuration option has been added "hide-to-bottom-layer". If set, the bar is sent to the bottom layer instead of the configured layer. Also a "bottom-layer" CSS class is then added to it. This enables eg. that the bar is shown on screens that do not have any windows open while hidden on monitors with shown windows.

Summary of config options:

```json
{
    "sway/hide": {
        "hide-on-startup": true,
        "hide-to-bottom-layer": true
    }
}
```

alebastr:
@emirror-de so I finally found time to complete my alternative implementation in #1244. I looked through all the comments here and ended up with pretty similar approach and functionality (but cleaner design, IMO). Sorry for doing that instead of continuing to review this PR :-(

Do you mind checking if there's something missing or behaving not as you expected in my branch?

emirror-de:
Hey @alebastr! This indeed looks much cleaner than an approach with an additional module. I will close this PR in favor of #1244. Discussion will also move to the other PR.

PR #1243, opened 2021-09-14
Unit-tests for configuration includes

alebastr:
* Extract config handling to a separate class
* Cover most of the config includes behavior with tests
* Try to make includes in multi-bar config behave more reasonable (breaking, see test config changes in https://github.com/Alexays/Waybar/commit/d935395f71862b9f2e01e3dc9d2337a30663e1ef)

@OskarCarl, @tiosgz, could you please take a look at the resulting behavior of multi-bar config includes and let me know if you disagree with that.


tiosgz:
Thanks for this ❤

I have only looked at the test configs (not even at the tests – but if you wish, I can have a closer look at the weekend), and it seems to be what I intended to get in #1157. The difference is only that now the included configs contain a single bar instead of all of them, is that right?

I don’t see the manpages in the diffs though I remember I described the multi-bar behaviour, can you update them please?

OskarCarl:
This is great, thank you!
I've had a *quick* look at the changes, and while I'm not all that firm in cpp, it looks good to me. I can't say anything about the test code, as I'm unfamiliar with testing in cpp.
The config files look like what I would expect to see in real use cases so I guess they should cover the important stuff.

Thanks for doing the work I wanted but couldn't find the time to do! This makes the feature way more reliable and intuitive.

alebastr:
Thanks for looking at the PR!

> The difference is only that now the included configs contain a single bar instead of all of them, is that right?

Yes. I found the behavior somewhat confusing and counter-intuitive, so I made that in multi-bar config `include` can only affect the containing bar object.

Another accidental difference is that if you have an array of included configs with the same property, previously the last value was respected. Now it's the first included config with the value that has priority.
I'd fix that, but I'm not sure which behavior makes more sense: top -> first included -> ... -> last included or top -> last included -> ... -> first included.

> I don’t see the manpages in the diffs though I remember I described the multi-bar behaviour, can you update them please?

Thanks for catching that, I missed that you had documented the current behavior! I'll update the man page later today.

> I can't say anything about the test code, as I'm unfamiliar with testing in cpp.

Well, all the test code here is about including the configs and checking if the merged json value looks as expected, that's it. The only important things to look at in the tests are assertions for conflicting values and other quirky cases.

Alexays:
Damn it tests on Waybar you rock @alebastr ❤️ 

> I'd fix that, but I'm not sure which behavior makes more sense: top -> first included -> ... -> last included or top -> last included -> ... -> first included.

I'll go for `top -> first included -> ... -> last included` IMO

Otherwise LGTM

tiosgz:
> > The difference is only that now the included configs contain a single bar instead of all of them, is that right?
> 
> Yes. I found the behavior somewhat confusing and counter-intuitive, so I made that in multi-bar config `include` can only affect the containing bar object.

It also makes more sense to me, but I didn’t feel like messing with the code more than what was really necessary.

> Another accidental difference is that if you have an array of included configs with the same property, previously the last value was respected. Now it's the first included config with the value that has priority.
> I'd fix that, but I'm not sure which behavior makes more sense: top -> first included -> ... -> last included or top -> last included -> ... -> first included.

I believe that the first value was respected, because the included configs got loaded first & then were overwritten with the earlier ones. But however it was, it doesn’t matter anymore.

Earlier declaration == higher priority (top, first included, ..., last included) is better IMO.

tiosgz:
> > > The difference is only that now the included configs contain a single bar instead of all of them, is that right?
> > 
> > 
> > Yes. I found the behavior somewhat confusing and counter-intuitive, so I made that in multi-bar config `include` can only affect the containing bar object.
> 
> It also makes more sense to me, but I didn’t feel like messing with the code more than what was really necessary.
> 
> > Another accidental difference is that if you have an array of included configs with the same property, previously the last value was respected. Now it's the first included config with the value that has priority.
> > I'd fix that, but I'm not sure which behavior makes more sense: top -> first included -> ... -> last included or top -> last included -> ... -> first included.
> 
> I believe that the first value was respected, because the included configs got loaded first & then were overwritten with the earlier ones. But however it was, it doesn’t matter anymore. // Edit: oh, nvm. `"include": ["foo", "bar"]` gave `bar` higher priority than `foo`, indeed.
> 
> Earlier declaration == higher priority (top, first included, ..., last included) is better IMO.

alebastr:
Latest revision:
 - Updated man and clarified the priorities
 - Fixed a corner case where explicit `null` value from the top config could be overwritten by includes. I.e. if someone wants to ignore the string option from included config by setting it to `null` in the parent, this should work now.

> I'll go for top -> first included -> ... -> last included IMO

Yep, that's how it works now. Seems like we all agree on the expected behavior.

Alexays:
LGTM, Thanks a lot @alebastr!
Can you update Github wiki for change in man/waybar.5.scd.in?

alebastr:
Thanks for merging!

> Can you update Github wiki for change in man/waybar.5.scd.in?

Done.

PR #1244, opened 2021-09-14
Yet another swaybar ipc implementation

alebastr:
This feature makes waybar a client for the Sway bar IPC protocol and allows controlling some aspects of the bar behavior with  `swaymsg`

Documentation: [i3 bar display modes](https://i3wm.org/docs/userguide.html#_display_mode), [`sway-bar(5)`](https://github.com/swaywm/sway/blob/master/sway/sway-bar.5.scd)

#### Prerequisites:
* Sway
* Single bar configuration block (i.e. `$XDG_CONFIG_HOME/config/waybar` is a JSON object)
* The bar is started from the sway config with `swaybar_command waybar`

#### Configuration
1.  Add `"ipc": true` to the waybar configuration
1. Set the desired `mode`, `hidden_state` and `modifier` in **the same** bar configuration block as `swaybar_command`   
    *  Sway will assign the default identifier `bar-0` to the only bar instance and run waybar as `/usr/bin/waybar -b bar-0`
    *  Waybar will read one bar configuration; create a bar; connect to the sway IPC with the bar id `bar-0` provided via commandline argument and get the initial configuration
    *  Only after that waybar will create a window with the mode specified in the sway config.
1. Control the bar mode/visibility with `swaymsg bar mode <mode> bar-0`, `swaymsg bar hidden_state <show|hide> bar-0` (note the trailing `bar_id` syntax)

## Multiple bars

Your `$XDG_CONFIG_HOME/config/waybar` is a JSON array with multiple bars defined.
* If you don't want a particular bar instance to be controlled via IPC, simply don't add the `ipc` option to the corresponding config block. It will default to disabled.
* If you want the bar instance to be controlled with the `id` from sway config (the one specified in the `bar` config block with `swaybar_command waybar`) - add only `"ipc": true`. Sway will pass the bar_id and the bar instance will use that value.
* If you want the bar instance to be controlled with a separate bar `id` (i.e. you have two bar configuration blocks in the sway config), set the `id` in the waybar config for the corresponding bar instance (`"id": "bar-1"`) and then use this identifier for the `swaymsg` command invocation: `swaymsg bar mode hide bar-1`

### Example configuration:
<details>
<summary>$XDG_CONFIG_HOME/sway/config</summary>

```
bar  {
        # will asssign id 'bar-0' by default
	swaybar_command /path/to/waybar
	mode dock
	position top
}

# the id can be set with `bar <id> { ... }` or with the `bar { id <id>; ... }`
# if you don't specify the id explicitly, sway will assign `bar-1`
bar bar-1 {
        # placeholder command, so that sway could create the configuration but won't run an actual second bar instance
	swaybar_command /bin/true
	mode hide
	position bottom
}
```
</details>

<details>
<summary>$XDG_CONFIG_HOME/waybar/config</summary>

```
[
{ // bar id is passed via `-b bar-0` commandline argument from sway
  // will react only to the events for "bar-0" or for any bar
	"ipc": true,
	"position": "top",
	...
},
{ // bar id is explicitly set to "bar-1"
  // will react only to the events for "bar-1" or for any bar
	"id": "bar-1",
	"ipc": true,
	"position": "bottom",
	...
}
]
```
</details>

The `bar_id` parameter is required for IPC subscription.

Test commands (note the trailing `bar_id` syntax. `bar <bar-id> mode ...` is broken in sway and will modify config for **all** the defined bars - https://github.com/swaywm/sway/pull/6680):
```sh
swaymsg bar mode dock bar-1
swaymsg bar mode hide bar-1
swaymsg bar hidden_state show bar-1
swaymsg bar hidden_state hide bar-1
swaymsg bar mode overlay bar-0
...
```

## Non-blocking issues:

 * [x] Defaults for `bar_id` - do we want to assume `bar-0` or to take the first entry from `get_bar_config` result if it's not set explicitly?
       `bar-0` is assumed if no other value is passed. The swaybar IPC client will try to get the initial configuration for the bar_id and disable itself if the sway config does not have the bar with that id.
 * [x] `setVisible` should be controlled either via signals or via IPC. Allowing both at the same time makes things messy and inconsistent.
       `setVisible` was changed to switch between `invisible` and `default` modes. That addressed most of the problems, the only remaining is to remember the last visible mode and use it instead of the `default`.
 * [x] Copy/move optimizations for SafeSignal
 * [ ] D-Bus method for changing mode or visibility when sway IPC is not available (but definitely not in this PR)
 * [x] Support for custom modes/mode configuration overrides; <details><summary>example:</summary>

   ```
   "modes": {
     "dock": {
       "layer": "top"
     },
     // show overlay bar over fullscreen applications
     "overlay": {
       "layer": "overlay"
     },
     // hide bar under the current window instead of making it invisible
     "invisible": {
       "passthrough": false,
       "visible": true
     } 
   }
   ```
   </details>

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -23,6 +23,35 @@ static constexpr const char* BAR_SIZE_MSG = "Bar configured (width: {}, height: 
 static constexpr const char* SIZE_DEFINED =
     "{} size is defined in the config file so it will stay like that";
 
+const Bar::bar_mode_map Bar::PRESET_MODES = {  //
+    {"dock",
+     {// Modes supported by the sway config; see man sway-bar(5)
+      .layer = bar_layer::BOTTOM,
+      .exclusive = true,
+      .passthrough = false,
+      .visible = true}},
+    {"hide",
+     {//
+      .layer = bar_layer::TOP,
+      .exclusive = false,
+      .passthrough = false,
+      .visible = true}},
+    {"invisible",
+     {//
+      .layer = bar_layer::BOTTOM,
+      .exclusive = false,
+      .passthrough = true,
+      .visible = false}},
+    {"overlay",
+     {//
+      .layer = bar_layer::TOP,
+      .exclusive = false,
+      .passthrough = true,
+      .visible = true}}};
+
+const std::string_view Bar::MODE_DEFAULT = "dock";
+const std::string_view Bar::MODE_INVISIBLE = "invisible";
+
 #ifdef HAVE_GTK_LAYER_SHELL
 struct GLSSurfaceImpl : public BarSurface, public sigc::trackable {
   GLSSurfaceImpl(Gtk::Window& window, struct waybar_output& output) : window_{window} {
@@ -533,28 +562,25 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
   }
 }
 
-void waybar::Bar::setMode(const std::string& mode) {
-  bool passthrough = false;
-  visible = true;
-  exclusive = true;
-  layer_ = bar_layer::BOTTOM;
-
-  if (mode == "hide") {
-    exclusive = false;
-    layer_ = bar_layer::TOP;
-    visible = false;
-  } else if (mode == "invisible") {
-    visible = false;
-  } else if (mode == "overlay") {
-    exclusive = false;
-    layer_ = bar_layer::TOP;
-    passthrough = true;
+void waybar::Bar::setMode(const std::string_view& mode) {
+  auto it = configured_modes.find(mode);
+  if (it != configured_modes.end()) {
+    last_mode_ = mode;
+    setMode(it->second);
+  } else {
+    spdlog::warn("Unknown mode \"{}\" requested", mode);
+    last_mode_ = MODE_DEFAULT;
+    setMode(configured_modes.at(MODE_DEFAULT));
   }
+}
 
+void waybar::Bar::setMode(const struct bar_mode& mode) {
+  layer_ = mode.layer;
+  exclusive = mode.exclusive;
   surface_impl_->setLayer(layer_);
   surface_impl_->setExclusiveZone(exclusive);
-  surface_impl_->setPassThrough(passthrough);
-  setVisible(visible);
+  surface_impl_->setPassThrough(mode.passthrough);
+  setVisible(mode.visible);
 }
 
 void waybar::Bar::onMap(GdkEventAny*) {
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -562,13 +562,21 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
 }
 
 void waybar::Bar::setMode(const std::string_view& mode) {
+  using namespace std::literals::string_literals;
+
+  auto style = window.get_style_context();
+  /* remove styles added by previous setMode calls */
+  style->remove_class("mode-"s + last_mode_);
+
   auto it = configured_modes.find(mode);
   if (it != configured_modes.end()) {
     last_mode_ = mode;
+    style->add_class("mode-"s + last_mode_);
     setMode(it->second);
   } else {
     spdlog::warn("Unknown mode \"{}\" requested", mode);
     last_mode_ = MODE_DEFAULT;
+    style->add_class("mode-"s + last_mode_);
     setMode(configured_modes.at(MODE_DEFAULT));
   }
 }
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -426,7 +426,6 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
     : output(w_output),
       config(w_config),
       window{Gtk::WindowType::WINDOW_TOPLEVEL},
-      layer_{bar_layer::BOTTOM},
       left_(Gtk::ORIENTATION_HORIZONTAL, 0),
       center_(Gtk::ORIENTATION_HORIZONTAL, 0),
       right_(Gtk::ORIENTATION_HORIZONTAL, 0),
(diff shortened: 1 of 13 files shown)


emirror-de:
I just tested this one and it looks and works great! I used the following config:

```
{
    "bar_id": "bar-0",
    "ipc": true,
    "position": "top",
    "mode": "hide",
    "hidden_state": "show"
    ...
}
```
This works as the expected behavior from sway-bar. Many thanks for that!

If I could make a wish, I would love to be the bar visible if there is no other window on the screen. Is this possible with your custom mode configuration overrides?
In additional, it would be great to have an additional parameter to set the initial value if a bar should be hidden or shown on start until the first MOD press. But as far as I understood, you already do this with the `hidden_state` parameter? I was not able to figure the configuration out yet.

alebastr:
> If I could make a wish, I would love to be the bar visible if there is no other window on the screen. Is this possible with your custom mode configuration overrides?

Something like `"hide": { "layer": "bottom", "exclusive": false, "visible": true }`? I.e. any opened window would cover the bar so you'll see it only on an empty workspace or through a transparent window.
And CSS bits to finish the trick and deal with transparency (`set_opacity` call in `Bar::setVisible` may interfere with that):
```
window#waybar.mode-hide {
   opacity: 0;
}
window#waybar.mode-hide.empty {
   opacity: 1;
}
```
I haven't tested if that works; for now you can try modifying the preset for "hide" in `bar.cpp`.

> In additional, it would be great to have an additional parameter to set the initial value if a bar should be hidden or shown on start until the first MOD press. But as far as I understood, you already do this with the hidden_state parameter? I was not able to figure the configuration out yet.

The behavior here is copied from swaybar and fully controlled by sway config.
I.e. the mode is received and applied before creating the window, and if it is set to `mode hide; hidden_state hide` then you won't see the bar on start

somini:
I testes this, but it doesn't work like #1241. The `$Mod` key cannot toggle the correct bar, it always toggle the first one.

alebastr:
> I testes this, but it doesn't work like #1241. The `$Mod` key cannot toggle the correct bar, it always toggle the first one.

Are you sure you assigned correct bar ids and modes? Check example configs in the pr description; with this configuration it should toggle only the second bar.

Also, there's a bug in sway that makes `swaymsg bar <bar_id> mode <mode>` change state for all the defined bars. If you ran this command, you may have incorrect internal state in the sway itself.

alebastr:
@somini the PR description now documents the feature, the configuration and the expected behavior.

Let me know if you still have problems or if you have any suggestions for making it easier to set up.

somini:
@alebastr Thanks, I re-read the docs and now I understood. Here's my setup (same test as #1241).

- One bar on bottom, always shown
- One bar on top, hidden, "peekable" by `$Mod`

Sway:
```
bar {
    swaybar_command /home/somini/Programming/Waybar/build/waybar
    mode dock
    hidden_state show
    position bottom
}
bar bar-1 {
    swaybar_command true
    mode hide
    hidden_state hide
    position top
}
```

Waybar:
```
[
{
    "position": "bottom",
    "name": "main",
    "modules-left": [
	"sway/workspaces",
	"sway/mode"
    ],
    "modules-center": [
	"tray"
    ],
    "modules-right": [
	"clock#localtime"
    ],
    "sway/workspaces": {
	"disable-scroll-wraparound": true
    },
    "sway/mode": {},
    "clock#localtime": {
	"format": "{:%d%b %H:%M:%S}",
	"interval": 1,
	"tooltip": true,
	"tooltip-format": "<big>{:%Y %B}</big>\n<tt>{calendar}</tt>"
    },
},
{
    "ipc": true,
    "id": "bar-1",
    "position": "top",
    "name": "info",
    "modules-center": [
	"clock#localtime"
    ],
    "modules-right": [
    ],
    "clock#localtime": {
	"format": "HIDE {:%d%b %H:%M:%S}",
	"interval": 1,
	"tooltip": true,
	"tooltip-format": "<big>{:%Y %B}</big>\n<tt>{calendar}</tt>",
	"on-click": "notify-send -t 1000 Example 'Body Text'",
    }
}
]
```

I see that #1241 was abandoned in favour of this, can confirm this works well too.

src/bar.cpp
```diff
@@ -505,15 +522,44 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
...
+  /* Init "default" mode from globals */
+  auto& default_mode = configured_modes[MODE_DEFAULT];
+  if (config["layer"] == "top") {
+    default_mode.layer = bar_layer::TOP;
+  } else if (config["layer"] == "overlay") {
+    default_mode.layer = bar_layer::OVERLAY;
+  }
+
+  if (config["exclusive"].isBool()) {
+    default_mode.exclusive = config["exclusive"].asBool();
```
Alexays, line 538: passthrough is not documented yet, right?

src/bar.cpp
```diff
@@ -505,15 +522,44 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
...
+  }
+
+  if (config["passthrough"].isBool()) {
+    default_mode.passthrough = config["passthrough"].asBool();
```
Alexays, line 542: passthrough is not documented yet, right?

Alexays:
LGTM, except one comment.
Amazing work, like always!
```
Defaults for bar_id - do we want to assume bar-0 or to take the first entry from get_bar_config result if it's not set explicitly?
```
Assuming bar-0 will be good enough for me

src/bar.cpp
alebastr: It's documented, both in the man and in the wiki. Although I think it shouldn't be now that we have modes support.

alebastr:
Rebased, added default bar id and connection error handling.
`BarIpcClient::onInitialConfig` intentionally doesn't swallow exceptions so that we can detect a wrong bar_id from the initial config request result and stop ipc.

Alexays (approved):

Issue #746, opened 2020-06-13
[feature request] Separate CPU cores 


apiraino:
Hi,

I'm not sure it is currently possible to split cpu load and have one icon per core. I'd love to see something similar to how the battery indicator is implemented:

```
    "cpu#cpu0": {
        "format": "{usage}% ",
        "tooltip": false
    },
    "cpu#cpu1": {
        "format": "{usage}% ",
        "tooltip": false
    },
...
```

thanks for Waybar!

apiraino:
was implemented in #1253

PR #1250, opened 2021-09-17
Add CPU usage for every core

Darkclainer:
This MR adds usage and icon for every CPU core. Use in format like `{usage0}`, `{usage1}` and etc. You can build nice graph for CPU with it:

image

P.S.
Sorry, I haven't write in C++ for years, so code can very bad.

UPD:
Found that there is relevant feature request: https://github.com/Alexays/Waybar/issues/746


Alexays:
Very nice!
Can you update man? :)

Darkclainer:
> Very nice!
> Can you update man? :)

Done.

Darkclainer:
Also, I have found eerie comment: https://github.com/Alexays/Waybar/blob/8da940f929bd8c06a30fd3d02de53e6fac2ac5f8/src/modules/cpu/common.cpp#L12

@Alexays, do you know what was it about?

Alexays:
> Also, I have found eerie comment:
> 
> https://github.com/Alexays/Waybar/blob/8da940f929bd8c06a30fd3d02de53e6fac2ac5f8/src/modules/cpu/common.cpp#L12
> 
> @Alexays, do you know what it was about?

It was for an older version of fmt I think.
It should be ok now.

Alexays:
Thanks!

Alexays:
Rollbacked as CI failed :/

Darkclainer:
> Rollbacked as CI failed :/

It didn't build on Fedora. I've checked tests of fmt lib and including `fmt/args.h` seems legit (https://github.com/fmtlib/fmt/blob/master/test/args-test.cc#L8). The problem seems to be, that waybar uses system's version of fmt.

I can try to create PR that will add fmt as dependency. Will you merge such PR, @Alexays?

alebastr:
It didn't built as Fedora (and a few more distributions according to [repology](https://repology.org/project/fmt/versions)) ships fmt 7.x in the stable releases. I'm pretty sure that if CI had reached Debian it would fail in the same way.

As a Fedora maintainer of waybar package, I'd prefer if we can keep compatibility with older fmt. There are policies that discourage us from having third party libraries bundled in the package source, so I'll have to deal with reverting the change downstream in Fedora.

As far as I can see, 7.x has required classes in the `<fmt/core.h>`. Maybe conditionally checking the fmt version and including the correct file for the installed fmt version would work for you (same way as it's already done around a few breaking fmt changes)?

PR #1253, opened 2021-09-19
Add cpu usage for every core

Darkclainer:
This is a second attempt to merge #1250. I have added conditional check on fmt version as was suggested by @alebastr.

I checked their commit history and found that separation of headers was made about time when 7.1.3 was released, but this change was actually introduced only in 8.0.0. Seems to be correct.

Also I moved header inclusion in cpp file, because this compliant with google-style mention in the README.


Alexays (approved):

Alexays:
Thanks for the fix!

PR #1256, opened 2021-09-21
dont escape essid in tooltip

WuerfelDev:
just like https://github.com/Alexays/Waybar/pull/834 but for the network name.


Alexays:
LGTM

Alexays:
CI fail

PR #1271, opened 2021-10-02
fix(tray): fix visibility of Passive items

alebastr:
`show_all` call from `Tray::update` attempts to walk the widget tree and
make every widget visible. Since we control individual tray item
visibility based on a `Status` SNI property, we don't want that to happen.

Modify `Tray::update` to control the visibility of a whole tray module
only and ensure that the children of `Item` are still visible when
necessary.

***
Fixes the issue described in https://github.com/Alexays/Waybar/pull/1145#issuecomment-922361054.
The code didn't change since the initial fix, I just was too ~~lazy~~busy to format the commit message and send the PR :smiley:


Alexays (approved):

Alexays:
Thanks @alebastr!

Issue #1272, opened 2021-10-02
sway/window uses empty style with only floating windows in workspace


RobertMueller2:
Using the 0.9.8 Debian package, the sway/window module and this CSS

    window#waybar.empty #window {
        padding: 0px;
        margin: 0px;
        border: 0px;
        background-color: transparent;
    }

    window#waybar.solo #window,
    #window {
        padding-left: 5px;
        padding-right: 5px;
        background-color: #cb4b16;
    }

![1](https://user-images.githubusercontent.com/29629287/135713513-7d3ed84a-b981-4a7e-8f43-11a6f1a7e137.png)
![2](https://user-images.githubusercontent.com/29629287/135713515-b241863a-c989-4df7-9686-4311db20a96b.png)

it appears that the empty style is used when there are only floating windows in a workspace. As soon as I open a tiled window, the module looks as expected, so it is not any app specific styling (which I don't have anyway). 

Looking at src/modules/sway/window.cpp, I think leafNodesInWorkspace only counts nodes but not floating_nodes. Do you agree and if so, should this be fixed?

PR #1274, opened 2021-10-02
sway/window: include floating_nodes when considering window count for class

RobertMueller2:
Quick fix suggestion for #1272 .

Tested briefly with solo, empty and normal and seemed to work so far.


Alexays (approved):

Alexays:
Thanks!

Issue #911, opened 2020-11-03
Empty string in clock module's 'timezones' option causes crash


Ghosthree3:
This is stated in man 5 waybar-clock for the 'timezones' option.
> Use "" to represent the system's local timezone.

It works when waybar is first started, but if using multiple timezones like so,
`"timezones": [
    "",
    "America/New_York"
],`
waybar will crash when rolling back around to the "" timezone when using the mouse wheel.

So the behavior is: waybar starts, local timezone is shown, scroll once, America/New_York timezone is shown, scroll once more, crash.
Resolved by setting "" to the local timezone literally.

rieje:
The `timezones` option doesn't seem to be working at all for me:

    "clock": {
        "timezone": "America/New_York",
        "tooltip-format": "<big>{:%Y %B}</big>\n<tt><small>{calendar}</small></tt>",
        "format":"{:%a, %b %d %H:%M %Z}",
        "format-alt": "{:%H:%M:%S}",
        "interval": 1,
        "timezones": [ "America/Chicago", "America/New_York", "Asia/Hong_Kong" ],
        "on-click-right": "alacritty --class '__bc' -e bc -ql"
    },

My understanding is that I just need to hover the mouse cursor over the time and mouse wheel scroll up/down to change the time according to the list of timezones defined. Am I missing something?

PR #1275, opened 2021-10-03
Fix Clock crash on empty string in timezones field

dartkron:
Fix: #911 

Also fixed timezones behavior: now Waybar starting with the first timezone in `timezones` list and falling back to the `timezone` field only if timezones omit or has no elements.


Alexays (approved):

Alexays:
Thanks buddy!

dartkron:
@Alexays thank you for the light-fast reaction! I've never seen so fast reaction in opensource.

PR #1276, opened 2021-10-03
Shared subprocesses for custom modules

rtoijala:
This is an implementation of a new feature: shared subprocesses for custom modules. The feature basically allows the user to configure per-bar commands to be exec'd. The output is a JSON object with module names as keys and output as values. The configuration is described in the updated man page.

The use case for this is that I have a complicated MPD bar with custom handling for several of the modules (so I can't use the builtin MPD module) and several clickable things (so they need to be separate modules). I have a custom MPD client that creates the output for the custom modules. Without shared subprocesses, each module runs its own instance of the MPD client, creating unnecessary CPU load. This feature allows me to run the script only once, outputting data for all of the modules. I also have separate "system status bar" where I use the same setup for various statistics about the system, also with many clickable things.

Since most of the subprocess handling is now shared between the custom module and the bar, I created a new helper class "WorkerThread" which handles the configuration, starts the process, runs the thread reading the output, and calls callbacks when output is read or the child exits.

(Side note: Even before my changes, trying to refresh modules by sending signals always terminates Waybar. I'm note sure what is wrong, but since it I just moved the signal handling from one place to another and I do not use the feature myself, I did not investigate much.)

Any feedback is welcome.


pocek:
In case anyone is interested, I've rebased this onto the current master: [pocek/Waybar@custom-execs](https://github.com/pocek/Waybar/tree/custom-execs). It works fine for me, but I have not tested anything related to signals.

JassonCordones:
Is this still open?

PR #1277, opened 2021-10-03
Refactor Clock: generalize multi timezones and single timezone cases

dartkron:
After this refactoring:
1. Timezones parses only once on start and the we refer to saved values. All time_zone.isString() checks gone to the constructor.
2. Single timezone case handling as case of multi timezoned logic.
3. Scroll event seems more clear now.
4. Tooltip template parses on start to check if there calendar placeholder or not. To do not calculate calendar_text() if not necessary.

This is preparations for another feature which I'll bring in separate PR.

Any feedback is appreciated.

### DIFF
--- a/include/modules/clock.hpp
+++ b/include/modules/clock.hpp
@@ -17,6 +17,8 @@ struct waybar_time {
   date::zoned_seconds ztime;
 };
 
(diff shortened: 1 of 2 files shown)


Alexays (approved):

Alexays:
LGTM thx

PR #1289, opened 2021-10-15
Add reverse scrolling config option for pulseaudio module

### LINKED ISSUE #1220: Allow users to disable reverse scrolling in pulseaudio module
When using a touchpad with natural (reverse) scrolling enabled via libinput, volume adjustment behavior by scrolling on the pulseaudio module also gets reversed. In other words, scrolling up decreases the volume while scrolling down increases it.

aashu:
When natural scrolling is enabled, the behavior of scrolling on pulseaudio
module is reversed, i.e. scrolling up reduces the volume, which is not expected.
This commit reverses the direction of scroll variable if "reverse-scrolling" is set to 1 in the config file like below.
`pulseaudio {
  "reverse-scrolling": 1
}`
resolves #1220

### DIFF
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -79,6 +79,13 @@ bool waybar::modules::Pulseaudio::handleScroll(GdkEventScroll *e) {
   if (dir == SCROLL_DIR::NONE) {
     return true;
   }
+  if (config_["reverse-scrolling"].asInt() == 1){
+    if (dir == SCROLL_DIR::UP) {
+      dir = SCROLL_DIR::DOWN;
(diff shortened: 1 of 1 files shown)


aashu:
Hello @Alexays, I understand you might be busy. But can you at least confirm if this pull request is useful or not? If it's not(meaning never going to be merged), maybe I will close it and add it on top of your future releases for my personal use. 
Or please let me know if there are better solution to my problem. Thanks.

Alexays:
Sorry for the delay 😢 
Yes, LGTM and thanks for the feature!

WhyNotHugo:
Thanks for this! I've added a wiki setting for this entry.

PR #1305, opened 2021-11-01
Add module group feature

njoyard:
This PR adds a module group feature. A module group contains multiple modules and renders them orthogonal to the bar direction (ie stacked vertically for a horizontal bar and vice versa).

My use case for it is that I want a vertical bar but I want some modules to share the same row on the bar. This may sound like a waste of screen space but I'm using waybar as an overlay that is most often hidden, and show on demand. A screenshot of my setup follows, volume/brightness bars are implemented each with two modules (one for the icon, one for the bar) placed in a group. Progress bars use CSS linear gradients as their background.

image

The way a group is configured is as follows:

```json
{
  "modules_left": [
    "module_a",
    "group/mygroup",
    "module_b"
  ],

  "group/mygroup": {
    "modules": [
      "module_c",
      "module_d"
    ]
  }
}
```

I chose to list modules in a "modules" key to leave space for future additional options for groups (eg. alignment and so on - which would probably be useful for right/bottom-positioned bars). The implementation does not prevent nested groups, but that's currently not very useful as a group direction is orthogonal to the bar, not to its parent.

Please note that I have no significant experience with c++ whatsoever, so my code may leave a lot to be desired. I've been using it as a daily driver for quite some time though so I'm hoping any remaining issues are mostly architecture/style/good practices related.

Let me know what you think.

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -73,7 +73,7 @@ class Bar {
  private:
   void onMap(GdkEventAny *);
   auto setupWidgets() -> void;
-  void getModules(const Factory &, const std::string &);
+  void getModules(const Factory &, const std::string &, Gtk::Box*);
   void setupAltFormatKeyForModule(const std::string &module_name);
   void setupAltFormatKeyForModuleList(const char *module_list_name);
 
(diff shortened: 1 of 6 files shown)


Alexays:
Nice addition!
Can you document this in man? :)

njoyard:
Added some documentation in the main manpage, and rebased on master.

PR #1312, opened 2021-11-09
Changed a setting to include {gwaddr} by default

mazunki:
After some update, I don't know exactly which, nor if it was waybar's or my kernel's (although from the change I did, I am guessing it was a waybar change), the pull request at commit 6938921e92798714d399a8a2f95c2fdbb4ac4d03 stopped working for me.

After some debugging, I realized that changing `want_route_dump(false)` to true made it work once again. Now that I look at it, I don't really understand how it worked earlier, since that line has been there since around commit c9bbaa7241bb40f06bed523c54faa9a8643f7ec2. 

If there's a reason to default it to false, this pull request doesn't make much sense, but since we are claiming to have {gwaddr} available, I don't see a reason for it to be inactive by default. 

Otherwise, how would I go by configuring a request to get the route dump from the config files?

Oh, and also, I added some checks for some file descriptors in some destructors so that my compiler would be quiet. Maybe we should handle the errors, but just having the checks there helps my compiler shut up.

### DIFF
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -68,7 +68,9 @@ inline int close(FILE* fp, pid_t pid) {
 inline FILE* open(const std::string& cmd, int& pid) {
   if (cmd == "") return nullptr;
   int fd[2];
-  pipe(fd);
+  if (pipe(fd) != 0){
+    throw std::runtime_error("Couldn't open a file descriptor");
+  }
 
   pid_t child_pid = fork();
 
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -69,7 +69,6 @@ inline FILE* open(const std::string& cmd, int& pid) {
   if (cmd == "") return nullptr;
   int fd[2];
   if (pipe(fd) != 0){
-    // std::runtime_error("Couldn't open a file descriptor");
   }
 
   pid_t child_pid = fork();
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -69,7 +69,7 @@ inline FILE* open(const std::string& cmd, int& pid) {
   if (cmd == "") return nullptr;
   int fd[2];
   if (pipe(fd) != 0){
-    throw std::runtime_error("Couldn't open a file descriptor");
+    // std::runtime_error("Couldn't open a file descriptor");
   }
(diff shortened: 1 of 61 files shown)


include/util/command.hpp
```diff
@@ -68,7 +68,9 @@ inline int close(FILE* fp, pid_t pid) {
 inline FILE* open(const std::string& cmd, int& pid) {
   if (cmd == "") return nullptr;
   int fd[2];
-  pipe(fd);
+  if (pipe(fd) != 0){
+    // std::runtime_error("Couldn't open a file descriptor");
```
Alexays, line 72: Remove comment

src/modules/sway/ipc/client.cpp
```diff
@@ -14,12 +14,16 @@ Ipc::~Ipc() {
 
   if (fd_ > 0) {
     // To fail the IPC header
-    write(fd_, "close-sway-ipc", 14);
+    if (write(fd_, "close-sway-ipc", 14) == -1) {
+       //std::runtime_error("Couldn't close Sway IPC through fd");
```
Alexays, line 18: Remove comment

src/modules/sway/ipc/client.cpp
```diff
@@ -14,12 +14,16 @@ Ipc::~Ipc() {
...
+    }
     close(fd_);
     fd_ = -1;
   }
   if (fd_event_ > 0) {
-    write(fd_event_, "close-sway-ipc", 14);
+    if (write(fd_event_, "close-sway-ipc", 14) == -1) {
+       //std::runtime_error("Couldn't close Sway IPC through fd_event");
```
Alexays, line 25: Remove comment

Alexays (changes requested):

src/modules/sway/ipc/client.cpp
mazunki: Are you saying I should remove `//`, or remove the line altogether? I was not sure which made more sense, so I left it commented in case anyone wants to handle it properly.

Alexays: You can remove the throw, not very useful here

PR #1328, opened 2021-11-27
build: Add experimental build option

Anakael:
As requested in https://github.com/Alexays/Waybar/pull/805


Alexays:
Perfect!

maximbaz:
It would be nice to also exclude installation of manpage for experimental modules, so that users are not confused by having docs for a module that they can't enable

PR #1329, opened 2021-11-27
mediaplayer.py: Exit properly on SIGPIPE

m-bdf:
Handle the Broken Pipe error thrown by `sys.stdout.flush()` (as Python ignores SIGPIPE by default), when piping `mediaplayer.py` to `head -n0` for example.


Alexays:
Thanks!

PR #1331, opened 2021-11-30
Allow sink in addition to source for pulse icon

xytovl:
A previous build fix changed the logic for per-device pulseaudio icon selection and used the source name instead of sink.

As some users might be using it based on source, I have added the sink in addition to the existing one.


Alexays:
lgtm, thx!

PR #1332, opened 2021-12-01
Escape ampersands

akaij:
When not escaped, artists or titles with '&' in them cause nothing to be printed.


akaij:
nevermind - I missed the existence of `escape` argument -_-

PR #1334, opened 2021-12-02
Adjust max wifi strength that is possible

bd-g:
By definition, the best possible signal strength is -30dBm, so having -20dBm here doesn't make sense

Reference (can Ctrl+F for -30dBm): 
https://eyenetworks.no/en/wifi-signal-strength/
https://www.metageek.com/training/resources/wifi-signal-strength-basics/

As a small note, signal strengths in the -30s and -40s dBm actually is __not__ good - it is too strong and can cause errors between hosts and a receiver. Often, the maximum desirable levels are in the -50s dBm. Thinking about creating a seperate pull request to adjust the signalStrength value to account for this? Thoughts?


Alexays:
Thanks!

PR #1335, opened 2021-12-02
Add battery left-pad for better formatting

sambhavsaggi:
Previously, if the minutes in the battery indicator was a single digit number, it would format strangely: "2:7" for two hours and seven minutes. This has been bugging me, so I have decided to fix it. I have added a very simple left-pad format into the function, so that two hours and seven minutes would appear as "2:07".


alebastr:
Wouldn't that make the default `format-time` look odd - `2 h 07 min`?
Also, I believe you can get the result you wanted with `"format-time": "{H:02}:{M:02}"`.

sambhavsaggi:
in that case, I'll create PR for the docs so other people don't get confused :)

PR #1338, opened 2021-12-03
bar: Fix crash when unplugging HDMI

jfred9:
There is a double delete situation which causes a SIGSEGV to happen
during destruction of bar.

This was introduced by the group feature patch.

The same object pointer is stored in two different vectors of
unique_ptr<AModule> element. Replace with shared_ptr to handle
reference counting correctly and avoid double delete.


Alexays:
Thanks!

PR #1341, opened 2021-12-05
Add logind feature, with its 'inhibitor' module

fhost:
The logind feature adds a new inhibitor module which allows to acquire
the inhibitor locks that logind presents.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -51,6 +51,9 @@
 #ifdef HAVE_LIBSNDIO
 #include "modules/sndio.hpp"
 #endif
+#ifdef HAVE_GIO_UNIX
+#include "modules/inhibitor.hpp"
+#endif
 #include "bar.hpp"
 #include "modules/custom.hpp"
 #include "modules/temperature.hpp"
--- a/include/modules/inhibitor.hpp
+++ b/include/modules/inhibitor.hpp
@@ -1,0 +1,27 @@
+#pragma once
+
+#include <memory>
+
+#include <gio/gio.h>
+
+#include "ALabel.hpp"
+#include "bar.hpp"
+
+namespace waybar::modules {
+
+class Inhibitor : public ALabel {
+ public:
+  Inhibitor(const std::string&, const waybar::Bar&, const Json::Value&);
+  ~Inhibitor() override;
+  auto update() -> void;
+  auto activated() -> bool;
+
+ private:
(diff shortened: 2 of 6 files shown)


alebastr:
Why do we need `sd-bus`? `gio-unix` is guaranteed to be there (unless you're trying to build waybar on Windows), and one DBus implementation is enough to make a simple method call.

Can anything break if we attempt to take multiple locks? Because each displayed bar in a multi-monitor setup will run its own instance of the module.

fhost:
> Why do we need `sd-bus`? `gio-unix` is guaranteed to be there (unless you're trying to build waybar on Windows), and one DBus implementation is enough to make a simple method call.

I did not know that `gio-unix` was guaranteed to be there. What guarantees it to be there? From the build options, I supposed that it could not be present (`dbusmenu-gtk` disabled). I did provide both as I was thinking that if the user has no gtk, as it is a logind module, it is guaranteed to have systemd or elogind (other the module is useless).

> Can anything break if we attempt to take multiple locks? Because each displayed bar in a multi-monitor setup will run its own instance of the module.

I'll try.

fhost:
> > Can anything break if we attempt to take multiple locks? Because each displayed bar in a multi-monitor setup will run its own instance of the module.
> 
> I'll try.

No issue.

alebastr:
> I did not know that gio-unix was guaranteed to be there. What guarantees it to be there?

One reason is that `gio-unix` is a platform-specific module built from the common glib source tree on *nix platforms. It'll be installed with glib headers, unless your distribution is splitting glib development packages more than necessary. The other reason is that it is used internally in the GTK.
But you are right, the build script requires it conditionally, which is easy to fix.

fhost:
> > I did not know that gio-unix was guaranteed to be there. What guarantees it to be there?
> 
> One reason is that `gio-unix` is a platform-specific module built from the common glib source tree on *nix platforms. It'll be installed with glib headers, unless your distribution is splitting glib development packages more than necessary. The other reason is that it is used internally in the GTK. But you are right, the build script requires it conditionally, which is easy to fix.

So you think I should remove the `sd-bus` implementation, and ask `gio-unix` for the logind feature in meson (which is, in fact, not bringing another dependency, ask the GTK part of waybar already draws it)?

fhost:
Update: Remove `sd-bus`, clean up some stuff and fix up a bit the coding style.

Alexays (approved):

Alexays:
LGTM, thanks!

PR #1353, opened 2021-12-14
Network Module Enhancements

bd-g:
This pull request contains a series of enhancements to the network module, separated out by commit. #3 and #4 on this list would require updates to the Wiki as well.

1. But first, a build dependency issue. Versions >=0.9.0 often do not build on Debian with the existing build dependencies. "xkbregistry" is required, but not supplied using the current list of Debian packages on the READme. I added the "libxkbregistry-dev" package to the READme under the command to install all required dependencies for Ubuntu.

2. Signal strength calculation. The previous calculated percentage was just a raw scale from -30dBm to -90dBm. While -30dBm is the strongest possible signal for wifi, if a signal is too strong, it can overwhelm receiving circuity that is designed to pick up and process a certain signal level. I changed the percentage calculation to set -45dBm as the optimal level, with signals that are strong or weaker than that calculated as less than 100% signal strength.

3. Signal strength application. Added a format replacement for the network module, "signalStrengthApp". This just displays what your connectivity is and what applications you may be able to handle at current network signal strength, i.e "streaming" or "web surfing".

4. Frequency. Wifi frequency is commonly referenced in GHz in the 802.11 IEEE Standard. I changed the frequency format replacement from MHz to GHz.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ libdbusmenu-gtk3 [Tray module]
 libmpdclient [MPD module]
 libsndio [sndio module]
 libevdev [KeyboardState module]
+xkbregistry
 ```
 
 **Build dependencies**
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM

Issue #1357, opened 2021-12-22
Select output by dimensions.


RyanGibb:

Alexays:
Implemented on master via PR #1358: bars can now select their output by resolution using the `output-dimensions` option (e.g. `"output-dimensions": "width > 1080"`, or an array of conditions), which makes configs portable across setups. Documented in waybar(5). Closing as resolved — please reopen if it does not behave as expected on a recent build from master.

PR #1358, opened 2021-12-22
added output-dimensions option

RyanGibb:
Addresses issue #1357 by adding an option to select outputs for different configurations based on the output's dimensions with "output-dimensions".

This is overridden by "output".


JassonCordones:
is this still open?

RyanGibb:
I don't use waybar anymore, happy for this to be merged/closed/forked

Vladimir-csp:
Selecting by position would be nice. In recent month I started to normalize everything to treat output at `0,0` as primary.

PR #1359, opened 2021-12-23
Allow to use custom play/pause icons

flipflop133:
Display album name in tooltip.

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -148,6 +148,8 @@
         "return-type": "json",
         "max-length": 40,
         "format-icons": {
+            "Playing": "\uf144",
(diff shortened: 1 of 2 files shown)


flipflop133:
For this to work proper configuration must be used in waybar config file by adding the following to the media section:
```
"format-icons": {
			"Playing": "\uf144",
			"Paused": "\uf28b"
		}
```

Alexays:
Can you update default config to include icon? :)

flipflop133:
Sorry for the late answer, I've been busy with university (and still am). Hope I made the right changes but it should be fine. :)

PR #1360, opened 2021-12-24
battery module usb-c reporting Plugged when discharging

hillyu:
Current battery adaptor check neglected the case when plugged a device in through a usb type-c shared with power supply, in this case even if the adaptor is online, if you don't plug the power in it will still be discharging. The current logic marked above case as "Plugged" thus module will think it is plugged even though it is actually discharging. Discharging should always be discharging regardless of the adaptor status. This patch fixed above issue.


Alexays:
Plugged does not necessarily mean that it is charging

PR #1369, opened 2022-01-04
Ellipsize song info when too long

KogasaPls:
When artist, title, album artist, or album are truncated, trim trailing whitespace & replace last 3 characters with "..." as long as they are longer than (char limit - 3).

Currently, "max_length" being set will cause the entire "format" string to be truncated & ellipsized, but there is no way to ellipsize the individual tags. This can make it hard to tell when strings are being truncated.

e.g. with artist-len == 13 and artist == "Queens of the Stone Age"
Old behavior: artist = "Queens of the"
New behavior: artist = "Queens of..."


Alexays:
Could you also update the man? :)

roland-rollo:
@KogasaPls do you still have time to make man?

PR #1376, opened 2022-01-06
fix: match sway workspace order with sway one

### LINKED ISSUE #802: Sway/workspace: Workspace names not in the same order as Sway internally stores them.
As a result, when running `workspace next`, the next workspace may not be the next one in the bar. Sway goes through them in the order that they are created, and Waybar sorts them into Alphabetical order.

This doesn't seem to be a problem for numbered workspaces.

This can be worked around by prefixing named workspaces with numbers, as Waybar orders those according to their prefixed numbers. I think it would be simpler and more consistent to have Waybar's workspace list match Sway's though, and if alphabetical order is desired, then to change the order of workspaces in Sway somehow.

My guess is that here is where Waybar sorts the workspace names, and where they go out of sync with Sway's internal list of workspaces: https://github.com/Alexays/Waybar/blob/master/src/modules/sway/workspaces.cpp#L118

EDIT:  As suggested by danshick in freenode/#sway, I removed lines 116 - 160 in         https://github.com/Alexays/Waybar/blob/5ebd3594e484baa085920dfe1f5312c57e5509d0/src/modules/sway/workspaces.cpp#L115, and the workspaces now appear in the correct order, so it is here where they are being sorted.

languitar:
Ensure that sway workspaces are always displayed in the same order as
used internally by sway. The previous sorting code always sorted
unnumbered workspaces lexicographically. This isn't the order used by
sway internally. Therefore, commands such as "workspace next" might have
jumped arbitrarily in waybar.

This commit reworks the sorting code such that the internal order is
always obeyed. Additionally, numbered persistent workspaces are inserted
at their natural position at the front of the workspace list while
unnumbered ones are appended. This should match the expectations of
workspace ordering known from sway's behavior.

The changes make the configuration property "numeric-first" unnecessary
as this will always be the case now. There's also no reasonable way
around this behavior now. Otherwise, persistent workspaces would jump
around in the visual representation as soon as they become known to
sway.

Fixes #802

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -2,6 +2,7 @@
 
 #include <spdlog/spdlog.h>
 
+#include <algorithm>
 #include <cctype>
 #include <string>
 
@@ -98,6 +99,7 @@ void Workspaces::onCmd(const struct Ipc::ipc_response &res) {
                   Json::Value v;
                   v["name"] = p_w_name;
                   v["target_output"] = bar_.output->name;
+                  v["num"] = convertWorkspaceNameToNum(p_w_name);
                   workspaces_.emplace_back(std::move(v));
                   break;
                 }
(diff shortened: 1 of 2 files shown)


src/modules/sway/workspaces.cpp
```diff
@@ -107,57 +109,60 @@ void Workspaces::onCmd(const struct Ipc::ipc_response &res) {
...
+          max_num = std::max(workspace["num"].asInt(), max_num);
+        }
+        for (auto & workspace : workspaces_) {
+          auto workspace_num = workspace["num"].asInt();
+          if (workspace_num > -1) {
+            workspace["sort"] = workspace_num;
+          } else {
+            workspace["sort"] = ++max_num;
+          }
+          std::cout << workspace["name"] << ": " << workspace["sort"] << std::endl;
```
Alexays, line 146: remove

Alexays:
Thanks!

languitar:
Is there a chance for a release including this fix?

Alexays:
Sure, I'll try to do a release this weekend!

djvs:
I notice wlr/workspaces has an ordering option but sway/workspaces doesn't.  I was a little confused today noticing my painstaking "use format-icons to alias my alphabetically ordered workspace names to the human readable names" hack was no longer working.  What option do I have now to sort workspaces?  

For context, I have 0-13, f1-f12, and m, in that order.  I previously had them as 00-13, f1-f12, and f99 for "m" for the actual workspace names.  Now f99/m is just showing up the middle again.

languitar:
> use format-icons to alias my alphabetically ordered workspace names to the human readable names

I don't get how your hack works. What do you want to achieve?

The basic idea of this patch was to make the ordering of workspaces in waybar more consistent with the ordering used by sway internally such that commands like "workspace next" are consistent with the ordering of workspaces visible in waybar. This wasn't the case before and iterating workspaces with multiple "workspace next" calls could actually have been a wild jumping through the workspace list in the bar.

So, at least any naming scheme that sorts the workspaces the way you want in sway should be sufficient here.

djvs:
I want to have a deterministic sorting by name or something.  Right now it seems to be sorted by sway's numeric workspace IDs which means my "mail" window will be at an arbitrary position in the list (I think based on order of workspace creation after sway startup) instead of guaranteed to be at the last position like before.

languitar:
I think it's impossible to fullfil both requirements (always last and being consistent with sway) without a configuration flag. The sway/i3 way of sorting would always be to name the workspace appropriately. But there's no such thing as "always the last workspace" in case unnumbered workspaces exist. Creation order is what counts there.

djvs:
Well, it's "always last" in terms of my specified workspaces alphabetically.  So alphabetical sort would take care of it for my needs.  I can do a PR for the config flag?

languitar:
> Well, it's "always last" in terms of my specified workspaces alphabetically. So alphabetical sort would take care of it for my needs. I can do a PR for the config flag?

The only thing I see critical about this is that there's no such behavior in sway itself.

@Alexays what do you think?

djvs:
I take it back, I was opening the file to change it and figured out how I'm supposed to be doing this.  

sway config: 
```
set {
  $ws0 "0:0"
  $ws1 "1:1"
  [...]
  $ws10 "10:10"
  [...]
  $wsf1 "101:f1"
  [...]
  $wsfm "120:fm"
}
[...]
bindsym {
  [...]
  $mod+grave exec /whatever/sway/switch_workspace.sh $ws0 current
  $mod+1 exec /whatever/sway/switch_workspace.sh $ws1 current
  $mod+2 exec /whatever/sway/switch_workspace.sh $ws2 current
  [...]
  $mod+Shift+grave move container to workspace $ws0
  $mod+Shift+1 move container to workspace $ws1
  $mod+Shift+2 move container to workspace $ws2
}
```

waybar config:
```
[...]
     "sway/workspaces": {
         "numeric-first": true,
         "disable-scroll": false,
         "disable-markup" : false,
         "all-outputs": true,
         "format": "  {icon}  ",
         "format-icons": {
             "0": "0",
             "1": "1",
             "2": "2",
              [...]
             "fm": "",
         }
     },
[...]
```

languitar:
You can drop the `numeric-first` option now. It has been removed with this PR.

Issue #1379, opened 2022-01-08
libfmt 8.1.0 breaks compilation


mazunki:
`src/modules.clock.cpp` fails during compilation if using `>=libfmt-8.1.0` or above: 

https://github.com/Alexays/Waybar/blob/d9b5c2595a08e481520d96d7ee30daab4ac07f1b/src/modules/clock.cpp#L236

due to libfmt making `specs` a private variable:

https://github.com/fmtlib/fmt/commit/f8542cd9888b0f07e40500ad5813d3b8ad4d9c68

We should probably use `libfmt`'s formatter altogether.

mazunki:
Here is a hotfix which seems to work for me.

https://github.com/mazunki/Waybar/commit/2301788a810b64be9361b15fbd8f510002c74cc4

ahmadraniri1994:
> Here is a hotfix which seems to work for me.
> 
> [mazunki@2301788](https://github.com/mazunki/Waybar/commit/2301788a810b64be9361b15fbd8f510002c74cc4)

Thanks, bro.

ahmadraniri1994:
> Here is a hotfix which seems to work for me.
> 
> [mazunki@2301788](https://github.com/mazunki/Waybar/commit/2301788a810b64be9361b15fbd8f510002c74cc4)

Dunno why, I tried your fork but my compilation still complains about fmt.

mazunki:
> Dunno why, I tried your fork but my compilation still complains about fmt.

Can you tell me what error you're getting? I can take a look tomorrow.

mazunki:
Closing this issue since there's an active pull request to solve it.

PR #1380, opened 2022-01-08
Fix build with fmt 8.1.0

### LINKED ISSUE #1368: Can't compile with FMT 8.1.0
When compiling Waybar 0.9.8 with newly updated fmt 8.1.0 I see this error:

```
../src/modules/clock.cpp:200:20: error: 'specs' is a private member of 'fmt::formatter<tm>'
DEBUG util.py:636:          auto& tm_format = specs;
DEBUG util.py:636:                            ^
DEBUG util.py:636:  /usr/include/fmt/core.h:715:64: note: in instantiation of function template specialization 'fmt::formatter<waybar::modules::waybar_time>::format<fmt::basic_format_context<fmt::appender, char>>' requested here
DEBUG util.py:636:      -> decltype(typename Context::template formatter_type<T>().format(
```

Full build log with more details is here: https://file-store.openmandriva.org/api/v1/file_stores/597775db70004e7530f2b3f4935fe56c68cd7fa4.log?show=true

Info: 
Waybar 0.9.8
FMT 8.1.0
OS: OpenMandriva Cooker x86_64
Compiler: LLVM/Clang 13.0.0
Linker: LLD

alebastr:
* Stop using `fmt::formatter<std::tm>` implementation details
 * Add tests for the `waybar_time` formatter

Tests might be a bit fragile as the `date::format` output depends on an underlying STL locale implementation. Any difference there is a bug though, and things are working on two stdlib implementations I tested.

Alpine build fails as the tests require timezone data. Dockerfile change in the top commit should address that.

Fixes #1368, #1379

### DIFF
--- a/Dockerfiles/alpine
+++ b/Dockerfiles/alpine
@@ -2,4 +2,4 @@
 
 FROM alpine:latest
 
-RUN apk add --no-cache git meson alpine-sdk libinput-dev wayland-dev wayland-protocols mesa-dev libxkbcommon-dev eudev-dev pixman-dev gtkmm3-dev jsoncpp-dev pugixml-dev libnl3-dev pulseaudio-dev libmpdclient-dev sndio-dev scdoc libxkbcommon
+RUN apk add --no-cache git meson alpine-sdk libinput-dev wayland-dev wayland-protocols mesa-dev libxkbcommon-dev eudev-dev pixman-dev gtkmm3-dev jsoncpp-dev pugixml-dev libnl3-dev pulseaudio-dev libmpdclient-dev sndio-dev scdoc libxkbcommon tzdata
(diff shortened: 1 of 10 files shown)


mazunki:
Thank you, @alebastr . This seems much better than my quick hotfix. 👍🏽 

I had to skip patches 8c41aaa and 766c8a1 for it to compile, since it was complaining about "can't find file to patch" on those. 

I figured my upstream didn't include tests, so it should be safe. Right?

alebastr:
@mazunki yes, tests were added after 0.9.8 so it's safe to skip these patches. You can even skip a couple more and take only https://github.com/Alexays/Waybar/pull/1380/commits/c0d84853ea2f3f8b016abf92458ba15378de6c91 and https://github.com/Alexays/Waybar/pull/1380/commits/7f6bef20496ca23e83692801f6ff2957d28dde76.

@Alexays thanks for merging. Can you please update the alpine image?

PR #1382, opened 2022-01-08
[Need help] Implement river un/focused output

GenesisTMS:
On multiple outputs i am missing possibility to visually distinguish focused from unfocused output.

Can anybody point me to right direction or finish this. ( I feel i am so close )
I was so sure that this will work and feel hopeless why it crash.

Thanks.


jfred9:
The reason for the crash is that you have omitted to provide a valid function pointer for .focused_view in your output_seat_listener_impl struct. The function will be called and if it does not point to anything valid, waybar will crash.

You can keep the function empty but it needs to be there.

Alexays:
Closing in favor of #1562

PR #1394, opened 2022-01-14
fix: Use locale when formatting clock

### LINKED ISSUE #1393: date format ignores locale when timezone not set
When the timezone is not set, the locale, either defined or system locale is ignored and it always uses mon, tue, thu, etc for the week name. I think this is because when the timezone is set, Waybar uses the fmt library. https://github.com/Alexays/Waybar/blob/81f0bcb3a3fe8c6d89d7a3f05bada16da5dedb6a/src/modules/clock.cpp#L98-L105

Is there some way to get Waybar to use the system time and the locale at the same time?

nullobsi:
Fixes #1393

When formatting the clock, pass the locale to the fmt function.


Alexays:
Thanks!

Issue #944, opened 2020-12-15
[Feature Request] Add support for custom Icons in custom modules


pan1997:
Is there interest in having support for icon images in custom modules? One possible way to support this would be to have a new field `icon` (like `text`) that can be accepted. The display can be handled as it is currently being done in `wlr/taskbar` module.

frank-bell:
This would really help me out as I am currently scraping some data off a web site, some of which is only available as image urls. In my scenario I'd ideally need more than one image (e.g. opponent, competition, channel), so support for an array of images would be better.

skysphr:
As a workaround, you may integrate your icons into a font and use it in your waybar config.

lilydjwg:
> As a workaround, you may integrate your icons into a font and use it in your waybar config.

You can't dynamically add icons in that way.

I want this because I'm going to decorate my custom task info with app icons.

autofool:
I was looking for a way to provide dynamic path for `image/module` with JSON output `"{"path": "/path/to/image"}"` which is currently not possible since there is no `exec` functionality for image module. If custom module can show image then it will make lot more sense with scripts providing dynamic values for output. I'm creating a simple launcher button with current user's icon theme but user's icon theme can change so hard coding path in config is not ideal.

PR #1397, opened 2022-01-16
Add image module

### LINKED ISSUE #1191: Showing an image
I am using `background-image: url("path/to/png")` to show a background image in a widget. This is especially useful, when showing plots that are updated not too frequently.

Of course this approach does not work, because the background image is set only once and is never updated. As far as I can see, GTK actually uses base64 encoded image. I wonder if this approach can somehow be fixed to work with re-rendering.

JakeStanger:
Resolves #1191, #944

Module which renders an image onto the bar.

Example paired with the MPD module:

image

Example config:

```json
"image/album-art": {
  "path": "/tmp/mpd_art",
  "size": 32,
  "interval": 5,
  "on-click": "mpc toggle"
}
```

A few problems:

- For some reason right click isn't working for me. All other events are.
- I don't think the image can be styled

This evening is my first dive into C++ and first dive into GTK so this probably isn't fantastic, so criticism is very welcome.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -56,6 +56,7 @@
 #endif
 #include "bar.hpp"
 #include "modules/custom.hpp"
+#include "modules/image.hpp"
 #include "modules/temperature.hpp"
 #if defined(__linux__)
 #  ifdef WANT_RFKILL
--- a/include/modules/image.hpp
+++ b/include/modules/image.hpp
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <fmt/format.h>
+#include <gtkmm/image.h>
+
+#include <csignal>
+#include <string>
+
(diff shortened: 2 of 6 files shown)


bd-g:
This looks really useful - any reason we haven't merged it yet?

bd-g:
Just kidding, there totally is a man page on this commit, I missed it.

yemouu:
I think it would be interesting if images could be integrated with other modules like the `pulseaudio` module instead of just being it's own module so that images can be displayed depending on the state the modules and not have to wait for an interval. Would be nice for those that would rather use images over icon fonts

Alexays:
Thanks!

Alexays:
Can you also update the Github Wiki?

Issue #1399, opened 2022-01-17
Options for modules/sway/window: empty, only_floating, only_tiling


dmyTRUEk:
Currently there are only 3 options in css for window: `empty`, `solo`, `<app_id>`.

In 0.9.8 `waybar::modules::sway::leafNodesInWorkspace` returned only the amount of tiling windows, but in 0.9.9 (changed in [this commit](https://github.com/Alexays/Waybar/commit/99723845976fb577a057b86e36b502c966bf254f#diff-8b4bd9b29db6e716716525789b945720af9e21335d52922fde1e4fc0ca422f3a)) it was changed to return the amount of tiling and floating windows.

On 0.9.8 i had this in my `style.css`:
```
window#waybar.empty {
    background-color: transparent;
}
```
and had transparent bar when there were no windows at all, or when there were only floating windows.

But now (on 0.9.9) it's impossible to configure bar color for the case when i don't have tiling windows, but have floating windows :(

I think css could have some more options (smt like this): `only_floating`, `only_tiling`, maybe even `solo_floating`, `solo_tiling`, which will be quite usable.

RobertMueller2:
Sorry, that was my fault, I didn't think there was a use case in ignoring floating windows. ;)

There are a lot of different possibilities:

empty
0 tiled and 1 floating
0 tiled and n floating
1 tiled and 0 floating
1 tiled and 1 floating
1 tiled and n floating
n tiled and 0 floating
n tiled and 1 floating
n tiled and n floating

Personally, I think when adjusting this, they should all somehow be covered. Either the code gets somewhat uglier, or we just add classes along the lines of 0-tiled or 1-tiled, or n-tiled as well as 0-float or 1-float or n-float, and then the user has to style the different combinations to their liking. Either way it inevitably breaks existing styles.

Even though I'm not terribly comfortable with or experienced in C++, I'm not totally unwilling to try and help here. (No promises on time though.)

But it would be good if @Alexays could point out a direction.

RobertMueller2:
Just another thought. Alternatively, there could be a config option to decide between old and new behaviour? Has the advantage of not breaking existing styles.

dmyTRUEk:
In my opinion, the best solution is to add new options, without changing behavior of an old options (comparing to 0.9.9 or 0.9.8?).

deuill:
[...]

PR #1405, opened 2022-01-25
add skip-floating-windows option to restore old behaviour of window c…

RobertMueller2:
Quick fix suggestion for https://github.com/Alexays/Waybar/issues/1399 .

This is not implementing any of the suggested new css classes. Instead, this adds a bool option
"skip-floating-windows" : true
that should restore the old behaviour if desired.

Seems to be working for both use cases.


RobertMueller2:
I guess I should add the option to the man page. Will do that in the evening.

RobertMueller2:
Closing in favour of https://github.com/Alexays/Waybar/pull/1419 .

PR #1408, opened 2022-01-27
Switch to more powerful, regex-based output matching

pocek:
This change mainly removes one limitation: it was not possible to have multiple negative matches. Now it's easy:

```json
"output": "(?!HDMI-A-1|eDP-1).*?/.*",
```

The format of the string matched is now *name*/*identifier*, because making two separate matches wouldn't make sense from logical perspective.

It also makes the code simpler.


Alexays:
Sorry for the delay, we need to keep current possibilities for breaking changes purpose.

​Bu why not if the string begin and end with / trigger this new feature?

JassonCordones:
is this still open?

PR #1410, opened 2022-01-28
Reverse the tray icon order

ErikReider:
At the moment when a new appindicator is added, the icon gets added to the end which isn't very good if the tray is positioned on the right of the bar.

With `reverse-direction` set to true, new app indicators get added to the start


Alexays:
Thanks, can you also update GitHub wiki? :)

ErikReider:
Ofc! :)

ErikReider:
The wiki is now updated

Issue #1414, opened 2022-01-29
Disabling clock tooltip has no effect


MultisampledNight:
This config:

```
{
	"modules-center": ["clock"],
	"clock": {
		"tooltip": false,
	}
}
```

fails with the tooltip still being visible at hover. While the option `tooltip` on the `clock` module is documented in the wiki, it is not in the manpage `waybar-clock`, so it may not even be implemented. Is this intended behavior? If yes, which reason had the non-implementation?

The tooltip is slightly annoying when switching screens over barrier/waynergy and the cursor stays at the top of the screen, but nothing too bad.

MultisampledNight:
Disabling the tooltip works now, thanks @gemmelos! Will this be documented in the manpage, too? I did not see any change to `man/waybar-clock.5.scd` in the PR.

gemmelos:
> Disabling the tooltip works now, thanks @gemmelos! Will this be documented in the manpage, too? I did not see any change to `man/waybar-clock.5.scd` in the PR.

Good catch, I forgot about that. #1421 

(I assumed this was a universal/global option that worked the same for each module, and was documented like this. Seems like it is documented for each module separately)

MultisampledNight:
Nice, thanks!

PR #1418, opened 2022-02-02
Fix disabled tooltip on clock module

gemmelos:
Seems like the tooltip text should only be set if both `if (tooltipEnabled())` and `if (config_["tooltip-format"].isString())` passes.

Addresses #1414


Alexays:
Thanks!

Issue #1349, opened 2021-12-11
sway: waybar background based on smart_gaps


primalmotion:
I've looked around but I could not find any issue related to this. I would like to be able to use a different style for the general background of waybar based on gaps. I use smart_gaps in sway that basically removes any gaps when there is only one window in a workspace. I usually set the waybar background transparent, but I would like to use a solid color when the windows actually "touch" waybar.

Is this something we can do and I missed it, or is it something we could add support for in waybar?

Awesome project otherwise!

WuerfelDev:
I don't think it's currently possible
My 'solution' was to use a black-to-transparent gradient.

This works well for me:

    background: linear-gradient(0deg, rgba(0,0,0,0.05) 0%, rgba(0,0,0,0.6) 100%);

(https://gitlab.com/WuerfelDev/dotfiles/-/blob/master/dots/waybar/.config/waybar/style.css#L10)

![screenshot](https://user-images.githubusercontent.com/12686819/148376689-65106c5c-f0e3-4229-9871-119174e2dc50.png)

WuerfelDev:
You should be able to change the `name` (which is essentially a css class) of the bar via IPC: https://github.com/Alexays/Waybar/wiki/Configuration#bar-config
Then in your style.css just add this: (assuming you set the name to 'singlewindow')
```css
window#waybar.singlewindow {
    background: black;
}
```


Then with a nice fade out transition:
```css
window#waybar {
    background: rgba(0,0,0,0);
    transition: background .3s;
}

window#waybar.singlewindow {
    background: rgba(0,0,0,1);
    transition: none;
}

```

primalmotion:
Oh. This feels so close, but I can't figure out how to change the bar name from IPC..

WuerfelDev:
I have not used IPC so I don't really know, but I have another idea.
From what I've read here is that you can change the mode via IPC so you could have two almost identical sway bars, one with solid color and one transparent. Then you have to toggle the modes for both bars between your default mode and invisible. However I am not sure if you can toggle modes individually per bar...

Or the hacky way: trigger a bash script that swaps the style.css with a copy that has a solid background bar. Then reload waybar `killall -SIGUSR2 waybar`

primalmotion:
It feels hacky. I will try to do dig a bit more on switching the bar name. thanks for the suggestions!

victordiaz:
I would love to have something like this.

[...]

PR #1419, opened 2022-02-02
Sway window newstyles

RobertMueller2:
A more sophisticated implementation for issue 1399, following the suggestions from https://github.com/Alexays/Waybar/issues/1399#issuecomment-1024913010

- Might break existing styles, but only if people relied on floating nodes being counted (introduced in commit 99723845976fb577a057b86e36b502c966bf254f -- so basically kinda breaks again what we fixed in that recent commit, but by using the new floating style, there should be a suitable workaround.)
- Provides classes empty, floating, tabbed, tiled, solo, stacked, and app_id. Some trace log output.
- Adds an "offscreen-css" bool option (default false), only effective when "all-outputs" is set. This provides styling according to windows on the given output when it's not focused, rather than showing the focused screens focused window and style.
- Adds an "offscreen-css-text" string option (default empty), only effective when "all-ouputs" and "offscreen-style" are set. This is shown as a text on the unfocused output.
- Adds a "show-focused-workspace" bool option (default false) to indicate the workspace name if the whole workspace is focused if nodes are present. If not set, empty text is shown but css classes according to nodes in the workspace are still applied.

Limitations:
- when the top-level layout changes (e.g. tabbed to stacked), there is no sway event (well, there might be a keybinding event, but that is too weak) therefore the implementation cannot react here. The only way around I could think of would be to poll get_tree on a timer. Asking that sway fire an event for this is unlikely, considering that i3 doesn't do it either. I think I tested that. If I remember correctly.
- I think the recursion could still be improved, getFocusedNode already iterates all leaves, and then leafNodesInWorkspace does it again for at least one workspace per output. It did not come across as a _problem_ during testing, but I think the code could be more concise. On the other hand, the given change is already big enough as it is, but I have as much as I think was possible tried to stay within the existing implementation. I wouldn't exactly feel comfortable with almost rewriting the whole thing. ;)
- while we're at it, I think the active workspace as well as the focused window could be passed as Json::Value instead of string content, then the update function can handle how the info is displayed, this could allow for different replacements for format, like marks, idle_inhibition status...

Review questions:
- I'm not sure if the case for immediateParent is good. It looks like parentWorkspace (which existed before), but I'm unsure if that's correct according to the style guideline.

Still to do:

- [x] FIXME: app_id should only be set when it's the only window. manpage says so, so keep that behaviour.
- [x] FIXME: there's still a problem with floating windows
- [x] Optimise code here and there
- ~~make output/workspace filtering optional~~ EDIT: Don't think this makes sense.
- [x] use clang-format on each of the commits to allow rebase again
- [x] consider "offscreen" style vs offscreen window situation, based on option
- [x] revisit code style
- ~~perhaps add "floating"/"tiled" styles depending on what the focused window is~~ EDIT: leaving this for a future feature
- [x] Man page update
- [x] ~~show-focused-workspace still isn't correct~~
- [ ] More testing after rebasing

Tested with this style:

```
window#waybar {
  background-color: #990000;
}

window#waybar.empty {
  background-color: transparent;
}

window#waybar.empty #window {
  padding: 0px;
  margin: 0px;
  border: 0px;
/*  background-color: rgba(66,66,66,0.5); */ /* transparent */
  background-color: transparent;
}

window#waybar.solo #window {
    padding-left: 5px;
    padding-right: 5px;
    color: #eee8d5; /* base2 */
    background-color: #073642; /*base02*/
}

window#waybar.floating #window {
    padding-left: 5px;
    padding-right: 5px;
    color: #eee8d5; /* base2 */
    background-color: #b58900; /*yellow*/
}

window#waybar.tiled #window {
    padding-left: 5px;
    padding-right: 5px;
    color: #eee8d5; /* base2 */
    background-color: #cb4b16; /* orange */

}
window#waybar.stacked #window {
    padding-left: 5px;
    padding-right: 5px;
    color: #eee8d5; /* base2 */
    background: #2aa196; /*cyan*/
}
window#waybar.tabbed #window {
    padding-left: 5px;
    padding-right: 5px;
    color: #eee8d5; /* base2 */
    background: #859900; /*green*/
}

window#waybar.code {
    background-color: #007ACC;
}

```

and this config (testing all combinations of all-output/offscreen-style, offscreen-text set or not set):

```
{
    "layer": "top", // Waybar at top layer
    "position": "top", // Waybar position (top|bottom|left|right)
    "height": 39, // Waybar height (to be removed for auto height)
    "modules-center": ["sway/window", "custom/media"],
    "sway/window": {
        "format": "{}",
        "all-outputs" : true,
        "offscreen-css" : true,
        "offscreen-css-text": "unfocused",
        "show-focused-workspace-name" : true,
        "max-length": 50,
        "icon": true,
        "rewrite": {
            "(.*) - Mozilla Firefox": " $1",
            "fish (.*)": "> [$1]"
        },
}
```

### DIFF
--- a/include/modules/sway/window.hpp
+++ b/include/modules/sway/window.hpp
@@ -19,10 +19,11 @@ class Window : public AIconLabel, public sigc::trackable {
   auto update() -> void;
 
  private:
+  void setClass(std::string classname, bool enable);
   void onEvent(const struct Ipc::ipc_response&);
   void onCmd(const struct Ipc::ipc_response&);
-  std::tuple<std::size_t, int, std::string, std::string, std::string, std::string> getFocusedNode(
-      const Json::Value& nodes, std::string& output);
+  std::tuple<std::size_t, int, int, std::string, std::string, std::string, std::string, std::string>
+  getFocusedNode(const Json::Value& nodes, std::string& output);
   void getTree();
   void updateAppIconName();
   void updateAppIcon();
@@ -32,12 +33,14 @@ class Window : public AIconLabel, public sigc::trackable {
   int windowId_;
   std::string app_id_;
   std::string app_class_;
+  std::string layout_;
   std::string old_app_id_;
   std::size_t app_nb_;
   std::string shell_;
   unsigned app_icon_size_{24};
   bool update_app_icon_{true};
   std::string app_icon_name_;
+  int floating_count_;
   util::JsonParser parser_;
   std::mutex mutex_;
   Ipc ipc_;
--- a/man/waybar-sway-window.5.scd
+++ b/man/waybar-sway-window.5.scd
@@ -66,6 +66,25 @@ Addressed by *sway/window*
 	default: true ++
 	Option to disable tooltip on hover.
 
+*all-outputs*: ++
+	typeof: bool ++
+	default: false ++
+	Option to show the focused window along with its workspace styles on all outputs.
+
+*offscreen-css*: ++
+	typeof: bool ++
+	default: false ++
+	Only effective when all-outputs is true. Adds style according to present windows on unfocused outputs instead of showing the focused window and style.
+
+*offscreen-css-text*: ++
+	typeof: string ++
+	Only effective when both all-outputs and offscreen-style are true. On screens currently not focused, show the given text along with that workspaces styles.
+
+*show-focused-workspace-name*: ++
+	typeof: bool ++
+	default: false ++
+	If the workspace itself is focused and the workspace contains nodes or floating_nodes, show the workspace name. If not set, text remains empty but styles according to nodes in the workspace are still applied.
+	
 *rewrite*: ++
 	typeof: object ++
 	Rules to rewrite window title. See *rewrite rules*.
@@ -117,6 +136,10 @@ Invalid expressions (e.g., mismatched parentheses) are skipped.
 # STYLE
 
 - *#window*
-- *window#waybar.empty* When no windows is in the workspace
-- *window#waybar.solo* When one window is in the workspace
+- *window#waybar.empty* When no windows are in the workspace, or screen is not focused and offscreen-text option is not set
+- *window#waybar.solo* When one tiled window is in the workspace
+- *window#waybar.floating* When there are only floating windows in the workspace
+- *window#waybar.stacked* When there is more than one window in the workspace and the workspace layout is stacked
+- *window#waybar.tabbed* When there is more than one window in the workspace and the workspace layout is tabbed
+- *window#waybar.tiled* When there is more than one window in the workspace and the workspace layout is splith or splitv
 - *window#waybar.<app_id>* Where *app_id* is the app_id or *instance* name like (*chromium*) of the only window in the workspace
--- a/src/modules/sway/window.cpp
+++ b/src/modules/sway/window.cpp
@@ -17,7 +17,7 @@
 namespace waybar::modules::sway {
 
 Window::Window(const std::string& id, const Bar& bar, const Json::Value& config)
-    : AIconLabel(config, "window", id, "{title}", 0, true), bar_(bar), windowId_(-1) {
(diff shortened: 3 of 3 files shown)


RobertMueller2:
Something I noticed, when switching the workspace layout (e.g. tabbed to splith), there is no immediate class change. Only after focus change.

I'm not exactly sure how we can deal with that, because the only event I saw was for the keybinding.

primalmotion:
This seems that can be a solution to my issue here too #1349 
However I can't get that patch to work. I have added the classes described above but it does not seems to change anything

RobertMueller2:
I think your use case should even work before the patch, I'll add a comment in #1349.

That makes me wonder if we should try and support "legacy" styles, I guess if I don't add that, this could well become a support nightmare.

Also I'm wondering if it helps for easier to understand styling if we also add classes to the label, not just the whole bar.

But I'm not sure why the patch doesn't do anything for you. Do you see any trace output from the module?

primalmotion:
As stated in the #1349, my issue was that I did not use the sway-window module. it works when I enable it. sorry for the noise :)

RobertMueller2:
> As stated in the #1349, my issue was that I did not use the sway-window module. it works when I enable it. sorry for the noise :)

no worries :)

RobertMueller2:
still to do:
- man page update
- more testing
- Consider what to do with https://github.com/Alexays/Waybar/issues/1399#issuecomment-1032961170 . It's a bigger change, though.

RobertMueller2:
Still to do:

- Optimise code here and there
- Man page update
- More testing

RobertMueller2:
Just noticed that not all nodes necessarily have the output property set. That requires more changes.

RobertMueller2:
Sadly, the current approach fails with several tiled windows. I was hoping not to change the underlying logic too much, but now I think it'll be more work. Could take a bit, too.

RobertMueller2:
I think I've done all I can here. I'll test the latest for a few more days and then remove the draft flag.

RobertMueller2:
While using this PR, I've noticed some behaviour which seems to be related to external screen connection, disconnection and reconnection.  The display of the focused window gets "stuck" when changing screen focus, because there is some node that is being iterated where there's no Type child node and/or string comparison is not possible.

I'm not sure if I can reproduce it just like that, but I feel I need to investigate further what went wrong, which, unfortunately, delays the PR even more. (And if possible I'd like to find out if the main branch already has the same issue.) 

Perhaps this is an issue in the get_tree, but even if it is, the module needs to be able to deal with it.

RobertMueller2:
I've had a look at the failing FreeBSD check and I'm not sure why it fails. It seems to be a timeout while installing packages rather than actual building. I've setup a FreeBSD VM and I can compile and run the test suite without problems.

RobertMueller2:
Rebasing is adding more and more pain. :D I don't think I can guarantee the sanity of each individual commit any longer, which defeats the purpose of keeping them, so I decided to squash them.

I'd like to get this out of the way, but unfortunately, it's difficult at the moment to find enough time to properly test this.

primalmotion:
I use this patch especially for the .floating and it works as expected for me

primalmotion:
Oh noes. conflicts :(

RobertMueller2:
> Oh noes. conflicts :(

rebased, at first glance still seems to work.

primalmotion:
Thanks! Yeah everything I use still seems to work

Alexays (approved):

RobertMueller2:
I've been using this for ages and have not encountered any issues, except one waybar crash in the context of replugging an external screen, where I was unable to replicate, and where I'm not even sure the reason was the module itself. 

That said, I'm a bit swamped at this time. It might be a few weeks before I'm able to look into any issues, should any be reported.

RobertMueller2:
Sure, done.

While doing so I've noticed a small error I've added to the manpage, will add a pull request in a min.

PR #1444, opened 2022-02-22
CPU Load Formatting Fix

### LINKED ISSUE #1443: CPU module shows load with high precision
I've been using waybar on Arch for 3 weeks now and though it hasn't been updated (0.9.9-1), CPU module with simple configuration:
```
...
    "cpu": {
         "format": "{usage}%  {load}",
         "tooltip": false
     },
...
```
lately shows load with high precision like with 10+ decimal digits, which wasn't the case in the first place. Looks like a representation bug to me, since LA is always shown with exactly 2 decimal digits.

bd-g:
Limit cpu load figure to 2 digits right of decimal point - standard formatting for cpu load

Fixes #1443


Alexays:
Thanks!

PR #1333, opened 2021-12-01
Feature Clock: show list of time in other timezones in a tooltip

dartkron:
Introducing new tooltip placeholder: {timezoned_time_list}. It will be replaced with the list of times in different time zones from the config.
I've found it useful to hover the mouse pointer on time and see time in all my timezones at once. Current timezone excluding for the list, so if you will scroll over the time module and change the active timezone, this timezone will be excluded from the list and the previous active zone will be added.

How it works for me:
output

I know that this feature wasn't requested, but I found it very useful for me and hope it will be useful for someone else, who working remotely or have another kind of interest in other timezones.
Refactoring in #1277 was done as a preparation for this feature =)
Any feedback will be highly appreciated.

PR #1454, opened 2022-03-06
Add a config to set a timeout for the idle_inhibitor module

### LINKED ISSUE #1288: [Feature Request] Standby Inhibitor timeout
I can't count the number of times I've hit the standby inhibitor and forgotten to turn it off later.
I generally want to inhibit standby for a set period of time (30 mins, 1 hour, 2 hours etc.)

Similar tools work by selecting a timeout, but could also work by clicking multiple times to cycle between a set of timeout periods.

I may get around to trying to implement this myself at some point... but wanted to put it here meanwhile in case I don't.

Psykar:
My C++ is almost non-existant, so I've probably made some rookie mistakes here, but this does appear to do the job.
Feel free to be brutal in review, I'm more than happy to improve this if required.

Fixes #1288


Psykar:
Looks like a recent merge causes the freebsd check to fail
https://github.com/Alexays/Waybar/pull/1333

```
In file included from ../include/factory.hpp:5:
../include/modules/clock.hpp:39:36: error: no member named '_V2' in namespace 'std::__1::chrono'
  auto timezones_text(std::chrono::_V2::system_clock::time_point *now) -> std::string;
                      ~~~~~~~~~~~~~^
```

src/modules/idle_inhibitor.cpp
```diff
@@ -23,6 +24,17 @@ waybar::modules::IdleInhibitor::IdleInhibitor(const std::string& id, const Bar&
   // Add this to the modules list
   waybar::modules::IdleInhibitor::modules.push_back(this);
 
+  thread_ = [this] {
```
Alexays, line 27: Check if timeout is defined before creating a useless thread

alebastr:
Actually, it should be possible to do that without creating a thread:

```c++
// not a real code but shows the general idea. not tested!
class IdleInhibitor {
// ...
private:
static sigc::connection timeout_;
}

sigc::connection waybar::modules::IdleInhibitor::timeout_ {};

bool waybar::modules::IdleInhibitor::handleToggle(...) {
  if (e->button == 1) {
    status = !status;

    if (timeout_) {
      /* cancel any already active timeout handler */
      timeout_.disconnect();
    }
    if (status && config_["timeout"].isNumeric()) {
      auto timeoutMins = config_["timeout"].asDouble();
      int timeoutSecs = timeoutMins * 60;
      timeout_ = Glib::signal_timeout().connect_seconds([]() {
        /* intentionally not tied to a module instance lifetime
         * as the output with `this` can be disconnected
         */
        spdlog::info("deactivating idle_inhibitor by timeout");
        status = false;
        for (auto const& module : waybar::modules::IdleInhibitor::modules) {
          module->update();
        }
        /* disconnect */
        return false;
      }, timeoutSecs);
    }
    
    // ...
  }
}
```

src/modules/idle_inhibitor.cpp
Psykar: Reworked to use a sigc::connection instead :)

Psykar:
> Actually, it should be possible to do that without creating a thread:

Thanks, not being familiar with gnome programming (or C++) this helped a lot! It was pretty close to what was needed it seems.

PR #1476, opened 2022-03-10
Fix freebsd action

Psykar:
I'm not clear on the implications of this change, but it seems to fix the freebsd github action error.
Fixes error introduced by https://github.com/Alexays/Waybar/pull/1333


dartkron:
@Psykar Thank you for the fix! Sorry I'm late =(

Issue #491, opened 2019-10-26
Configuration to display waybar over tiled and fullscreen windows?


l3s2d:
Is there configuration that will display `waybar` over all windows?  Similar to how floating windows are positioned when `"layer": "top"` is set. This would be used in conjunction with `killall -SIGUSR1 waybar` to toggle the bar.

Here is the config to achieve this with `swaybar`:
```
bar {
    mode hide
    position top
    modifier none
}

bindsym $mod+m bar hidden_state toggle
```

maximbaz:
> Similar to how floating windows are positioned when "layer": "top" is set.

Sorry for the offtopic, but @l3s2d could you please share some details on how to set this `"layer": "top"` for floating windows so that they appear on top of full-screen apps? I'm searching but seemingly can't find anything...

l3s2d:
@maximbaz I think I was referring to how floating windows can appear under waybar when `"layer": "top"` is set in waybar's config. To my knowledge there isn't a way to accomplish what you're asking. Sorry for the poor wording!

kit-ty-kate:
I would be really happy to have this feature as well

vfsoraki:
Is this implemented? In the meantime, I'm using this in `sway` config file to show/hide it
```
bindsym --no-repeat $mod+q exec pkill -USR1 waybar
```
But the thing is when I hit `$mod+q` and waybar shows, containers resize. It would be good to have waybar overlay on top of containers, like how `mode overlay` works for `sway-bar`.

victordiaz:
I'd love to have this too!

x10an14:
Is this what I'm experiencing when say, steam popups are partially covered by waybar?

barbiefan:
if you set "layer": "overlay" waybar is rendered on top of fullscreen windows (at least in hyprland)

TheHighGroundWins:
> if you set "layer": "overlay" waybar is rendered on top of fullscreen windows (at least in hyprland)

Does it still work when the bar is set to toggle.

On swaywm, it works until you setup waybar to be toggleable, at which point it doesn't show up in Fullscreen for some reason.

Alexays:
This is configurable: set `"layer": "overlay"` in your bar config and Waybar will render in front of tiled and fullscreen windows (`overlay` is a supported value alongside `top`/`bottom`, see `src/bar.cpp`):

```jsonc
{
    "layer": "overlay",
    // ...
}
```

[...]

PR #1479, opened 2022-03-12
Use user-defined mode when toggling visibility rather than resetting to default

### LINKED ISSUE #958: Height and overlay configuration?
I have a waybar that I'm running as my standard sway statusbar, but I'm trying to make an extra sidebar that I can call with a shortcut (correct me if this isn't really possible with waybar), and I'm running into 2 issues:
1. I'm trying to change the height (to make the top be below my existing waybar status bar), but setting "height" in my config hasn't done the trick. Can anyone help me figure out why this isn't working as intended?
2. I'd like it to be an overlay for what's already on the window, but right now it's shifting everything over as if it's a dock. Is this the only option or is there a way of changing this behavior?

Here are my config and stylesheet:
```json
{
    "layer": "top",
    "position": "left",
    "height": 500,
    "width": 320,
    "modules-left": ["clock#date","clock#time", "battery"],
    "clock#date": {
        "format": "{:%a %b %d, %Y}"
    },
    "clock#time": {
        "format": "{:%H:%M}"
    },
    "battery": {
        "format": "{icon}",
        "format-icons": ["", "", "", "" ],
        "format-charging": " {capacity}%",
        "states": {
            "happy": 100,
            "ok": 45,
            "sad": 20,
            "critical": 5
        }
    },
}
```

```css
@define-color bg #040501;
@define-color fg #cccccc;
@define-color color0 #2d2d2d;
@define-color color1 #f2777a;
@define-color color2 #99cc99;
@define-color color3 #ffcc66;
@define-color color4 #6699cc;
@define-color color5 #cc99cc;
@define-color color6 #66cccc;
@define-color color7 #cccccc;
@define-color color8 #999999;
@define-color color9 #f99157;
@define-color color10 #393939;
@define-color color11 #515151;
@define-color color12 #b4b7b4;
@define-color color13 #e0e0e0;
@define-color color14 #a3685a;
@define-color color15 #ffffff;

* {
    border: none;
    /* min-height: 0; */
    font-family: sans;
    font-size: 18px;
    background-color: @bg;
    color: @fg;
    border-radius: 0 80px 0 0;
}

#clock {
    font-family: display;
    font-size: 44pt;
    margin: 14pt 0
}

#clock.date {
    font-size: 14pt;
    margin-top: 44pt;
}

#battery {
    font-family: typicons;
    font-size: 64pt;
}

#battery.happy {
    color: @color2;
}

#battery.ok {
    color: @color3;
}

#battery.sad {
    color: @color1;
}

#battery.critical {
    color: @color1;
}

#battery.charging {
    color: black;
}
```

dorgnarg:
Fixes #958 and #491 (or at least [this comment](https://github.com/Alexays/Waybar/issues/491#issuecomment-799217823)), this way if the bar is configured to be in overlay mode, it will stay in that mode when it's re-shown after being hidden. If there's a better way to do this, let me know, I just reused the code elsewhere that retrieved the config mode in the first place

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -665,7 +665,11 @@ void waybar::Bar::onMap(GdkEventAny*) {
 
 void waybar::Bar::setVisible(bool value) {
   visible = value;
-  setMode(visible ? MODE_DEFAULT : MODE_INVISIBLE);
+  if (auto mode = config.get("mode", {}); mode.isString()) {
+    setMode(visible ? config["mode"].asString() : MODE_INVISIBLE);
+  } else {
+    setMode(visible ? MODE_DEFAULT : MODE_INVISIBLE);
+  }
 }
 
 void waybar::Bar::toggle() { setVisible(!visible); }


src/bar.cpp
```diff
@@ -665,7 +665,11 @@ void waybar::Bar::onMap(GdkEventAny*) {
 
 void waybar::Bar::setVisible(bool value) {
   visible = value;
-  setMode(visible ? MODE_DEFAULT : MODE_INVISIBLE);
+  if (auto mode = config.get("mode", {}); mode.isString()) {
+    setMode(visible ? config["mode"].asString() : MODE_INVISIBLE);
```
Alexays, line 669: Do you think it's worse to check if mode is available?
Against a hard-coded list, for example.

alebastr: `setMode` already does the check with fallback to default, no need to do it twice.

alebastr (approved):

PR #1406, opened 2022-01-25
sway/language: fix segmentation fault for layouts not found in XKBContext

Spiffyk:
This PR fixes a segmentation fault that happens on startup when there is a keyboard layout in use that does not appear in the `XKBContext` object. This may happen when one has a custom layout in e.g. `~/.xkb/symbols`.

Fixes #1387, possibly #1307

PR #1480, opened 2022-03-13
sway/language: Load exotic xkb rules too

avdv:
Hi.

I am using the `EurKEY` (`eu`) keyboard layout, which apparently is an "exotic" one.

Since waybar does not load the exotic ones, there is nothing displayed in the bar for the language module.

See my comment https://github.com/Alexays/Waybar/pull/1406#issuecomment-1059074832

### DIFF
--- a/src/modules/sway/language.cpp
+++ b/src/modules/sway/language.cpp
@@ -179,8 +179,7 @@ auto Language::init_layouts_map(const std::vector<std::string>& used_layouts) ->
 }
 
 Language::XKBContext::XKBContext() {
-  context_ = rxkb_context_new(RXKB_CONTEXT_NO_DEFAULT_INCLUDES);
-  rxkb_context_include_path_append_default(context_);
+  context_ = rxkb_context_new(RXKB_CONTEXT_LOAD_EXOTIC_RULES);
   rxkb_context_parse_default_ruleset(context_);
 }
 


src/modules/sway/language.cpp
```diff
@@ -179,8 +179,7 @@ auto Language::init_layouts_map(const std::vector<std::string>& used_layouts) ->
 }
 
 Language::XKBContext::XKBContext() {
-  context_ = rxkb_context_new(RXKB_CONTEXT_NO_DEFAULT_INCLUDES);
-  rxkb_context_include_path_append_default(context_);
+  context_ = rxkb_context_new(RXKB_CONTEXT_LOAD_EXOTIC_RULES);
```
avdv, line 182: BTW, I removed the `RXKB_CONTEXT_NO_DEFAULT_INCLUDES` flag (and the call to `rxkb_context_include_path_append_default`) since I found this to be quite ugly:
```
context_ = rxkb_context_new(static_cast<rxkb_context_flags>(RXKB_CONTEXT_NO_DEFAULT_INCLUDES | RXKB_CONTEXT_LOAD_EXOTIC_RULES));
```
and I see no reason to avoid loading the default rules at first, when the next statement adds them explicitly.

Alexays (approved):

Alexays:
Thx!

PR #1481, opened 2022-03-14
Battery module. Gamepads support

LukashonakV:
Standard battery module doesn't support of the gamepads such as PS Dualshock are connected over bluetooth . Moreover this module crashes waybar in case when wrong battery info is provided in the json .config. 
Current two commits bring next fixes:
1. Do not propagate raised exception when physical device disappeared in the system. When wrong config or the gamepad was unplugged and device is removed from the /sys/classes the module just is set to invisible mode. Once device is back, module will print info just after provided delay.
2. Gamepad devices are recognized in the linux such a power_supply class but with the limited information about devices. For example PS Dualshock provides the only capacity, status. For such devices module does not do capacity estimation and uses provided capacity by the gamepad controller as is.


Alexays (approved):

Alexays:
Thanks!

PR #1482, opened 2022-03-15
Possibility to use different format-icons profiles

LukashonakV:
The fix is pretty simple. In case when the module provides an array of icon profiles getIcon method does searching the current set of icons needs to be displayed. By default is used: format-icons.
Example:
ps_2022-03-15-13_32_14
Config example:
`"battery": {
        "bat": "BAT0",
        "states": {
            // "good": 95,
            "warning": 30,
            "critical": 15
        },
        "format-charging": "{icon} {capacity}% ",
        "format-plugged": "{capacity}% ",
        "format-full": "{capacity}% ",
        // "format-good": "", // An empty format will hide the module
        "format-icons": ["","","","",""],
        "format-icons-charging": ["","","","","","",""],
        "tooltip": true
    },`

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -80,6 +80,16 @@ std::string ALabel::getIcon(uint16_t percentage, const std::string& alt, uint16_
 
 std::string ALabel::getIcon(uint16_t percentage, const std::vector<std::string>& alts, uint16_t max) {
   auto format_icons = config_["format-icons"];
+  for(const auto& iconsSet : alts) {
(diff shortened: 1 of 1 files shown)


Alexays:
Why not using object?
```
{"format-icons": {"charging": ["","","","","","",""], "default":["","","","",""] }}
```

LukashonakV:
Let me check if it works. I thought it wasn't fixed.

LukashonakV:
Yeah. Sorry. Please decline PR.

PR #1489, opened 2022-03-18
feat: added network speed in Bytes

momokrono:
Hi, this is a simple PR to add B/s for network speed, since it's the standard for non french speakers.
Since an octet is a byte, the code was already there.


Alexays:
Thanks!

momokrono:
Thank you instead!
Should I also update the wiki page on the network module?

momokrono:
I updated it!

PR #1495, opened 2022-03-28
Battery module support for rk817

0cc4m:
I have a device with a battery driver that is not currently supported by your battery module. A PineNote with a Rockchip RK817 battery controller. The driver is not upstreamed yet, but I think the changes are generic enough for this not to matter.

```
» ls /sys/class/power_supply/rk817-battery/
charge_empty_design  charge_full_design  charge_type                  constant_charge_voltage_max  current_boot  hwmon0  present  subsystem  uevent       voltage_boot        voltage_min_design  wakeup4
charge_full          charge_now          constant_charge_current_max  current_avg                  device        power   status   type       voltage_avg  voltage_max_design  voltage_ocv
```

It does not have a capacity, so it fails the initial check, and it does not have `current_now` or `voltage_now`.
I added `charge_now` to the initial file check, since `capacity` is only needed if `charge_now` doesn't exist.
I also added a check whether `current_now` or `voltage_now` exist, and if not attempt to read `current_avg` and `voltage_avg`

Please let me know what you think.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -88,8 +88,9 @@ void waybar::modules::Battery::refreshBatteries() {
       auto dir_name = node.path().filename();
       auto bat_defined = config_["bat"].isString();
       if (((bat_defined && dir_name == config_["bat"].asString()) || !bat_defined) &&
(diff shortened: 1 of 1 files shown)


Alexays:
Thx!

PR #1510, opened 2022-04-07
Show hidden bar

towoe:
- Show bar if a workspace becomes urgent
- Clear urgency hint with modifier press
- Show bar on sway mode

Feedback on the coding is welcome, not sure with regards to the signal and ipc creation.

I followed the comment in [swaybar](https://github.com/swaywm/sway/blob/master/swaybar/ipc.c#L450) to implement the behavior to hide the bar again after the modifier is pressed.
But my implementation might differ a bit as it will show the bar again if multiple workspaces have the urgency hint, the bar is hidden again, and then one of the urgent workspaces is focused. For me this makes sense as it will give a notification that other workspaces sill have the urgency hint.

### DIFF
--- a/src/modules/sway/bar.cpp
+++ b/src/modules/sway/bar.cpp
@@ -23,7 +23,9 @@ BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
   signal_visible_.connect(sigc::mem_fun(*this, &BarIpcClient::onVisibilityUpdate));
   signal_urgency_.connect(sigc::mem_fun(*this, &BarIpcClient::onUrgencyUpdate));
 
-  ipc_.subscribe(R"(["bar_state_update", "barconfig_update", "workspace"])");
+  // Subscribe to non bar events to determine if the modifier key press is followed by another
+  // action.
+  ipc_.subscribe(R"(["bar_state_update", "barconfig_update", "workspace", "mode", "binding"])");
   ipc_.signal_event.connect(sigc::mem_fun(*this, &BarIpcClient::onIpcEvent));
   ipc_.signal_cmd.connect(sigc::mem_fun(*this, &BarIpcClient::onCmd));
   // Launch worker
@@ -79,7 +81,12 @@ void BarIpcClient::onIpcEvent(const struct Ipc::ipc_response& res) {
               ipc_.sendCmd(IPC_GET_WORKSPACES);
             }
           }
+          modifier_no_action_ = false;
         }
+        break;
+      case IPC_EVENT_MODE:
+      case IPC_EVENT_BINDING:
+        modifier_no_action_ = false;
         break;
       case IPC_EVENT_BAR_STATE_UPDATE:
       case IPC_EVENT_BARCONFIG_UPDATE:
@@ -133,6 +140,16 @@ void BarIpcClient::onConfigUpdate(const swaybar_config& config) {
 void BarIpcClient::onVisibilityUpdate(bool visible_by_modifier) {
   spdlog::debug("visibility update for {}: {}", bar_.bar_id, visible_by_modifier);
   visible_by_modifier_ = visible_by_modifier;
+  if (visible_by_modifier) {
+    modifier_no_action_ = true;
+  }
+  if (!visible_by_modifier_ && modifier_no_action_) {
+    // Modifier key was pressed and released without a different action.
+    // This signals an acknowledgment and should hide the bar again.
+    // Hide the bar and clear the urgency flag.
+    visible_by_urgency_ = false;
+  }
+
   update();
 }
 
--- a/src/modules/sway/bar.cpp
+++ b/src/modules/sway/bar.cpp
@@ -22,6 +22,7 @@ BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
   signal_config_.connect(sigc::mem_fun(*this, &BarIpcClient::onConfigUpdate));
   signal_visible_.connect(sigc::mem_fun(*this, &BarIpcClient::onVisibilityUpdate));
   signal_urgency_.connect(sigc::mem_fun(*this, &BarIpcClient::onUrgencyUpdate));
+  signal_mode_.connect(sigc::mem_fun(*this, &BarIpcClient::onModeUpdate));
 
   // Subscribe to non bar events to determine if the modifier key press is followed by another
   // action.
@@ -85,6 +86,11 @@ void BarIpcClient::onIpcEvent(const struct Ipc::ipc_response& res) {
         }
         break;
       case IPC_EVENT_MODE:
+        if (payload.isMember("change")) {
+          signal_mode_(payload["change"] != "default");
+          modifier_no_action_ = false;
+        }
+        break;
       case IPC_EVENT_BINDING:
         modifier_no_action_ = false;
         break;
@@ -137,6 +143,12 @@ void BarIpcClient::onConfigUpdate(const swaybar_config& config) {
   update();
 }
 
+void BarIpcClient::onModeUpdate(bool visible_by_mode) {
+  spdlog::debug("mode update for {}: {}", bar_.bar_id, visible_by_mode);
+  visible_by_mode_ = visible_by_mode;
+  update();
+}
+
 void BarIpcClient::onVisibilityUpdate(bool visible_by_modifier) {
   spdlog::debug("visibility update for {}: {}", bar_.bar_id, visible_by_modifier);
   visible_by_modifier_ = visible_by_modifier;
@@ -148,6 +160,7 @@ void BarIpcClient::onVisibilityUpdate(bool visible_by_modifier) {
     // This signals an acknowledgment and should hide the bar again.
     // Hide the bar and clear the urgency flag.
     visible_by_urgency_ = false;
+    visible_by_mode_ = false;
   }
 
   update();
@@ -160,7 +173,7 @@ void BarIpcClient::onUrgencyUpdate(bool visible_by_urgency) {
 }
 
 void BarIpcClient::update() {
-  bool visible = visible_by_modifier_ || visible_by_urgency_;
+  bool visible = visible_by_modifier_ || visible_by_mode_ || visible_by_urgency_;
   if (bar_config_.mode == "invisible") {
     visible = false;
   } else if (bar_config_.mode != "hide" || bar_config_.hidden_state != "hide") {
--- a/src/modules/sway/bar.cpp
+++ b/src/modules/sway/bar.cpp
@@ -3,6 +3,7 @@
 #include <fmt/ostream.h>
 #include <spdlog/spdlog.h>
 
+#include <sstream>
 #include <stdexcept>
 
 #include "bar.hpp"
@@ -19,6 +20,23 @@ BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
     handle.disconnect();
   }
 
+  Json::Value subscribe_events{Json::arrayValue};
+  subscribe_events.append("bar_state_update");
+  subscribe_events.append("barconfig_update");
+
+  bool has_mode = isModuleEnabled("sway/mode");
+  bool has_workspaces = isModuleEnabled("sway/workspaces");
+
+  if (has_mode) {
+    subscribe_events.append("mode");
+  }
+  if (has_workspaces) {
+    subscribe_events.append("workspace");
+  }
+  if (has_mode || has_workspaces) {
+    subscribe_events.append("binding");
+  }
+
   signal_config_.connect(sigc::mem_fun(*this, &BarIpcClient::onConfigUpdate));
   signal_visible_.connect(sigc::mem_fun(*this, &BarIpcClient::onVisibilityUpdate));
   signal_urgency_.connect(sigc::mem_fun(*this, &BarIpcClient::onUrgencyUpdate));
@@ -26,7 +44,9 @@ BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
 
   // Subscribe to non bar events to determine if the modifier key press is followed by another
   // action.
-  ipc_.subscribe(R"(["bar_state_update", "barconfig_update", "workspace", "mode", "binding"])");
+  std::ostringstream oss_events;
+  oss_events << subscribe_events;
+  ipc_.subscribe(oss_events.str());
   ipc_.signal_event.connect(sigc::mem_fun(*this, &BarIpcClient::onIpcEvent));
   ipc_.signal_cmd.connect(sigc::mem_fun(*this, &BarIpcClient::onCmd));
   // Launch worker
@@ -37,6 +57,19 @@ BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
       spdlog::error("BarIpcClient::handleEvent {}", e.what());
     }
   });
+}
+
+bool BarIpcClient::isModuleEnabled(std::string name) {
+  for (const auto& section : {"modules-left", "modules-center", "modules-right"}) {
+    if (const auto& modules = bar_.config.get(section, {}); modules.isArray()) {
+      for (const auto& module : modules) {
+        if (module.asString().rfind(name, 0) == 0) {
+          return true;
+        }
+      }
+    }
+  }
+  return false;
 }
 
 struct swaybar_config parseConfig(const Json::Value& payload) {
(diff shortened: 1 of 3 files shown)


src/modules/sway/bar.cpp
```diff
@@ -21,15 +21,39 @@ BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
...
-  ipc_.subscribe(R"(["bar_state_update", "barconfig_update"])");
+  // Subscribe to non bar events to determine if the modifier key press is followed by another
+  // action.
+  ipc_.subscribe(R"(["bar_state_update", "barconfig_update", "workspace", "mode", "binding"])");
   ipc_.signal_event.connect(sigc::mem_fun(*this, &BarIpcClient::onIpcEvent));
+  ipc_.signal_cmd.connect(sigc::mem_fun(*this, &BarIpcClient::onCmd));
+  ipc_mode_.subscribe(R"(["mode"])");
+  ipc_mode_.signal_event.connect(sigc::mem_fun(*this, &BarIpcClient::onEventMode));
+  ipc_workspace_.subscribe(R"(["workspace"])");
+  ipc_workspace_.signal_event.connect(sigc::mem_fun(*this, &BarIpcClient::onEventUrgency));
```
alebastr, line 36: You don't need to create multiple ipc connections to handle different event types; one connection is already enough.
```c++
void BarIpcClient::onIpcEvent(const struct Ipc::ipc_response& res) {
  switch(res.type) {
  case IPC_EVENT_BARCONFIG_UPDATE:
     ...
     break;
  case IPC_EVENT_BAR_STATE_UPDATE:
     ...
     break;
  case IPC_EVENT_WORKSPACE:
     ...
     break;
  case IPC_EVENT_MODE:
     ...
  }
```

src/modules/sway/bar.cpp
```diff
@@ -68,6 +119,9 @@ void BarIpcClient::onIpcEvent(const struct Ipc::ipc_response& res) {
     if (payload.isMember("visible_by_modifier")) {
       // visibility change for hidden bar
       signal_visible_(payload["visible_by_modifier"].asBool());
+    } else if (payload.isMember("change")) {
+      spdlog::trace("BarIpcClient::onIpcEvent: got event {}", payload["change"].asString());
+      signal_last_modifier_(false);
```
alebastr, line 124: What is the event type here?

src/modules/sway/bar.cpp
```diff
@@ -78,21 +132,68 @@ void BarIpcClient::onIpcEvent(const struct Ipc::ipc_response& res) {
...
 void BarIpcClient::update() {
-  bool visible = visible_by_modifier_;
+  bool visible = visible_by_mode_ || visible_by_modifier_ || visible_by_urgency_;
+  if (!visible_by_modifier_ && last_press_modifier_) {
+    // Modifier key was pressed and released without a different action.
+    // This signals an acknowledgment and should hide the bar again.
+    // Hide the bar and clear the mode and urgency flag.
+    visible = false;
+    visible_by_urgency_ = false;
+    visible_by_mode_ = false;
```
alebastr, line 194: That doesn't match with swaybar logic. Why do we need to track `last_press_modifier_`?
Also, you can set `visible_by_urgency_`/`visible_by_mode_` in onVisibilityUpdate.

src/modules/sway/bar.cpp
towoe: Thanks for the tip!

src/modules/sway/bar.cpp
towoe: I used this to store that the last modifier key press was followed by an event which was the "reason" for the key press, and not clear the visible status for the bar.

towoe: My reason to use a signal here was that I assumed that one `onIpcEvent` could be active while another one could be started, but not sure how `ipc_.signal_event.connect(...)` works.

alebastr: > not sure how ipc_.signal_event.connect(...) works

The logic is following:
1. everything connected to the `ipc_` signals directly will be executed in the ipc worker thread, sequentially (in order of receiving the events from the socket). But each ipc connection will have its own worker thread.
2. everything connected to SafeSignal instances will be queued for execution in the main (UI) thread, in the order of event emission.
3. SafeSignal.emit called from the main (UI) thread will bypass the queue and get processed synchronously.

tl;dr: event handlers in different threads may run in parallel, events in the same thread are serialized.

src/modules/sway/bar.cpp
towoe: Thanks for feedback!

> swaybar logic - reset `visible_by_urgency_`/`visible_by_mode_` when `visible_by_modifier_` is set by the IPC event.
> 

So this is this [comment](https://github.com/swaywm/sway/blob/master/swaybar/ipc.c#L450-L452). My interpretation of this was to "ignore" `visible_by_modifier` if another reason for visibility existed, as the bar is not visible by modifier but visible by e.g. urgency already.

In my setup the modifier key that shows the bar is the same as my "normal" modifier key for all the other key bindings, this might be the reason why I implemented the logic to wait for other events after the `visible_my_modifier` to make sure it was not followed by another event, meaning that the intended action of pressing the key was not meant to clear the visibility flag.

> I think I get why you believe it is buggy - new urgent or mode event may come while the modifier is still pressed and set the flag again.

Actually I did not think about this behavior, but I think if the event comes while the key is pressed, it is still shown after releasing the key.

My observation was the following (with the sway prefix key equal the modifier key for the bar):
- if the bar is visible by urgency, pressing and releasing the mod key hides the bar: okay
- if the bar is visible by urgency, pressing the mod key in combination with an workspace change keeps the bar visible after releasing the key: according to the referenced comment not correct, but from my point of view a correct behavior
- if the bar is visible by urgency, pressing the mod key to e.g. change the focus in the same workspace hides the bar: not good
  - changing the workspace will make the bar visible again


> Maybe pinging emersion to clarify swaybar behavior would be a good idea.

If you are now also in doubt of the intended behavior it might be good the get some input, but maybe it is just that I have some edge case setup for which is was not intended and needs a fix.
Not sure if you agree on this, but if sway does not make any rules how the bar should behave, might it not be okay if different bars behave differently?

src/modules/sway/bar.cpp
```diff
@@ -61,38 +66,108 @@ void BarIpcClient::onInitialConfig(const struct Ipc::ipc_response& res) {
...
+        break;
+      case static_cast<uint32_t>(IPC_EVENT_BINDING):
+        last_press_modifier_ = false;
+        break;
+      case static_cast<uint32_t>(IPC_EVENT_BAR_STATE_UPDATE):
+      case static_cast<uint32_t>(IPC_EVENT_BARCONFIG_UPDATE):
+        if (auto id = payload["id"]; id.isString() && id.asString() != bar_.bar_id) {
+          spdlog::trace("swaybar ipc: ignore event for {}", id.asString());
+          return;
+        }
```
Alexays, line 96: We should have this condition for all type, no?

src/modules/sway/bar.cpp
```diff
@@ -61,38 +66,108 @@ void BarIpcClient::onInitialConfig(const struct Ipc::ipc_response& res) {
...
-    }
-    if (payload.isMember("visible_by_modifier")) {
-      // visibility change for hidden bar
-      signal_visible_(payload["visible_by_modifier"].asBool());
-    } else {
-      // configuration update
-      auto config = parseConfig(payload);
-      signal_config_(std::move(config));
+    switch (res.type) {
+      case static_cast<uint32_t>(IPC_EVENT_WORKSPACE):
```
alebastr, line 70: Ouch. I made https://github.com/Alexays/Waybar/pull/1516 to fix the enum type.
Please, drop the casts once it's merged.

src/modules/sway/bar.cpp
alebastr: No, only these 2 events can be sent to a specific instance of a bar and may have bar_id.

src/modules/sway/bar.cpp
```diff
@@ -61,38 +66,108 @@ void BarIpcClient::onInitialConfig(const struct Ipc::ipc_response& res) {
...
+        if (payload.isMember("change")) {
+          // only check and send signal if the workspace update reason was because of a urgent
+          // change
+          if (payload["change"] == "urgent") {
+            // Retrieve status for all workspace in order to check if urgency is still active for at
+            // least one workspace
+            ipc_.sendCmd(IPC_GET_WORKSPACES);
+          }
+          last_press_modifier_ = false;
```
alebastr, line 79: You are changing the `last_press_modifier_` directly from the ipc worker thread. It would be preferable to use `std::atomic<bool>` or signal.

alebastr (commented):
One more thing to consider - this change shouldn't affect the bars that don't show mode indicator or workspaces.

Possible implementation:

```c++
bool isModuleEnabled(const Json::Value& config, std::string name) {
  for (const auto& section : {"modules-left", "modules-center", "modules-right"}) {
    if (const auto& modules = config.get(section, {}); modules.isArray()) {
      for (const auto& module : modules) {
        if (module.asString().rfind(name, 0) == 0) {
          return true;
        }
      }
    }
  }
  return false;
}

BarIpcClient::BarIpcClient(waybar::Bar& bar) : bar_{bar} {
  // ...
  Json::Value events{Json::arrayValue};
  events.append("bar_state_update");
  events.append("barconfig_update");

  bool has_mode = isModuleEnabled(bar.config, "sway/mode");
  bool has_ws = isModuleEnabled(bar.config, "sway/workspaces");
  if (has_mode) {
    events.append("mode");
  }
  if (has_ws) {
    events.append("workspace");
  }
  if (has_mode || has_ws) {
    events.append("binding");
  }

  // Subscribe to non bar events to determine if the modifier key press is followed by another
  // action.
  std::stringstream oss;
  oss << events;
  ipc_.subscribe(oss.str());
  // ...
}
```

src/modules/sway/bar.cpp
alebastr: I just confirmed that the `swaybar` behavior was copied mostly verbatim from `i3bar`. Any modifier press resets the visibility state of the bar, even if it was a part of the binding. So we can say it's intentional.

I'm fine with both options, with a slight preference towards swaybar/i3bar behavior.

> * if the bar is visible by urgency, pressing the mod key to e.g. change the focus in the same workspace hides the bar: not good
>   * changing the workspace will make the bar visible again

Actually, that second part sounds like a bug. i3 does not make the bar visible again on the workspace change, unless the last change in the `workspace` event was `"urgent"` and there are other workspaces with urgency hint set (and I can't figure out from the code why/how i3bar does that).

towoe:
> One more thing to consider - this change shouldn't affect the bars that don't show mode indicator or workspaces.
> 

That's a good point, I did not have that in mind. Thanks for the code, I added this check in a different commit so it is easier for review, can squash that later.

src/modules/sway/bar.cpp
towoe: Sorry, might have misunderstood your explanation before, I thought it will only be possible to write to this at one point.
Added the atomic, not sure if just changing the type was enough, though?

src/modules/sway/bar.cpp
towoe: Should have followed your suggestion literally :)

src/modules/sway/bar.cpp
towoe: I added a separate commit in which both options should be possible.
It is a bit messy, I think the bar config from the waybar config file will not be used if sway ipc is active? I guess it is not possible to see from the ipc message if a change occurs so we just use the values from sway?

One option is to reset the bar visibility right at pressing the mod key.
The other is the "release" part, which I defined as pressing the mod key and releasing it again, which for me makes more sense because it allows to start something without loosing the urgency hint and also to get rid of the bar if one intends to do so.
But I am open to feedback to adding a new config option.

I also changed the way the bar will reset visibility on a workspace change, it should now stay hidden if the urgency is cleared until a new event is received.

src/modules/sway/bar.cpp
```diff
@@ -4,12 +4,30 @@
...
+  for (const auto& section : {"modules-left", "modules-center", "modules-right"}) {
+    if (const auto& modules = config.get(section, {}); modules.isArray()) {
+      for (const auto& module : modules) {
+        active_modules.insert(module.asString());
+      }
+    }
+  }
+
+  for (auto& module : module_check) {
+    if (active_modules.find(module.first) != active_modules.end()) {
```
alebastr, line 25: Note that `sway/workspaces#bar-0` with an optional id after `#` is a valid module reference in the config. Exact string comparison will fail to find that, thus the prefix match with `module.asString().rfind(name,0)` in my PoC code.

To address that, you can try dropping the module ids (`#` and everything after) at line 19.

src/modules/sway/bar.cpp
```diff
@@ -61,38 +105,124 @@ void BarIpcClient::onInitialConfig(const struct Ipc::ipc_response& res) {
...
+  }
+}
+
 void BarIpcClient::onConfigUpdate(const swaybar_config& config) {
   spdlog::info("config update for {}: id {}, mode {}, hidden_state {}", bar_.bar_id, config.id,
                config.mode, config.hidden_state);
   bar_config_ = config;
+
+  // Get this option from the local configuration
+  bar_config_.modifier_reset = this->bar_.config.get("modifier-reset", "press").asString();
```
alebastr, line 184: This option cannot be modified by an IPC message. Swaybar IPC events are limited to what sway itself supports and cannot be expanded.

`bar_.config` is a waybar json config which doesn't support live reload, so you can read the option once in the constructor and move the property to the `BarIpcClient` class.

src/modules/sway/bar.cpp
alebastr: Yep, changing the type is enough. [`std::atomic<T>::operator=`](https://en.cppreference.com/w/cpp/atomic/atomic/operator%3D) is overloaded to take care of atomic assignments when we don't need to swap or check the current value.

> I thought it will only be possible to write to this at one point.

Change to non-atomic variable in one thread may be not (immediately) visible to a read operation in another thread; this behavior is undefined in the C++ memory model. That's why I insisted on some kind of synchronization.

src/modules/sway/bar.cpp
alebastr: Wasn't possible before the aforementioned PR; the compiler wouldn't allow you :)

src/modules/sway/bar.cpp
towoe: Was in the class for my first implementation :) but moved it to `bar_config_` so it would all be in the same place. Anyway, back in the class, looks cleaner as well.

src/modules/sway/bar.cpp
towoe: Your PoC code was cleaner than the one I added now. Any preference on how to do it differently? My thinking was to not do the loop multiple times...

src/modules/sway/bar.cpp
```diff
@@ -61,38 +111,119 @@ void BarIpcClient::onInitialConfig(const struct Ipc::ipc_response& res) {
...
-  spdlog::debug("visiblity update for {}: {}", bar_.bar_id, visible_by_modifier);
+  spdlog::debug("visibility update for {}: {}", bar_.bar_id, visible_by_modifier);
   visible_by_modifier_ = visible_by_modifier;
+  if (visible_by_modifier) {
+    last_press_modifier_ = true;
+  }
+
+  // On visible_by_modifier press, visible_by_modifier is true
+  // On visible_by_modifier release, visible_by_modifier is false
+  // Clear on either press or release depending on bar_.bar_config_.action value
```
alebastr, line 206: This comment looks outdated

alebastr (approved):

src/modules/sway/bar.cpp
alebastr: Well, that's a short loop (just how many modules can be placed on a single bar?) and it's done only once for each visible bar instance. I'd personally pick readability over the optimization here.

src/modules/sway/bar.cpp
towoe: Thanks for spotting that.

src/modules/sway/bar.cpp
towoe: Made the function into a member, looked nicer with the config access, hope that is not a problem?

towoe:
@alebastr thanks so much for all the review already.
I renamed the `last_press_modifier_` to `modifier_no_action_`, hopefully that makes it a bit easier to understand.

Alexays (approved):

Alexays:
Thx a lot!
Can you also update the github wiki?

PR #1516, opened 2022-04-10
refactor: change `enum ipc_command_type` to uint32_t

alebastr:
As we always use the enum to compare or initialize uint32_t values, it would be better to declare it with the right type. This way we could avoid `-Wnarrowing` warnings or unnecessary type casts.


alebastr:
@Alexays, clang-format-lint check is not running for this PR. If that wasn't your intention, you may want to change `on: [push]` to `on: [push, pull_request]`.

PR #1523, opened 2022-04-17
sway/window: Allow the user to change the size of the app icon

### LINKED ISSUE #1461: Possible to change the icon size of app icon in sway/window module?
It looks like it was recently added in the latest update. As much as how nice the feature is the icon seems quite big for my bar and is kinda uneven, is it possible to change its size so that it fits with other icons

FlexW:
Fixes #1461


Alexays (approved):

Alexays:
Thx!

PR #1524, opened 2022-04-17
bar: don't set layer-shell anchor for edges when width is not 'auto'

berbiche:
Don't set the anchor for certain edges when the width or the height is not set to a value of 'auto' (1).

When the bar is vertical, the top and bottom edges are not anchored otherwise the left and right edges are not anchored.

This resolves an issue wherein the width and height set for the layer-shell were ignored because the layer was set to anchor to all edges.

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -174,6 +174,17 @@ struct GLSSurfaceImpl : public BarSurface, public sigc::trackable {
                       GTK_LAYER_SHELL_EDGE_TOP, GTK_LAYER_SHELL_EDGE_BOTTOM}) {
       gtk_layer_set_anchor(window_.gobj(), edge, unanchored != edge);
     }
+
+    // Disable anchoring for other edges too if the width
+    // or the height has been set to a value other than 'auto'
+    // otherwise the bar will use all space
+    if (vertical_ && height_ > 1) {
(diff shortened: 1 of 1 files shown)


src/bar.cpp
```diff
@@ -174,6 +174,17 @@ struct GLSSurfaceImpl : public BarSurface, public sigc::trackable {
...
+    // Disable anchoring for other edges too if the width
+    // or the height has been set to a value other than 'auto'
+    // otherwise the bar will use all space
+    if (vertical_ && height_ > 1) {
+      gtk_layer_set_anchor(window_.gobj(), GTK_LAYER_SHELL_EDGE_BOTTOM, false);
+      gtk_layer_set_anchor(window_.gobj(), GTK_LAYER_SHELL_EDGE_TOP, false);
+    } else if (width_ > 1) {
+      gtk_layer_set_anchor(window_.gobj(), GTK_LAYER_SHELL_EDGE_LEFT, false);
+      gtk_layer_set_anchor(window_.gobj(), GTK_LAYER_SHELL_EDGE_RIGHT, false);
+    }
```
Alexays, line 187: Why not applying this logic inside the for loop?

berbiche: I'm not sure how to merge the if/else logic here with the `for` loop on line 173.
I think this is as concise and clear as it can be at the moment.

Issue #1473, opened 2022-03-10
sway/window: window icon appears are displayed on one machine but not on the other


tsdh:
I run two thinkpads (work and private) both with up-to-date Arch installs and waybar 0.9.10-1 with the very same waybar/sway configs.  On the private machine, the `sway/window` module shows the app's icon, on the other, it doesn't.  Or rather, on the private machine it shows the app icon for emacs, foot, and alacritty but not for firefoxdeveloperedition or gimp.  On the work machine, it never shows any icons including for emacs, foot, and alacritty.

How can I find out why it partially works on the private machine but not my work laptop?  Is there some optional dependency required for app icon support that my private machine happens to have?

Tray icons work just fine on both machines.  On the private machine, I've started using a self-built waybar from git master in order to debug the freeze issue (#1471 and #1464) which also happens for me.  I still get the same results as with 0.9.10-1, i.e., emacs, foot, and alacritty have working app icons while other programs I've tested don't.

RobertMueller2:
sway/window sets app_id_ to what sway's tree is reporting as app_id, falling back to instance if app_id is null. 

Then, it looks in all paths returned by Glib::get_system_data_dirs() (e.g. /usr/share/applications, $HOME/.local/share/applications) for a matching .desktop file. In the desktop file, there should be an Icon defined.

set_image_from_icon_name is a method provided by Gtk::Image, I suppose again it'll look in certain paths, but I have not confirmed that.

So I would first look at swaymsg -t get_tree output to find out app_id or instance, then try and locate a matching desktop file and then go from there. E.g. I noticed that Steam didn't have any icon. Well, Steam's app_id is null and Steam's instance is "Steam", but there was only a /usr/share/steam.desktop. Linking this to Steam.desktop then provided an icon.

tsdh:
Ok, my firefox has `"app_id": "firefoxdeveloperedition"` but the desktop file is `/usr/share/applications/firefox-developer-edition.desktop`, so I guess that's the problem why my firefox has no icon on the machine where icons generally work.

[...]

PR #1525, opened 2022-04-18
sway/window: Add heuristics for finding icons

FlexW:
This adds heuristics for finding the application's icon.

This will not find every icon, but a lot more than currently. I found issues with RenderDoc and JDownloader. But I guess these applications are legacy/X11 applications where icons can only be found with the help of Xlib.

Are you accepting this or would you rather like to adopt an approach like described [here](https://github.com/Alexays/Waybar/issues/1473#issuecomment-1065840022)? I'm unsure if we would have found the icons of the mentioned applications above with this approach.


Alexays:
LGTM using this solution!
Thx!

Issue #1392, opened 2022-01-14
0.9.9 - don't work "network" {bandwidth*}


0xIDENTITY:
**'Network speed' is working on 0.9.8.**
I don't know why. Now (0.9.9) always static 0 bit/s (or octet). Maybe I can try something step-by-step? Let me know, please.
My config:

    "network": {
        "format-wifi": " {bandwidthDowbBits}  {bandwidthUpBits} ",
        "format-ethernet": " {bandwidthDownBits}  {bandwidthUpBits} ",
        "format-linked": "(No IP)",
        "format-disconnected": "",
        "on-click": "alacritty -e nmtui",
        "tooltip": false,
        "interval": 5
    }

And, and, and. When "format-wifi" - nothing is displayed (empty, even without icons).

ibrokemypie:
I believe it was caused by commit 9d9f95976926d5817837d03344b9506d700cee16 which states that setting interface to "*" should result in the same as the old auto implementation. This isn't documented in the wiki though and should probably be the default since the old default no longer works.

bd-g:
I've done testing on that commit and setting the interface to "*" does __not__ have the same behavior as the old auto implementation. There probably wasn't enough testing done on different configurations prior to this commit, and I'm working on a fix. In the meantime, a PR to fix the wiki and reset the default would be helpful.

0x159BABA11227B902:
> I've done testing on that commit and setting the interface to "*" does __not__ have the same behavior as the old auto implementation.

How to set up multiple interfaces?

mvdan:
I don't think it's just the "auto" or wildcard detection; even when I set up the interface to `enp5s0`, my wired connection, I still get `0o/s` even though I can clearly see large numbers in /dev:

```
$ cat /proc/net/dev | grep enp5s0
enp5s0: 3793211018 3216162    0    4    0     0          0      1262 325850572 1492065    0    0    0     0       0          0
```

bd-g:
https://github.com/Alexays/Waybar/issues/610 was the initial issue that the commit was trying to fix
https://github.com/Alexays/Waybar/issues/1327 is a duplicate issue 

@ciarand Thanks for taking up the initial issue and pushing a fix! Before anyone else dives into the code, do you have an idea of why the above behavior would be happening?

droc12345:
I had another type problem (network related) so I looked at the code.
and between 0.98 and 0.99 only 2 changes were made to network.cpp (1 of which was a comment)
[...]

PR #1527, opened 2022-04-18
network: calculate bandwidth for current interface

siikamiika:
There was discussion on #1392 about implementing matching interface cycling to the network module, but as that is a more complex feature, this will patch the issue for the current single interface mode.

Tested working with `network.interface: <undefined>/null/"enp*"/"enp6s0"`


Alexays:
Thx!

PR #1540, opened 2022-05-02
feat: extend bluetooth module

### LINKED ISSUE #688: Feature request: more complete bluetooth module
I'm not using yet the current bluetooth module as it's not released yet but from what I see it only displays whether the adapter is activated or not.

It'd be nice to have some more information available to be displayed:
- list of connected devices
- whether the controller is discoverable
- whether the controller is pairable
-...

My personal use case is "I want to display the bluetooth icon only when some devices are connected to it".
I guess there would be several different "format-connected", "format-disconnected",... config keys like for other modules to allow this.

I'm currently using a custom module and script for this purpose using bluetoothctl. 
But a more robust native module would be welcome.

daangoossens22:
Resolves #688 
Goal is to show selected bluetooth adapter and its connections. Still a work in progress.

TODO:
- [x] adapter information
    - [x] use config field with name to select an adapter or select one randomly if config field not defined (adapter-alias)
    - [x] add format replacements for label and tooltip
    - [x] add new styling classes
- [x] connected devices
    - [x] add format replacement for label for 1 connected device
        - [x] select either the last connected device for this or from a predefined config option to rank which connected device to display (format-device-preference)
    - [x] make possible to show information of all connected devices in the tooltip
    - [x] experimental bluez battery percentage
        - [x] add format replacement for connected device battery percentage (if available)
        - [x] show as experimental feature in waybar-bluetooth man page
- [x] update the module only when signaled from the bluez D-Bus bus or when the rfkill state has changed (if enabled).
- [x] update doc when all features finished
- [x] fix style so it matches the project style

I also noticed that there is no entry for the bluetooth module in the wiki. So that should also maybe be added.

Possible future improvements not part of this merge request:
- show icon field of the connected bluetooth device when available using a config defined icon theme/size
- have format-icons for the connected device battery percentage
- change the selected connected device that appears in the main label changeable via scrolling or clicking?? Don't know how that would interact with the current format-device-preferences or show the latest connection model.
- also enumerate disconnected devices in the tooltip (e.g. by adding a tooltip-format-enumerate-paired option)

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -55,8 +55,8 @@
 #include "modules/sndio.hpp"
 #endif
 #ifdef HAVE_GIO_UNIX
-#include "modules/inhibitor.hpp"
 #include "modules/bluetooth.hpp"
+#include "modules/inhibitor.hpp"
 #endif
 #include "bar.hpp"
 #include "modules/custom.hpp"
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -56,15 +56,11 @@
 #endif
 #ifdef HAVE_GIO_UNIX
 #include "modules/inhibitor.hpp"
+#include "modules/bluetooth.hpp"
 #endif
 #include "bar.hpp"
 #include "modules/custom.hpp"
 #include "modules/temperature.hpp"
-#if defined(__linux__)
-#ifdef WANT_RFKILL
-#include "modules/bluetooth.hpp"
-#endif
-#endif
 
 namespace waybar {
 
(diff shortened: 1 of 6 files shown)


ErikReider:
Oh cool! I was just about to do this! :)

daangoossens22:
I think its ready to be reviewed.

This is my first time using meson, so hopefully I didn't make any beginner mistakes there.

Also for the battery percentage feature, I just put it in a seperate experimental section in the waybar-bluetooth man page, since wrapping the code in an experimental build flag would make the code more complicated in my opinion. The code should work in all cases (whether bleuz experimental features are turned on or off and whether a device provides its battery percentage or not). But if it should be wrapped with an experimental build flag, let me know.

daangoossens22:
Forgot to hide the module when its empty from the linked feature request, so I quickly added it.

Alexays:
LGTM, thx!

PR #1551, opened 2022-05-11
Add classes to individual keyboard state labels

loserMcloser:
Allow labels for numlock/capslock/scrolllock to be styled independently by adding classes to them.

_Use case:_ unknowingly having caps lock on while entering a password is annoying --- I'd like to have the caps lock state indicator styled in red when stated is locked, but don't want num lock state indicator to also be red when state is locked.

### DIFF
--- a/man/waybar-keyboard-state.5.scd
+++ b/man/waybar-keyboard-state.5.scd
@@ -79,4 +79,9 @@ The following *format-icons* can be set.
 - *#keyboard-state*
 - *#keyboard-state label*
 - *#keyboard-state label.locked*
-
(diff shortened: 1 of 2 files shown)


Alexays:
Thx!
Can you also update the man?

loserMcloser:
Sorry! I didn't think to look in the man directory. I've forced pushed an update to the commit. Not sure is you wanted me to be that verbose about the different style class combinations....

loserMcloser:
OK, now I've also updated the commit message. ;)

Alexays:
Thx!

Issue #1552, opened 2022-05-12
Calendar module: LC_TIME


LukashonakV:
@tmpm697 > Can we use different LC_TIME for it (when display {calendar})? As the language to display will depends on it.

LukashonakV:
@tmpm697, see [PR1555](https://github.com/Alexays/Waybar/pull/1555). Please check PR description.

LukashonakV:
> So still no support for local `LC_TIME`? I saw some revert commits.

LC_TIME is supported by the glibc library. So it works well on my machine.
![ps_2022-05-24-19_40_56](https://user-images.githubusercontent.com/23121044/170088296-57cc2e12-e0be-425c-96ed-b2415c7edc60.png)
![ps_2022-05-24-19_40_24](https://user-images.githubusercontent.com/23121044/170088302-c605cf57-14d2-442d-b779-d958561e994f.png)

PR #1555, opened 2022-05-13
Calendar module: localization issue #1552

LukashonakV:
1. Fixed week numbers right alignment when are used different locales(week is started from the SU or MO)
2. Hardcode prefix 'W' is removed in order to provide user an opportunity to use his own prefix via pango markup. Examples:
`"format-calendar-weeks": "<span color='#99ffdd'><b>W{}</b></span>"`
`"format-calendar-weeks": "<span color='#99ffdd'><b>WEEK{}</b></span>"`


tmpm697:
@Alexays can you give a check?

PR #1562, opened 2022-05-19
River Window Module

trevnels:
This PR adds a module to report the currently focused window under river using the `zriver_seat_status_v1` interface.

- The module can be styled depending on if it is located on the currently focused output or not.
- Instances of this module on unfocused outputs will retain their output's last focused window title (the window that _will_ be focused if focus moves to that output).
- There is no support for icons since river's protocol only supplies window titles.

Since this is the first time I've written code for Wayland, I'm not entirely sure if this is the best way to do things, but it seems to work as expected.

swappy-20220519_160158

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -18,6 +18,7 @@
 #endif
 #ifdef HAVE_RIVER
 #include "modules/river/tags.hpp"
+#include "modules/river/window.hpp"
 #endif
 #if defined(__linux__) && !defined(NO_FILESYSTEM)
 #include "modules/battery.hpp"
--- a/include/modules/river/window.hpp
+++ b/include/modules/river/window.hpp
@@ -1,0 +1,33 @@
+#pragma once
+
+#include <gtkmm/button.h>
+#include <wayland-client.h>
+
+#include "ALabel.hpp"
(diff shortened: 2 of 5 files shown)


trevnels:
This also provides similar functionality to #1382 since the window title styling can be used to indicate output focus.

Alexays (approved):

Alexays:
LGTM
Can you also update Github wiki? 
Thx!

trevnels:
> I only want it to represent `Google Chrome`

For that you would need [rewrite rules](https://github.com/Alexays/Waybar/wiki/Module:-Sway#rewrite-rules) to apply a regex to chrome's reported window title. Right now this is only an option for `sway/window` but I can probably add similar functionality to `river/window` in the near future.

niksingh710:
> > I only want it to represent `Google Chrome`
> 
> For that you would need [rewrite rules](https://github.com/Alexays/Waybar/wiki/Module:-Sway#rewrite-rules) to apply a regex to chrome's reported window title. Right now this is only an option for `sway/window` but I can probably add similar functionality to `river/window` in the near future.

Would be nice to see feature like this that will make it more flexible.

PR #1571, opened 2022-05-25
fix: vertical bar not anchored when width is set

### LINKED ISSUE #1568: Waybar 0.9.13 position lleft with width appears in the middle of the screen.
After upgrading 0.9.12 -> 0.9.13 Waybar will appear in the middle of the screen when position left (or right) if the width is set to a non-default value other than 0.

top/bottom works wine with height.

Relevant lines from the config. When setting the width to 0 or commenting the line the bar will appear at the correct position.

```
"position": "left", // Waybar position (top|bottom|left|right)
"width": 60, // Waybar width 
```

Sway, Arch Linux.

daangoossens22:
Fixes #1568 

Don't unanchor from the side where the bar is supposed to anchor.


Alexays:
Thx!

PR #1575, opened 2022-05-28
Add total bandwidth formatting options

cosandr:
This PR adds total bandwidth (`bandwidthTotalBits`, `bandwidthTotalOctets` and `bandwidthTotalBytes`) to the network module.

I have decided against adding a variable at the top of the `update()` method, mostly because it seems calculations are being done on demand if the format is actually requested, but this can be moved if I am interpreting it wrong.

I have also updated the man page, which was missing bytes up/down too.


Alexays:
Thx, can you also update the github wiki? :)

cosandr:
Sure, done.

Issue #1565, opened 2022-05-22
format-calendar-weeks won't accept format strings


hboetes:
I've tried quite a few variations, for example:
```
"format-calendar-weeks": "{:%V}",
```

But only the default works.
```
"format-calendar-weeks": "{}",
```

Using the latest version from git.

hboetes:
Thanks!

With this config it looks perfect now.

```
        "format-calendar-weeks": "<span color='#888888'><b>{}</b></span>",
        "format-calendar-weeks": "{:%V}",
        "tooltip-format": "<big><u>{:%B %Y}</u></big>\n\n<tt>{calendar}</tt>",
        "today-format": "<span color='#ff6699'><b>{}</b></span>",
```
![screenshot_2022-07-14_21-35-56_002215327](https://user-images.githubusercontent.com/2285225/179069667-8afd5c28-8055-428b-b600-2dc8308a9e4b.png)



But with this config, it looks wrong, did I make a mistake?
```
        "format-calendar-weeks": "<span color='#888888'><b>{}</b></span>",
        "format-calendar-weeks": "{}",
        "tooltip-format": "<tt>{calendar}</tt>",
        "today-format": "<span color='#ff6699'><b>{}</b></span>",
```
![image](https://user-images.githubusercontent.com/2285225/179069358-d0fa99ac-463d-4e26-a410-617108abbdcb.png)

PR #1587, opened 2022-06-13
Valid format options

LukashonakV:
* Weeknum is represented via the last day of the week

Implements [ISSUE#1565](https://github.com/Alexays/Waybar/issues/1565)
Week number now is shown is based on the last day of the week (with the correction if SU or MO is the first day of week).

Preferable config is:
`"format-calendar-weeks": "<span color='#99ffdd'><b>W{:%V}</b></span>",`

### DIFF
--- a/Dockerfiles/alpine
+++ b/Dockerfiles/alpine
@@ -2,4 +2,4 @@
 
 FROM alpine:latest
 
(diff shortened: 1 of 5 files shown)


hboetes:
Thanks!

PR #1588, opened 2022-06-13
Retain instance name css class for custom modules

qubidt:
When adding a custom module with a name, e.g.:

```jsonc
{
  ...,
  "custom/foo#bar": { },
  ...
}
```

The custom module does not retain the `bar` class as it should, because all the classes are replaced with the runtime output:

https://github.com/Alexays/Waybar/blob/1b4a7b02f46c1bce2eafdde2a76f4a3adec6620e/src/modules/custom.cpp#L141-L147

Avoid removing the module instance name class so css class behavior is consistent between all modules.

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -30,6 +30,7 @@ class Custom : public ALabel {
 
   const std::string name_;
   std::string text_;
+  std::string id_;
   std::string alt_;
   std::string tooltip_;
   std::vector<std::string> class_;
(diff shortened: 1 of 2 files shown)


Alexays:
LGTM, can you just lint your code :)
Thanks!

qubidt:
I've fixed the lines in my commit. The CI's linting step is still failing on files that are not part of this PR (`src/modules/network.cpp`). Should I fix and commit them anyways?

Alexays:
Thanks, i'll lint `src/modules/network.cpp`

PR #1589, opened 2022-06-13
group module: configurable orientation

qubidt:
currently, the orientation of group modules is always the opposite of the bar. Change it so that:

* the default orientation of the group module is always the opposite of its parent, even for nested groups
* the orientation can be overridden in the config
* css ID and class are set for the group element

Motivation:

I would like to set styling for sets of modules so that they are visually grouped. Some of this can be simulated by enumerating style settings for each module, but many things are not possible when all the modules' elements are siblings of each other 

The module group functionality is a bit is a bit inflexible and only really viable for a specific use case. This hopefully makes it more generally useful

### DIFF
--- a/include/group.hpp
+++ b/include/group.hpp
@@ -12,7 +12,7 @@ namespace waybar {
 
 class Group : public AModule {
  public:
-  Group(const std::string&, const Bar&, const Json::Value&);
+  Group(const std::string&, const std::string&, const Json::Value&, bool);
   ~Group() = default;
   auto update() -> void;
   operator Gtk::Widget&();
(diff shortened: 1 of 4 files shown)


khaneliman:
This is something I am finding myself really needing... I'm trying to get a parent div around modules for visually grouping but this feature breaks the layout because it creates a vertical div instead of a matching horizontal like the other modules-left, right, center.

khaneliman:
Builds fine for me and looks like it works how i'd expect...  screenshot on top is with current waybar, second is built with this PR.

khaneliman (approved):

Alexays:
LGTM

PR #1590, opened 2022-06-14
pulseaudio: avoid retaining outdated form factor

qubidt:
when the module fails to get the pulseaudio device form factor, the module persists the existing value, resulting in the incorrect format-icon being used to format the label on device changes.

reset the form factor value so that the icon lookup properly falls back to "default" when missing


Alexays:
Thanks!

PR #1598, opened 2022-06-23
All docker containers: added fmt, spdlog

LukashonakV:
In order to use more actual components version, needs to update all five docker containers. This blocks ISSUE#1562. Please see successful run using fixed containers from the my own docker repository https://github.com/LukashonakV/Waybar/actions/runs/2549601107

### DIFF
--- a/Dockerfiles/alpine
+++ b/Dockerfiles/alpine
@@ -2,4 +2,4 @@
 
 FROM alpine:latest
 
-RUN apk add --no-cache git meson alpine-sdk libinput-dev wayland-dev wayland-protocols mesa-dev libxkbcommon-dev eudev-dev pixman-dev gtkmm3-dev jsoncpp-dev pugixml-dev libnl3-dev pulseaudio-dev libmpdclient-dev sndio-dev scdoc libxkbcommon tzdata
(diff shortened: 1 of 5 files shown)


alebastr:
The expected way to fix your dependency issue is to update [subprojects/fmt.wrap](https://github.com/Alexays/Waybar/blob/master/subprojects/fmt.wrap) and [subprojects/spdlog.wrap](https://github.com/Alexays/Waybar/blob/master/subprojects/spdlog.wrap). This way we can ensure that waybar compiles even if system dependencies are missing or outdated and detect breakages for this scenario early in CI.

You can get updated wrap files for the versions you need from https://mesonbuild.com/Wrapdb-projects.html.

LukashonakV:
> The expected way to fix your dependency issue is to update [subprojects/fmt.wrap](https://github.com/Alexays/Waybar/blob/master/subprojects/fmt.wrap) and [subprojects/spdlog.wrap](https://github.com/Alexays/Waybar/blob/master/subprojects/spdlog.wrap). This way we can ensure that waybar compiles even if system dependencies are missing or outdated and detect breakages for this scenario early in CI.
> 
> You can get updated wrap files for the versions you need from https://mesonbuild.com/Wrapdb-projects.html.

Thank you @alebastr , I've already tried to check that on official git repository and for fmt utility there were not any updates.. but yes, this issue can be fixed in different ways. My way is to refresh docker containers instead of building component like subproject during the main building. In the same time all popular distributives have got fmt and spdlog in their repos. Why can't we use this native opportunity?  I mean what are advantages and disadvantages of refreshing containers/refreshing wrappers  ?

PR #1617, opened 2022-07-12
fix: adapt to fmt 9.0.0 breaking changes

alebastr:
Addresses the following breaking change:

> Disabled automatic `std::ostream` insertion operator (`operator<<`) discovery when `fmt/ostream.h` is included to prevent ODR violations. You can get the old behavior by defining `FMT_DEPRECATED_OSTREAM` but this will be removed in the next major release. Use `fmt::streamed` or `fmt::ostream_formatter` to enable formatting via `std::ostream` instead.

The patch is good enough to fix builds in distributions updating to fmt 9. WIP because it's already late and I need to give it another look tomorrow; there's one more deprecation warning I want to address:

```
../subprojects/fmt/include/fmt/core.h:1711:43: warning: 'map<mpd_idle, 0>' is deprecated [-Wdeprecated-declarations]           
  const auto& arg = arg_mapper<Context>().map(std::forward<T>(val));                                                           
                                          ^                                                                                                                                                                                                                    
../subprojects/fmt/include/fmt/core.h:1451:3: note: 'map<mpd_idle, 0>' has been explicitly marked deprecated here
  FMT_DEPRECATED FMT_CONSTEXPR FMT_INLINE auto map(const T& val)                                                               
  ^                                                                                                                                                                                                                                                                                                                                                                                                                                       
...
../src/modules/mpd/state.cpp:81:11: note: in instantiation of function template specialization 'spdlog::debug<mpd_idle &>' requested here
  spdlog::debug("mpd: Idle: recv_idle events -> {}", events);
```

### DIFF
--- a/include/modules/keyboard_state.hpp
+++ b/include/modules/keyboard_state.hpp
@@ -1,11 +1,6 @@
 #pragma once
 
-#include <fmt/format.h>
-#if FMT_VERSION < 60000
-#include <fmt/time.h>
-#else
 #include <fmt/chrono.h>
-#endif
 #include <gtkmm/label.h>
 
 #include "AModule.hpp"
--- a/include/modules/simpleclock.hpp
+++ b/include/modules/simpleclock.hpp
@@ -1,11 +1,7 @@
 #pragma once
 
-#include <fmt/format.h>
-#if FMT_VERSION < 60000
-#include <fmt/time.h>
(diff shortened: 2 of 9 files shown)


Alexays (approved):

Alexays:
Thanks @alebastr!

PR #1651, opened 2022-08-10
Feature: sway/window can show 'shell' parameter

TheRealLorenz:
This is a very simple pull request, which grants the sway/window module the capability to display the 'shell' parameter from sway tree.

I personally find it very handy whenever I need to check if a certain window is running natively on Wayland or not.

I was wondering if I should also edit the manpage file for sway/window and the wiki.

Please feel free to correct my mistakes and formatting.

### DIFF
--- a/include/modules/sway/window.hpp
+++ b/include/modules/sway/window.hpp
@@ -21,7 +21,7 @@ class Window : public AIconLabel, public sigc::trackable {
  private:
   void onEvent(const struct Ipc::ipc_response&);
   void onCmd(const struct Ipc::ipc_response&);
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM, and yes if you can update the man <3

TheRealLorenz:
I'm not entirely sure about what should I put as default settings, btw

Alexays:
Yes if you can also update the wiki that would be perfect

TheRealLorenz:
> Yes if you can also update the wiki that would be perfect

Done!

PR #1653, opened 2022-08-11
add JACK module

kennypm:
This module monitors the state of the JACK server: connection state, load percentage, buffer size, sample rate, latency, and xruns. It works with both jack2 and PipeWire, though PW users will need to build PW from git until the next release because this module depends on a fix that was just committed today.

I haven't written documentation yet but I'm happy to if the code is good to merge.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -4,14 +4,14 @@ on: [ push, pull_request ]
 
 jobs:
   clang:
-    # Run actions in a FreeBSD vm on the macos-10.15 runner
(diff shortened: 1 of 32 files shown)


Alexays:
LGTM, just the documentation.
Thx for the feature!

kennypm:
[Wiki page](https://github.com/Alexays/Waybar/wiki/Module:-JACK)

Wasn't sure if you wanted me to add it to the sidebar yet so I didn't.

kennypm:
@Alexays I think this should be all good to go, let me know if you need anything else from me!

Alexays:
LGTM, yes you can add it to the sidebar :)

PR #1659, opened 2022-08-20
Add 'max-volume' option to pulseaudio

### LINKED ISSUE #1607: Allow volume above 100%
The pulseaudio module limits the volume to 100% but in some cases it makes sense to raise the volume above 100%. The most common is when using a monitor or a laptop computer with a lousy speaker. Or that could simply be because you are listening to a bad audio stream. 

The pulseaudio api has no practical limit (except for the 32 bit integer used to store the value) but it probably does not make sense to allow for very high volumes. Pavucontrol allows up to 150% which seems reasonable to me.  

So I propose to add a configuration option `max-volume`.

TheRealLorenz:
Resolves #1607.

Added option to override the maximum volume in the pulseaudio module, defaults at 100.

Edited the manpage.

### DIFF
--- a/src/modules/pulseaudio.cpp
+++ b/src/modules/pulseaudio.cpp
@@ -90,12 +90,16 @@ bool waybar::modules::Pulseaudio::handleScroll(GdkEventScroll *e) {
   double volume_tick = static_cast<double>(PA_VOLUME_NORM) / 100;
   pa_volume_t change = volume_tick;
   pa_cvolume pa_volume = pa_volume_;
+  int max_volume = 100;
   // isDouble returns true for integers as well, just in case
   if (config_["scroll-step"].isDouble()) {
     change = round(config_["scroll-step"].asDouble() * volume_tick);
   }
+  if (config_["max-volume"].isInt()) {
+    max_volume = std::min(0, config_["max-volume"].asInt());
+  }
   if (dir == SCROLL_DIR::UP) {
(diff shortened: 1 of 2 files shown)


src/modules/pulseaudio.cpp
```diff
@@ -90,12 +90,16 @@ bool waybar::modules::Pulseaudio::handleScroll(GdkEventScroll *e) {
   double volume_tick = static_cast<double>(PA_VOLUME_NORM) / 100;
   pa_volume_t change = volume_tick;
   pa_cvolume pa_volume = pa_volume_;
+  int max_volume = 100;
   // isDouble returns true for integers as well, just in case
   if (config_["scroll-step"].isDouble()) {
     change = round(config_["scroll-step"].asDouble() * volume_tick);
   }
+  if (config_["max-volume"].isInt()) {
+    max_volume = std::min(0, config_["max-volume"].asInt());
```
TheRealLorenz, line 99: Idk if the std::min is necessary, but it seamed reasonable to me to have some type of "validation"

TheRealLorenz, line 99: Idk if the std::min is necessary, but it seamed reasonable to me to have some type of "validation".

Alexays:
thx can you also update the github wiki?

TheRealLorenz:
> thx can you also update the github wiki?

I'll do it asap

TheRealLorenz:
Done!

Issue #1584, opened 2022-06-06
Keyboard state not work with external keyboard


vanarok:
I get a cyclic error if the external keyboard has been disconnected and even if plugged back in. I even set the device-path to the external keyboard, but still the same error. The workaround is to restart the waybar.
```[error] keyboard-state: Failed to sync evdev device: ENODEV No such device```

asas1asas200:
This is because the keyboard device path is assigned at the `keyboard-state` module initialization:
https://github.com/Alexays/Waybar/blob/fd24d7bcf63476b8d3063e877752eedd0b661976/src/modules/keyboard_state.cpp#L122-L151
A temporary solution is, set the `device-path` in the config file.
You can find your device in `/dev/input/by-id` or `/dev/input/by-path`, these paths won't change after reconnecting devices.
But it still doesn't work if you don't connect it again.
The better solution is using the internal keyboard in `device-path`.

vanarok:
> This is because the keyboard device path is assigned at the `keyboard-state` module initialization:
> 
> https://github.com/Alexays/Waybar/blob/fd24d7bcf63476b8d3063e877752eedd0b661976/src/modules/keyboard_state.cpp#L122-L151
> 
> 
> A temporary solution is, set the `device-path` in the config file.
> You can find your device in `/dev/input/by-id` or `/dev/input/by-path`, these paths won't change after reconnecting devices.
> But it still doesn't work if you don't connect it again.
> The better solution is using the internal keyboard in `device-path`.

Here is the problem. the thing is that the device-path is set and it does not change on the keyboard, and the waybar cannot read the state  (see error and already mentioned) after I re-plug the external keyboard.

asas1asas200:
@vanarok
My latest commit implemented the hotplug function, hope this works for you.

PR #1661, opened 2022-08-22
Improve keyboard response time and enable hotplug support

asas1asas200:
The `keyboard-state` module used the polling mechanism to get the keyboard status by the interval.
But we more like to see the indicator update immediately, just like the `pulseaudio` module.

So, I use libinput keyboard event to listen to the specific key events update when it has been triggered.
And the `thread_` now is an event loop, so the `interval` value has no effect.

This patch is still in progress, such as the wiki page and the man page aren't finished yet.

I will also add the keyboard removal function in this patch because libinput can get the remove event.
But a prerequisite for this is all the devices that needed be listened should be connected before launching waybar, or you need 
to restart waybar, see related issue #1584 .

I need some opinions about this change.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -21,7 +21,8 @@ jobs:
           pkg install -y git #  subprojects/date
           pkg install -y catch evdev-proto gtk-layer-shell gtkmm30 jsoncpp \
             libdbusmenu libevdev libfmt libmpdclient libudev-devd meson \
-            pkgconf pulseaudio scdoc sndio spdlog wayland-protocols upower
+            pkgconf pulseaudio scdoc sndio spdlog wayland-protocols upower \
+            libinotify
         run: |
           meson build -Dman-pages=enabled
           ninja -C build
--- a/include/modules/keyboard_state.hpp
+++ b/include/modules/keyboard_state.hpp
@@ -3,12 +3,15 @@
 #include <fmt/chrono.h>
 #include <gtkmm/label.h>
 
+#include <unordered_map>
+
 #include "AModule.hpp"
 #include "bar.hpp"
 #include "util/sleeper_thread.hpp"
 
 extern "C" {
 #include <libevdev/libevdev.h>
+#include <libinput.h>
 }
 
 namespace waybar::modules {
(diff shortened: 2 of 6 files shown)


alebastr:
Have you considered using Gio classes instead of the raw OS facilities?
E.g. [`Gio::FileMonitor`](https://developer-old.gnome.org/glibmm/stable/classGio_1_1FileMonitor.html) or Glib event loop?

asas1asas200:
> Have you considered using Gio classes instead of the raw OS facilities? E.g. [`Gio::FileMonitor`](https://developer-old.gnome.org/glibmm/stable/classGio_1_1FileMonitor.html) or Glib event loop?

I have tried the sample code below listening to the `/dev/input`: https://gist.github.com/simu/0ed1e9b12871e20cec48efe73eeeb8e1
The result is as follows:
```
monitoring /dev/input
/dev/input/event3 contents changed
/dev/input/event7 contents changed
/dev/input/event14 contents changed
/dev/input/event13 contents changed
/dev/input/event1 contents changed
/dev/input/event2 contents changed
/dev/input/event4 contents changed
/dev/input/event3 contents changed
/dev/input/event7 set of changes done

/dev/input/event14 set of changes done

/dev/input/event13 set of changes done

/dev/input/event1 set of changes done

/dev/input/event2 set of changes done

/dev/input/event4 set of changes done

/dev/input/event3 set of changes done

/dev/input/event3 contents changed
/dev/input/event7 contents changed
/dev/input/event14 contents changed
/dev/input/event13 contents changed
/dev/input/event1 contents changed
/dev/input/event2 contents changed
/dev/input/event4 contents changed
/dev/input/event3 contents changed
/dev/input/event7 contents changed
/dev/input/event14 contents changed
/dev/input/event13 contents changed
/dev/input/event1 contents changed
/dev/input/event2 contents changed
/dev/input/event4 contents changed
/dev/input/event3 contents changed
/dev/input/event7 contents changed
/dev/input/event14 contents changed
/dev/input/event13 contents changed
```
Even if I don't add or delete any device, it will get the unneeded events because the `G_FILE_MONITOR_WATCH_MOVE` attribute will lead to recording various events.

asas1asas200:
The current version has hotplug support and event-based keyboard lock status detection, problems in the first comment have been resolved.
It can auto-detect new devices and won't crash when keyboards have been removed.

Alexays:
Thx!
Can you also add the deprecated interval info into the github wiki?

Issue #1662, opened 2022-08-24
[wlr/taskbar] only display minimized applications


ErikReider:
I've been using SwayFX (sway fork) that sends Windows to the scratchpad if they're minimized so only displaying minimized would be a cool new feature :)

PR #1667, opened 2022-08-28
[sway/scratchpad] New module for scratchpad of sway

asas1asas200:
Scratchpad indicator for swaywm.

Related issue: #1662 .

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -9,6 +9,7 @@
 #ifdef HAVE_SWAY
 #include "modules/sway/language.hpp"
 #include "modules/sway/mode.hpp"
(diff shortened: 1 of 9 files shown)


ErikReider:
Does this show each application icon just like the wlr/taskbar module or just the count?

asas1asas200:
I only put them in the tooltip, it will show when the cursor hovered.
It shows too much information if displays all apps.
If you want to see all the scratchpad apps on the bar I can write like systray show only app icon on that, what do you think?

ErikReider:
Yeah. An option for displaying the app icons sounds good! :)

Clicking the icons to raise the application would also be a cool addition

Alexays:
LGTM :)

PR #1668, opened 2022-08-29
#1591 river escape window and mode

### LINKED ISSUE #1591: Unescaped ampersand in `river/window`
```
(waybar:203895): Gtk-WARNING **: 09:26:25.466: Failed to set text 'Nerd Fonts - Iconic font aggregator, glyphs/icons collection, & fonts patcher - qutebrowser' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

alex-courtis:
resolves #1591 

The window and mode text from river is escaped. The user's format is not, however the error will be clear.

### DIFF
--- a/src/modules/river/mode.cpp
+++ b/src/modules/river/mode.cpp
@@ -103,7 +103,7 @@ void Mode::handle_mode(const char *mode) {
     }
 
     label_.get_style_context()->add_class(mode);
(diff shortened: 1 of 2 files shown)


alex-courtis:
Thanks for the quick review

PR #1679, opened 2022-09-05
Some logic issues in battery.cpp

tomcharnock:
Some of the logic in the battery module does not cover all possibilities. I've tried to reconcile this by changing the logic flow. Before if some of the `/sys/.../battery/...` files were available it was assumed that others would be although I found this to not be the case. Instead the module now checks for the existence of all files.

I've also added some extra calculations of possible variables which can be calculated from other variables (like how `charge_now` or `energy_now`, etc. can be calculated) which allows `time_remaining` to be calculated when `energy_now` or `power_now` are missing in most cases. Hopefully there isn't anything wrong in my logic, but I've been using it for a few days without any problems.


Alexays:
Thanks!

PR #1687, opened 2022-09-15
Escape text in custom module

sespiros:
When having the following line in a custom module:
```
        "exec": "echo 'test1 & test2'",
```
running waybar results in:
```
(waybar:3775754): Gtk-WARNING **: 01:21:03.702: Failed to set text 'test1 & test2 ' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

### DIFF
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -125,7 +125,7 @@ auto waybar::modules::Custom::update() -> void {
     } else {
       parseOutputRaw();
     }
-    auto str = fmt::format(format_, text_, fmt::arg("alt", alt_),
+    auto str = fmt::format(format_, Glib::Markup::escape_text(text_).raw(), fmt::arg("alt", alt_),
                            fmt::arg("icon", getIcon(percentage_, alt_)),
                            fmt::arg("percentage", percentage_));
     if (str.empty()) {


lilydjwg:
The fix doesn't look right. It disables the ability to pass markup even `escape` is set to `false`.

sespiros:
@lilydjwg Upon further look `text_` seems to be escaped with `escape_text()` in other parts of the code in `custom.cpp`. The case I reported seems to be mitigated by escaping `&` as `&amp;` as the warning suggests but I had another case where that didn't work that I cannot remember. Feel free to revert if it's causing issues and I will revisit it.

lilydjwg:
Do you still need to escape `&` as `&amp;` if you add `"escape": true` (this config is false by default)?

sespiros:
With `"escape": true`, I don't, you are right.

ErikReider:
I don't know if this helps but [this](https://github.com/ErikReider/SwayNotificationCenter/blob/1ef364c3342b0ddf90033a661a57996974a2b4a5/src/notification/notification.vala#L260-L286) worked for me in SwayNC

gadgetmg:
I can confirm that this breaks pango markup on a custom module of mine. I think this should be reverted, especially if `"escape": true` solves the original problem.

Issue #1596, opened 2022-06-22
Workspace 10 shows before 2-9 with sort-by-name


R0dn3yS:
On the "wlr/workspaces" module workspace 10 shows before 2-9 with sort-by-name.

CharlesEkkel:
> Because it reads 1 then 0. So its like Workspace 1, 0 instead of 10. Technically correct. If you want, you can try renaming 10 to 92 or something

I agree, this seems like expected behaviour for sort-by-name. However, personally, I have the same issue yet I still use `sort-by-name` anyway since `sort-by-coordinates` doesn't appear to have any effect (even with `sort-by-name` set to false), and otherwise workspaces are sorted by order of creation (which I assume is sorting by id).

Can anyone confirm if `sort-by-coordinates` actually does anything compared to being turned off?

stomar117:
> Can anyone confirm if `sort-by-coordinates` actually does anything compared to being turned off?

I have tried to manually disable `sort-by-name` and then testing out `sort-by-coordinates`, I can say that it did not yield any difference.

But disabling `sort-by-name` actually fixed the issue with `10` appearing before `1`

This works for me,
```jsonc
    "wlr/workspaces": {
        //"sort-by-coordinates": true, //It did not have an effect
        "sort-by-name": false,
        "on-click": "activate",
        "format": "{name} {icon}",
        "format-icons": {
            "default": "",
            "active": "",
            "urgent": ""
        }
    },
```

JoshuaFern:
Same issue here, 20 workspaces are sorted like:
1, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 2, 20, 3, 4, 5, 6, 7, 8, 9

Makes sense why it sorts this way, but it's not how I want my bar to look.

PR #1691, opened 2022-09-19
Concept for an option to sort by number

Dries1234:
As documented in #1596  , When workspace names are numbered, workspace 10 shows up before 2 - 9.
Adding a new config option seems like the preferred solution here.
I've created a working concept using stoi().
Very much open to feedback as I am not an experienced C++ developer.

### DIFF
--- a/include/modules/wlr/workspace_manager.hpp
+++ b/include/modules/wlr/workspace_manager.hpp
@@ -152,6 +152,7 @@ class WorkspaceManager : public AModule {
 
   bool sort_by_name_ = true;
   bool sort_by_coordinates_ = true;
+  bool sort_by_number_ = true;
   bool all_outputs_ = false;
   bool active_only_ = false;
   bool creation_delayed_ = false;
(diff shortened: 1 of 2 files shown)


bgreggsmith:
What you're looking for is pretty much this https://github.com/Alexays/Waybar/pull/178

I'm also feeling the pain...

Dries1234:
> What you're looking for is pretty much this #178
> 
> I'm also feeling the pain...

This doesn't feel like it solves the same thing?
Maybe I'm missing something

primalmotion:
I tried this patch but it does not seem to do anything for me :/ Also do you mind squashing your commits? The https://github.com/Alexays/Waybar/pull/1691.patch is horrendous :)

primalmotion:
My bad I had a typ in sort by coordinate, making it take precedence. Patch fixes the issue! :)

Alexays:
Closed in favor of #1721

PR #1702, opened 2022-09-30
FreeBSD: Add support to temperature

lbartoletti:
This PR aims to propose a FreeBSD to gain temperature support using sysctl on hw.acpi.thermal.tz0.temperature.

Screenshot (with cpu and memory):
image

@jbeich maybe you have a better idea about this fix?

### DIFF
--- a/src/modules/temperature.cpp
+++ b/src/modules/temperature.cpp
@@ -2,8 +2,19 @@
 
 #include <filesystem>
 
+#if defined(__FreeBSD__)
+// clang-format off
+#include <sys/types.h>
+#include <sys/sysctl.h>
+// clang-format on
+#endif
+
 waybar::modules::Temperature::Temperature(const std::string& id, const Json::Value& config)
     : ALabel(config, "temperature", id, "{temperatureC}°C", 10) {
+
+#if defined(__FreeBSD__)
+// try to read sysctl?
+#else
   if (config_["hwmon-path"].isString()) {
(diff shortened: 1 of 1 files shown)


src/modules/temperature.cpp
```diff
@@ -65,6 +77,17 @@ auto waybar::modules::Temperature::update() -> void {
 }
 
 float waybar::modules::Temperature::getTemperature() {
+#if defined(__FreeBSD__)
+  int temp;
+  size_t size = sizeof temp;
+
+  if (sysctlbyname("hw.acpi.thermal.tz0.temperature", &temp, &size, NULL, 0) != 0) {
+    throw std::runtime_error("sysctl hw.acpi.thermal.tz0.temperature or dev.cpu.0.temperature failed");
```
jbeich, line 85: `dev.cpu.0.temperature` doesn't seem to be used here.

jbeich: Hmm, Linux conditional supports more than one `thermal-zone` but here you're hardcoding the first (i.e., `tz0`). No clue if there's a better way to count `tz%d` than to loop `sysctl` until it fails.

jbeich: Ah, you don't need to count. `thermal-zone` number is provided by the user via Waybar config, see man/waybar-temperature.5.scd

lbartoletti: Thanks,
Done in https://github.com/Alexays/Waybar/pull/1705

PR #1703, opened 2022-09-30
FreeBSD: Add support to battery

lbartoletti:
This commit aims to propose a FreeBSD to gain battery support using sysctl on hw.acpi.battery.*

Discharging
image

Plugged
image

cc @jbeich

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -179,27 +179,27 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
     /* Allocate state of battery units reported via ACPI. */
     int battery_units = 0;
     size_t battery_units_size = sizeof battery_units;
-    if( sysctlbyname("hw.acpi.battery.units", &battery_units, &battery_units_size, NULL, 0) != 0) {
+    if (sysctlbyname("hw.acpi.battery.units", &battery_units, &battery_units_size, NULL, 0) != 0) {
       throw std::runtime_error("sysctl hw.acpi.battery.units failed");
     }
 
-    if(battery_units < 0) {
+    if (battery_units < 0) {
       throw std::runtime_error("No battery units");
     }
 
     int capacity;
     size_t size_capacity = sizeof capacity;
-    if (sysctlbyname("hw.acpi.battery.life", &capacity, &size_capacity, NULL,0) != 0) {
+    if (sysctlbyname("hw.acpi.battery.life", &capacity, &size_capacity, NULL, 0) != 0) {
       throw std::runtime_error("sysctl hw.acpi.battery.life failed");
     }
     int time;
     size_t size_time = sizeof time;
-    if (sysctlbyname("hw.acpi.battery.time", &time, &size_time, NULL,0) != 0) {
+    if (sysctlbyname("hw.acpi.battery.time", &time, &size_time, NULL, 0) != 0) {
       throw std::runtime_error("sysctl hw.acpi.battery.time failed");
     }
     int rate;
     size_t size_rate = sizeof rate;
-    if (sysctlbyname("hw.acpi.battery.rate", &rate, &size_rate, NULL,0) != 0) {
+    if (sysctlbyname("hw.acpi.battery.rate", &rate, &size_rate, NULL, 0) != 0) {
       throw std::runtime_error("sysctl hw.acpi.battery.rate failed");
     }
 
@@ -217,13 +217,13 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
       capacity = 100.f;
     }
     uint8_t cap = round(capacity);
-    if (cap == 100 && status=="Plugged") {
+    if (cap == 100 && status == "Plugged") {
       // If we've reached 100% just mark as full as some batteries can stay
       // stuck reporting they're still charging but not yet done
       status = "Full";
     }
 
-    //spdlog::info("{} {} {} {}", capacity,time,status,rate);
+    // spdlog::info("{} {} {} {}", capacity,time,status,rate);
     return {capacity, time, status, rate};
 
 #elif defined(__linux__)
@@ -538,7 +538,7 @@ const std::string waybar::modules::Battery::getAdapterStatus(uint8_t capacity) c
 #if defined(__FreeBSD__)
   int state;
   size_t size_state = sizeof state;
-  if (sysctlbyname("hw.acpi.battery.state", &state, &size_state, NULL,0) != 0) {
+  if (sysctlbyname("hw.acpi.battery.state", &state, &size_state, NULL, 0) != 0) {
     throw std::runtime_error("sysctl hw.acpi.battery.state failed");
   }
   bool online = state == 2;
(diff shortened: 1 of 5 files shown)


src/modules/battery.cpp
```diff
@@ -1,9 +1,16 @@
...
+#include <sys/types.h>
+#include <sys/sysctl.h>
+// clang-format on
+#endif
 #include <spdlog/spdlog.h>
 
+#include <iostream>
 waybar::modules::Battery::Battery(const std::string& id, const Json::Value& config)
     : ALabel(config, "battery", id, "{capacity}%", 60) {
+#if defined(__Linux__)
```
jbeich, line 13: Should be lowercase:
```suggestion
#if defined(__linux__)
```
Also, it's not complete without the same conditional around
https://github.com/Alexays/Waybar/blob/246e377404b2c189d7e78f4b4038ecee3e602dc4/include/modules/battery.hpp#L9

src/modules/battery.cpp
```diff
@@ -19,11 +26,12 @@ waybar::modules::Battery::Battery(const std::string& id, const Json::Value& conf
   if (global_watch < 0) {
     throw std::runtime_error("Could not watch for battery plug/unplug");
   }
-
+#endif
   worker();
 }
 
 waybar::modules::Battery::~Battery() {
+#if (__Linux__)
```
jbeich, line 34: Witthout `defined` may trigger `-Wundef` on FreeBSD and should be lowercase:
```suggestion
#if defined(__linux__)
```

src/modules/battery.cpp
```diff
@@ -68,9 +83,11 @@ void waybar::modules::Battery::worker() {
     refreshBatteries();
     dp.emit();
   };
+#endif
 }
 
 void waybar::modules::Battery::refreshBatteries() {
+#if (__Linux__)
```
jbeich, line 90: Witthout `defined` may trigger `-Wundef` on FreeBSD and should be lowercase:
```suggestion
#if defined(__linux__)
```

src/modules/battery.cpp
```diff
@@ -156,6 +174,47 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
   std::lock_guard<std::mutex> guard(battery_list_mutex_);
 
   try {
+#if defined(__FreeBSD__)
+    int capacity;
+    size_t size_capacity = sizeof capacity;
+    if (sysctlbyname("hw.acpi.battery.life", &capacity, &size_capacity, NULL,0) != 0) {
```
jbeich, line 180: Hmm, [top(1)](https://man.freebsd.org/top/1) seems to check `hw.acpi.battery.units` before `hw.acpi.battery.life`. https://github.com/freebsd/freebsd-src/commit/fc8ae86a8f25

src/modules/battery.cpp
```diff
@@ -156,6 +174,47 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
...
+    if (cap == 100 && status=="Plugged") {
+      // If we've reached 100% just mark as full as some batteries can stay
+      // stuck reporting they're still charging but not yet done
+      status = "Full";
+    }
+
+    //spdlog::info("{} {} {} {}", capacity,time,status,rate);
+    return {capacity, time, status, rate};
+
+#else
```
jbeich, line 217: Cannot use `#else` here when `inotify_init1()` is hidden behind `#if defined(__Linux__)`.
```suggestion
#elif defined(__linux__)
```

src/modules/battery.cpp
```diff
@@ -275,24 +334,36 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
...
     return {0, 0, "Unknown", 0};
   }
 }
 
 const std::string waybar::modules::Battery::getAdapterStatus(uint8_t capacity) const {
+#if defined(__Linux__)
   if (!adapter_.empty()) {
     bool online;
     std::ifstream(adapter_ / "online") >> online;
+#else
```
jbeich, line 349: Cannot be `#else` here given `#include <sys/sysctl.h>` is hidden behind `#if defined(__FreeBSD__)`.
```suggestion
#elif defined(__linux__)
```

src/modules/battery.cpp
```diff
@@ -275,24 +334,36 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
...
     return {cap, time_remaining, status, total_power / 1e6};
+#endif
   } catch (const std::exception& e) {
     spdlog::error("Battery: {}", e.what());
     return {0, 0, "Unknown", 0};
   }
 }
 
 const std::string waybar::modules::Battery::getAdapterStatus(uint8_t capacity) const {
+#if defined(__Linux__)
```
jbeich, line 345: Should be lowercase:
```suggestion
#if defined(__linux__)
```

src/modules/battery.cpp
```diff
@@ -275,24 +334,36 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
...
+    bool online = state == 2;
+#endif
     if (capacity == 100) {
       return "Full";
     }
     if (online) {
       return "Plugged";
     }
     return "Discharging";
+#if defined(__Linux__)
```
jbeich, line 364: Better inject `{` at the end of the FreeBSD conditional. Also, order conditionals consistently within the file: either Linux first then FreeBSD or vice versa.

src/modules/battery.cpp
```diff
@@ -313,10 +384,12 @@ const std::string waybar::modules::Battery::formatTimeRemaining(float hoursRemai
 }
 
 auto waybar::modules::Battery::update() -> void {
+#if __Linux__
```
jbeich, line 387: Witthout `defined` may trigger `-Wundef` on FreeBSD and should be lowercase:
```suggestion
#if defined(__linux__)
```

src/modules/battery.cpp
```diff
@@ -1,9 +1,16 @@
 #include "modules/battery.hpp"
-
+#if defined(__FreeBSD__)
+// clang-format off
+#include <sys/types.h>
```
jbeich, line 4: Did you try without `<sys/types.h>`? It can be often omitted due to being included by other headers. See also https://github.com/freebsd/freebsd-src/commit/56ee5c551f89

jbeich (commented):
Sorry, I don't have a laptop to test this, so can only comment on style.

src/modules/battery.cpp
lbartoletti: Thanks! Indeed, it works without.
`<sys/types.h> removed.

src/modules/battery.cpp
lbartoletti: Done, thanks.h

src/modules/battery.cpp
lbartoletti: Thanks.
I added a test and throw an error if there is no battery units.

src/modules/battery.cpp
lbartoletti: Done, thanks.

src/modules/battery.cpp
```diff
@@ -1,9 +1,15 @@
 #include "modules/battery.hpp"
-
+#if defined(__FreeBSD__)
+// clang-format off
```
jbeich, line 3: `clang-format` lines can be dropped now as there's only one header, so no need to interfere with sorting by name.

Alexays:
Can you fix build errors, c.f. github action?
Thx :)

Alexays:
Thanks!

PR #1704, opened 2022-09-30
Fixed pulseaudio max-volume configuration

gunslingerfry:
Fixed issue where volume stepping would cause the max volume to go above the max and never reach 0.


Alexays:
Thanks!

PR #1705, opened 2022-10-04
[FreeBSD] Use thermal-zone

lbartoletti:
The zone was hardcoded in #1702.
This commit allows to use the "thermal-zone"
variable.

Follow up #1702


src/modules/temperature.cpp
```diff
@@ -4,14 +4,12 @@
 
 #if defined(__FreeBSD__)
 // clang-format off
```
jbeich, line 6: Can be dropped now as there's only one header, so no need to interfere with sorting by name.

Alexays:
Thx, can you solve the conflict and we'll be good to merge

PR #1710, opened 2022-10-09
mpd: encode "&" signs to "&amp;" for gtk

m-braunschweig:
What does this PR do?:
Replace all ``&`` signs with ``&amp;`` since GTK will otherwise be unable to render it and log a warning.

Why did I write it?:
I noticed that my filename property had a problem showing ``&`` signs. I copied the ``replaceAll`` from ``language.cpp`` and hardcoded it to only replace ``&`` with ``&amp;``. I used it on all the ``getTag`` calls too since I *assume* that it could be a problem there too, but it should be tested first. A better solution would probably be to have that lambda as function (since there were already 2 replace calls in the code)

The GTK error is this btw.:
```
(waybar:12268): Gtk-WARNING **: 19:04:05.824: Failed to set text '3/3: Nightcore - Up & Down (Marnik) [k7tH72AEBZw].mp3: 00:01/02:18' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity * escape ampersand as &amp;
```


ErikReider:
Does this cover `&lt;`, `&gt;`, `&apos;`, etc?

m-braunschweig:
No. I didn't even knew that GTK has some reserved chars until I saw the Error. It's currently just a quickfix :/

m-braunschweig:
Can you send me a link to the documentation that specifies all encodings?

m-braunschweig:
Ok. I found out what has to be replaced. I will update this PR in the next few days

ErikReider:
> Can you send me a link to the documentation that specifies all encodings?

https://wiki.ubuntu.com/NotifyOSD#Treatment_of_the_Desktop_Notifications_Specification_1.1
if you scroll down a bit, you'll se a table with the conversions.

m-braunschweig:
@ErikReider I rewrote the PR to now encode all of these chars: ``<>&"'`` and it is in the ``waybar::util`` namespace. I confirmed that the function is working by using the mpd filename formatter, but I didn't test the ``hyprland`` changes. I hope the way I wrote it is good enough in terms of the coding style (especially that I used a C array instead of a C++ array since I couldn't get the initalization working in a C++ array)

Alexays:
Thx!

PR #1718, opened 2022-10-13
Add "thermal-zone-type" to Temperature Module

bd-g:
## Motivation

Currently, the temperature module only supports specifying an integer to get temperature from `/sys/class/thermal` (e.g. 1 for `/sys/class/thermal/thermal_zone1/temp`). This is problematic because thermal zones reset on every boot. If the user wants a temperature module to just track their CPU temperature, the CPU package may be thermal zone 3 on one boot and thermal zone 7 on the next boot. `hwmon-path` also resets on boot, so this can't be used as a workaround.

There should be a config option available that allows for tracking a specific thermal zone type, like a CPU or hard drive, across boots.

## Implementation

Adds a `thermal-zone-type` option to the temperature module config. On module creation, match this option against each file matching the pattern `/sys/class/thermal/thermal_zone[0-9]+/type`. If a match is found, use that thermal zone. Otherwise, default to `thermal-zone` config option or thermal zone 0.

Note: `hwmon-path` and `hwmon-path-abs` still take highest priority.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,5 @@
 ---
 BasedOnStyle: Google
-AlignConsecutiveDeclarations: true
-BinPackArguments: false
+AlignConsecutiveDeclarations: false
 ColumnLimit: 100
 ...
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -4,14 +4,14 @@ on: [ push, pull_request ]
 
 jobs:
   clang:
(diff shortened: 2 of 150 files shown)


bd-g:
To-do: If PR is merged, update wiki page for temperature module.

bd-g:
I'm a nube, didn't realize that hwmon were symbolic links to devices that seem to persist across boot. So you can have the behavior of this PR with `hwmon-path-abs`, making this redundant. 

Don't want redundant code!

Issue #1716, opened 2022-10-12
[hyprland/window] Option to show active window on a per monitor basis


herlev:
Add an option to allow hyprland/window to show the active window of the monitor the bar belongs to, instead of the currently focused window like it does at the moment.

PR #1719, opened 2022-10-13
[hyprland/window] Add option to show active window on a per monitor basis

herlev:
Implements #1716

Example config:
```json
"hyprland/window": {
    "separate-outputs": true
}
```


Alexays:
Thx, can you solve the conflict?
I'll merge the PR then

Issue #1585, opened 2022-06-11
Idle Inhibitor default state


fuzunspm:
Is it possible to set a default state for idle inhibitor? I mostly turned it on or forgot to do it and screen gets locked during game play.

HarHarLinks:
You may be able to automate idle inhibitor e.g. with sway `for_window [app_id="somethingsomething"] inhibit_idle visible`

bd-g:
@fuzunspm Can this be closed? #1720 addresses this with a `start-activated` boolean config value.

PR #1720, opened 2022-10-14
Add config value for inhibitor default state.

IanManske:
This addresses #1585.
The config value is tentatively named `start-activated`.
Let me know if anything looks wrong / needs fixing.


Alexays:
LGTM, can you also update the github wiki?

PR #1721, opened 2022-10-14
Add option to wlr/workspaces to sort workspaces by number

### LINKED ISSUE #1596: Workspace 10 shows before 2-9 with sort-by-name

herlev:
#1691 didn't work for me. This pull request fixes #1596, by adding the option "sort-by-number" to wlr/workspaces, which takes precedence over any other sort-by option if set to true.

Example config
```json
"wlr/workspaces": {
  "sort-by-number": true,
}
```
The icons can still be changed using
```json
"format-icons": {
  "1": "a",
  "2": "b",
  "3": "c",
  "urgent": "d",
  "active": "e",
  "default": "f"
}
```
and the sorting will still be done according to the actual workspace numbers.

### DIFF
--- a/include/modules/wlr/workspace_manager.hpp
+++ b/include/modules/wlr/workspace_manager.hpp
@@ -152,6 +152,7 @@ class WorkspaceManager : public AModule {
 
   bool sort_by_name_ = true;
   bool sort_by_coordinates_ = true;
+  bool sort_by_number_ = false;
   bool all_outputs_ = false;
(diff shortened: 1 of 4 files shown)


Alexays:
Thanks!
Can you update the man to add the change and also update the Github wiki? :)

herlev:
I have updated both the man page and the wiki now.

PR #1722, opened 2022-10-15
CustomList module

grahnen:
A module adapted from the custom module, that allows the exec process to output a json formatted structure:

`{"data": [ {"name": "a", "text": "label", "class": "css classes", "onclick": "command", "tooltip": "tooltip" }, ... ] }`

that creates a button for each value in data, and formats it according to the input. Let me know if I need to change anything or submit an example script.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -73,6 +73,7 @@
 #endif
 #include "bar.hpp"
 #include "modules/custom.hpp"
+#include "modules/custom_list.hpp"
 #include "modules/temperature.hpp"
 
 namespace waybar {
--- a/include/modules/custom_list.hpp
+++ b/include/modules/custom_list.hpp
@@ -1,0 +1,58 @@
+#pragma once
+
+#include <fmt/format.h>
+
+#include <csignal>
+#include <string>
+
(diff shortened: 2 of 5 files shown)


Alexays:
Why not modify the custom module to add this feature?
I think it could be simpler and avoid this duplication of code :)

grahnen:
That's a good idea. My initial thought was for my changes to be too intrusive. I'd hate to break someones' configuration, which is why I simply created a new module. 

I moved everything into the custom module, and tried my best to keep the behaviour unchanged whenever there is no "data" field in the output json.

grahnen:
The custom module uses ALabel::getIcon. When reparenting to AModule (to allow for having a box of buttons instead of just one label), this functionality needs to be moved. For now, I copied the function to custom.cpp as seems to have been done to for the sway/workspaces module. Should this be done in some other way?

JassonCordones:
Is this still open?

grahnen:
I am not working on it anymore. I think it was ready-ish for merging back when I did? Now it definitely needs to be rebased before that, if this feature is something you want. 

Otherwise feel free to close it

PR #1724, opened 2022-10-16
Fix build with catch2>=3.0.0

llyyr:


Alexays:
Does that still work with older version?

llyyr:
It does not. I guess this fix was too shortsighted, you'd have to use conditional preprocessors I guess

Alexays:
Can you take a look? if you have some time :)

PR #1728, opened 2022-10-18
pulseaudio: disconnect on destruction

### LINKED ISSUE #1115: pulseaudio module leaks connections
This probably won't affect too many people, but an unrelated problem with my setup has exposed this.
In my case, I have a display that keeps disconnecting/reconnecting while in standby. This results in waybar constantly removing and recreating bars on that display, and adding new pulseaudio connections, sadly without removing the old ones. This becomes noticeable when applications suddenly can't output any sound, and logs reveal pulseaudio rejecting new connections because the limit of 64 concurrent connections has been reached.

lilydjwg:
This fixes #1115 when many bars are created and destroyed, there will be too many PulseAudio connections.


Alexays:
thanks!

PR #1729, opened 2022-10-18
Remove border effect on hover

pinselimo:
Moves the ``border = none;`` attribute from workspace buttons to the global scope. The hover effects on all buttons are now consistent in the default stylesheet.

Resolves problem discussed in #1120.


Alexays:
Thanks!

PR #1730, opened 2022-10-18
[hyprland/window] Implement rewrite option like in sway/window

### LINKED ISSUE #1727: Provide rewrite option for hyprland/window
Currently, the hyprland/window lacks the very useful rewrite option like what's in sway/window.

This should probably be relatively easy to add, as it should just be following the same logic.

herlev:
Fixes #1727

### DIFF
--- a/include/modules/hyprland/window.hpp
+++ b/include/modules/hyprland/window.hpp
@@ -19,7 +19,6 @@ class Window : public waybar::ALabel {
  private:
   uint getActiveWorkspaceID(std::string);
   std::string getLastWindowTitle(uint);
-  std::string rewriteTitle(const std::string&);
   void onEvent(const std::string&);
 
   bool separate_outputs;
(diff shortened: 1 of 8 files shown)


herlev:
The code should probably be refactored, moving `std::string Window::rewriteTitle(const std::string& title)` to some common place and let it take the rules as an additional argument. Because right now there is code duplication between sway/window and hyprland/window.

Maybe a file somewhere in `include/util`?

Alexays:
Yep, move this func to a common place :)
Thanks!

herlev:
Refactoring done. I have tested it to be working on hyprland and assume it to work on sway as well since it's only a minor change.

Btw, any way to run the linter check locally before pushing?

Alexays:
Yes you can run clang-format :)

Alexays:
Thx!

PR #1734, opened 2022-10-18
feat(backlight): hide if the display is powered off

### LINKED ISSUE #393: Conditionally execute backlight module
I typically have a laptop connected to an external monitor and disable the laptop's display. When I do this the backlight module shows `0%` as expected but it would be nice to have an option to disable displaying this module if the output is disabled. Maybe something like `exec-if`?

I don't know if this is something that has already been discussed, I couldn't find it easily. Hopefully this may be of interest to others too.

baltitenger:
Probably closes #393.

I know you want a more universal / robust solution for hiding any module, but this does what I need and takes less effort :P

### DIFF
--- a/include/modules/backlight.hpp
+++ b/include/modules/backlight.hpp
@@ -18,12 +18,14 @@ class Backlight : public AButton {
   class BacklightDev {
    public:
     BacklightDev() = default;
-    BacklightDev(std::string name, int actual, int max);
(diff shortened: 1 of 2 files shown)


Alexays:
Thx!

baltitenger:
uhm, sorry, I didn't bother to rebase the latest changes properly as of yet, so now #1773 is basically reverted...

PR #1737, opened 2022-10-19
Set default minimal width of buttons to zero

pinselimo:
Seems like GTK labels have no intrinsic minimal width while buttons do. Probably because they should stay clickable with an empty label.
As a result, even if all margins, padding and borders of buttons are removed the label inside the buttons may still be padded if they are too short. Setting the minimal width of buttons to zero fixes this issue.

This PR sets the ``min-width`` of buttons to zero in the default config, resolving #1731 for future adopters of Waybar.

### DIFF
--- a/resources/style.css
+++ b/resources/style.css
@@ -40,6 +40,8 @@ button {
     /* Avoid rounded borders under each button name */
     border: none;
     border-radius: 0;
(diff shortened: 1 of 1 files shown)


Alexays:
Thx

PR #1738, opened 2022-10-20
Fix button padding

pinselimo:
Buttons come with an intrinsic min-width but lack a method to alter this property. Setting the requested size to zero has also no effect on it.
The only way I found to work is to hard code the CSS into the button, added in this PR.


Alexays:
Nice!
Thx a lot @pinselimo

PR #1762, opened 2022-11-03
build: require jsoncpp >= 1.9.2

jpalus:
with #1719 Waybar started using Json::Value.as() available since jsoncpp 1.9.2.


Alexays:
thx

Issue #1560, opened 2022-05-18
status "plugged" is not detected


name-snrl:
I want to see when my laptop plugged. A similar issue is described in #253.
My device is a Thinkpad t440s with NixOS unstable on board. I use tlp to set charging thresholds

![Untitled](https://user-images.githubusercontent.com/72071763/168993458-1807b13c-d53b-43a4-b8ef-926eccdd0703.png)

I also tried using "format-plugged" but nothing changed.

clownhelmet:
Same issue here. I'm not using TLP but it seems it uses the same interface `/sys/class/power_supply/BAT0/charge_control_end_threshold`. 

`cat /sys/class/power_supply/BAT0/status
Not charging`

`cat /sys/class/power_supply/AC0/online
1`

dtk9289:
bump

Alexays:
Should be fixed with #1679

ericliou:
@Alexays the fix didn't work for me, Arch, Lenovo laptop.
I think the problem might be on line 473 below. 
When my laptop is plugged, the value of `/sys/class/power_supply/BAT0/status` is `Not Charging`. It wouldn't enter the `if` in my case.
https://github.com/Alexays/Waybar/blob/92cc01f40149292d315456f7e5bbdad6dc099caf/src/modules/battery.cpp#L473-L479

Another question that might be related. Why does L477 reads `/status` from the adapter? It seems that my adapter doesn't have the `/status` node, only the battery has it.
```
ls /sys/class/power_supply/ADP0/
device	hwmon0	online	power  subsystem  type	uevent	wakeup26
```

geogiraux:
hi, sorry to make this "revive" but I have the same problem with waybar 0.11.0. Could @clownhelmet or @name-snrl confirm it is solved please ? If it's ok for you I will fill a new issue and provide my  own informations

Issue #253, opened 2019-03-29
Add battery status for plugged in, but not charging


norseghost:
I'd like to distinguish from

- battery plugged in, but not charging (I set charging thresholds with TLP)
- battery charging
- battery discharging

preferably with different format_icons for each state (ex, discharging -> battery levels; charging -> battery levels with power symbol; plugged in/not charging -> power plug)

sharethewisdom:
I'd like `#battery` to also have successive classes based on the states, e.g. `#battery:not(.charging)._30` would allow me to style an 'unloading bar' background animation spanning 30% of the module's total width.

Alexays:
@sharethewisdom What do you mean? The state key is added as a class.

Alexays:
Should be done with 5b3402e110eca2bb3f99c6c81fa78106f7a51d29

sharethewisdom:
sorry, I was confusing states with statuses :confused:

so it was already possible to define `states = { "_90": 90 }` and use it as `#battery:not(.charging)._90 { animate: ... }` and include a css transition to change the background-size for `._80`, `._70`, ... classes.

norseghost:
How would I go about defining formaticons for the charging state?

PR #1769, opened 2022-11-04
Battery: "Plugged" status for TLP settings

ericliou:
When using TLP for threshold, "Not charging" is still being shown, instead of "Plugged".
I don't think it was fixed as mentioned in https://github.com/Alexays/Waybar/issues/1560#issuecomment-1293083222

This should cover:
https://github.com/Alexays/Waybar/issues/253
https://github.com/Alexays/Waybar/issues/1560

Side note: my adapter does not have `/status`, so `current_status` on [L477](https://github.com/Alexays/Waybar/blob/master/src/modules/battery.cpp#L477) will always be empty. This logic is also present at two other places, so I'm not sure if I understand it.I can open a separate issue if it's relevant (Lenovo laptop, Arch).
```cpp
std::getline(std::ifstream(adapter_ / "status"), current_status);
```

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -470,7 +470,9 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
       }
     }
 
-    if (!adapter_.empty() && status == "Discharging") {
+    // Give `Plugged` higher priority over `Not charging`.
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks, and yes go open another issue :)

PR #1773, opened 2022-11-07
Make backlight module respect format-<state> config

kj:
The backlight module should respect `format-<state>` config like other modules.


Alexays:
Thx!

Issue #1681, opened 2022-09-07
exec-one-event executes 'exec' concurrently to event.


ruanformigoni:
Sample snippet:

```json
  "custom/camera": {
    "exec": "notify-send two | echo yoy",
    "interval": "once",
    "on-click": "notify-send one && sleep 2",
    "on-click-right": "notify-send one && sleep 2",
    "on-click-middle": "cheese",
    "exec-on-event": true,
    "tooltip": false
  },
```

According to the docs:
> If an event command is set (e.g. on-click or on-scroll-up) then re-execute the script after executing the event command.

But that does not happen after, it does concurrently, is there a way for `exec` to wait for the event to finish?

jbirnick:
@Alexays Could you please fix this? (/ merge the pull request)

This is an extremely important issue.

Usually, the `on-click` events do some sort of update/change to the underlying software, and then `exec-on-event` is used to make sure that the representation on waybar is updated immediately after the update/change.
However, when you `exec-on-event` doesn't fire _after_ the update/change is completed, then it triggers a data race and potentially (in fact, most likely) fetches old data. The problem becomes even more severe when the underlying software doesn't have a data-race-safe update mechanism.

I even claim that `exec-on-event` is completely useless if fires simultaneously to the event and not after.

This issue affects many packages, please fix.

PR #1784, opened 2022-11-10
exec runs after on-* events

ruanformigoni:
This is a fix for issue #1681. `exec` now runs after `on-click`, `on-click-right`, `on-*` events.

Test sample:

```json
  "custom/camera": {
    "exec": "notify-send exec; echo exec",
    "interval": "once",
    "on-click": "notify-send on-click && sleep 5",
    "exec-on-event": true,
    "tooltip": false,
    "format": "{icon}",
    "format-icons": ["→"]
  },
```

`on-click` runs, and after 5 seconds `exec` runs.

### DIFF
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -25,6 +25,7 @@ class AModule : public IModule {
   SCROLL_DIR getScrollDir(GdkEventScroll *e);
   bool tooltipEnabled();
 
+  std::vector<int> pid_children_;
   const std::string name_;
   const Json::Value &config_;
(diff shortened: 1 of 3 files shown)


andresilva:
Any reason not to merge this? It's quite useful to be able to use `interval = once` for some custom widgets and be assured of execution ordering.

whitelynx:
Should the part of the `exec-on-event` description [in the wiki](https://github.com/Alexays/Waybar/wiki/Module:-Custom) that points to this issue be removed, since this has been merged?

PR #1788, opened 2022-11-12
normalize capacity by number of batteries

grfrederic:
On laptops with more than one battery the calculation of capacity is broken: the sum of capacities of batteries is displayed (and if this turns out to be bigger than a 100%, it gets clamped to 100%).

The `total_capacity` is computed as a sum over individual capacities of batteries, which can be in the range from 0% to _number of batteries_ * 100%. If `total_capacity` was calculated, the `calculated_capacity` is set to the same value. However, all values over 100% are clamped, since, as explained in the code comment:
> `// This can happen when the battery is calibrating and goes above 100%`

I propose the `calculated_capacity` should be normalized by the number of batteries.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -495,7 +495,7 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
     float calculated_capacity{0.0f};
     if (total_capacity_exists) {
       if (total_capacity > 0.0f)
-        calculated_capacity = (float)total_capacity;
+        calculated_capacity = (float)total_capacity / batteries_.size();
       else if (total_energy_full_exists && total_energy_exists) {
         if (total_energy_full > 0.0f)
           calculated_capacity = ((float)total_energy * 100.0f / (float)total_energy_full);


alebastr:
Interesting. This highlights the issue with the battery module which existed way before the change.

My hardware (ThinkPad X1Y6) has only one battery. But there's two batteries reported:
```
% cat /sys/class/power_supply/BAT0/capacity
78
% cat /sys/class/power_supply/wacom_battery_0/capacity
100
% cat /sys/class/power_supply/wacom_battery_0/type
Battery
```
The latter one is a built-in Wacom stylus, which in its dormant state reports capacity 0. If you use it even once since boot, it'll start reporting the actual capacity and affect the total (don't know if it's a firmware or a kernel bug, doesn't matter for this discussion). I don't use it often, so it was on 0 most of the time.

So after the change I see either 39% or 89% instead of the actual 78% and will have to configure acceptable battery names in the waybar config :)
...which seems to be a good idea overall. I expect that some complaints could be fixed by limiting the module to `BAT*` or even a set of common battery name patterns.

grfrederic:
Thank you @alebastr, this might very much be related to @Estebiu s issue from #1792! I like the idea of filtering down to `BAT*` by default.

grfrederic:
@Estebiu, could you post your `ls /sys/class/power_supply/`?

jzs:
For what it's worth, i also have the issue with battery percentage getting clamped. I just tried building master and this PR has indeed fixed it for me :)

I'm on an Lenovo X260 with two batteries

My /sys/class/power_supply looks like this.
```
ls -la /sys/class/power_supply
total 0
drwxr-xr-x  2 root root 0 Nov 17 07:44 .
drwxr-xr-x 72 root root 0 Nov 17 07:44 ..
lrwxrwxrwx  1 root root 0 Nov 17 07:44 AC -> ../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:00/PNP0C09:00/ACPI0003:00/power_supply/AC
lrwxrwxrwx  1 root root 0 Nov 17 07:44 BAT0 -> ../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:00/PNP0C09:00/PNP0C0A:00/power_supply/BAT0
lrwxrwxrwx  1 root root 0 Nov 17 07:44 BAT1 -> ../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:00/PNP0C09:00/PNP0C0A:01/power_supply/BAT1
```

Estebiu:
Hey, sorry for the late reply. Busy week.
As @alebastr it's saying, I think that I'm also encountering his problem. 
I have a laptop with a wacom pen- thing is, it's never been able to report it's battery correctly. Even on GNOME it says that it's battery it's low- 0%.
So waybar it's probably counting the pen's battery togheter with the laptop one.

grfrederic:
@Estebiu nice! That means ignoring batteries different than BAT* and CMB* when autodetecting should work.

Speaking of which, I restored the branch, rebased and added a commit that should do just that. Could someone test it? Sadly I don't have any unusual batteries like you guys, so I'm coding in the dark.

grfrederic:
Hm, sad thing is I think github is no longer tracking changes on that branch, maybe bc the PR is closed. Should I open a new one? Kinda don't wanna lose this discussion. Anyway, here's a direct link to the branch:

https://github.com/grfrederic/Waybar/tree/normalize-battery-capacity

grfrederic:
I just tried a "half test": if I set the string to ignore batteries other than BAT**1**, it works correctly on my machine, that is it only uses BAT**0** when autodetecting. I am now relatively confident that the fix should work for other people. If no one has the time to try it next 1-2 days I'll open a new PR with the fix.

alebastr:
I'm just going to make things worse and tell you that CMBx/BATx are the names specific for Linux ACPI battery driver.
Ex. Pinebook: `/sys/class/power_supply/axp20x-battery`, PBP: `/sys/class/power_supply/cw2015-battery`, M1: `/sys/class/power_supply/macsmc-battery`.
I also suspect there are some ancient laptops where the battery is not routed through ACPI 😞

Maybe a list of globs would work? Something like `"bat": ["BAT*", "CMB*", "*battery"]`, with these values as the defaults.

abcdw:
@grfrederic I built your branch and it shows correct battery capacity now.

grfrederic:
@alebastr i think it's a good idea, it should work out of the box most times and give a lot of leeway to fix things in configuration in trickier cases.

@abcdw nice! i'll check then if i can work in @alebastr's suggestion easily and we're good to go.

grfrederic:
i pushed code that allows for regexes, @alebastr wdyt?

also, how do we set a default? do we just update the one in the reference config and assume that that's the one users will copy over?

grfrederic:
btw i notice that your suggested `*battery` regex would exactly match @Estebiu 's pen...

grfrederic:
i'm thinking maybe the regexes are a bit overkill after all, we could go back to BAT* and CMB* if battery is not given, and besides that allow for a string (single battery) or list of strings (for multiple batteries). i don't think people will often have to configure this, and if they do, they can just write out the batteries they want.

alebastr:
I've just found some time and figured out how UPower distinguishes system batteries from various peripheral devices - #1862.
If that works, all the regex/name checks I suggested are unnecessary. Sorry about that.

@Estebiu can you please check if `/sys/class/power_supply/hid-0018:04F3:2BB3.0001-battery/scope` exists and contains "Device"? If it is, my PR should address the issue.

Estebiu:
Yes, ´/sys/class/power_supply/hid-0018:04F3:2BB3.0001-battery/scope´ exist and contains 'Device'.

grfrederic:
Thanks @alebastr, thats a neat solution ;)

PR #1797, opened 2022-11-17
feat: wireplumber support

smoak:
Adds basic support for showing volume via wireplumber. Allows specifying the node-id or falling back to the default Audio/Sink node id if node-id is not set. If tooltip on hover is enabled, will show `{node_name}` by default otherwise `tooltip-format`.

Format replacements:

`{volume}` - Volume in percentage
`{node_name}` - The node's nickname (`node.nick` property)

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -72,6 +72,9 @@
 #ifdef HAVE_LIBJACK
 #include "modules/jack.hpp"
 #endif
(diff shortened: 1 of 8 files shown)


Alexays:
LGTM, can you add the documentation to the github wiki?

Alexays:
Thx!

smoak:
> LGTM, can you add the documentation to the github wiki?

Done :blush:

Issue #1811, opened 2022-11-24
wireplumber module doesn't work


roland-rollo:
I installed `Waybar v0.9.16` and added `Wireplumber` module to my config file, but if I reload bar it's crash:
```
waybar                                                                                                     22:20
[2022-11-24 22:20:11.296] [info] Using configuration file /home/roland/.config/waybar/config
[2022-11-24 22:20:11.296] [info] Using CSS file /home/roland/.config/waybar/style.css
[2022-11-24 22:20:11.303] [info] config update for bar-0: id bar-0, mode hide, hidden_state hide
[2022-11-24 22:20:11.310] [warning] For a functional tray you must have libappindicator-* installed and export XDG_CURRENT_DESKTOP=Unity
[2022-11-24 22:20:11.310] [warning] As using a timezone, some format args may be missing as the date library haven't got a release since 2018.
[2022-11-24 22:20:11.445] [error] wireplumber: argument not found
[2022-11-24 22:20:11.562] [warning] Requested height: 15 is less than the minimum height: 22 required by the modules
[2022-11-24 22:20:11.562] [info] Bar configured (width: 1920, height: 22) for output: eDP-2
[2022-11-24 22:20:11.562] [error] Item '': No icon name or pixmap given.
C 22:20:11.588532                Gtk (null):(null):(null): gtk_widget_set_accel_path: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
C 22:20:11.588553                Gtk (null):(null):(null): gtk_widget_add_accelerator: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
C 22:20:11.589156                Gtk (null):(null):(null): gtk_widget_set_accel_path: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
C 22:20:11.589164                Gtk (null):(null):(null): gtk_widget_add_accelerator: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
fish: Job 1, 'waybar' terminated by signal SIGSEGV (Address boundary error)
```

I use Sway on Arch Linux and have `ibappindicator-gtk3` and `wireplumber` :wireplumber
Compiled with libwireplumber 0.4.12
Linked with libwireplumber 0.4.12`

Alexays:
@smoak Do you have some time to check?

roland-rollo:
I deleted old config file and copy new from xdg waybar folder:
```
[...]

PR #1836, opened 2022-11-26
feat: add icon support to the wireplumber module

smoak:
Adds basic icon support for the wireplumber module.

This can be achieved by using `{icon}` in the `format` config and pairing it with the `format-icons` config as well.

Example:

```
"wireplumber": {
    "format": "{volume}% {icon}",
    "format-icons": ["", "", ""]
}
```

Adding because #1811 indicated it wasn't supported.


roland-rollo:
May Linus Torvalds bless you.

PR #1845, opened 2022-11-29
Don't search "Keyboard at" from hyprland/language

adamant-pwn:
The current output form of `hyprctl devices` is like this:
```
        Keyboard at 6f80ad70:
                ITE Tech. Inc. ITE Device(8910) Keyboard
                        rules: r "", m "", l "us,ru", v "", o "grp:alt_shift_toggle"
                        active keymap: Russian
                        main: no
```

That is, `Keyboard at` goes _before_ the keyboard name, so looking for `Keyboard at` only makes it skip to the keyboard _after_ the one that the user specified (or crash if it's the last keyboard in the list).

### DIFF
--- a/src/modules/hyprland/language.cpp
+++ b/src/modules/hyprland/language.cpp
@@ -85,7 +85,6 @@ void Language::initLanguage() {
 
   try {
     auto searcher = INPUTDEVICES.substr(INPUTDEVICES.find(KEEBNAME) + KEEBNAME.length());
-    searcher = searcher.substr(searcher.find("Keyboard at"));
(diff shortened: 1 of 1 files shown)


adamant-pwn:
By the way, I don't know if it's just me, but [wiki](https://github.com/Alexays/Waybar/wiki/Module:-Hyprland) suggests to use the name that starts with `AT Traanslated Set...`, but at least on my machine the keyboard of interest has a different name (and `AT Traanslated Set 2 Keyboard` is also there, but it doesn't change layout to Russian, unlike `ITE Tech.`).

Alexays:
Thx!

PR #1854, opened 2022-11-30
fix AIconLabel spacing between image and label

Dordovel:


src/AIconLabel.cpp
```diff
@@ -10,16 +10,15 @@ AIconLabel::AIconLabel(const Json::Value &config, const std::string &name, const
...
+  if (iconEnabled()) {
+    box_.set_spacing(8);
+    box_.add(image_);
+  }
   box_.add(label_);
   event_box_.add(box_);
 }
 
-auto AIconLabel::update() -> void {
-  image_.set_visible(image_.get_visible() && iconEnabled());
```
Alexays, line 20: Why removing this line?

Dordovel: Hidden picture has size, and i added
```
if (iconEnabled()) {
    box_.set_spacing(8);
    box_.add(image_);
  },
```
and I added it so as not to add a picture at all
Screenshot_20221202_0816

PR #1858, opened 2022-12-02
fix: use getaddrinfo() instead of gethostbyname()

uncomfyhalomacro:
Small fix for https://build.opensuse.org/build/X11:Wayland/openSUSE_Tumbleweed/x86_64/waybar/_log. This complies with openSUSE's packaging guidelines.

Reference: https://en.opensuse.org/openSUSE:Packaging_checks#binary-or-shlib-calls-gethostbyname


uncomfyhalomacro:
Not an expert of Cpp nor knowledgeable of it. :smile:

PR #1860, opened 2022-12-02
Hide module if text resolves to being empty. Issue #699

### LINKED ISSUE #699: Hide module if format "resolves" to being empty
Here is the config for my "capture" module
```
"pulseaudio#capture": {
  "format": "{format_source}",
  "format-source": "",
  "format-source-muted": ""
}
```
Its goal is to display the microphone icon only when it is not muted.
It works as intended. However when it is muted, the module is still present in the bar, showing nothing but still taking its space (as tested by setting a different background color for it)

My mpd module has the same issue
```
"mpd": {
  "format": "{stateIcon}",
  "format-stopped": "",
  "format-disconnected": "",
  "state-icons": {
    "paused": "",
    "playing": ""
  }
}
```
The module is still present when the server is stopped.

From what I see other modules I have seems correctly hidden when their format is empty.
My guess here is that the problem comes from the fact that the "format" key is not directly empty but "resolves" to it after some logic is applied. I believe the modules should be completely hidden in such cases as well.

prohornikitin:
Hides mdp/pulseaudio/sndio module if it's text resolves to being empty. 
Should close #699.

### DIFF
--- a/src/modules/mpd/mpd.cpp
+++ b/src/modules/mpd/mpd.cpp
@@ -96,7 +96,13 @@ void waybar::modules::MPD::setLabel() {
     auto format = config_["format-disconnected"].isString()
                       ? config_["format-disconnected"].asString()
                       : "disconnected";
-    label_.set_markup(format);
+    if(format.empty()) {
+      label_.set_markup(format);
+      label_.show();
+    } else {
+      label_.hide();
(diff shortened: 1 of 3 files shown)


Alexays:
Thanks!

PR #1862, opened 2022-12-04
fix(battery): ignore non-system power supplies

alebastr:
Linux power_supply sysfs interface allows checking if the battery powers the whole system or a specific device/tree of devices with `scope` attribute[1]. We can use it to skip the non-system power supplies in the battery module and avoid adding HIDs or other peripheral devices to the total.

The logic is based on UPower, where it is assumed that "Unknown" devices or devices without a `scope` are system power supplies.

[1]: https://lore.kernel.org/lkml/alpine.LNX.2.00.1201031556460.24984@pobox.suse.cz/T/

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -107,6 +107,15 @@ void waybar::modules::Battery::refreshBatteries() {
         std::ifstream(node.path() / "type") >> type;
 
         if (!type.compare("Battery")) {
+          // Ignore non-system power supplies unless explicitly requested
+          if (!bat_defined && fs::exists(node.path() / "scope")) {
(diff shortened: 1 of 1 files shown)


Alexays:
Sorry, I just wanted to approve the PR 🥲

Alexays (approved):

Alexays:
@alebastr tell me when you think it's ready to be merged! :)

alebastr:
With the confirmation from https://github.com/Alexays/Waybar/pull/1788#issuecomment-1338084036 I think it's ok to merge.

Alexays:
Thanks!

PR #1865, opened 2022-12-05
fix AIconLabel spacing between image and label

Dordovel:
Fix https://github.com/Alexays/Waybar/pull/1854#issue-1469577538


Alexays:
Thanks!

PR #1867, opened 2022-12-05
feat(battery): add macsmc-battery time remaining support for asahi

cyrinux:
Hi,

This use time_to_empty_now and time_to_full_now, this is usefull to get a pretty waybar back on asahi linux on macbook.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -233,6 +233,10 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
     bool total_energy_full_design_exists = false;
     uint32_t total_capacity = 0;
     bool total_capacity_exists = false;
+    uint32_t time_to_empty_now = 0;
+    bool time_to_empty_now_exists = false;
+    uint32_t time_to_full_now = 0;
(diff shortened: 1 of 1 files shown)


src/modules/battery.cpp
```diff
@@ -481,8 +495,15 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
     }
 
     float time_remaining{0.0f};
-    if (status == "Discharging" && total_power_exists && total_energy_exists) {
+    if (status == "Discharging" && time_to_empty_now_exists) {
+      if (time_to_empty_now != 0) time_remaining = (float)time_to_empty_now / 1000.0f;
```
jpalus, line 499: What is divisor `1000` based on exactly? According to: https://docs.kernel.org/power/power_supply_class.html both `time_to_empty` as well as `time_to_full` are values in seconds and I think `time_remaining` is expected to be in hours so divisor should be `3600` no?

cyrinux: @jpalus I'm agree something wrong.

upower: time to empty:       3.8 hours
waybar: 12h14

PR #1870, opened 2022-12-06
feat(backlight): added percentexp as format option

nova-nowiz:
Humans don't perceive light linearly, so an exponential curve is more appropriate to gauge the perceived brightness of a screen.
This is fully tested and should work everywhere (Tested with Nix).
The new format would need to be documented on the wiki.
The `icon` markup is based on the normal `percent` markup for now but I can make an `iconexp`.
Also, I don't know the syntax for markup, so if it needs to be `percent_exp` and `icon_exp` (which is more readable), I can do that as well


Narice:
I actually did what I said preemptively:
`percentexp` -> `percent_exp`
added `icon_exp`
Tested and `icon_exp` is actually quite nice to have

btw, I use `brightnessctl` to be able to change my brightness with a certain exponent:
```
brightnessctl -n --exponent=2.718 s +1%
```

Alexays:
Just fixed the conflict, sorry for the delay, could you update the man and the Wiki? 🙏

PR #1871, opened 2022-12-06
dev: Added Nix Flake support

Narice:
- Enables Nix users to get the git version of waybar
- Enables Nix users to develop waybar easily
- Adds a fully reproducible development environment
  - The user only has to install Nix, no other depencencies
  - Automatic dev env on directory entry through .envrc


michaeladler (commented):
Just saw this by chance. Please do not check in `.envrc` files. Either provide `.envrc.sample` or something similar or just omit the file since it's trivial.

Alexays:
Just renamed .envrc to .envrc.sample :)

Narice:
Thanks! Sorry, I didn't see the review before but thanks you for the merge!

mohon8156 (commented):
Drims Analysis works is good.

PR #1868, opened 2022-12-06
fix(calendar week numbers)

prohornikitin:
Fix #1802.

PR #1877, opened 2022-12-07
Calendar week numbers

### LINKED ISSUE #1874: format-calendar-weeks won't accept format strings
Reopened. Is based on [Issue 1565](https://github.com/Alexays/Waybar/issues/1565).
Reason: [Commit](https://github.com/Alexays/Waybar/commit/c5babb4c4455a42d0818d401a7f9ab89dc2c3501) breaks the functionality

prohornikitin:
I'm sorry, but my previous PR #1868 introduced more bugs than fixed. Luckily @LukashonakV told me that the issue itself was wrong and there have to be custom format for a week number in config. 
So, I reverted almost all changes and fixed week number calculation bugs (both mine and previous). 
And now it should work fine for any locale. If there is no custom format, it will apply default format which is %U or %V depending on the first day of the week.
It fix #1874 and other bugs that have not been reported yet.
Sorry again for the previous PR. I hope this is better :)

### DIFF
--- a/include/modules/clock.hpp
+++ b/include/modules/clock.hpp
@@ -28,6 +28,7 @@ class Clock : public ALabel {
   date::year_month_day calendar_cached_ymd_{date::January / 1 / 0};
   date::months calendar_shift_{0}, calendar_shift_init_{0};
   std::string calendar_cached_text_;
+  std::string weeks_format_;
   bool is_calendar_in_tooltip_;
   bool is_timezoned_list_in_tooltip_;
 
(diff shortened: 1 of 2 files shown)


Alexays:
@LukashonakV is it ok?

LukashonakV:
My apologies for late answer. For some reason I didn't get any notifications via email. For me it looks OK. 
The only thing I'm thinking of about calculation of the week number. It's implicit relates to the [ISSUE#762](https://github.com/HowardHinnant/date/issues/762) of the date library. Out current approach is to provide week number is based on the last week day of the week . Actually looking at example I am inclined to think of changing the approach and provide week number is based on the first day of week. For the first calendar row it should be calculated on the first day of the month. What do you think @Alexays , @prohornikitin  ?

LukashonakV:
Let's check https://github.com/Alexays/Waybar/pull/1882 as a summarized changes with defaults formats and some improvements.

PR #1882, opened 2022-12-10
ISSUE#1877 Calendar week numbers

LukashonakV:
1. Let's do code simpler
2. Week format using regexp. Needs when user provide additional characters in format string and need to align week days according
3. Week format has got default formats: ":%U",":%V"
4. Week number is based on the first day of the week now. The output is the same as of date library now.
5. Avoiding of unnecessary operations
----
Waybar_clock

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -240,7 +240,7 @@ auto waybar::modules::Clock::calendar_text(const waybar_time& wtime) -> std::str
 
       os << '\n';
 
-      print_wd = {ym / d};
+      print_wd = (ym / d);
 
       if (weeks_pos == WeeksSide::LEFT) {
         os << fmt::format(weeks_format_, print_wd) << ' ';
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -5,8 +5,8 @@
 
 #include <ctime>
 #include <iomanip>
-#include <sstream>
 #include <regex>
+#include <sstream>
 #include <type_traits>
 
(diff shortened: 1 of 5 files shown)


LukashonakV:
@prohornikitin is it ok for you ?

LukashonakV:
Addition fixed small Linter's notifications

src/modules/clock.cpp
```diff
@@ -74,6 +75,13 @@ waybar::modules::Clock::Clock(const std::string& id, const Json::Value& config)
     locale_ = std::locale("");
   }
 
+  if (config_["format-calendar-weeks"].isString()) {
+    weeks_format_ = std::regex_replace(config_["format-calendar-weeks"].asString(), std::regex("\\{\\}"), (first_day_of_week() == date::Monday) ? "{:%V}" : "{:%U}");
+    weeks_format_left_gaps = std::regex_replace(weeks_format_, std::regex(".*<b>|</b>.*|\\{.?+\\}"), "").length();
```
prohornikitin: I suggest changing regex because format string may not contain `<b>`, or contain other tags inside it, like this: `<b><span color='#99ffdd'>{:%U}</span></b>`.
For example `"</?[^>]+>|\\{.?+\\}"` can be used.

LukashonakV: Thank you @prohornikitin , I'll check tonight,

LukashonakV: @prohornikitin  good point. Thanks. Took into the PR

PR #1887, opened 2022-12-14
format-calendar-weeks won't accept format strings #1874.

LukashonakV:
During the investigation there was found an issue


jlaunay:
> 

I confirm that commit fix the right padding for week numbers.
(Waybar v0.9.13-355-g995802e)
Thks

Alexays:
Thanks guys!

PR #1888, opened 2022-12-15
Fix hyprland language initialization issues

### LINKED ISSUE #1699: hyprland/language does not obey format on bar start
The `hyprland/language` module does not obey the `format` option on the initialization, only after it receives the first event. This can be confirmed by looking at the code.

eneshecan:
Hi, thanks a lot for the great project!

This PR fixes two problems in `hyprland/language` module on initialization:
* Sets the layout name when the `keyboard-name` option is not set by getting the first keyboard device from hyprland config
* Modifies the layout name according to the `format-<lang>` option (Fixes #1699)

Please let me know if something else is needed.


Alexays:
Thanks!

PR #1898, opened 2022-12-21
custom: Change external binary output format to TAB instead of newline

red-9m:
Custom command output is handled with:
`getline(&buff, &len, fp_) == -1)`
it means that output cannot contain parameters with newline delimiter. So, no matter to have parsing with:
`while (getline(output, line)) {`
and new line changed to TAB

### DIFF
--- a/man/waybar-custom.5.scd
+++ b/man/waybar-custom.5.scd
@@ -124,11 +124,11 @@ This should look like this:
 
 The *class* parameter also accepts an array of strings.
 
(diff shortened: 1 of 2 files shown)


Alexays:
Sorry for the delay, in order to not break actual config, this should be a setting

red-9m:
I suppose that nobody is using raw (non json) return type for now, as parseOutputRaw() is broken an cannot parse output with newlines, as getline() consumes all of them. This means that no configs should be broken with this update.

PR #1899, opened 2022-12-21
[WIP] Module for (tabbed) sway windows

juliag2:
This is an extremely WIP module for tabbed sway windows.

I'm a huge fan of the "tabbed" layout, but I dislike the default title bar and wish to change it. However, sway doesn't offer a whole lot of configuration for that (and it's obviously out of scope for sway) so, to match with the style of my other waybar, I wish to add a new module for sway windows. It's mainly thought for the tabbed layout but it's probably viable for other layouts as well.

Right now, it's very rudimentary. There's a bunch of stuff that I'd like to change:

- Merge my new module with the existing "window" module, keeping the default config in a way that would not alter the current functionality, just "optionally expand" it, if you know what I mean.
- Improve the "getIcon" method, as it currently relies on icon fonts and the user specifying every single icon.
- Add "multi-layout"-support, especially support for floating windows.

It's usable right now, though I wouldn't really advise it. It's basically a ripped-off version of the workspaces class.

Please excuse my poor C++ code, this is one of the first things I'm actually writing in C++. If you find any issues, feel more than free to tell me.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -11,6 +11,7 @@
 #include "modules/sway/mode.hpp"
 #include "modules/sway/scratchpad.hpp"
 #include "modules/sway/window.hpp"
+#include "modules/sway/windows.hpp"
 #include "modules/sway/workspaces.hpp"
 #endif
 #ifdef HAVE_WLR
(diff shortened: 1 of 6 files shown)


Anakael:
Can you provide some screenshots how it looks? :)
Also, I think "workspaces" is not good name, because it can be confused with "workspace" module.

PR #1905, opened 2022-12-26
[WirePlumber Module] Improvement, Added source node

Arisa-Snowbell:
I made this for myself but when I am at it, I thought I could share it with the community as I enjoy using waybar.

### DIFF
--- a/include/modules/wireplumber.hpp
+++ b/include/modules/wireplumber.hpp
@@ -20,9 +20,12 @@ class Wireplumber : public ALabel {
   void loadRequiredApiModules();
   void prepare();
   void activatePlugins();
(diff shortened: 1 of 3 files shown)


Arisa-Snowbell:
Explanation why duplicated the methods: I tried to do deduplication of the methods with passing enum value as "Source" and "Sink" to update methods but that failed, it caused memory corruption and I think it's because wireplumber api, no idea what to do with that as I am not C++ dev so I just made function for each and it just works without memory corruption, super weird... to me at least

E-D-W-I-N:
Gentle ping :)

This stuff is good, please consider to merge it

Levantado:
@Arisa-Snowbell can you fix it?

PR #1911, opened 2022-12-28
modules/clock: improve ux when calendar_shift is used

cdump:
I've tried to improve ux in the clock module when calendar_shift is used:

1. Alter only the date, not the time (see the difference in the videos below).
2. Use shifted values only in the tooltip; when scrolling, the date will change only there.
3. When the mouse leaves (the popup closes), the shift is reset.

**current waybar 0.9.16**

https://user-images.githubusercontent.com/449375/209773492-5af428ba-6e22-4ddd-9085-5c181590162c.mp4

**after proposed changes**

https://user-images.githubusercontent.com/449375/209773501-fae54cc5-7ad6-43e1-bbd9-73ebeaea0c27.mp4


Alexays:
Thx!

PR #1918, opened 2023-01-02
sway,feat: allow alphabetical sort

ldelossa:
for users who do not utilize any form of "workspace prev/next" commands, it can be very handle to sort the workspaces alphabetically.

this commit adds a new "alphabetical_sort" to the `sway/workspaces` module which allows the module to alway sort workspaces alphabetically.

this docs are updated to warn the user of the implications involved.

Signed-off-by: Louis DeLosSantos <louis.delos@gmail.com>

I understand that this PR may not be welcome, since the comments in the code look like a good amount of effort was put into having Waybar *not* act like this. Feel free to push back and I'll just maintain this in my fork.


Alexays:
LGTM, as it's just an option.

ldelossa:
Thanks. How often do you release? Only asking because ill run my fork until this change gets to the fedora repo

PR #1932, opened 2023-01-10
wlr-taskbar: allow sorting by app_id

ldelossa:
some users (maybe only myself) may want to sort the task bar by app_id which then places occurrences of the same task next to each other.

there is an alternative method to do this, making TaskBar's `tasks_` vector public, and having each Task find itself in `tbar_->tasks_` and setting their button in the right place, but it "feels" better to have TaskBar decide to place where its Task(s) should be within itself. 

Signed-off-by: Louis DeLosSantos <louis.delos@gmail.com>

### DIFF
--- a/include/modules/wlr/taskbar.hpp
+++ b/include/modules/wlr/taskbar.hpp
@@ -40,6 +40,8 @@ class Task {
     FULLSCREEN = (1 << 3),
     INVALID = (1 << 4)
   };
+  // made public so TaskBar can reorder based on configuration.
+  Gtk::Button button;
 
  private:
   static uint32_t global_id;
@@ -53,7 +55,6 @@ class Task {
 
   uint32_t id_;
 
-  Gtk::Button button_;
   Gtk::Box content_;
   Gtk::Image icon_;
   Gtk::Label text_before_;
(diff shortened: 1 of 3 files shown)


include/modules/wlr/taskbar.hpp
```diff
@@ -40,6 +40,8 @@ class Task {
     FULLSCREEN = (1 << 3),
     INVALID = (1 << 4)
   };
+  // made public so TaskBar can reorder based on configuration.
+  Gtk::Button button_;
```
Alexays, line 44: button_ really need to be public?
If yes please remove the _ :)

ldelossa: Button on this Task class needs to be public or else TaskBar class cannot access it to sort and move the button. 

There is an alternative way which I mention in the PR comments, but I thought this was cleaner. 

Ill remove the underscore as well, missed that.

ldelossa: updated.

ldelossa:
@Alexays fixed linting errors.

ldelossa:
@Alexays do you think this is OK to merge ?

Alexays:
Sorry again for the delay, your branch seems broken, can you rebase it on master?

ldelossa:
@Alexays no worries at all, rebased.

Issue #1446, opened 2022-02-27
Crash with Steam


manujedi:
Simple and short, when i start Steam the waybar crashes. 
I have no experience with anything gui/graphical related so i hope that i can provide enough Infos for the Bug. It also happens with an application named Vivado but Steam is a way more popular so i will post the log to Steam.
I recompiled waybar with debug symbols so i can post a stacktrace.
I normally use the aur version of waybar but it also happens with the default waybar from archlinux. I don't have anything special or custom things configured with waybar.
gdb output:

```
Thread 1 "waybar" received signal SIGSEGV, Segmentation fault.
0x00007ffff79a24fa in Gio::DesktopAppInfo::get_startup_wm_class[abi:cxx11]() const () from /usr/lib/libgiomm-2.4.so.1
(gdb) bt
#0  0x00007ffff79a24fa in Gio::DesktopAppInfo::get_startup_wm_class[abi:cxx11]() const () at /usr/lib/libgiomm-2.4.so.1
#1  0x0000555555655417 in waybar::modules::wlr::get_desktop_app_info(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (app_id="Steam") at ../src/modules/wlr/taskbar.cpp:139
#2  0x0000555555655636 in waybar::modules::wlr::Task::set_app_info_from_app_id_list(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (this=0x555555d20d70, app_id_list="Steam") at ../src/modules/wlr/taskbar.cpp:162
#3  0x00005555556573e5 in waybar::modules::wlr::Task::handle_app_id(char const*) (this=0x555555d20d70, app_id=0x5555559b75cc "Steam") at ../src/modules/wlr/taskbar.cpp:434
#4  0x0000555555655fe3 in waybar::modules::wlr::tl_handle_app_id(void*, zwlr_foreign_toplevel_handle_v1*, char const*) (data=0x555555d20d70, handle=0x5555558c89b0, app_id=0x5555559b75cc "Steam") at ../src/modules/wlr/taskbar.cpp:243
#5  0x00007ffff64dbd4a in  () at /usr/lib/libffi.so.8
#6  0x00007ffff64db267 in  () at /usr/lib/libffi.so.8
#7  0x00007ffff7f76323 in  () at /usr/lib/libwayland-client.so.0
#8  0x00007ffff7f725ab in  () at /usr/lib/libwayland-client.so.0
[...]

PR #1935, opened 2023-01-11
get_desktop_app_info: fix crash on failed DesktopAppInfo::create

multun:
Even though it makes little sense for this call to fail, it sometimes randomly does, and takes down waybar with it.

Fixes the crash from #1446, but does not bring icons back

As the issue occurs randomly, I plan on running this patch for a while to see if the issue goes away. I will mark the pull request as ready when confident the patch indeed works.


multun:
@Alexays this PR does seem to fix the _crash_ but icons are still missing
It also seems like it only occurs with multiple monitors, and it's not deterministic

Alexays:
As it's just a check, yep, Thx!

PR #1937, opened 2023-01-13
fix: battery time remaining

cyrinux:
time is reported in second and should be divided by 3600 and not 1000.

should fix: https://github.com/Alexays/Waybar/pull/1867#pullrequestreview-1246756355


Alexays:
Thx!

PR #1942, opened 2023-01-13
fix(wireplumber): waybar crashes when default node changes

### LINKED ISSUE #1907: [bug] Wireplumber module crashes on audio context change
The wireplumber module causes waybar to crash when the wireplumber context changes.
This is probably due to the fact that `node.id` is not a stable identifier for nodes.

## How to replicate

Sway v1.7
Waybar v0.9.16
libwireplumber v0.4.13

**config.conf:**
```json
{ "modules-center": ["wireplumber"] }
```
**style.css** empty

There are several ways to trigger the issue.

### Disconnect sink
If you have a USB soundcard, a bluetooth headset or anything that you can disconnect:
1. plug it
2. set it as default sink with `wpctl set-default <id>`
3. start waybar
4. unplug it
5. waybar will crash on next refresh of the module, you can force it by changing your volume for example

### Change TTY
For some reasons, wireplumber reassign a new id to every node when I open a new tty.
1. start waybar
(opt) check nodes id with `wpctl status`
2. switch to a new tty (no need to log in)
3. switch back to sway
(opt) check nodes id again

## Output
```
> waybar -c config.conf -s style.css -l trace
[2022-12-27 10:14:43.318] [info] Using configuration file config.conf
[2022-12-27 10:14:43.318] [info] Using CSS file style.css
[2022-12-27 10:14:43.324] [debug] Output detection done: eDP-1 (Sharp Corporation 0x1453 0x00000000)
[2022-12-27 10:14:43.345] [debug] GTK widget tree:
window#waybar.background.eDP-1..mode-default:dir(ltr)
  decoration:dir(ltr)
  box.horizontal:dir(ltr)
    box.horizontal.modules-left:dir(ltr)
    box.horizontal.modules-center:dir(ltr)
      widget:dir(ltr)
        label#wireplumber:dir(ltr)
    box.horizontal.modules-right:dir(ltr)

[2022-12-27 10:14:43.346] [info] Bar configured (width: 1920, height: 19) for output: eDP-1
[2022-12-27 10:14:47.439] [error] Node 31 does not support volume
> echo $?
1
```
Note 1: I triggered the error by switching tty2 at ~10:14:47.4
Note 2: `Node 31` was the default Audio/Sink node when waybar started

## Related issue
- If the sink changes, due to hot-plug of a device for example, the module keep showing the volume of the old default sink.
- Doesn't seem linked to #1852 

## Suggestion
- Re-fetch the default node each time the module is updated
- Add a callback on relevant WP signal (maybe Object Manager `objects-changed`?)
- Use the `node.name` instead of `node.id` (doesn't fix the related issue though)

smoak:
In order to fix the issue, the default node name is cached rather than the default node id. This is due to ids being unstable. So now when the object manager is installed (ie ready), the default node name is retrieved and stored for later.

Now when the mixer changed signal is emitted, the id of the changed node is used to get the node from the object manager. The nodes name is grabbed off that node and compared against the default node name, if they match the volume is updated. Some safeguarding has been added such that if the node cannot be found off the object manager, it's ignored.

Additionally, the "changed" signal on the default nodes api is now utilized to update the default node name if it has changed. This way if the default node changes, the module will be updated with the correct volume and node.nick.

This adds additional debug logging for helping diagnose wireplumber issues.

This also adds the wireplumber man page entry to the main waybar supported section.

Fixes #1907

### DIFF
--- a/include/modules/wireplumber.hpp
+++ b/include/modules/wireplumber.hpp
@@ -20,15 +20,19 @@ class Wireplumber : public ALabel {
   void loadRequiredApiModules();
   void prepare();
   void activatePlugins();
-  static void updateVolume(waybar::modules::Wireplumber* self);
-  static void updateNodeName(waybar::modules::Wireplumber* self);
-  static uint32_t getDefaultNodeId(waybar::modules::Wireplumber* self);
+  static void updateVolume(waybar::modules::Wireplumber* self, uint32_t id);
+  static void updateNodeName(waybar::modules::Wireplumber* self, uint32_t id);
   static void onPluginActivated(WpObject* p, GAsyncResult* res, waybar::modules::Wireplumber* self);
   static void onObjectManagerInstalled(waybar::modules::Wireplumber* self);
+  static void onMixerChanged(waybar::modules::Wireplumber* self, uint32_t id);
+  static void onDefaultNodesApiChanged(waybar::modules::Wireplumber* self);
 
   WpCore* wp_core_;
   GPtrArray* apis_;
   WpObjectManager* om_;
+  WpPlugin* mixer_api_;
+  WpPlugin* def_nodes_api_;
+  gchar* default_node_name_;
   uint32_t pending_plugins_;
   bool muted_;
   double volume_;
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM

PR #1959, opened 2023-01-18
GCC 13 and C++20 support

alebastr:
GCC 13 is coming to Fedora Rawhide and it brings incomplete implementation of C++20 P0355R7 (`<chrono>` Calendars and Time Zones). It's not possible to use that with an unmodified GCC snapshot build right now, but all the code we need is already there.

To compile waybar with C++20, I had to deal with [compile-time format string validation](https://fmt.dev/latest/api.html#compile-time-format-string-checks) in fmtlib, either by wrapping the strings with `fmt::runtime` or by making the parameter `constexpr`.

We could also use `fmt::vformat` in some places and reuse `fmt::format_args`, but that would be prone to all kinds of lifetime bugs - `fmt::arg` does not extend the lifetime of a temporary object passed as an argument.

***
Draft because `fmt::runtime` conversion was pretty mechanical and I want to take another look at the changes. Otherwise it's ready for review.


.github/workflows/linux.yml
```diff
@@ -13,16 +13,23 @@ jobs:
           - fedora
           - opensuse
           - gentoo
+        cpp_std: [c++17]
+        include:
+          - distro: fedora
+            cpp_std: c++20
+          - distro: alpine
+            ignore-test-errors: true
```
Alexays, line 21: Is this because of this https://github.com/Alexays/Waybar/actions/runs/3943499703/jobs/6748390098#step:6:122
?

Alexays (approved):

.github/workflows/linux.yml
alebastr: Yes. LC_TIME support on Alpine [is broken](https://wiki.musl-libc.org/open-issues.html#Locale-limitations), but it used to pass because `std::locale` outright reported that `en_US`/`en_GB` were unavailable and the tests were skipped.

Something has changed in Alpine 3.17 (GCC 12?) and now it's possible to get locale facets, except that the localized date formatting is still unsupported. Thus, we have to ignore or skip the tests on Alpine.

Alexays (approved):

Alexays:
Thanks a lot!

Alexays:
I'll update dockerfiles

PR #1975, opened 2023-01-24
Image module CSS fix

bd-g:
The image module was a great contribution, I love having it in my bar. However, it didn't match the style of the rest of the repo and had a few issues that this PR addresses.

1. The current implementation does not properly attach CSS tags to itself. By running `env GTK_DEBUG=interactive waybar` one can see that the image module does not populate a name or attach any relevant style classes. This PR fixes that and brings it in parallel with styling of other waybar modules. You can now address it as `#image` in `styles.css`, and if you want duplicate modules you can define them as `image#img1`,`image#img2` and address them as `#image-img1`,`#image-img2`.
2. Documentation was not up to speed. The man page has been updated in this PR, and I'm adding the relevant wiki page as we speak.

### DIFF
--- a/include/modules/image.hpp
+++ b/include/modules/image.hpp
@@ -7,6 +7,7 @@
 #include <string>
 
 #include "ALabel.hpp"
+#include "gtkmm/box.h"
 #include "util/command.hpp"
 #include "util/json.hpp"
 #include "util/sleeper_thread.hpp"
@@ -15,7 +16,7 @@ namespace waybar::modules {
 
 class Image : public AModule {
  public:
-  Image(const std::string&, const std::string&, const Json::Value&);
+  Image(const std::string&, const Json::Value&);
   auto update() -> void;
   void refresh(int /*signal*/);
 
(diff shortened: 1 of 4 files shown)


Alexays:
Thanks!

bd-g:
> This works well, many thanks.
> 
> I did however notice that defining my module as "image#im1" in config, and then trying to address it as #image-img1 did not work. Instead I defined it as simply #image in css which worked.
> 
> Not an issue unless trying to use multiple images.

Hmm, that should work, it’s working for me. Could you send over the version of waybar you are running, and your current config file?

MrPenguin07:
> > This works well, many thanks.
> > I did however notice that defining my module as "image#im1" in config, and then trying to address it as #image-img1 did not work. Instead I defined it as simply #image in css which worked.
> > Not an issue unless trying to use multiple images.
> 
> Hmm, that should work, it’s working for me. Could you send over the version of waybar you are running, and your current config file?

Version is current master,
config
`"image#gentoo": {
        "path": "/usr/share/icons/Papirus/48x48/apps/start-here-gentoo.svg",
        "size": 24,
        "on-click": "rofi -show drun -theme arthur -icon-theme Papirus -show-icons"
 },`

style.css
` #image {
          margin-left: 8px;
 }`

If I change style.css to #image-gentoo the css isn't being attached. It works simply with #image.

PR #1978, opened 2023-01-25
Set Image path from executable file (Image Module)

kpanuragh:
I just made a small tweak in the image module to get the image path from an executable file. Here is a sample configuration.

```
 "image/album-art": {
	  "exec":"~/.config/waybar/custom/spotify/album_art.sh",
	  "size": 32,
	  "interval": 5,  
    },
```
If there is a path config exist, then it will load the image from the path config; otherwise, it will check for an exec configuration and execute the script, which should return a file path, and then it will load the image.
Here is a simple executable file
```
#!/bin/bash
album_art=$(playerctl -p spotify metadata mpris:artUrl)
if [[ -z $album_art ]] 
then
   # spotify is dead, we should die to.
   exit
fi
curl -s  "${album_art}" --output "/tmp/cover.jpeg"
echo "/tmp/cover.jpeg"

```

### DIFF
--- a/include/modules/image.hpp
+++ b/include/modules/image.hpp
@@ -27,6 +27,7 @@ class Image : public AModule {
   std::string path_;
   int size_;
   int interval_;
+  util::command::res output_;
 
   util::SleeperThread thread_;
 };
--- a/include/modules/image.hpp
+++ b/include/modules/image.hpp
@@ -27,7 +27,6 @@ class Image : public AModule {
   std::string path_;
   int size_;
   int interval_;
-  util::command::res output_;
 
   util::SleeperThread thread_;
 };
--- a/src/modules/image.cpp
+++ b/src/modules/image.cpp
@@ -9,7 +9,6 @@ waybar::modules::Image::Image(const std::string& name, const std::string& id,
 
   dp.emit();
 
-  //path_ = config["path"].asString();
   size_ = config["size"].asInt();
 
   interval_ = config_["interval"].asInt();
--- a/src/modules/image.cpp
+++ b/src/modules/image.cpp
@@ -44,10 +44,14 @@ auto waybar::modules::Image::update() -> void {
   {
     path_ = config_["path"].asString();
   }
-  else
+  else if(config_['exec'].isString())
   {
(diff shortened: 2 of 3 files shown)


kpanuragh:
@Alexays any update ?

src/modules/image.cpp
```diff
@@ -41,7 +40,15 @@ void waybar::modules::Image::refresh(int sig) {
 
 auto waybar::modules::Image::update() -> void {
   Glib::RefPtr<Gdk::Pixbuf> pixbuf;
-
+  if(config_["path"].isString())
+  {
+    path_ = config_["path"].asString();
+  }
+  else
```
Alexays, line 47: Please add if `config_["exec"].isString()` otherwise throw an error :)

kpanuragh: added else if statement to check exec is string

kpanuragh:
@Alexays  is it okey

include/modules/image.hpp
```diff
@@ -27,6 +27,7 @@ class Image : public AModule {
   std::string path_;
   int size_;
   int interval_;
+  util::command::res output_;
```
Alexays, line 30: Remove that variable

src/modules/image.cpp
```diff
@@ -41,7 +40,19 @@ void waybar::modules::Image::refresh(int sig) {
...
 auto waybar::modules::Image::update() -> void {
   Glib::RefPtr<Gdk::Pixbuf> pixbuf;
-
+  if(config_["path"].isString())
+  {
+    path_ = config_["path"].asString();
+  }
+  else if(config_["exec"].isString())
+  {
+    output_ = util::command::exec(config_["exec"].asString());
```
Alexays, line 49: create a local variable for output instead of class one

kpanuragh:
@Alexays

Alexays:
LGTM, can you update the wiki and the github one?
And i'll merge the PR :)

kpanuragh:
@Alexays  updated man page and  wiki

kpanuragh:
@Alexays  any update

Issue #1321, opened 2021-11-19
calendar day-of-week header is misaligned in Chinese


lilydjwg:
Just see the screenshot:

![calendar-misalign](https://user-images.githubusercontent.com/440661/142644018-6218180b-5f7d-467f-9a88-190e8ceda1de.png)

Those characters are: 一 二 三 四 五 六 日.

LukashonakV:
As far as I found out it's not an module issue. There is a general question how to align Chinese numbers with the Cyrillic numbers. This issue depends on many factors. Please see an explanation: https://stackoverflow.com/questions/22525779/how-to-print-a-chinese-character + https://stackoverflow.com/questions/4655250/difference-between-utf-8-and-utf-16
And there is a good explanation: https://wiki.archlinux.org/title/Fonts - see Chinese section.  So it seems you should setup correct locale and choose appropriate font.

lilydjwg:
Oh, it seems to be a font issue. The tooltip doesn't specify a monospace font so it's not aligned as expected. Do you know how to specify font for this tooltip? I don't want to use monospace for all text on the bar (and even I tried, I failed to change font for the tooltip).

LukashonakV:
@lilydjwg please see the last comment in PR. If it will be confirmed, calendar can become a little bit more customizable.

lilydjwg:
Thanks, I can confirm that this is a font issue.

In my fontconfig for the monospace font I prefer `DejaVu Sans Mono` and then `文泉驿等宽正黑`. The former doesn't contain Chinese characters so for Chinese it will fallback to the latter. Unfortunately they have different widths so it's not aligned perfectly.

When I set it to use the same font it's better aligned:

```
"tooltip-format": "<big>{:%Y %B}</big>\n<span font=\"文泉驿等宽正黑 11\">{calendar}</span>",
```

However it's still not perfectly aligned at some sizes (11 is a best one I can find around that size). there seems to be some rounding errors.

LukashonakV:
Hi @lilydjwg , can you check [PR#1971](https://github.com/Alexays/Waybar/pull/1971). It should solve your issue

lilydjwg:
No, it doesn't.

(Please ignore the blurriness; I configured the wrong widget on the secondary monitor.)

Single month without `<tt>` (`"      {:%Y %B}\n\n{calendar}"`)
[...]

PR #1982, opened 2023-01-27
Clock. Formating simplify

LukashonakV:
Signed-off-by: Viktar Lukashonak <myxabeer@gmail.com>

Hi @Alexays , during [ISSUE#1321](https://github.com/Alexays/Waybar/issues/1321) investigation I've found one small minor bug. In order to avoid unmanaged string formatting in called functions... string formatting moved to the caller functionality + bug with the initial clock configuration (when weeks position are defined and weeks format not and vise versa) is fixed.


Alexays:
Thanks!

PR #1991, opened 2023-01-31
add river/layout

alex-courtis:
Similar functionality to river/window but shows layout.

Updated river output status protocol 2->4

Tested OK:
* hotplug monitor
* river output status version 2 - module does not load
* css classes
* all configuration options


src/modules/river/layout.cpp
```diff
@@ -0,0 +1,174 @@
...
+    spdlog::error("river_status_manager_v1 not advertised");
+    return;
+  }
+
+  if (!seat_) {
+    spdlog::error("wl_seat not advertised");
+  }
+
+  label_.hide();
+  ALabel::update();
```
alex-courtis, line 124: This does not initially hide the label. I did notice the same behaviour from window... any ideas?

alex-courtis:
It looks like there was an infrastructural issue with the builds. Can I re-run the builds?

Should I push a dummy commit to force a re-run?

alex-courtis:
Merged master and updated to obey `cpp_std=c++20`

alex-courtis:
I'll add to the wiki after merge.

src/modules/river/layout.cpp
Alexays: Maybe because the label isn't empty, and so ALabel::update re-show it 🤔

pinpox:
@alex-courtis How do I make this work with `river-luatile`? Where do I have to specify the layout name?

alex-courtis:
> @alex-courtis How do I make this work with `river-luatile`? Where do I have to specify the layout name?

It's up to `river-luatile` to specify the layout name each time it pushes/commits.

The name seems to be specified here:
https://github.com/MaxVerevkin/river-luatile/blob/0ae8e1ca68417878674469def3c861a148fac867/src/main.rs#L55

PR #1996, opened 2023-02-01
Backlight: add brightness control

### LINKED ISSUE #1411: Feature Suggestion: Backlight Scroll
Hello. The backlight module is nice, but there is no scroll by default. This is easily fixable by creating a helper script  like so:
```fish
#!/bin/fish
set backlight_device $argv[1]
set action $argv[2]
set dvalue $argv[3]

echo Backlight device: $backlight_device Bumping: $action By: $dvalue

set current_value (cat $backlight_device)

if test $action = up
    echo (math $current_value + $dvalue) > $backlight_device
else if test $action = down
    echo (math $current_value - $dvalue) > $backlight_device
else
    exit 2
end
```
And prepending the module config with somthing like
```json
        "interval" : 0.2,
        "device": "intel_backlight",
        "on-scroll-down": "backlight_delta /sys/class/backlight/intel_backlight/brightness down 50",
        "on-scroll-up": "backlight_delta /sys/class/backlight/intel_backlight/brightness up 50",
```
You are unable to see it from the screencast, but the level is actually changing: 
https://mega.nz/file/ubgk1baZ#01vxMsjlZ4KRBI_weng8qg7SrIG7xHbCeNupLXd5xv0
It is probably even possible to minify/hardcode some stuff and make an easily insertable oneliner

Kuruyia:
This adds the possibility of setting the brightness by scrolling up or down on the backlight module.
Code was adapted from the PulseAudio module, so the `reverse-scrolling` and `scroll-step` options were ported over.

Closes #1411.

### DIFF
--- a/include/modules/backlight.hpp
+++ b/include/modules/backlight.hpp
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "ALabel.hpp"
+#include "giomm/dbusproxy.h"
 #include "util/json.hpp"
 #include "util/sleeper_thread.hpp"
 
@@ -62,5 +63,7 @@ class Backlight : public ALabel {
   std::vector<BacklightDev> devices_;
   // thread must destruct before shared data
   util::SleeperThread udev_thread_;
+
+  Glib::RefPtr<Gio::DBus::Proxy> login_proxy_;
 };
 }  // namespace waybar::modules
--- a/include/modules/backlight.hpp
+++ b/include/modules/backlight.hpp
@@ -50,6 +50,8 @@ class Backlight : public ALabel {
   template <class ForwardIt, class Inserter>
   static void enumerate_devices(ForwardIt first, ForwardIt last, Inserter inserter, udev *udev);
 
+  bool handleScroll(GdkEventScroll* e);
+
   const std::string preferred_device_;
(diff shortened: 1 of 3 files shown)


t-8ch:
Instead of directly writing the sysfs value which requires privileges this could use systemds  `org.freedesktop.login1.Session:SetBrightness()` dbus API which does not require privileges.

Kuruyia:
I was wondering about using D-Bus, and if it would affect compatibility with systems that do not have (e)logind installed (most notably FreeBSD).
If it's not a problem to do it that way then yes, I'd be happy to switch to D-Bus instead (or try D-Bus and fallback to udev if not available).

(On Arch Linux at least,) the user only needs to be part of the `video` group for them to have write permission on the brightness setting.

t-8ch:
> (On Arch Linux at least,) the user only needs to be part of the video group for them to have write permission on the brightness setting.

This doesn't seem to be right. sysfs does not have support for changing permissions or ACLs.

Kuruyia:
Nice find, in my case the group/permission has been changed by the "brillo" package.
In that case, I'll update this PR so it uses D-Bus instead as it seems to be the better solution.

Kuruyia:
@Alexays Any update?

Alexays:
LGTM, thx!

PR #2016, opened 2023-02-19
Module for controlling gamma

manueldiagostino:
This simple module controls the gamma of the display as mentioned [here](https://github.com/Alexays/Waybar/discussions/2008). I have listed the available options in the associated wiki.

I added also the `wlr_hyprland` option to make clicking on workspaces work in Hyprland.


alebastr:
Wouldn't this be better as a standalone application?
As far as I can see there's almost no interaction with the bar, so custom module with "on-click": "new-gamma-control-settings-app" is a much cleaner option.

Arnaudv6:
do you know about https://github.com/jeremija/wl-gammarelay?

manueldiagostino:
> Wouldn't this be better as a standalone application? As far as I can see there's almost no interaction with the bar, so custom module with "on-click": "new-gamma-control-settings-app" is a much cleaner option.

@alebastr yes, I believe you are right; I was thinking that having buttons to click directly on the bar would have been nicer to see.


> do you know about https://github.com/jeremija/wl-gammarelay?

@Arnaudv6 thanks for the suggestion, I didn't know it.

PR #2017, opened 2023-02-19
feat(sway): add workspace css class

pascal-huber:
From what I can see, it is currently not possible to individually and consistently style workspace buttons using css (unless all workspaces are visible at all times such that `:nth-child` can be used). This PR adds a css class to every workspace button using the workspace name. This the allows to use selectors such as `#workspaces button.workspace-2.focused`.

### Notes
 - The css classes are not sanitized, special symbols have to be escaped by the user. For example, if my workspace has the name `"@x$ f"`, the selector is `.workspace-\@x\$\ f`.
 - The wiki would need to be updated.


Alexays:
LGTM, you can update the github wiki! :)

Issue #1977, opened 2023-01-25
[improvement] Add support for middle click on calendar/clock module


jlaunay:
Following [this pull request](https://github.com/Alexays/Waybar/pull/1956) from @LukashonakV it would be nice to  add `"on-click-middle": "mode"` support on `calendar`.
That would be useful if left and right clicks are already mapped to another functions for clock (show alternative label, launch a calendar app etc).

PR #2019, opened 2023-02-20
ISSUE#1977. AModule implements module actions call

LukashonakV:
Hi @Alexays 
This PR is based on the https://github.com/Alexays/Waybar/issues/1977
The main idea - to stop of the code duplication in every modules when the module has to extend it's functionality via internal module member-functions.
So
1. IModule interface is extended by the doAction method
2. AModule has got it's own base doAction:  1. defines internal mapping between event name and module action name. 2. Calls override doAction.
3. Each module can provide it's own doAction which actually calls internal module functionality. Module action name is provided through function parameter
4. If Module overrides doAction and implements all necessary functionality via member-functions and has got internal mapping ... it could drop override handleToggle and handleScroll methods.
---

Es an example... module Clock now provides all necessary functionality and implement doAction which provide to user an ability to redefine events for intenral actions.(If PR is OK, all actions will be described in WIKI)
```json
 "clock": {
        "format": "{:%H:%M}  ",
        "format-alt": "{:%A, %B %d, %Y (%R)}  ",
        "tooltip-format": "\n<small><tt>{calendar}</tt></small>",
        "calendar": {
                    "mode"          : "year",
                    "mode-mon-col"  : 3,
                    "weeks-pos"     : "right",
                    "on-scroll"     : 1,
                    "format": {
                              "months":     "<span color='#ffead3'><b>{}</b></span>",
                              "days":       "<span color='#ecc6d9'><b>{}</b></span>",
                              "weeks":      "<span color='#99ffdd'><b>W{}</b></span>",
                              "weekdays":   "<span color='#ffcc66'><b>{}</b></span>",
                              "today":      "<span color='#ff6699'><b><u>{}</u></b></span>"
                              }
                    },
        "actions": {
                   "on-click-right": "mode",
                   "on-click-forward": "tz_up",
                   "on-click-backward": "tz_down",
                   "on-scroll-up": "shift_up",
                   "on-scroll-down": "shift_down"
                    }
```
So for now:

1. AModule is responsible for config parsing ("actions" section)
2. AModule handles users events (press buttons, wheel)
3. AModule when it needs calls module doAction in order to call module action
4. Clock got from the AModule which action should be executed
5. Clock defines internal function through function pointer and executes necessary functionality
---
This PR provides an opportunity to any module to focus on internal action implementation. And don't care about event functionality implementation via overriding of the handleToggle, handleScroll functions

### DIFF
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -11,16 +11,20 @@ namespace waybar {
 
 class AModule : public IModule {
  public:
-  AModule(const Json::Value &, const std::string &, const std::string &, bool enable_click = false,
-          bool enable_scroll = false);
   virtual ~AModule();
   virtual auto update() -> void;
   virtual auto refresh(int) -> void{};
   virtual operator Gtk::Widget &();
+  virtual auto doAction(const std::string& name) -> void;
 
   Glib::Dispatcher dp;
 
  protected:
+  // Don't need to make an object directly
+  // Derived classes are able to use it
+  AModule(const Json::Value &, const std::string &, const std::string &, bool enable_click = false,
+          bool enable_scroll = false);
+
   enum SCROLL_DIR { NONE, UP, DOWN, LEFT, RIGHT };
 
   SCROLL_DIR getScrollDir(GdkEventScroll *e);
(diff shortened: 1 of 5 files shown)


LukashonakV:
Hi @Alexays , any update here?

LukashonakV:
@Alexays , up )

Alexays:
Sorry for the delay, LGTM thx :)

Issue #2030, opened 2023-02-28
calendar: mode action not working without format-alt


Linus789:
Without the `format-alt` option specified in the clock, right clicking to change the mode does not work.

Config (right click does not work here):
```json
"clock": {
	"format": "{:%Y-%m-%d}",
	"tooltip-format": "<tt><small>{calendar}</small></tt>",
	"tooltip": true,
	"calendar": {
		"mode": "year",
		"mode-mon-col": 3,
		"weeks-pos": "right",
		"on-scroll": 1,
		"on-click-right": "mode",
		"format": {
			"months": "<span color='#ffead3'><b>{}</b></span>",
			"days": "<span color='#ecc6d9'><b>{}</b></span>",
			"weeks": "<span color='#99ffdd'><b>W{}</b></span>",
			"weekdays": "<span color='#ffcc66'><b>{}</b></span>",
			"today": "<span color='#ff6699'><b><u>{}</u></b></span>"
		}
	}
},
```

LukashonakV:
Hi @Linus789 , redefining of the module action is implemented and waiting for approval. Once it's merged you can specify any button for actions are supported by Clock module. See https://github.com/Alexays/Waybar/pull/2019

LukashonakV:
Hi @Linus789 #2019 is merge into the master.
So now your are able to redefine Clock actions on your wish. 
Please see: 
1. [Module actions config](https://github.com/Alexays/Waybar/wiki/Configuration#module-actions-config)
2. [Wiki clock](https://github.com/Alexays/Waybar/wiki/Module:-Clock)

If it's Ok for you , please close an issue.
Thank you

Linus789:
@LukashonakV Now I’m not even able to change the mode or scroll to change the year. My config:
```json
"clock": {
	"format": "{:%Y-%m-%d}",
	"format-alt": "{:%Y-%m-%d}",
	"tooltip-format": "<tt><small>{calendar}</small></tt>",
	"tooltip": true,
	"calendar": {
		"mode": "year",
		"mode-mon-col": 3,
		"weeks-pos": "right",
		"on-scroll": 1,
		"format": {
			"months": "<span color='#ffead3'><b>{}</b></span>",
			"days": "<span color='#ecc6d9'><b>{}</b></span>",
			"weeks": "<span color='#99ffdd'><b>W{:%V}</b></span>",
			"weekdays": "<span color='#ffcc66'><b>{}</b></span>",
[...]

PR #2035, opened 2023-03-01
Issue #2030. AModule enable_click & enable_scroll

LukashonakV:
Hi @Alexays . This PR supplement modules action and solves the Issue #2030 . The issue here is module (like Clock) when format-alt is not provided can call ALabel contstructor with enable_click = false or enable_scroll = false.  In the same time user can use redefining action functionality. As a result even user redefined actions AModule don't catch events. So AModule first check this. And in case when user redefined actions, AModule initializes event catching.


Alexays:
Thanks!

Issue #878, opened 2020-10-08
clock module should reuse default timezone (`TZ` or `/etc/localtime`)


lanodan:
After updating to Waybar 0.9.3 (with plus rfkill+sndio) I ended up with the following error and the clock module simply being absent from the bar:
```
[2020-10-08 05:49:13.955] [error] clock: Could not get current timezone
```

I fixed it by copying my timezone into the config but this isn't something I should have to do, it should fallback to the `TZ` environment variable and then `/etc/localtime`.

LukashonakV:
Hi @lanodan can you check again please? It was implemented at https://github.com/Alexays/Waybar/pull/2001

LukashonakV:
Hi @lanodan , up

lanodan:
Doesn't seems to be fixed, at least I'm getting the same error on f29a400c97d6cc339c5be2ca6c5467c04323b766

That is:
	[2023-03-02 01:59:29.303] [warning] module clock: Disabling module "clock", Could not get current timezone

(At least the logs do follow the timezone information)

LukashonakV:
Hi @lanodan , can you please:
1. Provide waybar config
2. Provide Linux OS you're using
Thanks

PR #2036, opened 2023-03-02
Clock tz try catch

LukashonakV:
Hi @Alexays , this PR should (but not sure 😄  ) solve #878 . At least there's a potential issue: 
1. User defines timezones as an array
2. User defines timezone as a single value

In both cases there no any check to ensure timezone can be initialize properly. 

This PR does:
1. TZ initialization is wrapped into try catch. If there is an exception waybar prints it as a warning,
2. If array of necessary timezones is empty waybar tries to use default system timezone


Alexays:
You're on fire, thanks!

PR #2037, opened 2023-03-02
Happy clang

LukashonakV:
Hi @Alexays , the goal of this PR is to:
1. Reduce warnings during clang compiling
2. To bring more order into code

Some points here:
1. According general useful recommendations when we have a deal with the inheritance, destructors in derived classes should be marked as virtual
2. Virtual methods on the most base classes should be marked as virtuals
3. Methods of derived classes should me marked as override

This PR quite huge but should impact any functionality.


Alexays:
Thanks!

PR #2043, opened 2023-03-05
[WIP] Sway workspaces: Configure output from which workspaces are shown

Gobbel2000:
This change allows configuring a sway workspaces module that only shows workspaces from one specific output, even if it's not the current output.

## Use Case

My use case is that I have 3 monitors but only one bar on one of the monitors. That bar shows all workspaces, but to have it organized better, I wanted to visually differentiate 3 workspace groups so that it is easily identifiable what workspace is on what output. I have 3 workspace modules configured something like this:

```json
    "sway/workspaces#out1": { 
        "for-output": "DP-2"
    },
    "sway/workspaces#out2": {
        "for-output": "HDMI-A-1"
    },
    "sway/workspaces#out3": {
        "for-output": "DP-1"
    },
```

This style gives each group a different underline color:

```css
#workspaces.out1 button {
    box-shadow: inset 0 -3px #dc0024;
}

#workspaces.out2 button {
    box-shadow: inset 0 -3px #00b41e;
}   

#workspaces.out3 button {
    box-shadow: inset 0 -3px #dcdc00;
}
```
### Screenshot
workspaces

## Further comments

If there is a desire to include this feature, then there probably is a bit more to be done before it can be merged:

* Documentation
* Possibly the same in other workspace modules, not just sway (where would that apply?)

Finally, I'm not that experienced with C++, so if you take issue with anything I've written, don't hesitate to say so.

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -30,6 +30,7 @@ class Workspaces : public AModule, public sigc::trackable {
 
   void onCmd(const struct Ipc::ipc_response&);
   void onEvent(const struct Ipc::ipc_response&);
+  bool isShown(const Json::Value&);
   bool filterButtons();
   Gtk::Button& addButton(const Json::Value&);
   void onButtonReady(const Json::Value&, Gtk::Button&);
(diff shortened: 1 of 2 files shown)


Alexays:
Sorry for this long delay 😓
Nice addition!
If you have time to add it to the man and Github wiki, aswell as checking if it's relevant to add this feature to other workspaces modules.

PR #2048, opened 2023-03-08
Wireplumber: Free the default node name

### LINKED ISSUE #2041: Possible memory leak in Wireplumber module
The old default_node_name_ should be freed with `g_free` here:

https://github.com/Alexays/Waybar/blob/b34d325a7114a0f84154d904a5c349b2006647f9/src/modules/wireplumber.cpp#L177

LDAP:
The `default_node_name_` is reassigned without calling `g_free` on the old string.

Fixes #2041


Alexays:
Thanks!

PR #2053, opened 2023-03-13
config:output:Allow multiple exclusions & wildcard

paul-ri:
As described in 90206f55be01c12be9c1b239b54bc3b346821742:

Covers the use case where needing to exclude more than 1 output but still include all other displays.

e.g. I have 3 monitors: laptop + HD + 4K; and 3 bar types:
- The main bar is on the laptop. `output: "laptop-monitor-id"`
- The 4K has a specific waybar bar-1 configuration. `output: "4K-monitor-id"`
- I want all other displays (3rd HD monitor / any HDMI output when presenting) to have a plain bar: `output: ["!laptop-monitor-id", "!4k-monitor-id", "*"]`

### Testing done
Added tests in the code and tried with an updated CSS:

```css
/* 4K screen */
.4k * {
    font-size: initial;
    background: pink;
}
```
The laptop on the right has the extra modules, the 4K screen in the middle is pink, the HD screen (and hopefully any other screen) displays the workspaces & date

image

### Note

If the star `'*'` is added anywhere before the end in the array, all other outputs will be ignored

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,4 +29,4 @@ jobs:
       - name: build
         run: ninja -C build
       - name: test
-        run: meson test -C build --no-rebuild --verbose --suite waybar
+        run: make test
--- a/Makefile
+++ b/Makefile
@@ -19,5 +19,9 @@ run: build
 debug-run: build-debug
 	./build/waybar --log-level debug
 
+test:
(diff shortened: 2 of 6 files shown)


paul-ri:
Ah damn, I just saw #1408. It's more than a year old and has breaking changes, so this current PR should be a good workaround for now.

Alexays:
Yep, LGTM thanks!

PR #2064, opened 2023-03-19
[hyprland/language] Add more format replacements.

en3wton:
This adds codes like {short} {variant} etc.  It uses the same codes as the ones for sway/language, I didn't bother adding the flags. Existing configs should still work fine.

I don't know C++ so please tell me if this is terrible.


en3wton:
I removed the use of "new" on the layout struct, in case that causes a memory leak? idk

Alexays:
LGTM

PR #2080, opened 2023-03-31
modules/clock: fix calendar shift in months with 31 days

cdump:
**Reproducer:** set the date to 31 March 2023, try scrolling the calendar (`"on-scroll-up": "shift_down"`, ...)
**Expected:** All months
**Actual:** Months with 31 days only

Debug print shows `2023-02-31 is not a valid year_month_day`

In this MR I set day to 1 if shit is not zero to fix this.


LukashonakV (commented):
I didn't check, but. The old code did shift date with year/month precision. Does current do the same? @cdump ?

cdump:
@LukashonakV [cldCurrShift_](https://github.com/Alexays/Waybar/blob/master/include/modules/clock.hpp#L52) is still `date::months`
I've tested shift in both `moth` and `year` calendar mode and everything looks good

PR #2092, opened 2023-04-06
mpris: fix non-dynamic tags not being html escaped

### LINKED ISSUE #2091: certain songs with characters cause mpris title too not update
was working on turning my current hyprland theme into something shareable and i noticed that the mpris module was spewing this into the log, not sure how to prevent this on my end, is their a way too have the title not treated as special by gtk?

`
(waybar:114900): Gtk-WARNING **: 15:43:58.688: Failed to set text 'spotify    BLXXDY NXSE (feat. Scarlxrd & nascar aloe)  Jasiah  01:50/03:07' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
`

chayleaf:
Closes #2091


Alexays:
Thx!

PR #2114, opened 2023-04-14
Split cpu module

mmhat:
This PR contains three different modules that were derived from the original cpu module: cpu_frequency, cpu_usage and load.
The reasoning behind that split is that it is impossible to define states that severe all three purposes equally well, e.g. say I want a different color if the cpu usage goes above 50 percent and a different format if the load goes above 10.
As another improvement the load module now also shows the 5 minute and the 15 minute average.


Alexays:
Hey!
Sorry for the delay, can you move common parts of these modules in shared files to avoid duplication?

mmhat:
@Alexays I deduplicated the code; I didn't move the shared code to own files though, the cpu module uses the functions of the cpu_frequency, cpu_usage and load module. Is that ok? If not, where should those shared files live? In utils?

Alexays:
There's some errors in freebsd tests, can you check?

mmhat:
@Alexays I fixed the freebsd code and the CI is green again. Can you approve the workflows?

Alexays:
Sorry again for the delay, do you think it's possible to do another PR to update man and Github Wiki?
Thanks again!

anarcat:
hmm... it looks like the wiki docs haven't been updated to reflect the changes here:

https://github.com/Alexays/Waybar/wiki/Module:-CPU

Issue #2126, opened 2023-04-20
[Feature request] Display battery percentage as weighted average of battery levels


bruceblore:
Right now, if you are using a laptop with multiple batteries, it appears that the displayed capacity percentage is just an average of the capacity levels of the individual batteries. This is fine if the batteries are of a similar size, but if the batteries are of significantly different sizes, this can result in the reported percentage not being an accurate representation of how much battery life remains.

For example, it is common for thinkpads to have a 24 WHr internal battery and a 72 WHr removable battery, and to use the removable battery first. The first 50% reported by Waybar will last a long time because the first battery is large. But the second 50% will not last nearly as long. At the moment where 50% is displayed, the laptop really only has 25% of its overall capacity left. In the opposite scenario with the internal battery drained but the removable battery completely full, Waybar still displays 50%, but the laptop has 75% of its overall capacity left.

By using the weighted average of battery capacity levels, with the battery capacities as weights, the battery percentage displayed by Waybar would be much more accurate. It would probably be best to make this optional, in case someone prefers the existing behavior. And it would probably be good to use the `design-capacity` option to determine whether current capacity or design capacity gets used in the calculation.

bruceblore:
I just got some free time and decided to try implementing this myself in pull request #2134. Should I close this issue?

Icey-Glitch:
@bruceblore i was wondering you provide an example of how to use what you added. i have a thinkpad and avg of the 2 bats would be great. thanks!

bruceblore:
"battery": {
            "weighted-average": true
    }

PR #2134, opened 2023-04-23
Display battery percentage as weighted average of battery levels

bruceblore:
A few days ago, I opened issue #2126 requesting this feature. I had some free time recently and decided to try making a pull request. I determined that weighted average of percentages with capacities as weights was a little complex, and that a better way to calculate the percentage that I wanted would be to just compute total_energy * 100 / total_energy_full. No change is required for the condition where the user sets design-capacity to true, as that is already total_energy * 100 / total_energy_full_design.

Right now, the option to enable this is called weighted-average. Is that an appropriate name, or should I call it something else?

Additionally, I discovered that it total_capacity is not greater than zero, Waybar already falls back to doing this calculation. (And still would, even if weighted-average is set to false.) However, it does not fall back if total_capacity could not be calculated at all. Is this the intended behavior? I have a feeling that the following code a few lines before my changes:
```
if (total_capacity_exists) {
  if (total_capacity > 0.0f)
    calculated_capacity = (float)total_capacity / batteries_.size();
  else if (total_energy_full_exists && total_energy_exists) {
    if (total_energy_full > 0.0f)
      calculated_capacity = ((float)total_energy * 100.0f / (float)total_energy_full);
  }
}
```
should have been this:
```
if (total_capacity_exists && total_capacity > 0.0f) {
  calculated_capacity = (float)total_capacity / batteries_.size();
} else if (total_energy_full_exists && total_energy_exists) {
  if (total_energy_full > 0.0f)
    calculated_capacity = ((float)total_energy * 100.0f / (float)total_energy_full);
}
```
but I wasn't quite confident enough to make the change myself. Is that correct?

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -533,6 +533,13 @@ const std::tuple<uint8_t, float, std::string, float> waybar::modules::Battery::g
       }
     }
 
+    // Handle weighted-average
+    if ((config_["weighted-average"].isBool() ? config_["weighted-average"].asBool() : false) &&
+        total_energy_exists && total_energy_full_exists) {
+      if (total_energy_full > 0.0f)
+        calculated_capacity = ((float)total_energy * 100.0f / (float)total_energy_full);
+    }
+
     // Handle design-capacity
     if ((config_["design-capacity"].isBool() ? config_["design-capacity"].asBool() : false) &&
         total_energy_exists && total_energy_full_design_exists) {


hashworks:
Could you add this to the [wiki](https://github.com/Alexays/Waybar/wiki/Module:-Battery)?

bruceblore:
> Could you add this to the [wiki](https://github.com/Alexays/Waybar/wiki/Module:-Battery)?

I wasn't aware that I was able to edit the wiki. Thanks for pointing that out. I have added it.

Issue #2133, opened 2023-04-22
waybar: error while loading shared libraries: libcava.so: cannot open shared object file: No such file or directory


prathameshk23:
I get this error while running waybar it runs fine from the ./build

LukashonakV:
Hi @prathameshk23, can you try `sudo ldconfig`, please

prathameshk23:
Still the same

LukashonakV:
> Hi @LukashonakV Still the same

Can you please check:
1. `ninja -C build install`
2. `sudo ldconfig`

prathameshk23:
> > Hi @LukashonakV Still the same
> 
> Can you please check:
> 1. `ninja -C build install`
> 2. `sudo ldconfig`
> 3. and try run waybar as usual

The ninja -C build is not completing throw error while compiling. I tried installing it from the aur still getting the libcava.so error

LukashonakV:
> > > Hi @LukashonakV Still the same
> > 
> > Can you please check:
> > 1. `ninja -C build install`
> > 2. `sudo ldconfig`
> > 3. and try run waybar as usual
> 
> The ninja -C build is not completing throws error while compiling. I tried installing it from the aur still getting the libcava.so error

Can you share aur repo, link, package?

prathameshk23:
> > > > Hi @LukashonakV Still the same
> > > 
> > > Can you please check:
> > > 1. `ninja -C build install`
> > > 2. `sudo ldconfig`
> > > 3. and try run waybar as usual
> > 
> > The ninja -C build is not completing throws error while compiling. I tried installing it from the aur still getting the libcava.so error
> 
> Can you share aur repo, link, package?

https://archlinux.org/packages/community/x86_64/waybar/

LukashonakV:
> > > > Hi @LukashonakV Still the same
> > > 
> > > Can you please check:
> > > 1. `ninja -C build install`
> > > 2. `sudo ldconfig`
> > > 3. and try run waybar as usual
> > 
> > The ninja -C build is not completing throws error while compiling. I tried installing it from the aur still getting the libcava.so error
> 
> Can you share aur repo, link, package?

`sudo ninja -C build install` installs waybar application you just built. It also should install libcava. Then `sudo ldconfig` . Can you try not just build but install too

prathameshk23:
> > > > > Hi @LukashonakV Still the same
> > > > 
> > > > Can you please check:
> > > > 1. `ninja -C build install`
[...]

Issue #2130, opened 2023-04-21
module cava: Unknown module: cava


XxRiKxX:
so I copy-pasted the cava config from the wiki to my config file and when I start waybar, it says that the module is not found. I am using hyprland with waybar-hyprland-git from aur

pbosab:
seems an aur package problem or am I wrong?
https://aur.archlinux.org/cgit/aur.git/log/?h=waybar-hyprland-git

LukashonakV:
It might depend on whether aur contain particular waybar package. Cava was released a couple of days ago.

XxRiKxX:
don't know why they did it, but I will just install the non-git version. Also does the default waybar arch repository works without any problems in hyprland? Thanks!

LukashonakV:
> seems an aur package problem or am I wrong?
> https://aur.archlinux.org/cgit/aur.git/log/?h=waybar-hyprland-git
> 
> *edit: last edit message is: [Remove cava module](https://aur.archlinux.org/cgit/aur.git/commit/?h=waybar-hyprland-git&id=227ac1a905e1429f528614c537fea90a417b9df1)

Thank you, at least now the reason is clear for me. I'll check next week how to handle the issue properly

maxmx03:
I tried to build, but I was getting these errors. I installed Cava, but I still kept getting the error:

`44:16: ERROR: C++ shared or static library 'cava' not found.`

I had to modify line 345 in the meson.build file:

```
cava = compiler.find_library('cava',
                             required: false)
```

distro opensuse - hyprland.

thyeun:
for waybar-hyprland-git version at aur arch linux, during installation/compile, you can see a line mentioned that cava been disable.

Not sure related to this or not.

LukashonakV:
> for waybar-hyprland-git version at aur arch linux, during installation/compile, you can see a line mentioned that cava been disable.
> 
> Not sure related to this or not.

This log exact means waybar was built without cava module support

XxRiKxX:
why did they left cava out though? and also the waybar package from official repo does not work fully. Like the workspaces module.

thyeun:
@XxRiKxX if your WM is under `hyprland` and your config is using `wlr/workspaces` than you need `waybar-hyprland-git` version to make the workspaces work, but if your workspace is under sway (not with wlroots), than you need only `waybar-git`. i think they separate it for cover up wlroot and others, instead that they can combine into one, to have a `mode setting` for wlroot or others.

Issue #2129, opened 2023-04-21
[warning] module cava: Unknown module: cava


shasherazi:
`cava` works fine in terminal, there is no other error when starting `waybar`, just this.

Here is my config
```json
{
  "height": 33,
  "spacing": 10,
  "layer": "top",
  "margin-left": 5,
  "margin-right": 5,
  "modules-left": [
    "cava",
    "clock"
  ],
  "modules-center": [
    "wlr/workspaces"
  ],
  "modules-right": [
    "memory",
    "cpu",
    "pulseaudio",
    "battery",
    "network",
    "tray"
  ],
  "tray": {
    "icon-size": 21,
    "spacing": 10
  },
  "battery": {
    "bat": "BAT0",
    "interval": 5,
    "states": {
      "warning": 50,
      "critical": 20
    },
    "format": "{capacity}% {icon}",
    "format-charging": "{capacity}% 󰂄",
    "format-time": "{H}h {M}min",
    "format-icons": [
      " ",
      " ",
      " ",
      " ",
      " "
    ]
  },
  "clock": {
    "format": "{:%H:%M}",
    "interval": 60,
    "format-alt": "{:%A, %B %d, %Y (%R)}",
    "tooltip-format": "<tt><small>{calendar}</small></tt>",
    "calendar": {
      "mode": "month",
      "mode-mon-col": 3,
      "weeks-pos": "right",
      "on-scroll": 1,
      "on-click-right": "mode",
      "format": {
        "months": "<span color='#ffead3'><b>{}</b></span>",
        "days": "<span color='#ecc6d9'><b>{}</b></span>",
        "weeks": "<span color='#99ffdd'><b>W{}</b></span>",
        "weekdays": "<span color='#ffcc66'><b>{}</b></span>",
        "today": "<span color='#ff6699'><b><u>{}</u></b></span>"
      }
    },
    "actions": {
      "on-click-right": "mode",
      "on-click-forward": "tz_up",
      "on-click-backward": "tz_down",
      "on-scroll-up": "shift_up",
      "on-scroll-down": "shift_down"
    }
  },
  "wlr/workspaces": {
    "format": "{icon}",
    "format-icons": {
      "default": "",
      "active": ""
    },
    "on-click": "activate",
    "sort-by-number": true,
    "persistent_workspaces": {
      "1": [],
      "2": [],
      "3": [],
      "4": [],
      "5": [],
      "6": [],
      "7": [],
      "8": [],
      "9": [],
      "10": []
[...]

PR #2144, opened 2023-04-28
cava: shared library is not found

LukashonakV:
Hi @Alexays , 
Arch users noticed cava fork compiles both and library and application. So once waybar is installed it overrides existing cava application what is not they want. So in the libcava fork I refreshed building logic and by default provide libcava as a single library. In addition new  libcava release now provides pkg-config manifest. So waybar can use dependency meson method now.
This PR should solve [Arch, ...] users complaining about cava module with the error:  shared library is not found
Relates: #2133, #2130, #2129


AFAgarap:
This issue persists in EndeavourOS after cloning the latest repository for the fix for #2351

LukashonakV:
Hi @AFAgarap , please look into #2328.

PedrelliMath:
@LukashonakV thanks, its solved the problem for me

PR #2156, opened 2023-05-05
Add persistent format-icon to wlr/workspaces

### LINKED ISSUE #2155: Persistent workspace format-icon in wlr/workspaces
There's no way to set an specific icon to persistent workspaces, making it so you cant differentiate it from non empty workspaces that are not active

theRealCarneiro:
This PR introduces the "persistent" format-icon to the wlr/workspaces module, changing the icon of empty persistent workspaces
Fixes #2155


theRealCarneiro:
Should i just change persistent to empty instead?

Alexays:
We should keep `persistent ` in order to not produce breaking changes.
LGTM

PR #2175, opened 2023-05-15
Add a new start_hidden flag

maqrrr:
Add a new `start_hidden` flag to set visibility on launch.

I'm using `"mode": "hide"` on Hyprland.

To make the bar hide on launch, I currently have to do `waybar & sleep 1 && killall -SIGUSR1 waybar` to hide waybar on launch.  (I think the sleep is required because it takes a little under a second for the signal handling thread to become available.)

With this change, I can add this to my waybar config and have it hide on launch:
```json    
"mode": "hide",
"start_hidden": true,
```


Alexays:
LGTM, can you also update the man? thx :)

PR #2185, opened 2023-05-24
Update to new ipc version

xdavidel:
Since [MadcowOG](https://github.com/MadcowOG) change the IPC patch for DWL - I also updated the DWL module to match it.

This targets [version 2](https://github.com/djpohly/dwl/compare/main...madcowog:ipc-v2.patch) of the path


Alexays:
LGTM thx!

Issue #2186, opened 2023-05-25
hyprland/window hide when its empty.


nimaaskarian:
I'm using AUR's `waybar-hyprland-git`.  I want my window module to get hidden when its empty. But I couldn't get it to work.  
This is my hyprland/window configuration:
``` 
"hyprland/window":{
    "format": "<span>{}</span>",
    "separate-outputs": true,
    "max-length": 25,
  },
```

Issue #1995, opened 2023-02-01
[request] format-empty for hyprland/window module?


Fxzzi:
Hi all, is there a possibility that a format-empty option could be added which allows us to format a module when the it is empty? I'll show an example here:

![](https://cdn.upload.systems/uploads/kTwNcAFu.png)

the module center-right is the hyprland/window module in question. You can see in this screenshot it looks fine.

![](https://cdn.upload.systems/uploads/XySx89lj.png)

In this screenshot, you can see that nothing is focused. This causes my bar to look a little off as there is a | seperator with nothing after it. Is there a possibility that this could be added to alleviate issues like these? I have my | seperators set up as a custom module.

This will also probably occur on the tray module I have on the right.

gmr458:
I support this, something similar happens when using borders.
![20230201_15h41m36s_grim](https://user-images.githubusercontent.com/53343401/216158665-6d49e7b9-ffa3-428d-a1f5-92ef6712cd60.png)

fybx:
This issue possibly has the exact root cause as #1934.

haShinui:
Hello so for me just adding ```"hyprland/window"{
"format": "{}"
"rewrite" : {
        /*stuff you have like discord Firefox and so on */
       " ": "TEST"
}
}, ``` 
This worked for me and if its an empty window it display TEST or whaever you want else it display my discord and so on. So adding `"": "TEST"` is what you need to add

SilvanVL:
"hyprland/window": {
	"format": "{}",
	"rewrite": {
		"(.+)": "|$1"
	},
	"max-length": 36,
	"separate-outputs": true
},

if title length is 1+ symbols long it will add an "|" symbol to its start. You can also remove paddings and use " " instead to completely hide the module. Hope it will help

Babkock:
Damn these are some sweet rices!

Issue #1934, opened 2023-01-11
window label is shown when no window is focused in Hyprland module


fybx:
> Did you have a look at #1448 or #1349?
> 
> I think you could style
> ```
> window#waybar.empty #window {  
>  background:none;  
> }  
> ```
> or similar.
_Originally posted by @RobertMueller2 in https://github.com/Alexays/Waybar/issues/1507#issuecomment-1090619823_
      

This solution for Sway doesn't work for Hyprland/window module.

RobertMueller2:
Not much idea of the hyprland/window module, but replying here because I was tagged. ;) I think you can stop trying around with CSS at this point. ;) In general, it can be helpful to use GTK_DEBUG=interactive environment variable with a test config to get an insight to CSS names and classes. Here is a comparison between hyprland/window and sway/window modules:

![swappy-20230112_061506](https://user-images.githubusercontent.com/29629287/211983740-293af1a9-a156-4aa7-a6e0-13183119f90a.png)

sway/window adds CSS classes to the parent GTK Window with the name waybar (window#waybar), which you can use to style the GTK Label with the name window.  hyprland/window doesn't do any of this.

Now whether it SHOULD, or whether it should do things the same way, I can't say. ;)

fybx:
In Hyprland, the label element gets the ID `#window`, but classes do not reflect state changes as it does on Sway.

![20230112_21h33m58s_grim](https://user-images.githubusercontent.com/41127439/212150307-f1f37b05-d51a-442d-a297-09c20b3f35db.png)

Maybe an update regarding this will solve it, thanks for responding :)

PR #2196, opened 2023-06-01
[hyprland/window] Add css class for empty window name

calvinchd:
Adds an `empty` css class when window name is empty.

Select with `#window.empty`.

Related #2186, #1995, #1934
### Notes
- Assumes `windowName` is not null.
- Matches window with an empty name.

### DIFF
--- a/src/modules/hyprland/window.cpp
+++ b/src/modules/hyprland/window.cpp
@@ -92,7 +92,8 @@ void Window::onEvent(const std::string& ev) {
   if (windowName == lastView) return;
 
   lastView = windowName;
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

beh-10257:
this doesn't work anymore 
#window.empty {
    background-color: #ffffff;
}
this doesn't work per example

calvinchd:
> this doesn't work anymore #window.empty { background-color: #ffffff; } this doesn't work per example

Try replacing `#window.empty` with `window#waybar.empty #window`

PR #2203, opened 2023-06-03
mpris: dynamic tag ordering and separator customization

rodrgz:
This PR makes it possible:
 - the choice of tag orders in dynamic mode;
 - the choice of the separator;
 - the non-inclusion of the tag if not declared in "dynamic-order".
 
It is observed that "dynamic-priority" was renamed to "dynamic-importance-order" to be more intuitive and not confused with "dynamic-order".

### DIFF
--- a/man/waybar-mpris.5.scd
+++ b/man/waybar-mpris.5.scd
@@ -71,11 +71,26 @@ The *mpris* module displays currently playing media via libplayerctl.
 	something less than or equal to this value, so the title will always be ++
 	displayed.
 
+*dynamic-order*: ++
+	typeof: []string ++
+	default: ["title", "artist", "album", "position", "length"] ++
+	Order of the tags shown by Dynamic tag. The position and length tags ++
+	will always be combined in the format [{position}/{length}]. The order ++
+	of these tags in relation to other tags will be determined based on the ++
+	declaration of the first among the two tags. Absence in this list means ++
+	force exclusion.
+
 *dynamic-priority*: ++
 	typeof: []string ++
 	default: ["title", "length", "position", "artist", "album"] ++
 	Priority of the tags when truncating the Dynamic tag (absence in this
 	list means force inclusion).
+
+*dynamic-separator*: ++
+	typeof: string ++
+	default: " - " ++
+	These characters will be used to separate two different tags, except ++
+	when one of these tags is position and length.
 
 *truncate-hours*: ++
 	typeof: bool ++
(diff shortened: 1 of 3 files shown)


rodrgz:
@chayleaf @robertgzr

man/waybar-mpris.5.scd
```diff
@@ -71,12 +71,27 @@ The *mpris* module displays currently playing media via libplayerctl.
...
+*dynamic-order*: ++
+	typeof: []string ++
+	default: ["title", "artist", "album", "position", "length"] ++
+	Order of the tags shown by Dynamic tag. The position and length tags ++
+	will always be combined in the format [{position}/{length}]. The order ++
+	of these tags in relation to other tags will be determined based on the ++
+	declaration of the first among the two tags. Absence in this list means ++
+	force exclusion.
+
+*dynamic-truncation-priority*: ++
```
Alexays, line 83: We need to keep `dynamic-priority` for backward compability

chayleaf: I think it's fine to rename it in the manual, but check if the old name is set in code

rodrgz: > I think it's fine to rename it in the manual, but check if the old name is set in code

done

man/waybar-mpris.5.scd
```diff
@@ -71,12 +71,27 @@ The *mpris* module displays currently playing media via libplayerctl.
...
-	default: ["title", "length", "position", "artist", "album"] ++
-	Priority of the tags when truncating the Dynamic tag (absence in this
+	default: ["title", "artist", "album", "position", "length"] ++
+	Order of the tags shown by Dynamic tag. The position and length tags ++
+	will always be combined in the format [{position}/{length}]. The order ++
+	of these tags in relation to other tags will be determined based on the ++
+	declaration of the first among the two tags. Absence in this list means ++
+	force exclusion.
+
+*dynamic-truncation-priority*: ++
```
chayleaf, line 83: at the risk of bikeshedding here, I don't really like the name `dynamic-truncation-priority` because it implies that high priority = gets truncated first, while it's precisely the opposite here (high priority = gets truncated last)

rodrgz: What I understood was that the first on the list will be truncated before. Do you have another name suggestion?

chayleaf: maybe dynamic-truncation-order?

chayleaf: maybe dynamic-importance?

Alexays:
LGTM

Alexays:
Can you also update the Github wiki? thx 🙏

PR #2207, opened 2023-06-04
fix(dwl-module): fix typo to avoid null pointer

dejor:
Fixes an issue that occurs after starting Waybar with the `dwl/tags` module.

Error:
`[0000-00-00 00:00:00.000] [error] dwl_status_manager_v2 not advertised`


xdavidel:
I confirm the problem.
This commit fixes a typo the remained from the old IPC patch.

PR #2212, opened 2023-06-05
wireplumber: Support for scrolling

evyatark2:
This adds the ability to scroll the WirePlumber widget like the PulseAudio one.

- Default scroll action is to change the volume using WirePlumber's mixer API.
- max-volume and scroll-step values can be used to configure how volume adjustments works
- ~The default scroll-step is provided by the mixer API~
- Customizable scroll actions

### DIFF
--- a/include/modules/wireplumber.hpp
+++ b/include/modules/wireplumber.hpp
@@ -27,6 +27,8 @@ class Wireplumber : public ALabel {
   static void onMixerChanged(waybar::modules::Wireplumber* self, uint32_t id);
   static void onDefaultNodesApiChanged(waybar::modules::Wireplumber* self);
 
+  bool handleScroll(GdkEventScroll* e) override;
+
   WpCore* wp_core_;
   GPtrArray* apis_;
   WpObjectManager* om_;
@@ -36,6 +38,7 @@ class Wireplumber : public ALabel {
   uint32_t pending_plugins_;
   bool muted_;
   double volume_;
+  double step_;
   uint32_t node_id_{0};
   std::string node_name_;
 };
--- a/include/modules/wireplumber.hpp
+++ b/include/modules/wireplumber.hpp
@@ -38,7 +38,7 @@ class Wireplumber : public ALabel {
   uint32_t pending_plugins_;
   bool muted_;
   double volume_;
-  double step_;
+  double min_step_;
   uint32_t node_id_{0};
   std::string node_name_;
 };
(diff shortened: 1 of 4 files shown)


evyatark2:
I am not sure as to what to put in the documentation for the default value of scroll-step.
I am using the value that the mixer plugin is providing but I am not sure what that is, but **I think** it is the minimum value of stepping the device supports.
Should I change my implementation to just use a default value of 1% like the PulseAudio module does?

evyatark2:
Ok, after reading WirePlumber's and PipeWire's source code step is in fact the minimum step, so I used 1% as a default step.
I added manpages and will update the wiki now, you can merge this

yangyingchao:
With some simple modification, issue #2253 could be fixed :)
```diff
From 97dcf3a9e2ecfbfb17a94830894832c51d8e4ce3 Mon Sep 17 00:00:00 2001
From: yangyingchao <yang.yingchao@qq.com>
Date: Mon, 26 Jun 2023 17:17:10 +0800
Subject: [PATCH] auto unmute when scroll while sink is muted

---
 src/modules/wireplumber.cpp | 8 ++++++++
 1 file changed, 8 insertions(+)

diff --git a/src/modules/wireplumber.cpp b/src/modules/wireplumber.cpp
index b2d9b39d..d3ef2b82 100644
--- a/src/modules/wireplumber.cpp
+++ b/src/modules/wireplumber.cpp
@@ -350,6 +350,14 @@ bool waybar::modules::Wireplumber::handleScroll(GdkEventScroll* e) {
     GVariant* variant = g_variant_new_double(new_vol);
     gboolean ret;
     g_signal_emit_by_name(mixer_api_, "set-volume", node_id_, variant, &ret);
+
+    if (muted_) {
+      g_auto(GVariantBuilder) b = G_VARIANT_BUILDER_INIT(G_VARIANT_TYPE_VARDICT);
+      GVariant* variant = nullptr;
+      g_variant_builder_add(&b, "{sv}", "mute", g_variant_new_boolean(false));
+      variant = g_variant_builder_end(&b);
+      g_signal_emit_by_name(mixer_api_, "set-volume", node_id_, variant, &ret);
+    }
   }
   return true;
 }
-- 
2.39.3


```

yangyingchao:
> I think leaving it like this is better as this is how the Pulseaudio module works. You can use `on-scroll-up` and `on-scroll-down` to do that.

I think it would be nice if you explain how to do it in wiki or related manual, then leave some comments in issue #2253 :)

ticoombs:
@evyatark2 You made my day btw. Thanks so much for this.

PR #2213, opened 2023-06-05
DBUS IPC architecture

LukashonakV:
Hi @Alexays , this PR is the try to make/organize first base service-client architecture among Waybar modules. 
There are new base classes  - DBusService + DBusClient.  These classes are developed base on the DBus concept where service is a singleton. 
Cava and Clock modules are working now under the new architecture.
Communication via DBus is organized through asynchronous calls(including making the DBus path registration)
Waybar introspection for DBus is presented in IDBUS interfaces. Necessary services are constructed by the DBusService using this introspection.
So.. once Waybar i starting Cava and Clock are registering their services on DBus. All next raised Waybar instances are trying to register DBus names but due to they are already busy by the current instance, new services are not working. 
In the same time all clients are alive and can send necessary requests.
For example once Waybar started even user can ask DBus to provide result for Clock and Cava now
ps_2023-06-06-00_25_38

---
Even now concept is - one service <-> many clients such architecture doesn't reduce resource consumption for the heavy modules like cava under high load (but the difference is really is not significant). 
In the same time such functionality exposes Waybar API for external modules using DBus. Even user can check what exact result is per services/module


src/services/DBusService.cpp
```diff
@@ -0,0 +1,105 @@
...
+namespace waybar::services {
+
+DBusService::DBusService(const Json::Value& config, const std::string& name, const std::string& id,
+                         const std::string& format)
+    : Config(config, format),
+      name_(std::move(name)),
+      introspection_data_{Gio::DBus::NodeInfo::create_for_xml(
+          Glib::ustring::sprintf(DBUS_INTROSPECTION_XML, name_))} {
+  owner_id_ = Gio::DBus::own_name(
+      Gio::DBus::BusType::BUS_TYPE_SESSION, Glib::ustring::sprintf(DBUS_NAME, name_),
```
alebastr, line 12: Service name does not include `id` or bar id. What does it do when you have multiple instances of the same module? I.e. `clock#first`, `clock#second`.

What would it do when you have two bar configs with differently configured clock modules? I.e.
```json
[
  {
    "position": "top",
    "modules-center": ["clock"],
    "clock": {
        "timezone": "America/Los_Angeles"
    }
  },
  {
    "position": "bottom",
    "modules-center": ["clock"],
    "clock": {
        "timezone": "UTC"
    }
  }
]
```

include/interfaces/IDBus.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace waybar {
+
+static constexpr auto DBUS_OBJECT_PATH{"/fr/arouillard/waybar/services/%s"};
+static const Glib::ustring DBUS_NAME{"fr.arouillard.waybar.services.%s"};
+static const Glib::ustring DBUS_INTROSPECTION_XML =
+    "<!DOCTYPE node PUBLIC \"-//freedesktop//DTD D-BUS Object Introspection 1.0//EN\""
+    "\"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd\">"
+    "<node>"
+    "  <interface name='fr.arouillard.waybar.services.%s'>"
+    "    <method name='doAction'>"
```
alebastr, line 18: This interface is not really useful for creating external modules, as it requires a module code in waybar that knows the list of available actions.

The module API has a fixed set of actions and signals. Wouldn't it be better to have these in the interface? Add capabilities - supports click/doubleClick/scroll/etc, and you'll make it possible to do a generic `"custom-dbus#something": { "name": "xxx", "object_path": "/yyy" }`.

alebastr (commented):
That doesn't seem like a good idea overall. DBus communication will negate any possible performance benefits from the singletons and increase IO. Like some people say, DBus makes everything worse.

Have you considered just splitting the module presentation and logic and storing module backends in a global map `bar id + module name + module id -> weak_ptr(module backend)`?
I had that idea for a while, but I don't have time to work on it :cry:

src/services/DBusService.cpp
LukashonakV: Agree. Totally forgot about this . Implementation took huge time.

include/interfaces/IDBus.hpp
LukashonakV: Actually all logic about method is inherited under doAction method which is fully responsive about incoming requests

Anyway I'm still hesitating about this PR useful

LukashonakV:
> That doesn't seem like a good idea overall. DBus communication will negate any possible performance benefits from the singletons and increase IO. Like some people say, DBus makes everything worse.
> 
> Have you considered just splitting the module presentation and logic and storing module backends in a global map `bar id + module name + module id -> weak_ptr(module backend)`? I had that idea for a while, but I don't have time to work on it cry

This is exactly what happened after development was finish. In the end I realized DBus layer doesn't bring some significant improvement. Moreover there is some not significant resource penalty
However ... when the system is not under the pressure work - new architecture even works a little efficient
ps_2023-06-06-12_30_51
It consumes less CPU...
But unfortunately under the loading... consuming CPU changes to the opposite
ps_2023-06-06-12_34_29


@alebastr  thank you a lot for you reviewing and opinion. Actually I had doubts about publishing this PR... due to I don't see significant benefits.... And I'm still thinking of .... is it really useful to split modules functionalities on the back and front....

PR #2217, opened 2023-06-06
Tray module causes: Invalid id passed to g_bus_unwatch_name()

LukashonakV:
When Waybar exit in the log I see:
ps_2023-06-06-23_18_51
`(waybar:25493): GLib-GIO-WARNING **: 22:57:40.825: Invalid id 2 passed to g_bus_unwatch_name()`

Investigation shows: Tray module constructor calls :
1. watcher constructor
2. host constructor
In the same time both base classes(watcher and host) keeps information about cross resources. Watcher instance has got information about host and watcher either. The same for host class..
So when time to destruct instances, both watcher and host are trying to destroy cross resources. In such case : watcher may kill host resource earlier then host destruction is called and vise versa for host class.
This PR splits resource destruction between base classes. Host now destruct only host objects, watcher - watchers objects.

Actually I didn't find any other dependencies where these base classes are used separately without each other.

### DIFF
--- a/src/modules/sni/host.cpp
+++ b/src/modules/sni/host.cpp
@@ -22,10 +22,6 @@ Host::~Host() {
     Gio::DBus::unwatch_name(bus_name_id_);
     bus_name_id_ = 0;
   }
-  if (watcher_id_ > 0) {
-    Gio::DBus::unwatch_name(watcher_id_);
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks @LukashonakV you rock!

PR #2219, opened 2023-06-07
Wake all sleeping threads when leaving suspend

xytovl:
std::condition_variable::wait_for does not count time spent in sleep mode, resulting in longer than expected waits.


Alexays:
Thanks!

Issue #2188, opened 2023-05-25
sway/workspaces: Focused and default icons not displayed


d4rios:
Hi

In my waybar configuration, the module "sway/workspaces" does not show the **focused** workspace and **default** workspace **icons**. I have only been able to change the color of the workspace used, using the style.css file. Otherwise it would be very difficult for me to know which workspace I am in.

Does anyone know why this might be happening?. 

❯ sway -v
sway version 1.8.1

❯ waybar -v
Waybar v0.9.17

sway/workspaces module config:

    "sway/workspaces": {
        "all-outputs": true,
        "disable-scroll": true,
        "format": "{icon}",
        "format-icons": {
        "1": "",
        "2": "",
        "3": "",
        "4": "",
        "5": "",
        "6": "",
        "7": "",
        "8": "",
        "9": "",
        "10": "",
        "urgent": "",
        "focused": "",
        "default": ""
        },

        "persistent_workspaces": {
        "1": [],
        "2": [],
        "3": [],
        "4": [],
        "5": [],
        "6": [],
        "7": [],
        "8": [],
        "9": [],
        "10": []
        }
 },

[My Waybar](https://tinypic.host/images/2023/05/25/Screenshot_2023-05-25-07-35-23_1920x10800A1920x1080.png)

Anakael:
Default icon doesn't work because you have config for all 10 workspaces. Just try to remove some numbered workspaces from "format-icons" section and it should be fine.

Anakael:
And the same reason for missing `focused` icon. You can reproduce it by:
1) remove some numbered workspaces from config (for example, starting from 5)
2) create workspace 6 and launch terminal on it
3) create workspace 7 and stay on it
Then you will have 6-default and 7-focused.
For `focused` icon behavior it is definitely a bug. It should have priority over existing configs.

d4rios:
Thanks for your answer, Anakael.

Now the waybar shows me the default icon in the last workspace also shows 6 workspaces when I only have 5 defined. 
I don't understand anything.

sway_output
![imagen](https://github.com/Alexays/Waybar/assets/80931135/a4571d34-d7af-4378-a91f-081579fb9841)

sway/workspaces
![imagen](https://github.com/Alexays/Waybar/assets/80931135/b442295a-97f2-4bc4-8e0a-f2e5d798812a)

waybar
![imagen](https://github.com/Alexays/Waybar/assets/80931135/92fcf544-90dd-4423-ac2b-799827da7aa7)

[...]

PR #2223, opened 2023-06-08
[sway/workspaces] fix: change icons priority

Anakael:
Resolves: https://github.com/Alexays/Waybar/issues/2188


Samu1808:
I think this was a fix, which does not made things better for other configuration using special icons for every workspace.
See on #2188 and please reopen it again.

PR #2234, opened 2023-06-13
sway/window: app_id on unfocused workspaces

kris7t:
This patch tries to make `app_id` specific styling from `sway/window` more useful if the `smart_borders` and `smart_gaps` Sway options are used in conjunction with the `offscreen-css` option of the module.

My main use case for this is styles like `window#waybar.solo.kitty`, which can make Waybar blend in with the solo window on the workspace (gapless and borderless thanks to the Sway options). For example, you may set a semi-transparent background for Waybar only if a terminal with semi-transparent background is the only window on the workspace, but use a fully opaque background for any other (opaque) window.

Unfortunately, `app_id`-specific styling is not very useful with multiple monitors. Even if `offscreen-css` is enabled, the CSS class corresponding to `app_id` is only applied for the focused window. Thus, Waybar can only ever blends with focused windows.

This patch also applies the `app_id` for unfocused workspaces when `offscreen-css` is set, but only if there is a single, non-floating window. This should catch situations where `smart_borders` and `smart_gaps` are useful. Technically, gaps and borders are collapsed also when the workspace itself is set to tabbed/stacking with multiple windows, but in that case titlebars (for tabs/stacks) are always present, which would prevent the bar from seamlessly blending in with the visible window anyways.

Since this changes established behavior and might break some configs, a more prudent alternative approach would be to add another configuration option to be used in conjunction with `offscreen-css` to enable this behavior.

### DIFF
--- a/src/modules/sway/window.cpp
+++ b/src/modules/sway/window.cpp
@@ -251,6 +251,39 @@ std::pair<int, int> leafNodesInWorkspace(const Json::Value& node) {
   return {sum, floating_sum};
 }
 
+std::optional<std::reference_wrapper<const Json::Value>> getSingleChildNode(const Json::Value& node) {
+  auto const& nodes = node["nodes"];
+  if (nodes.empty()) {
+    if (node["type"].asString() == "workspace")
+      return {};
+    else if (node["type"].asString() == "floating_con") {
+      return {};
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #2241, opened 2023-06-17
support multiple items in hwmon-path of temperature module

yangyingchao:
So user can share configuration file among different machines with different hardware configurations.


Alexays:
Thanks!

Issue #2211, opened 2023-06-05
[Feature Request] Workspace styles for Hyprland


donovanglover:
Similar to the [Sway module](https://github.com/Alexays/Waybar/wiki/Module:-Sway#style-1), there should be custom styles for [Hyprland workspaces](https://github.com/Alexays/Waybar/wiki/Module:-Workspaces).

Personally, I would love to change the style of the active workspace if it has a fakefullscreen window.

For those that need this feature *now*, you can use Hyprland's [`renameworkspace` dispatcher](https://wiki.hyprland.org/Configuring/Dispatchers/#list-of-dispatchers) with a custom script.

gardenappl:
I'm working on a PR that implements customization similar to Sway, but I'm not sure what the use-case is for `fakefullscreen`. Should it display a custom style if the focused window is `fakefullscreen` (I can kinda imagine this being useful), or should it do it if there's just one such window anywhere on the workspace (why?).

donovanglover:
I use `fakefullscreen` to temporarily show windows in fullscreen while still having access to waybar. Sometimes I forget that there's another window on the same workspace, so an indicator would help me remember that the active window is fullscreen.

Currently I make fullscreen windows a different color with [`hyprland-autoname-workspaces`](https://github.com/hyprland-community/hyprland-autoname-workspaces) and [HTML tags](https://github.com/donovanglover/nix-config/blob/209fb4d9cad8ff52da0187e45673683413495224/home/hyprland-autoname-workspaces.nix#L48), although a waybar solution would be nice as well.

gardenappl:
Interesting, I thought fakefullscreen didn't cover other windows.

I might implement this but only checking the current focused window.

donovanglover:
You're right, I was confusing `fakefullscreen` with `fullscreen, 1`, which "maximizes and keeps gaps and bars". Sorry for the confusion.

And I agree, checking fullscreen for the current focused window would be great.

gardenappl:
Oh that clears it up, thanks.

gardenappl:
The PR was merged and I updated the wiki. (but right now you have to use the -git version to see changes)

donovanglover:
It works great thank you! Very cool that the styles can apply to the entire bar. Lots of possibilities when it comes to customization.

PR #2245, opened 2023-06-19
hyprland/window: Add CSS customization similar to sway/window

gardenappl:
This PR reworks the `hyprland/window` module a bit, to allow the following CSS classes:

- `window#waybar.empty` When no windows are in the active workspace (or monitor's workspace if `separate-outputs` is enabled)
- `window#waybar.solo` When one tiled window is in the workspace
- `window#waybar.<app_id>` Where `<app_id>` is the class name (e.g. `chromium`) of the only window in the workspace

These are similar to the Sway module, except without the layout modes, because Hyprland's layout modes are different (and afaik it doesn't expose them to IPC anyway)

The "rework" was done because Hyprland IPC doesn't have one clear event for when the amount of windows on a workspace changes, so we just listen to `activewindow`, `closewindow` and `movewindow` and query the current workspace every time. I did not extensively test this but this seems like it should cover most use cases.

Also I migrated to using IPC for everything instead of executing `hyprctl`.

### DIFF
--- a/include/modules/hyprland/window.hpp
+++ b/include/modules/hyprland/window.hpp
@@ -1,7 +1,5 @@
 #include <fmt/format.h>
 
-#include <tuple>
-
 #include "ALabel.hpp"
 #include "bar.hpp"
 #include "modules/hyprland/backend.hpp"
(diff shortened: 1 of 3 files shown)


gardenappl:
Also added `floating` class (identical to Sway module) and `fullscreen` class.

The fullscreen class is useful since Hyprland allows to put a window into fullscreen without hiding the bar. This obscures other windows in the same workspace, so being able to indicate this via Waybar would be useful (see #2211 )

Alexays:
LGTM thx!

primalmotion:
Sorry to necrobump but how should I handle classes containing dots (like most flatpaks)? For instance 

`window#waybar:not(.floating).firefox{}` works as intended but not `window#waybar:not(.floating).de.haeckerfelix.Shortwave {}`

gardenappl:
I haven't thought about this since I just copied the idea from the Sway module. Got curious and a quick search for "css class containing dot" shows a StackOverflow answer, saying you can escape the dot with a backslash. Just tested this and it works:

```
window#waybar.org\.telegram\.desktop {
    color: #383a42;
    background-color: #000000;
}
```
By the way you shouldn't need the `:not(.floating)` selector, the app_id thing works only for tiled windows.

PR #2247, opened 2023-06-20
More robust Hyprland backend

gardenappl:
Basically I was testing #2245 and noticed that Waybar would randomly crash when there are too many windows in a workspace. Turns out the current IPC implementation only reads 8192 bytes of response and discards the rest. This PR is a quick fix.


Alexays:
Thanks!

PR #2248, opened 2023-06-21
Add Auto-Hide feature

varbhat:
@Alexays ,
I implemented Auto-Hide feature in Waybar.  It allows Bar to autohide itself when cursor is no longer over it and makes itself visible when the cursor touches the edge of the screen to which Bar is positioned. It's incredibly useful in Small screen displays.  It's working properly for me. I wish to have this merged. It adds three more configuration options to the config.

```
*autohide* ++
	typeof: bool ++
	default: false ++
	Option to specify whether bar must autohide.
	Use *autohide-delay* and *autohide-starthidden* to customize autohide.

*autohide-delay* ++
	typeof: integer ++
	default: 0 ++
	Time in milliseconds after which the bar must autohide if cursor is no longer over the bar.

*autohide-starthidden* ++
	typeof: bool ++
	default: false ++
	Option to specify whether the bar must start in hidden state if *autohide* option is true.
```

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -10,6 +10,8 @@
 
 #include <memory>
 #include <vector>
+#include <atomic>
+#include <mutex>
 
 #include "AModule.hpp"
 #include "xdg-output-unstable-v1-client-protocol.h"
@@ -83,6 +85,7 @@ class Bar {
   void setVisible(bool visible);
   void toggle();
   void handleSignal(int);
+  void setupAutohide();
 
   struct waybar_output *output;
   Json::Value config;
@@ -90,6 +93,7 @@ class Bar {
   bool visible = true;
   bool vertical = false;
   Gtk::Window window;
+  Gtk::Window hotspotWindow;
 
 #ifdef HAVE_SWAY
   std::string bar_id;
@@ -102,12 +106,16 @@ class Bar {
   void setupAltFormatKeyForModule(const std::string &module_name);
   void setupAltFormatKeyForModuleList(const char *module_list_name);
   void setMode(const bar_mode &);
+  static void showMainbar(GtkWidget * widget, const gchar* h,  gpointer data);
+  static void hideMainbar(GtkWidget * widget, const gchar* h, gpointer data);
+  static void hideMainbarCallback(gpointer data);
 
   /* Copy initial set of modes to allow customization */
   bar_mode_map configured_modes = PRESET_MODES;
   std::string last_mode_{MODE_DEFAULT};
 
   std::unique_ptr<BarSurface> surface_impl_;
+  std::unique_ptr<BarSurface> hotspotsurface_impl_;
   Gtk::Box left_;
   Gtk::Box center_;
   Gtk::Box right_;
@@ -120,6 +128,10 @@ class Bar {
   std::unique_ptr<BarIpcClient> _ipc_client;
 #endif
   std::vector<std::shared_ptr<waybar::AModule>> modules_all_;
+
+  std::atomic<unsigned int> autohide_delay_ms = 0;
+  std::mutex autohide_mutex;
+  std::atomic<guint> autohide_timeout_handle;
 };
 
 }  // namespace waybar
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -106,9 +106,9 @@ class Bar {
   void setupAltFormatKeyForModule(const std::string &module_name);
   void setupAltFormatKeyForModuleList(const char *module_list_name);
   void setMode(const bar_mode &);
-  static void showMainbar(GtkWidget * widget, const gchar* h,  gpointer data);
-  static void hideMainbar(GtkWidget * widget, const gchar* h, gpointer data);
-  static void hideMainbarCallback(gpointer data);
+  void showMainbar(GdkEventCrossing* ev);
+  void hideMainbar(GdkEventCrossing* ev);
+  bool hideMainbarCallback();
 
   /* Copy initial set of modes to allow customization */
   bar_mode_map configured_modes = PRESET_MODES;
@@ -129,9 +129,8 @@ class Bar {
 #endif
   std::vector<std::shared_ptr<waybar::AModule>> modules_all_;
 
-  std::atomic<unsigned int> autohide_delay_ms = 0;
-  std::mutex autohide_mutex;
-  std::atomic<guint> autohide_timeout_handle;
+  unsigned int autohide_delay_ms = 0;
+  sigc::connection autohide_connection;
 };
 
 }  // namespace waybar
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -10,8 +10,6 @@
 
 #include <memory>
 #include <vector>
-#include <atomic>
-#include <mutex>
 
 #include "AModule.hpp"
 #include "xdg-output-unstable-v1-client-protocol.h"
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -481,6 +481,7 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
     : output(w_output),
       config(w_config),
       window{Gtk::WindowType::WINDOW_TOPLEVEL},
+      hotspotWindow{Gtk::WindowType::WINDOW_TOPLEVEL},
       left_(Gtk::ORIENTATION_HORIZONTAL, 0),
       center_(Gtk::ORIENTATION_HORIZONTAL, 0),
       right_(Gtk::ORIENTATION_HORIZONTAL, 0),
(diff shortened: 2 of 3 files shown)


alebastr:
Have you considered cropping the surface with [wp_viewporter](https://wayland.app/protocols/viewporter) instead of introducing an additional hotspot surface?

include/bar.hpp
```diff
@@ -120,6 +128,10 @@ class Bar {
   std::unique_ptr<BarIpcClient> _ipc_client;
 #endif
   std::vector<std::shared_ptr<waybar::AModule>> modules_all_;
+
+  std::atomic<unsigned int> autohide_delay_ms = 0;
+  std::mutex autohide_mutex;
```
alebastr, line 133: All the UI rendering happens in a single thread. Why do you need this lock?

src/bar.cpp
```diff
@@ -579,6 +587,13 @@ waybar::Bar::Bar(struct waybar_output* w_output, const Json::Value& w_config)
   // GTK layer shell anchors logic relying on the dimensions of the bar.
   surface_impl_->setPosition(position);
 
+  hotspotsurface_impl_->setMargins(margins_);
```
alebastr, line 590: Note that the bar doesn't have to be adjacent to the edge of the screen. Non-zero margin on an anchored edge may break your hotspot logic.

src/bar.cpp
```diff
@@ -683,6 +710,59 @@ void waybar::Bar::setVisible(bool value) {
...
+    curbar->autohide_timeout_handle.store(g_timeout_add_once(curbar->autohide_delay_ms.load(),GSourceOnceFunc(hideMainbarCallback),curbar));
+  }
+  curbar->autohide_mutex.unlock();
+}
+
+void waybar::Bar::setupAutohide() {
+  this->autohide_timeout_handle.store(0);
+  this->autohide_delay_ms.store(config["autohide-delay"].isUInt() ? config["autohide-delay"].asUInt() : 0);
+
+  gtk_layer_set_layer(hotspotWindow.gobj(), GTK_LAYER_SHELL_LAYER_OVERLAY);
```
alebastr, line 760: gtk-layer-shell is an optional dependency, and Waybar can be compiled without it. Even if it doesn't make much sense nowadays.

varbhat:
@alebastr , I have not used that protocol. But, some alternative bars have followed the additional hotspot surface method to implement auto-hide. Is there any disadvantage of using additional hotspot surface? For me, it's working properly.

src/bar.cpp
```diff
@@ -683,6 +710,59 @@ void waybar::Bar::setVisible(bool value) {
...
+
+auto waybar::Bar::hideMainbar(GtkWidget* widget, const char* h, gpointer user_data) -> void {
+  auto curbar = static_cast<Bar*>(user_data);
+  curbar->autohide_mutex.lock();
+  auto tag = curbar->autohide_timeout_handle.load();
+  if(tag !=0 && g_source_remove(tag)) {
+    spdlog::debug("removed timeout handle in hideMainbar: {}",tag);
+    curbar->autohide_timeout_handle.store(0);
+  } else {
+    curbar->autohide_timeout_handle.store(g_timeout_add_once(curbar->autohide_delay_ms.load(),GSourceOnceFunc(hideMainbarCallback),curbar));
```
alebastr, line 751: This way of setting the timeout is unaware of the object lifetimes and may trigger the callback after the bar instance is destroyed. Please, use glibmm and sigc::mem_fun.

alebastr:
> @alebastr , I have not used that protocol. But, some alternative bars have followed the additional hotspot surface method to implement auto-hide. Is there any disadvantage of using additional hotspot surface? For me, it's working properly. 

Interaction between two exclusive layer surfaces anchored to the same edge of the screen is a compositor-defined behavior. I.e. you shouldn't expect any specific behavior, especially the one you consider correct.

src/bar.cpp
varbhat: So, should i enable autohide when only waybar is compiled with gtk-layer-shell?

src/bar.cpp
varbhat: @alebastr , when waybar is docked right or left and when it doesn't occupy whole screen height, I thought that margins helped in that case. Will correct it.

include/bar.hpp
varbhat: Will remove it

src/bar.cpp
varbhat: @alebastr , I have corrected the codebase. It uses sigc::mem_fun now.

src/bar.cpp
varbhat: @alebastr , have removed it

include/bar.hpp
varbhat: @alebastr , have removed it. Please review current corrected PR.

varbhat:
@alebastr ,Please review now.

src/bar.cpp
varbhat: @alebastr , done.

varbhat:
@alebastr @Alexays , does this PR look good?

alebastr:
Sorry, I'll return to review this weekend. Busy week at work :(

From a brief look: I don't think last commit fixes compilation with
-Dgtk-layer-shell=disabled. Have you tried using setLayer/setExclusiveZone
on hotspot surface instead of direct calls to gtk-layer-shell?

There's also a WIP branch in my fork where I tried to implement hotspot
with a single surface and wp-viewporter. Seems like this could work, but
needs some refactoring of surface abstractions.

donovanglover:
This is pretty cool and allows waybar to be used on-demand with fullscreen applications. I previously tried to toggle waybar with [waycorner](https://github.com/AndreasBackx/waycorner) but the hot corners would end up behind waybar.

One issue I ran into while using your branch with Hyprland is that the bar appears to be re-drawn every time a different module is hovered over. Another issue is that taskbar icons have different padding when autohiding is enabled.

Is there a way to toggle the autohide feature of waybar while it's running? For example, if you need to click something where the bar would show, is it possible to change the bar from `auto-hiding non-exclusive` to `don't autohide + exclusive`?

varbhat:
@alebastr , Did what you said. It can now compile with  gtk-layer-shell disabled. I too experimented with wp_viewporter yesterday but it was segfaulting. I think that wp-viewporter solution brewing in your branch could be better if that works.

varbhat:
@donovanglover ,

Toggling while running -> Can be done. But, i probably will try to implement it later.
re_drawn -> yeah, it flashes for me too. Maybe show_all is redrawing everything? I need to check that.
different padding -> I didn't face this issue. I need to dive more into this.


Can you look into PR and try to improve it? It mustn't be too difficult.

varbhat:
@alebastr any updates? also, what can be done with wp-viewporter?

alebastr:
> @alebastr any updates? also, what can be done with wp-viewporter?

Sorry for delay. I've updated my [vp-viewporter branch](https://github.com/alebastr/Waybar/tree/wp-viewporter) with a working code. Feel free take the first commit as is and use the second one as an example.

The main benefit of viewporter solution, IMO, is that the change applies to the same surface and takes just one Wayland commit. I.e. that's pretty much atomic change with no flickering.

The configuration for fullscreen applications could look like
```json
{
   "mode": "hide", // or "hotspot" to start hidden
   "modes": { // modify builtin modes
     "hide" { "layer": "overlay" },
     "hotspot": { "layer": "overlay", "hotspot": 2 }
   },
   "autohide-timeout": 1000 // ms
}
```

I'm not happy with how that introduces third way to control bar visibility (`SIGUSR1`, Sway IPC and this), but getting all 3 to work together with sane interaction model requires more refactoring than I have time for :(

Also, not entirely convinced that a dedicated hotspot mode is necessary. It's possible to do `"modes": { "invisible": { "hotspot": 1, "layer": "top", "passthrough": false } }` to get the same effect. Or apply the changes to `bar_.configured_modes[Bar::MODE_INVISIBLE]` when `autohide` is set.

bretello:
@alebastr I tested your code and it seems to be working fine. As you mentioned, it's a bit confusing to have this many methods for controlling bar visibility, can you elaborate a bit on what the refactoring you were talking about would look like?

Alexays (approved):

droc12345:
Just tried this on the git version, and it works well.
I'm running hyprland with a top and bottom bar, 
where the top is window title, workspace # and music info
and bottom is app names, sensor info, and date/time

I've set the bottom to auto hide and the top to always stay visible.
Working well, so far.

nightah:
@varbhat and @alebastr was this PR closed because the `wp-viewporter` branch is the preferred approach or just because this wasn't merged?

This patch seems to not cleanly apply to the current HEAD as changes have been made in the interim. Really keen on this feature though keen to understand what we should be doing?

varbhat:
@nightah , Although my approach works, wp-viewporter seems to be the best approach to solve this.

* I got an external monitor and thus i don't need autohide feature (it's desirable on small screens like that of laptop)
* I also moved back to just swaybar (not that waybar is bad; it's great btw)

As i no longer need this feature, I've stopped working on this. Someone can pickup where i left (I consider this to be complete) and get this merged or continue with wp-viewporter approach.

SnaetWarre:
can we reopen this by any chance? would love to see a auto-hide feature implemented

PR #2251, opened 2023-06-24
[wlr/workspaces] Improve persistent_workspaces by allowing fixed nr of workspaces per monitor

### LINKED ISSUE #2243: [Feature request] Automatic persistent workspaces for Hyprland
## The current situation

I'm using [this plugin](https://github.com/Duckonaut/split-monitor-workspaces) to automatically manage persistent workspaces on Hyprland. It sets up keybinds so for example if I press super+3, it will go to the 3rd workspace on the current active monitor. 

This works great with @MonstrousOgre's [persistent_workspaces feature](https://github.com/Alexays/Waybar/pull/2089), where we can tell waybar which workspaces should appear on which monitor.

Say I have 3 monitors, the plugin will create 5 workspaces on every monitor and I have to look at which numbers it assigned to which monitor, so I can update my config.json with those numbers: https://github.com/zjeffer/dotfiles/blob/main/.config/hypr/apps/waybar/config.jsonc#L40

However, the problem is that if you connect a new monitor, waybar won't know which workspaces have to be assigned to it. I have to change my config.json every time a new monitor is connected. If I disconnect a monitor, the plugin might assign different numbers to the monitors that are still connected, so I need to update the config.json again to make sure they match.

## My feature request

I would like:
* Every monitor to show 5 workspaces each
* Connecting or disconnecting monitors should update the waybar workspaces automatically
* In the waybar config, instead of a map `persistent_workspaces`, the ability to specify a fixed number of workspaces for every monitor

## A relatively easy fix

The plugin holds [a map of all monitors and their respective workspaces](https://github.com/Duckonaut/split-monitor-workspaces/blob/main/src/main.cpp#L18), which updates automatically at startup or when a connect or disconnect happens. If we could write a way to get that map from the plugin and pass it to waybar, we could map the workspaces automatically. You wouldn't even need to specify the number of workspaces in the waybar config, waybar could just get the correct amount from the map.

What do you guys think of this idea? Before I look into making changes to the plugin's & waybar's code, is there a better way to achieve what I want?

zjeffer:
WIP, attempts to fix #2243.

This PR improves the `persistent_workspaces` option in `wlr/workspaces`. Currently, for every workspace you want to add, you have to specify an array of monitors where that workspace should show up.

With this PR, you can now also specify an amount of workspaces a monitor should have. It will automatically calculate what the workspace names should be.

Example that will add 3 workspaces for every monitor, but add 5 for the monitor with name "eDP-1":
```json
"persistent_workspaces": {
  "*": 3,
  "eDP-1": 5
}
```

**TODO:**
- [ ] Find a way to get the monitor ID. In Hyprland, every monitor gets a name (string) and an ID (integer).
  -  I'm using [this plugin](https://github.com/Duckonaut/split-monitor-workspaces) to automatically assign workspaces based on the monitor ID
  - Say the amount of workspaces is 5, and the ID is 0, that monitor should have workspaces 1-5. If the ID is 2, that monitor should have workspaces 11-15. 
  - I could get this using hyprctl, but I'm looking for a more general wlroots solution.
- [x] Test with more monitors. I wrote this on a single monitor, will test more when I have access to a 2nd monitor.
- [ ] Add documentation

### DIFF
--- a/src/modules/wlr/workspace_manager.cpp
+++ b/src/modules/wlr/workspace_manager.cpp
@@ -216,7 +216,7 @@ auto WorkspaceGroup::fill_persistent_workspaces() -> void {
     for (const std::string &key : keys) {
       const Json::Value &value = pWorkspaces[key];
 
-      if (value.isNumeric()) {
+      if (value.isNumeric() && value.asInt() > 0) {
         // value == amount of workspaces this workspace should have
         uint32_t monitorId = 0;  // TEMP: how to get monitor ID from output name, that matches Hyprland's ID?
         if ((key == "*" && std::find(keys.begin(), keys.end(), bar_.output->name) == keys.end()) ||
--- a/src/modules/wlr/workspace_manager.cpp
+++ b/src/modules/wlr/workspace_manager.cpp
@@ -12,6 +12,8 @@
 #include "client.hpp"
 #include "gtkmm/widget.h"
 #include "modules/wlr/workspace_manager_binding.hpp"
+#include "modules/hyprland/backend.hpp"
+#include "util/json.hpp"
 
 namespace waybar::modules::wlr {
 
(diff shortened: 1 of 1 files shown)


src/modules/wlr/workspace_manager.cpp
```diff
@@ -210,22 +210,34 @@ WorkspaceGroup::WorkspaceGroup(const Bar &bar, Gtk::Box &box, const Json::Value
...
-        for (const Json::Value &output : p_w) {
+    const Json::Value &pWorkspaces = config_["persistent_workspaces"];
+    const std::vector<std::string> keys = pWorkspaces.getMemberNames();
+
+    for (const std::string &key : keys) {
+      const Json::Value &value = pWorkspaces[key];
+
+      if (value.isNumeric() && value.asInt() > 0) {
+        // value == amount of workspaces this workspace should have
+        uint32_t monitorId = 0;  // TEMP: how to get monitor ID from output name, that matches Hyprland's ID?
```
zjeffer, line 221: I'd like some help with this: Hyprland has monitor IDs I can get using `hyprctl`, but I don't know how to do it with native wlroots commands. Does this even make sense for non-Hyprland compositors?

zjeffer: For now, I implemented it with hyprctl, as an example of how I can get the IDs. I'd love to hear how I can make it work for other wlroots compositors.

zjeffer:
Closing in favour of #2271, I think it's better to have a dedicated module for workspaces on hyprland.

Issue #2214, opened 2023-06-06
wlr/taskbar: hide taskbar when no open windows


portus-zephyr:
here is my bar with open windows:
![Screenshot_2023-06-06-10-39-55](https://github.com/Alexays/Waybar/assets/66728045/4c77b678-85d3-4c07-af6e-0e7e220cc2fe)

here is my bar with no open windows:
 
![Screenshot_2023-06-06-10-40-20](https://github.com/Alexays/Waybar/assets/66728045/74904b81-c67e-4373-b8c7-e30a31b5c5ca)

it would be nice to remove the black line in the second image when no windows are open.

Anakael:
I'll try to fix that

Anakael:
Now you can use "empty" class for taskbar in order to set `background: transparent`

emalsidog:
Hi. Seem like "empty" class does not work

fluid46:
```css
#wlr-taskbar:empty {
  background: transparent;
}
```
returns:
style.css:48:13Invalid name of pseudo-class

PR #2258, opened 2023-06-26
[wrl/taskbare] fix: use empty class for empty taskbar

Anakael:
Resolves: https://github.com/Alexays/Waybar/issues/2214


Alexays:
Thanks!

Issue #1140, opened 2021-06-25
pulseaudio scrolling goes in what feels the wrong direction, negative scroll-step not supported


chris-morgan:
waybar 0.9.7, Arch Linux, “natural scrolling” enabled on my laptop’s touchpad.

Two parts to this:

1. On a pulseaudio widget, swiping two fingers upwards causes volume to go *down*. I expect this action to make it go *up*. (If I disable natural scrolling, it does.)

    I will note that `<input type=number>` in Firefox treats scrolling the same way that waybar-pulseaudio is. I find this just as disconcerting and wrong. I haven’t tried anything else to see if it’s universal or not.

    It’s distinctly possible that this sort of control should be one of the rare things that actually ignores the natural scrolling setting.

2. In order to work around this, I try setting scroll-step to a negative value, which should certainly work. But instead it causes scrolling to just do nothing.

OJFord:
This is (probably - it is in my case) a consequence of having 'natural scroll' enabled, for example in swayconfig:
```
input type:touchpad {
  natural_scroll enabled
}
```

This is unexpected, because in the case of volume the default (`natural_scroll disabled`) behaviour is more, er, _natural_.

JayceFayne:
fixed by https://github.com/Alexays/Waybar/pull/1289

OJFord:
That's just the mentioned work-around by a different method though really, doesn't help when you sometimes use a `natural_scroll` device and sometimes not - probably common among 'clamshell' users, a mouse when docked and a possibly-natural touchpad otherwise.

korylprince:
This has been bugging me, so I've been investigating ways to fix this.

The type of input that is scrolling can be detected from the GTK scroll event.

`GdkEventScroll.get_source_device().input_source` can be checked to see if the input is a mouse, touchpad, or trackpoint. (GTK-wayland code [here](https://github.com/GNOME/gtk/blob/42a704fefbde473e6fecc842482fa7f322abf9fd/gdk/wayland/gdkseat-wayland.c#L2494-L2548))

I propose two solutions to allow intuitive scrolling for mice and touchpads simultaneously:

* Ignore the `"reverse-scrolling"` option when the scroll event is generated by a mouse
  * I can't think of why anyone would actually want "natural scrolling" on a mouse wheel, and this would be an easier change...
* Allow configuring `"reverse-scrolling"` for mice and touchpads separately
[...]

PR #2267, opened 2023-07-01
Don't apply reverse scrolling to mice wheels

korylprince:
This is a fix for #1140.

Currently reverse-scrolling is applied globally. This can be annoying for those who use both a mouse and touchpad with "natural scrolling". In the case of scrolling to adjust volume, brightness, etc, scrolling up on a touchpad feels "natural" to adjust those things up.

However, this is the opposite of what "natural scrolling" does because it was designed for scrolling documents, which _does_ feel natural to reverse up/down.

So what this means for Waybar is that if we leave reverse-scrolling = false, the mouse wheel feels natural, but the touchpad feels backwards. If we set reverse-scrolling = true, the touchpad feels natural, but the mouse wheel feels backwards.

This PR does two things to address this issue:

- Fixes a regression introduced in #2232 that prevents reverse-scrolling from being applied on smooth scroll events (e.g. most touchpad events)
- Disables reverse-scrolling on mouse wheel scroll events, even if reverse-scrolling = true

I can't see any good reason someone would want to apply "natural scrolling" to a mouse wheel. If you are concerned that [someone might](https://xkcd.com/1172/), I could refactor this to have independent `reverse-mouse-scrolling` and `reverse-touchpad-scrolling` config options, keeping the existing `reverse-scrolling` as a global option.

### DIFF
--- a/src/AModule.cpp
+++ b/src/AModule.cpp
@@ -99,11 +99,12 @@ bool AModule::handleToggle(GdkEventButton* const& e) {
 AModule::SCROLL_DIR AModule::getScrollDir(GdkEventScroll* e) {
   // only affects up/down
   bool reverse = config_["reverse-scrolling"].asBool();
+  bool reverse_mouse = config_["reverse-mouse-scrolling"].asBool();
 
   // ignore reverse-scrolling if event comes from a mouse wheel
   GdkDevice* device = gdk_event_get_source_device((GdkEvent *)e);
   if (device != NULL && gdk_device_get_source(device) == GDK_SOURCE_MOUSE) {
-    reverse = false;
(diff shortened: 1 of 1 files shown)


Alexays:
LGTM

Alexays:
Maybe we should add another option to force reverse ?

korylprince:
I added a commit that adds `reverse-mouse-scrolling`, so that reverse scrolling can be independently enabled for mice scroll wheels. Does that work?

I can also update the wiki once approved.

Alexays:
Perfect!

korylprince:
The Backlight and PulseAudio wiki pages have been updated to reflect the changes in this PR.

PR #2268, opened 2023-07-01
Fixes for hyprland/workspaces

zjeffer:
Workspaces were not being deleted from the box


zjeffer:
@Anakael I haven't found the bug yet but this is something else that was bothering me.

Anakael:
> @Anakael I haven't found the reason it segfaults yet but this is something else that was bothering me.

Yes, I found it too.

PR #2269, opened 2023-07-02
wlr/taskbar: add support for rewrite rules

Arcus92:
Hello,

I'm using the _wlr/taskbar_ and I was annoyed by some long application names like "Firefox Web Browser". I would prefer to just read "Firefox", so the button doesn't take so much space. 

I simply implemented the rewrite logic from sway/window and _hyprland/window_ for _wlr/taskbar_. A simple `waybar::util::rewriteString` was added and the man page was updated.

**Here is an example:**
grafik
The top waybar is using the rewrite rules. The bottom waybar is the old version with the full names.

You see, that the taskbar is way shorter by removing unnecessary parts from the application name.

```
        "wlr/taskbar": {
            "format": "{icon} {name}",
            "icon-size": 20,
            "tooltip": false,
            "on-click": "activate",
            "on-click-middle": "close",
            "on-click-right": "minimize-raise",
            "rewrite": {
                "(\\s*)Firefox Web Browser": "$1Firefox",
                "(\\s*)Foot Server": "$1Terminal",
                "(\\s*)CLion.*": "$1CLion",
                "(\\s*)IntelliJ IDEA.*": "$1IDEA"
            }
        },
```
*I had to include the whitespace between the {icon} and {name}.*

Sure, you could simply change the `.desktop` file for these application, but they may reset after an update. I hope this change is useful for some of you.

### DIFF
--- a/man/waybar-wlr-taskbar.5.scd
+++ b/man/waybar-wlr-taskbar.5.scd
@@ -142,9 +142,9 @@ Invalid expressions (e.g., mismatched parentheses) are skipped.
 		"firefoxdeveloperedition": "firefox-developer-edition"
 	},
 	"rewrite": {
-      "Firefox Web Browser": "Firefox",
-      "Foot Server": "Terminal"
-    }
+		"Firefox Web Browser": "Firefox",
+		"Foot Server": "Terminal"
+	}
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Alexays:
Nice!
Thanks!

Alexays:

Arcus92:
@Alexays Done :-)

PR #2276, opened 2023-07-03
Stop MPRIS module from updating every ~20ms

skligys:
I noticed that while MPRIS module is running and music is playing, all tooltips on other modules stop working.  When music is paused or stopped, they work again.

Upon investigation, MPRIS onPlayerMetadata(), onPlayerPlay() callbacks get triggered every ~20ms and call Mpris::update() without regard for update interval configured.  The Mpris label gets updated every time and interferes with showing tooltips for other modules.

The fix is to ignore any Mpris::update() calls until the configured update interval elapses from the previous non-ignored call.


Alexays:
Thanks!

PR #2278, opened 2023-07-03
[hyperland/workspaces] fix: segfault on create/destroy

Anakael:


Anakael:
@zjeffer 
Can you please test it to ensure? :)

Alexays:
I prefer to merge it now.
Thanks!

zjeffer:
Awesome that you found it! I can test it this evening.

zjeffer:
I tested it out, the usual segfault is fixed, but it *very rarely* crashes here: https://github.com/Alexays/Waybar/blob/master/src/modules/hyprland/workspaces.cpp#L107. 

When it doesn't find the workspace (because it already got deleted, getting the id is impossible (workspace->get() is a nullptr) so we should just print the id we get from the function. I'll change it and test it out.

I'm testing this by scrolling incredibly rapidly through hundreds of workspaces a second, using my mouse's scrollwheel that can free scroll (Logitech G502). I don't think this happens through normal use ;)

Anakael:
Great tests! :)

PR #2282, opened 2023-07-04
Fixed build warnings

zjeffer:
Maybe we should force the compilation to fail if there are warnings? :)


Alexays:
Yep we can do it on the CI side at least.
Thanks!

PR #2293, opened 2023-07-08
fix, default to true, sway/workspaces: warp-on-scroll

sigboe:
Sorry about that, @q234rty correctly pointed out that my previous already merged pull request for a new feature did not default to the old behavior. 

While I don't understand why anyone wouldn't use this feature if they already are using mouse_warp, since I can't read out the current status of the mouse_warp option, having this default to being enable disturbs people who don't use mouse_warp also, which I totally see is a preferential subject. So until there is a way to read the current status of mouse_warp, this has to default to the old behavior.

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -327,7 +327,7 @@ bool Workspaces::handleScroll(GdkEventScroll *e) {
       return true;
     }
   }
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Can we have a new release with this fix?

The regression cause by this bug is massive and makes GUI applications near unusable for me.

sigboe:
@Alexays please note whynothugo's comment

PR #2294, opened 2023-07-08
`hyprland/window` expose more data

### LINKED ISSUE #2289: [Feature Request] More data in hyprland/window
I'd like to have more data available in the `hyprland/window` module than just the current title. Especially I'm interested in the `initialTitle` of the window as it has the nicest output most of the time (see `hyprctl clients`) and the title of the window (the only data currently accessible) is often long and doesn't always include the name of the program

Mr-Pine:
This exposes more data as format arguments in the `hyprland/window` module. Most importantly you can now use `"{initialTitle}"` which often shows a nice String describing the program, other than the current state of the program (i.e. for a Firefox window it is "Mozilla Firefox" not "Comparing Alexays:master...Mr-Pine:hyprland-window-data · Alexays/Waybar — Mozilla Firefox")

Fixes #2289

### DIFF
--- a/include/modules/hyprland/window.hpp
+++ b/include/modules/hyprland/window.hpp
@@ -24,8 +24,20 @@ class Window : public waybar::ALabel, public EventHandler {
     static auto parse(const Json::Value&) -> Workspace;
   };
 
+  struct WindowData {
+    bool floating;
+    int monitor;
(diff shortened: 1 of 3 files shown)


Alexays:
Thanks!

PR #2296, opened 2023-07-09
clock: handle timezone changes (again)

m-bdf:
Add `nullptr` to the timezones vector, so that the current timezone is fetched on each `update()`, and changes to it (e.g. with `timedatectl set-timezone`) are propagated.


Alexays:
Thanks!

Issue #1925, opened 2023-01-06
[hyprland/window] Feature request: Application icon


MonstrousOgre:
Sway's window module has the application icon now. Would it be too much work to add it to hyprland too?

Alexays:
Closing — this is fixed/implemented on current master and will be in the next release. Please reopen if it still occurs after updating. 🙏

PR #2303, opened 2023-07-10
App Icons for `hyprland/window`

Mr-Pine:
This PR introduces App Icons for the `hyprland/window` module (as requested in #1925)

Since this required almost no changes from the code for `sway/window` and I didn't want to duplicate the code, I introduces a new `AAppIconLabel` class which now both `hyprland/window` and `sway/window` inherit from. The sway implementation uses `app_class` if `app_id` is not set (appears to be the case if the window is running under xwayland) but I wasn't really able to find a suitable replacement for that with hyprland (`app_id` seems to correspond to hyprlands `class` though). 
Tested on both hyprland and sway

### DIFF
--- a/include/AAppIconLabel.hpp
+++ b/include/AAppIconLabel.hpp
@@ -16,7 +16,8 @@ class AAppIconLabel : public AIconLabel {
   auto update() -> void override;
 
  protected:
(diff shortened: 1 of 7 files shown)


Alexays:
Nice!
Thanks :)

PR #2308, opened 2023-07-12
hyprland/window: rename .hidden to .swallowing (and fix grouped windows)

gardenappl:
Sorry, I know I submitted #2307 just half an hour ago, but I changed my mind. Rather than renaming `.hidden` to some other class name, I have renamed it to `.swallowing`. After some more discussion in the comments to https://github.com/Alexays/Waybar/commit/cdece498c1e3849ea6e8ff7795ddce1ca8265803#commitcomment-121410267 I realized that instead of excluding so many edge cases for it, it's easier to check specifically for window swallowing, as that's the only use case that makes sense and is not covered by other means.


Alexays:
Don't worry, thx!

Issue #2281, opened 2023-07-04
[Error] Installing Waybar with Cava module


wowblvck:
Every time I try to install Waybar with Cava module support I get the following error:

<details><summary>Click to show error</summary>
<p>

FAILED: subprojects/cava-0.8.4/libcava.so 
cc  -o subprojects/cava-0.8.4/libcava.so subprojects/cava-0.8.4/libcava.so.p/src_cavacore.c.o subprojects/cava-0.8.4/libcava.so.p/src_common.c.o subprojects/cava-0.8.4/libcava.so.p/src_config.c.o subprojects/cava-0.8.4/libcava.so.p/src_input_common.c.o subprojects/cava-0.8.4/libcava.so.p/src_input_fifo.c.o subprojects/cava-0.8.4/libcava.so.p/src_input_shmem.c.o subprojects/cava-0.8.4/libcava.so.p/src_output_common.c.o subprojects/cava-0.8.4/libcava.so.p/src_output_terminal_noncurses.c.o subprojects/cava-0.8.4/libcava.so.p/src_output_raw.c.o subprojects/cava-0.8.4/libcava.so.p/src_output_noritake.c.o subprojects/cava-0.8.4/libcava.so.p/src_output_terminal_ncurses.c.o subprojects/cava-0.8.4/libcava.so.p/src_input_alsa.c.o subprojects/
[...]

PR #2316, opened 2023-07-15
hyprland/workspaces improvements

MightyPlaza:
Some improvements to the hyprland/workspaces modules

>adds workspace clicking support (except for the non named special workspace which seems to be an issue on hyprland's side)

>adds "all-ouptuts" and "show-special" to config

>listens to "focusedmon" and "moveworkspace" events

>actually implements "{name}" format which was  present on the wiki but wasn't implemented (opposite of  "{id}")

>fixes issues with previous event implementation on named workspaces

>adds more variables to the workspaces

>sorts workspaces in the order normal -> named -> special -> named special, instead of id

didn't find any other issues/bugs 

will also ping @Anakael @zjeffer in case I accidentally broke something

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -11,7 +11,7 @@ namespace waybar::modules::hyprland {
 
 class Workspace {
  public:
-  Workspace(const Json::Value& value);
+  Workspace(const Json::Value& workspace_data);
   std::string& select_icon(std::map<std::string, std::string>& icons_map);
   Gtk::Button& button() { return button_; };
 
@@ -42,7 +42,7 @@ class Workspace {
 class Workspaces : public AModule, public EventHandler {
  public:
   Workspaces(const std::string&, const waybar::Bar&, const Json::Value&);
-  virtual ~Workspaces();
+  ~Workspaces() override;
   void update() override;
   void init();
 
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -54,7 +54,7 @@ class Workspaces : public AModule, public EventHandler {
  private:
   void onEvent(const std::string&) override;
   void sort_workspaces();
-  void create_workspace(const Json::Value& value);
+  void create_workspace(Json::Value& value);
   void remove_workspace(std::string name);
 
   bool all_outputs_ = false;
@@ -65,7 +65,7 @@ class Workspaces : public AModule, public EventHandler {
   bool with_icon_;
   std::string active_workspace_name;
   std::vector<std::unique_ptr<Workspace>> workspaces_;
-  std::vector<int> workspaces_to_create_;
+  std::vector<Json::Value> workspaces_to_create_;
   std::vector<std::string> workspaces_to_remove_;
   std::mutex mutex_;
   const Bar& bar_;
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -9,27 +9,30 @@
 
 namespace waybar::modules::hyprland {
 
-struct WorkspaceDto {
-  int id;
-
-  static WorkspaceDto parse(const Json::Value& value);
-};
-
 class Workspace {
  public:
-  Workspace(int id);
-  Workspace(WorkspaceDto dto);
-  int id() const { return id_; };
-  int active() const { return active_; };
+  Workspace(const Json::Value& value);
   std::string& select_icon(std::map<std::string, std::string>& icons_map);
-  void set_active(bool value = true) { active_ = value; };
   Gtk::Button& button() { return button_; };
+
+  int id() const { return id_; };
+  std::string name() const { return name_; };
+  std::string monitor() const { return monitor_; };
+  int active() const { return active_; };
+  bool is_special() const { return is_special_; };
+
+  auto handle_clicked(GdkEventButton* bt) -> bool;
+  void set_active(bool value) { active_ = value; };
 
   void update(const std::string& format, const std::string& icon);
 
  private:
   int id_;
+  std::string name_;
+  std::string monitor_;
+  int windows_;
   bool active_;
+  bool is_special_;
 
   Gtk::Button button_;
   Gtk::Box content_;
@@ -43,19 +46,27 @@ class Workspaces : public AModule, public EventHandler {
   void update() override;
   void init();
 
+  auto all_outputs() const -> bool { return all_outputs_; }
+  auto show_special() const -> bool { return show_special_; }
+
+  auto get_bar_output() const -> std::string { return bar_.output->name; }
+
  private:
   void onEvent(const std::string&) override;
   void sort_workspaces();
-  void create_workspace(int id);
-  void remove_workspace(int id);
+  void create_workspace(const Json::Value& value);
+  void remove_workspace(std::string name);
+
+  bool all_outputs_ = false;
+  bool show_special_ = false;
 
   std::string format_;
   std::map<std::string, std::string> icons_map_;
   bool with_icon_;
-  int active_workspace_id;
+  std::string active_workspace_name;
   std::vector<std::unique_ptr<Workspace>> workspaces_;
   std::vector<int> workspaces_to_create_;
-  std::vector<int> workspaces_to_remove_;
+  std::vector<std::string> workspaces_to_remove_;
   std::mutex mutex_;
   const Bar& bar_;
   Gtk::Box box_;
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -17,19 +17,19 @@ class Workspace {
 
   int id() const { return id_; };
   std::string name() const { return name_; };
-  std::string monitor() const { return monitor_; };
+  std::string output() const { return output_; };
   int active() const { return active_; };
   bool is_special() const { return is_special_; };
 
   auto handle_clicked(GdkEventButton* bt) -> bool;
-  void set_active(bool value) { active_ = value; };
+  void set_active(bool value = true) { active_ = value; };
 
   void update(const std::string& format, const std::string& icon);
 
  private:
   int id_;
   std::string name_;
-  std::string monitor_;
+  std::string output_;
   int windows_;
   bool active_;
   bool is_special_;
--- a/man/waybar-hyprland-workspaces.5.scd
+++ b/man/waybar-hyprland-workspaces.5.scd
@@ -50,7 +50,6 @@ Additional to workspace name matching, the following *format-icons* can be set.
 		"active": "",
 		"default": ""
 	},
-	"sort-by-number": true,
 	"all-outputs": false,
         "show-special": false
 }
--- a/man/waybar-hyprland-workspaces.5.scd
+++ b/man/waybar-hyprland-workspaces.5.scd
@@ -21,6 +21,16 @@ Addressed by *hyprland/workspaces*
 	typeof: array ++
 	Based on the workspace id and state, the corresponding icon gets selected. See *icons*.
 
+*show-special*: ++
+	typeof: bool ++
+	default: false ++
+	If set to true special workspaces will be shown.
+
+*all-outputs*: ++
+	typeof: bool ++
+	default: false ++
+	If set to false workspaces group will be shown only in assigned output. Otherwise all workspace groups are shown.
+
 # FORMAT REPLACEMENTS
 
 *{id}*: id of workspace assigned by compositor
--- a/man/waybar-hyprland-workspaces.5.scd
+++ b/man/waybar-hyprland-workspaces.5.scd
@@ -25,6 +25,8 @@ Addressed by *hyprland/workspaces*
 
 *{id}*: id of workspace assigned by compositor
 
+*{name}*: workspace name assigned by compositor
+
 *{icon}*: Icon, as defined in *format-icons*.
 
 # ICONS
@@ -48,7 +50,9 @@ Additional to workspace name matching, the following *format-icons* can be set.
 		"active": "",
 		"default": ""
 	},
-	"sort-by-number": true
+	"sort-by-number": true,
+	"all-outputs": false,
+        "show-special": false
 }
 ```
 
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -102,20 +102,20 @@ void Workspaces::onEvent(const std::string &ev) {
   } else if (eventName == "focusedmon") {
     active_workspace_name = payload.substr(payload.find(",") + 1);
 
-  } else if (eventName == "moveworkspace") {
+  } else if (eventName == "moveworkspace" && !all_outputs()) {
     std::string workspace = payload.substr(0, payload.find(","));
-    std::string new_monitor = payload.substr(payload.find(",") + 1);
-    if (all_outputs()) {
-      if (bar_.output->name == new_monitor) {  // TODO: implement this better
-        const Json::Value workspaces_json = gIPC->getSocket1JsonReply("workspaces");
-        for (auto &workspace_json : workspaces_json) {
-          if (workspace_json["name"].asString() == workspace &&
-              bar_.output->name == workspace_json["monitor"].asString())
-            create_workspace(workspace_json);
+    std::string new_output = payload.substr(payload.find(",") + 1);
+    if (bar_.output->name == new_output) {  // TODO: implement this better
+      const Json::Value workspaces_json = gIPC->getSocket1JsonReply("workspaces");
+      for (auto &workspace_json : workspaces_json) {
+        if (workspace_json["name"].asString() == workspace &&
+            bar_.output->name == workspace_json["monitor"].asString()) {
+          create_workspace(workspace_json);
+          break;
         }
-      } else {
-        workspaces_to_remove_.push_back(workspace);
       }
+    } else {
+      workspaces_to_remove_.push_back(workspace);
     }
   }
 
@@ -166,7 +166,7 @@ Workspaces::~Workspaces() {
 Workspace::Workspace(const Json::Value &value)
     : id_(value["id"].asInt()),
       name_(value["name"].asString()),
-      monitor_(value["monitor"].asString()),  // TODO:allow using monitor desc
+      output_(value["monitor"].asString()),  // TODO:allow using monitor desc
       windows_(value["id"].asInt()) {
   active_ = true;
   is_special_ = false;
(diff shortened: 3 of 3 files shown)


zjeffer:
Great, thanks for this! I'll do some tests with this tomorrow with multiple monitors.

I just compiled and ran it on a single monitor, and I had a segfault after switching to the fourth workspace, when it creates the new workspace button. I can't seem recreate it again, so I'll do some more testing.

include/modules/hyprland/workspaces.hpp
```diff
@@ -9,27 +9,30 @@
...
   Gtk::Button& button() { return button_; };
 
+  int id() const { return id_; };
+  std::string name() const { return name_; };
+  std::string monitor() const { return monitor_; };
+  int active() const { return active_; };
+  bool is_special() const { return is_special_; };
+
+  auto handle_clicked(GdkEventButton* bt) -> bool;
+  void set_active(bool value) { active_ = value; };
```
Anakael, line 25: I can suggest put `value = true` here.

src/modules/hyprland/workspaces.cpp
```diff
@@ -56,55 +68,73 @@ auto Workspaces::update() -> void {
...
   } else if (eventName == "createworkspace") {
-    int new_workspace_id;
-    std::from_chars(payload.data(), payload.data() + payload.size(), new_workspace_id);
-    workspaces_to_create_.push_back(new_workspace_id);
+    const Json::Value workspaces_json = gIPC->getSocket1JsonReply("workspaces");
+    for (auto &workspace_json : workspaces_json) {
+      if (workspace_json["name"].asString() == payload &&
+          (all_outputs() || bar_.output->name == workspace_json["monitor"].asString()) &&
+          (workspace_json["name"].asString().find("special:") != 0 || show_special()))
+        create_workspace(workspace_json);
```
Anakael, line 99: You can't create gtk entities here. Only in `update` method.

Anakael (changes requested):

src/modules/hyprland/workspaces.cpp
zjeffer: This is probably the reason for the segfault

src/modules/hyprland/workspaces.cpp
```diff
@@ -193,7 +249,20 @@ std::string &Workspace::select_icon(std::map<std::string, std::string> &icons_ma
   if (default_icon_it != icons_map.end()) {
     return default_icon_it->second;
   }
-
   return icons_map[""];
 }
+
+auto Workspace::handle_clicked(GdkEventButton *bt) -> bool {
+  if (id() > 0) {  // normal
+    system(("hyprctl dispatch workspace " + std::to_string(id()) + " &> /dev/null").c_str());
```
zjeffer, line 257: Instead of using system(), we should do this through hyprland's IPC.

zjeffer: I made a PR with this change here: https://github.com/MightyPlaza/Waybar/pull/1

src/modules/hyprland/workspaces.cpp
Anakael: Yes, exactly.

man/waybar-hyprland-workspaces.5.scd
```diff
@@ -48,7 +50,8 @@ Additional to workspace name matching, the following *format-icons* can be set.
 		"active": "",
 		"default": ""
 	},
-	"sort-by-number": true
+	"all-outputs": false,
+        "show-special": false
```
Anakael, line 54: Can you please explain in doc what it is?

include/modules/hyprland/workspaces.hpp
```diff
@@ -9,27 +9,30 @@
...
+
   void update(const std::string& format, const std::string& icon);
 
  private:
   int id_;
+  std::string name_;
+  std::string monitor_;
+  int windows_;
```
Anakael, line 33: What it is?

include/modules/hyprland/workspaces.hpp
```diff
@@ -43,19 +46,27 @@ class Workspaces : public AModule, public EventHandler {
   void update() override;
   void init();
 
+  auto all_outputs() const -> bool { return all_outputs_; }
+  auto show_special() const -> bool { return show_special_; }
```
Anakael, line 50: I think it's better to put it in Workspaces. Not individual workspace. Because it's value is same for all workspaces.

include/modules/hyprland/workspaces.hpp
```diff
@@ -43,19 +46,27 @@ class Workspaces : public AModule, public EventHandler {
...
+  void remove_workspace(std::string name);
+
+  bool all_outputs_ = false;
+  bool show_special_ = false;
 
   std::string format_;
   std::map<std::string, std::string> icons_map_;
   bool with_icon_;
-  int active_workspace_id;
+  std::string active_workspace_name;
```
Anakael, line 66: What's the reason for this change and for workspaces to remove?
Comparing integers is faster than strings.

include/modules/hyprland/workspaces.hpp
```diff
@@ -43,19 +46,27 @@ class Workspaces : public AModule, public EventHandler {
...
+
+  auto get_bar_output() const -> std::string { return bar_.output->name; }
+
  private:
   void onEvent(const std::string&) override;
   void sort_workspaces();
-  void create_workspace(int id);
-  void remove_workspace(int id);
+  void create_workspace(const Json::Value& value);
+  void remove_workspace(std::string name);
```
Anakael, line 58: Why change id to name here?

include/modules/hyprland/workspaces.hpp
```diff
@@ -43,19 +46,27 @@ class Workspaces : public AModule, public EventHandler {
...
+  auto show_special() const -> bool { return show_special_; }
+
+  auto get_bar_output() const -> std::string { return bar_.output->name; }
+
  private:
   void onEvent(const std::string&) override;
   void sort_workspaces();
-  void create_workspace(int id);
-  void remove_workspace(int id);
+  void create_workspace(const Json::Value& value);
```
Anakael, line 57: I would prefer leave Dto struct. Json value hints nothing about what's inside it, but Dto does.
And the reason why id was here -- is because it's clearly to see what's going on inside `onEvent` methond and `update` method. Push Json value through methods -- it' a mess. You can compare current solution with sway solution.

src/modules/hyprland/workspaces.cpp
```diff
@@ -56,55 +68,73 @@ auto Workspaces::update() -> void {
     create_workspace(workspace_to_create);
   }
 
-  workspaces_to_create_.clear();
-
-  for (std::unique_ptr<Workspace> &workspace : workspaces_) {
-    workspace->set_active(workspace->id() == active_workspace_id);
-
+  for (auto &workspace : workspaces_) {
+    workspace->set_active(workspace->name() == active_workspace_name);
```
Anakael, line 72: It's better to leave it with integers comparision.

src/modules/hyprland/workspaces.cpp
```diff
@@ -56,55 +68,73 @@ auto Workspaces::update() -> void {
...
+
   } else if (eventName == "createworkspace") {
-    int new_workspace_id;
-    std::from_chars(payload.data(), payload.data() + payload.size(), new_workspace_id);
-    workspaces_to_create_.push_back(new_workspace_id);
+    const Json::Value workspaces_json = gIPC->getSocket1JsonReply("workspaces");
+    for (auto &workspace_json : workspaces_json) {
+      if (workspace_json["name"].asString() == payload &&
+          (all_outputs() || bar_.output->name == workspace_json["monitor"].asString()) &&
+          (workspace_json["name"].asString().find("special:") != 0 || show_special()))
```
Anakael, line 98: Can you please explain what special workspace is?
Didn't found any info in Hyprland wiki.

include/modules/hyprland/workspaces.hpp
```diff
@@ -9,27 +9,30 @@
...
+
+  auto handle_clicked(GdkEventButton* bt) -> bool;
+  void set_active(bool value) { active_ = value; };
+
   void update(const std::string& format, const std::string& icon);
 
  private:
   int id_;
+  std::string name_;
+  std::string monitor_;
```
Anakael, line 32: I can suggest name it `output`, as in wayland ecosystem.

src/modules/hyprland/workspaces.cpp
```diff
@@ -56,55 +68,73 @@ auto Workspaces::update() -> void {
...
+        create_workspace(workspace_json);
+    }
+
+  } else if (eventName == "focusedmon") {
+    active_workspace_name = payload.substr(payload.find(",") + 1);
+
+  } else if (eventName == "moveworkspace") {
+    std::string workspace = payload.substr(0, payload.find(","));
+    std::string new_monitor = payload.substr(payload.find(",") + 1);
+    if (all_outputs()) {
```
Anakael, line 108: You can revert `if` statement and remove one level of indention.

src/modules/hyprland/workspaces.cpp
```diff
@@ -56,55 +68,73 @@ auto Workspaces::update() -> void {
...
+      if (bar_.output->name == new_monitor) {  // TODO: implement this better
```
Anakael, line 109: You can do the same trick with indention here.

src/modules/hyprland/workspaces.cpp
```diff
@@ -136,13 +163,27 @@ Workspaces::~Workspaces() {
...
+  is_special_ = false;
+
+  if (name_.find("name:") == 0) {
+    name_ = name_.substr(5);
+  } else if (name_.find("special") == 0) {
+    name_ = id_ == -99 ? name_ : name_.substr(13);
+    is_special_ = 1;
+  }
+  // spdlog::info("\nNew workspace:\n\tid:({})\n\tname:({})\n\tmonitor:({})\n\tspecial:({})\n", id_,
+  // name_, monitor_, is_special_ ? "yes" : "no"); //make less noisy
```
Anakael, line 181: I think it should be eather removed or changed to debug.

src/modules/hyprland/workspaces.cpp
```diff
@@ -161,14 +202,29 @@ void Workspace::update(const std::string &format, const std::string &icon) {
...
+                if ((a->is_special()) ^ (a->is_special())) {
+                  return a->id() > b->id();
+                } else {
+                  return a->id() < b->id();
+                }
+              }
+              if ((a->id() > 0) ^ (b->id() > 0)) {
+                return a->id() > b->id();
+              }
+              spdlog::error("huh!!?");
```
Anakael, line 226: ?))

Anakael (changes requested):

include/modules/hyprland/workspaces.hpp
MightyPlaza: it is in workspaces

Anakael:
Can you please explain what named, special and named special workspaces are?

include/modules/hyprland/workspaces.hpp
MightyPlaza: what you were getting from ipc was not an int it was a string, which for the normal workspaces is "id", but for named or special it won't work

MightyPlaza: trying to use the use the name and get the id to the work with it would probably be worse

include/modules/hyprland/workspaces.hpp
MightyPlaza: same as above (IPC)

include/modules/hyprland/workspaces.hpp
Anakael: Oh, right. Sorry.

src/modules/hyprland/workspaces.cpp
MightyPlaza: same (IPC gives name not id)

include/modules/hyprland/workspaces.hpp
Anakael: I see.

include/modules/hyprland/workspaces.hpp
MightyPlaza: sure, was using hyprland's name

src/modules/hyprland/workspaces.cpp
MightyPlaza: should never get there
will probably make it unreachable

Anakael:
> Can you please explain what named, special and named special workspaces are?

I've found it in wiki.

MightyPlaza:
there seems to be an issue with dispatcher that makes giving no args use the dispatcher as arg, that's why clicking the non named special workspace doesn't work.
currently trying to fix it and other issues on hyprland's side

src/modules/hyprland/workspaces.cpp
MightyPlaza: used it for debugging at start, probably will remove

man/waybar-hyprland-workspaces.5.scd
MightyPlaza: done

src/modules/hyprland/workspaces.cpp
MightyPlaza: I need the 2nd if for the else

src/modules/hyprland/workspaces.cpp
MightyPlaza: removed

src/modules/hyprland/workspaces.cpp
MightyPlaza: fixed

include/modules/hyprland/workspaces.hpp
MightyPlaza: done

src/modules/hyprland/workspaces.cpp
MightyPlaza: fixed

zjeffer:
This might not be a big problem but I notice the workspace creation is much slower than the wlr/workspaces implementation. Here's a video showing my bar with wlr/workspaces at the top, and your implementation on the bottom:

https://github.com/Alexays/Waybar/assets/4633209/ace456bc-e77d-406c-95e4-b270582538f2

I quickly scroll through the workspaces, and have to wait every once in a while to let the bottom bar update. 

I built the latest master and noticed it doesn't have this problem, although it has another problem where it sometimes doesn't destroy the workspaces, resulting in something like this:

MightyPlaza:
tested spamming the mouse wheel and it seems to update at the same speed as wlr. Also seems to be updating at the same speed as the IPC socket.
the workspaces not getting deleted should be fixed soon, despite not being able to repro

MightyPlaza:
fixed one of the issues for clicking the main special workspace on hyprland's side.
should be in the main branch soon
Edit: also had to change previous waybar code for the intended implementation

src/modules/hyprland/workspaces.cpp
Anakael:
```c++
if (bar_.output->name == new_monitor) {
  workspaces_to_remove_.push_back(workspace);
  continue;
}

// rest code
```

src/modules/hyprland/workspaces.cpp
```diff
@@ -43,68 +53,92 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value
...
+
   } else if (eventName == "createworkspace") {
-    int new_workspace_id;
-    std::from_chars(payload.data(), payload.data() + payload.size(), new_workspace_id);
-    workspaces_to_create_.push_back(new_workspace_id);
+    const Json::Value workspaces_json = gIPC->getSocket1JsonReply("workspaces");
+    for (Json::Value workspace_json : workspaces_json) {
+      if (workspace_json["name"].asString() == payload &&
+          (all_outputs() || bar_.output->name == workspace_json["monitor"].asString()) &&
+          (workspace_json["name"].asString().find("special:") != 0 || show_special())) {
```
Anakael, line 100: Let's put `show_special` check at first in order to skip comparing strings if flag is set to true.

Anakael: And on 155 line too.

src/modules/hyprland/workspaces.cpp
```diff
@@ -136,13 +167,25 @@ Workspaces::~Workspaces() {
...
+      output_(value["monitor"].asString()),  // TODO:allow using monitor desc
+      windows_(value["id"].asInt()) {
+  active_ = true;
+  is_special_ = false;
+
+  if (name_.find("name:") == 0) {
+    name_ = name_.substr(5);
+  } else if (name_.find("special") == 0) {
+    name_ = id_ == -99 ? name_ : name_.substr(13);
+    is_special_ = 1;
```
Anakael, line 182: `is_special = true`? )

Anakael (changes requested):

src/modules/hyprland/workspaces.cpp
zjeffer: I changed this here: https://github.com/MightyPlaza/Waybar/pull/3

src/modules/hyprland/workspaces.cpp
MightyPlaza: not inside a loop, can't use continue

src/modules/hyprland/workspaces.cpp
MightyPlaza: removed it

src/modules/hyprland/workspaces.cpp
MightyPlaza: done

MightyPlaza:
I suggest this gets merged now, since the only issue is the non removed workspaces which only happen on spamming the mouse wheel.
That should be fixed, in my opinion, by using a vector of combined DTO to create/destroy, which will likely be a major rewrite, so will take some time.
What do you guys think?

zjeffer:
> since the only issue is the non removed workspaces which only happen on spamming the mouse wheel.

To clarify: this problem only occured for me on the master branch, not on this PR's branch. On this PR's branch, it's just a little slower when creating & destroying hundreds of workspaces per second. Not a big deal.

I agree it can be merged now.

include/modules/hyprland/workspaces.hpp
MightyPlaza: this and a combined destroy/remove queue is something that should be done, but maybe in the future, since it'll require a rewrite of most of the code.

Anakael (approved):

zjeffer (approved):

zjeffer:
@Alexays  can this be merged? I added persistent workspaces on my own fork and would like to open a separate PR for this.

Alexays:
Yes, LGTM

sudg6:
on-click: activate is not working, nor can I click to switch to any workspace button!

[Edit]
`pacman` manager is still hosting an outdated version of Waybar.
I compiled it under my Arch Linux machine and had an issue with loading the libcanva.so library but manually fixed it as discussed [here](https://github.com/Alexays/Waybar/issues/2281#issuecomment-4818689311).

zjeffer:
> on-click: activate is not working, nor can click to switch to any workspace button!

This has been fixed in https://github.com/Alexays/Waybar/pull/5013, @Alexays just needs to make a new release, or you can compile it yourself.

zjeffer:
@sudg6 your edit is incorrect 

>  pacman manager is still hosting an outdated version of Waybar.

this is not Arch's or pacman's fault. There simply isn't a release yet that includes the fix.

PR #2322, opened 2023-07-16
Use C++20 by default

zjeffer:
Is there a reason we're still on C++17?

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -13,7 +13,7 @@ jobs:
           - fedora
           - opensuse
           - gentoo
-        cpp_std: [c++17]
+        cpp_std: [c++20]
         include:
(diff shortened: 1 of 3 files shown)


LukashonakV:
Hi @zjeffer I'm not sure.. but it might possible there are some OS where c++20 is not supported out of the box... Debian for example. But need to check. + Maybe some CPU architectures different to x86_64 ....

zjeffer:
The CI builds seem to pass, though, [which includes a Debian build](https://github.com/Alexays/Waybar/blob/master/.github/workflows/linux.yml#L12). 

I'll try building it on my Raspberry Pi 4

zjeffer:
Seems to build on RPi4. Some warnings due to an older GCC version but it does build fine.

PR #2323, opened 2023-07-16
Mediaplayer improvements

zjeffer:
* Uses a class to manage the players (PlayerManager)
* Automatically shows the most important player:
  * show the currently playing player
  * or else show the first paused player
  * or else show nothing
  * => If the current active player is closed, the next player in the list (sorted by most recent) will be shown instead.


Alexays:
Thanks!

PR #2326, opened 2023-07-17
hyprland/language: Differentiating keyboard layout variants

czM1K3:
This improves hyprland/language module, so you can specify keyboard variant for formatting.
So you are able to have multiple keyboard variants of same language and differentiate between them, for example, you can define "format-cs-qwerty" and "format-cs-coder".


Alexays:
Thx!

PR #2336, opened 2023-07-20
[sway/workspaces] feat: add high-priority-named optiion

Anakael:
Resolves: comments in https://github.com/Alexays/Waybar/issues/2188

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -28,6 +28,11 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value 
     : AModule(config, "workspaces", id, false, !config["disable-scroll"].asBool()),
       bar_(bar),
       box_(bar.vertical ? Gtk::ORIENTATION_VERTICAL : Gtk::ORIENTATION_HORIZONTAL, 0) {
+  if (config["format-icons"]["high-priority-named"].isArray()) {
+    for (auto &it : config["format-icons"]["high-priority-named"]) {
+      high_priority_named_.push_back(it.asString());
+    }
+  }
   box_.set_name("workspaces");
   if (!id.empty()) {
     box_.get_style_context()->add_class(id);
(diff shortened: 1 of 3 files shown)


src/modules/sway/workspaces.cpp
```diff
@@ -279,9 +284,24 @@ Gtk::Button &Workspaces::addButton(const Json::Value &node) {
 }
 
 std::string Workspaces::getIcon(const std::string &name, const Json::Value &node) {
-  std::vector<std::string> keys = {"urgent", "focused", name, "visible", "default"};
```
Alexays, line 282: Normal to remove visible?

src/modules/sway/workspaces.cpp
```diff
@@ -279,9 +284,24 @@ Gtk::Button &Workspaces::addButton(const Json::Value &node) {
...
+
+      it = std::find_if(high_priority_named_.begin(), high_priority_named_.end(),
+                        [&](const std::string &member) {
+                          return trimWorkspaceName(member) == trimWorkspaceName(name);
+                        });
+      if (it != high_priority_named_.end()) {
+        return config_["format-icons"][trimWorkspaceName(name)].asString();
+      }
+    }
+    if (key == "focused" || key == "urgent") {
```
Alexays, line 304: Same here

Anakael:
There is no documentation about that key neather in .scd or in wiki. So it's literally undocumented option.And it seems to overlap with `default` key.  21.07.2023, 10:41, "Alexis Rouillard" ***@***.***>:
@Alexays commented on this pull request.



In src/modules/sway/workspaces.cpp:
> @@ -279,9 +284,24 @@ Gtk::Button &Workspaces::addButton(const Json::Value &node) {
 }
 
 std::string Workspaces::getIcon(const std::string &name, const Json::Value &node) {
-  std::vector<std::string> keys = {"urgent", "focused", name, "visible", "default"};

Normal to remove visible?

src/modules/sway/workspaces.cpp
Anakael: There is no documentation about that key neather in .scd or wiki. So it's literally undocumented option.
And it seems to overlap with `default` key.

Alexays:
Thanks!

Issue #2331, opened 2023-07-19
Calendar shows wrong week numbers


J0nnyMak0:
I'm using the default calendar format  from `man waybar-clock` and I think the week numbers are wrong. Unless I'm missing something, the week number of the last week of the month should repeat at the first week of the following month (unless the first day of the month is a Sunday). However, currently the week number of the first and second weeks of the month repeat (if month doesn't start on a Sunday) and every month gets a new first week number.

```
      "calendar": {
                  "mode"          : "year",
                  "mode-mon-col"  : 3,
                  "weeks-pos"     : "right",
                  "on-scroll"     : 1,
                  "on-click-right": "mode",
                  "format": {
                              "months":     "<span color='#ffead3'><b>{}</b></span>",
                              "days":       "<span color='#ecc6d9'><b>{}</b></span>",
                              "weeks":      "<span color='#99ffdd'><b>W{}</b></span>",
                              "weekdays":   "<span color='#ffcc66'><b>{}</b></span>",
                              "today":      "<span color='#ff6699'><b><u>{}</u></b></span>"
                              }
            }
```

![waycal](https://github.com/Alexays/Waybar/assets/6559121/2ff7e996-b758-40e6-abf7-05e4b6fc7bfe)

LukashonakV:
Hi @J0nnyMak0 , can you please share additional info:

1. `date '+%c'`
2. `echo $LANG`
3. `waybar --version`
4. `uname -a`
5. Please provide here the whole block for `clock` module

J0nnyMak0:
> Hi @J0nnyMak0 , can you please share additional info:
> 
> 1. `date '+%c'`

Wed 19 Jul 2023 10:13:02 AM CDT

> 2. `echo $LANG`

en_US.UTF-8

> 3. `waybar --version`

Waybar v0.9.19-39-g1e2b9cb5 (branch 'master')

> 4. `uname -a`

Linux jonmhome 6.4.3-arch1-2 #1 SMP PREEMPT_DYNAMIC Sat, 15 Jul 2023 19:25:49 +0000 x86_64 GNU/Linux

> 5. Please provide here the whole block for `clock` module

```
        "clock": {
                "timezone": "America/Chicago",
                "format": "{:%a}",
                "tooltip-format": "<tt><small>{calendar}</small></tt>",
                "calendar": {
                  "mode"          : "year",
                  "mode-mon-col"  : 3,
                  "weeks-pos"     : "right",
[...]

Issue #2254, opened 2023-06-26
Calendar tooltip renders invalid month on first scroll, and highlights invalid day in year mode


ovk:
Clock module configuration:

```
    "clock": {
        "tooltip-format": "<tt>{calendar}</tt>",
        "calendar": {
            "mode": "month",
            "mode-mon-col": 3,
            "on-scroll": 1,
            "format": {
                "months":   "<span color='#d8dee9'><b>{}</b></span>",
                "days":     "<span><b>{}</b></span>",
                "weekdays": "<span color='#ebcb8b'><b>{}</b></span>",
                "today":    "<span color='#ebcb8b'><b><u>{}</u></b></span>"
            }
        },
        "actions":  {
            "on-click-right": "mode",
            "on-scroll-up": "shift_down",
            "on-scroll-down": "shift_up"
        }
    }

```

When tooltip with the calendar first appears, it looks correct and the "today" day highlighted correctly (doing this test on June 25):
![20230625_20h26m51s_grim](https://github.com/Alexays/Waybar/assets/693072/33e0f86d-3162-431a-9e19-3891c48c279e)

However, after first mouse wheel scroll the following appears, which is clearly incorrect:
![20230625_20h31m36s_grim](https://github.com/Alexays/Waybar/assets/693072/167b58a9-6c7f-4ec1-8a3d-b7130e841b05)

Surprisingly, the next scroll finally shows July (which is correct):
![20230625_20h31m40s_grim](https://github.com/Alexays/Waybar/assets/693072/77349044-242f-431e-ae78-4a6bb6dd8400)

The issue also appears in "year" mode in slightly different (and very confusing) way: if I switch mode after scrolling for one time , and then scroll back and forth in "year" mode, it will highlight invalid "today" date.

LukashonakV:
Hi , @ovk . Is it reproducible or happened once?

ovk:
It's fully reproducible on two different machines. I'm also wiling to troubleshoot it if needed.

LukashonakV:
Actually, I'm on vacation.  So, I would be able to check this issue in 2 weeks.

ovk:
@LukashonakV looks like #2340 does indeed fix the issue. Switching from 0.9.19 and main branch I can see this issue disappear.

LukashonakV:
Great, @ovk. If there are no more issues so far, please close it. Thank you

PR #2340, opened 2023-07-23
time conversion between time zones

LukashonakV:
Hi @Alexays , this PR should fix 

1. #2331
2. probably (#2254) .  This need to check and ask for issue author to provide conclusion

The problem was here: when calendar used to print information through fmt::format it used sys_days function which doesn't care about timezones and calculates date time in UTC. 
For now: the whole clock is moved to zoned logic. Even user doesn't configure his local timezone, system time zone will be used to provide info about date, time, calendar.  When user configures array of timezones and switches between each other calendar will also been recalculated with applying of the picked timezone.

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -24,8 +24,8 @@ waybar::modules::Clock::Clock(const std::string& id, const Json::Value& config)
     for (const auto& zone_name : config_["timezones"]) {
       if (!zone_name.isString()) continue;
       if (zone_name.asString().empty())
-        // nullptr means that local time should be shown
-        time_zones_.push_back(nullptr);
+        // local time should be shown
+        time_zones_.push_back(date::current_zone());
       else
         try {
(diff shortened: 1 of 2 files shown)


src/modules/clock.cpp
```diff
@@ -48,20 +47,22 @@ waybar::modules::Clock::Clock(const std::string& id, const Json::Value& config)
   // If all timezones are parsed and no one is good
   if (!time_zones_.size()) {
     // nullptr means that local time should be shown
```
Alexays, line 49: remove comment

Alexays (approved):

src/modules/clock.cpp
LukashonakV: Done + fixed one small lost logic )

Alexays:
Thanks!

m-bdf:
This reverts #2296, is this intentional?

PR #2341, opened 2023-07-24
Persistent workspaces in hyprland/workspaces

zjeffer:
Adds `persistent_workspaces` to the hyprland/workspaces module, based on the implementation from wlr/workspaces.

Apart from what's possible with wlr/workspaces (specifying an array of monitors a workspace should be shown on), you can also specify a fixed number of workspaces each monitor should have, like this:

```jsonc
"persistent_workspaces": {
  "*": 3,    // 3 workspaces by default
  "DP-1": 5 // but DP-1 gets 5 workspaces
}
```

CC: @MonstrousOgre @Anakael @MightyPlaza

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -55,6 +55,9 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value 
   gIPC->registerForIPC("destroyworkspace", this);
   gIPC->registerForIPC("focusedmon", this);
   gIPC->registerForIPC("moveworkspace", this);
+  gIPC->registerForIPC("openwindow", this);
+  gIPC->registerForIPC("closewindow", this);
+  gIPC->registerForIPC("movewindow", this);
 }
 
 auto Workspaces::update() -> void {
@@ -71,7 +74,7 @@ auto Workspaces::update() -> void {
   workspaces_to_create_.clear();
 
   for (auto &workspace : workspaces_) {
-    workspace->set_active(workspace->name() == active_workspace_name);
+    workspace->set_active(workspace->name() == active_workspace_name_);
     std::string &workspace_icon = icons_map_[""];
     if (with_icon_) {
       workspace_icon = workspace->select_icon(icons_map_);
@@ -87,7 +90,7 @@ void Workspaces::onEvent(const std::string &ev) {
   std::string payload = ev.substr(eventName.size() + 2);
 
   if (eventName == "workspace") {
-    active_workspace_name = payload;
+    active_workspace_name_ = payload;
 
   } else if (eventName == "destroyworkspace") {
     workspaces_to_remove_.push_back(payload);
@@ -97,14 +100,14 @@ void Workspaces::onEvent(const std::string &ev) {
     for (Json::Value workspace_json : workspaces_json) {
       if (workspace_json["name"].asString() == payload &&
           (all_outputs() || bar_.output->name == workspace_json["monitor"].asString()) &&
-          (show_special() || workspace_json["name"].asString().find("special") != 0)) {
+          (show_special() || workspace_json["name"].asString().starts_with("special"))) {
         workspaces_to_create_.push_back(workspace_json);
         break;
       }
     }
 
   } else if (eventName == "focusedmon") {
-    active_workspace_name = payload.substr(payload.find(',') + 1);
+    active_workspace_name_ = payload.substr(payload.find(',') + 1);
 
   } else if (eventName == "moveworkspace" && !all_outputs()) {
     std::string workspace = payload.substr(0, payload.find(','));
@@ -121,13 +124,47 @@ void Workspaces::onEvent(const std::string &ev) {
     } else {
       workspaces_to_remove_.push_back(workspace);
     }
+  } else if (eventName == "openwindow" || eventName == "closewindow" || eventName == "movewindow") {
+    update_window_count();
   }
 
   dp.emit();
 }
 
+void Workspaces::update_window_count() {
+  const Json::Value workspaces_json = gIPC->getSocket1JsonReply("workspaces");
+  for (auto &workspace : workspaces_) {
+    auto workspace_json = std::find_if(
+        workspaces_json.begin(), workspaces_json.end(),
+        [&](Json::Value const &x) { return x["name"].asString() == workspace->name(); });
+    if (workspace_json != workspaces_json.end()) {
+      try {
+        workspace->set_windows((*workspace_json)["windows"].asUInt());
+      } catch (const std::exception &e) {
+        spdlog::error("Failed to update window count: {}", e.what());
+      }
+    } else {
+      workspace->set_windows(0);
+    }
+  }
+}
+
 void Workspaces::create_workspace(Json::Value &value) {
-  workspaces_.push_back(std::make_unique<Workspace>(value));
+  // replace the existing persistent workspace if it exists
+  auto workspace = std::find_if(
+      workspaces_.begin(), workspaces_.end(), [&](std::unique_ptr<Workspace> const &x) {
+        auto name = value["name"].asString();
+        return x->is_persistent() &&
+               ((name.starts_with("special:") && name.substr(8) == x->name()) || name == x->name());
+      });
+  if (workspace != workspaces_.end()) {
+    // replace workspace, but keep persistent flag
+    workspaces_.erase(workspace);
+    value["persistent"] = true;
+  }
+
+  // create new workspace
+  workspaces_.emplace_back(std::make_unique<Workspace>(value));
   Gtk::Button &new_workspace_button = workspaces_.back()->button();
   box_.pack_start(new_workspace_button, false, false);
   sort_workspaces();
@@ -139,6 +176,12 @@ void Workspaces::remove_workspace(std::string name) {
                                 [&](std::unique_ptr<Workspace> &x) { return x->name() == name; });
 
   if (workspace == workspaces_.end()) {
+    // happens when a workspace on another monitor is destroyed
+    return;
+  }
+
+  if ((*workspace)->is_persistent()) {
+    // don't remove persistent workspaces, create_workspace will take care of replacement
     return;
   }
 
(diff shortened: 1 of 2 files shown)


MonstrousOgre:
This is great! Once this is merged, I can start using this over wlr/workspaces

MightyPlaza:
I was not using window_, so I just created the var and parsed it on start.
We need to listen to windowCreate and windowDestroy (or more) to keep track of window count

MightyPlaza:
It's possible to listen to (openwindow, movewindow, closewindow) but you'd have to keep track of which window are where for it to work.

zjeffer:
> Seems to work pretty well, though.

Nevermind, still a small bug when moving a window to another workspace: the workspace is not seen as empty until I go back to the empty workspace.

For example, here I move a window from workspace 1 to 4, one by one:

https://github.com/Alexays/Waybar/assets/4633209/51be6e13-2073-4315-97d3-2cb04aff3f33

A white background means the workspace is not empty, grey means persistent and empty.

zjeffer:
Ok I think I fixed it for real this time: now it will update the window counts on `openwindow`, `closewindow`, and `movewindow`, and set it to 0 when the json received from hyprland's IPC does not include the workspace (meaning it is a persistent workspace).

Ready for review.

MonstrousOgre:
What about listening for `destroyworkspace` to know if it's empty? Don't empty workspaces get destroyed when not focused?

src/modules/hyprland/workspaces.cpp
```diff
@@ -139,15 +172,90 @@ void Workspaces::remove_workspace(std::string name) {
                                 [&](std::unique_ptr<Workspace> &x) { return x->name() == name; });
 
   if (workspace == workspaces_.end()) {
+    // happens when a workspace on another monitor is destroyed
+    return;
+  }
+
+  if ((*workspace)->is_persistent()) {
     return;
```
zjeffer, line 180: @MonstrousOgre If I just add `(*workspace)->set_windows(0);` here (which gets called when a persistent workspace gets a destroyworkspace event), the whole thing works without listening to the open/close/move-window events. The only problem is the windows_ variable does not show the correct amount: it always shows 0 or 1. Adding more windows does not update the window count.

I can imagine in the future, someone would want to make use of the windows_ variable for something other than knowing whether the workspace is empty or not. With this PR's current implementation, the windows_ variable shows the actual window count.

MonstrousOgre: That makes sense

src/modules/hyprland/workspaces.cpp
```diff
@@ -139,15 +172,90 @@ void Workspaces::remove_workspace(std::string name) {
...
+      }
+    }
+  }
+}
+
+void Workspaces::create_persistent_workspaces() {
+  for (const std::string &workspace_name : persistent_workspaces_to_create_) {
+    Json::Value new_workspace;
+    try {
+      new_workspace["id"] = workspace_name.starts_with("special") ? -99 : std::stoi(workspace_name);
```
MightyPlaza, line 228: don't use id
this won't work for named workspaces or named special workspaces

MightyPlaza (changes requested):

src/modules/hyprland/workspaces.cpp
```diff
@@ -202,6 +316,7 @@ void add_or_remove_class(const Glib::RefPtr<Gtk::StyleContext> &context, bool co
 void Workspace::update(const std::string &format, const std::string &icon) {
   auto style_context = button_.get_style_context();
   add_or_remove_class(style_context, active(), "active");
+  add_or_remove_class(style_context, is_empty(), "persistent");
 
   label_.set_markup(fmt::format(fmt::runtime(format), fmt::arg("id", id()),
```
MightyPlaza, line 321: persistent workspaces makes implementing label id impossible, so this should be removed and doc updated

MightyPlaza (changes requested):

src/modules/hyprland/workspaces.cpp
```diff
@@ -284,10 +284,10 @@ Workspace::Workspace(const Json::Value &workspace_data)
       windows_(workspace_data["windows"].asInt()),
       active_(true),
       is_special_(false) {
-  if (name_.find("name:") == 0) {
+  if (name_.starts_with("name:")) {
     name_ = name_.substr(5);
-  } else if (name_.find("special") == 0) {
-    name_ = id_ == -99 ? name_ : name_.substr(13);
```
zjeffer, line 180: @MightyPlaza why was this substr(13) and not 8?

src/modules/hyprland/workspaces.cpp
zjeffer: Do you mean just remove the `fmt::arg("id", id())` from this line?

MightyPlaza: yes

src/modules/hyprland/workspaces.cpp
MightyPlaza: yes it should (maybe I was thinking "special:name:<name_here>" but that isn't what happens)

[special.txt](https://github.com/Alexays/Waybar/files/12177873/special.txt)
this diff fixes it and issues I found in the the on_click function (no idea how I didn't notice this before)

src/modules/hyprland/workspaces.cpp
```diff
@@ -139,23 +172,95 @@ void Workspaces::remove_workspace(std::string name) {
...
+        }
+      }
+    }
+  }
+}
+
+void Workspaces::create_persistent_workspaces() {
+  for (const std::string &workspace_name : persistent_workspaces_to_create_) {
+    Json::Value new_workspace;
+    new_workspace["id"] = workspace_name.starts_with("special") ? -99 : 0;
```
MightyPlaza, line 227: only give id as 0 if name is not int, otherwise use name as id

also workspace_name == "special" since named special workspaces have the same issue as the normal named workspaces

src/modules/hyprland/workspaces.cpp
```diff
@@ -139,23 +172,95 @@ void Workspaces::remove_workspace(std::string name) {
...
+    create_workspace(new_workspace);
+  }
+}
+
 void Workspaces::init() {
-  active_workspace_name = (gIPC->getSocket1JsonReply("activeworkspace"))["name"].asString();
+  active_workspace_name_ = (gIPC->getSocket1JsonReply("activeworkspace"))["name"].asString();
+
+  // get monitor ID from name (used by persistent workspaces)
+  monitor_id_ = 0;
```
MightyPlaza, line 241: why use id here instead of desc?

src/modules/hyprland/workspaces.cpp
```diff
@@ -215,11 +326,19 @@ void Workspaces::sort_workspaces() {
                 return a->id() < b->id();
               }
               if (a->id() < 0 && b->id() < 0) {
-                if ((a->is_special()) ^ (a->is_special())) {
+                if ((a->is_special()) ^ (b->is_special())) {
                   return a->id() > b->id();
                 }
                 return a->id() < b->id();
               }
+              if ((a->id() == 0) || (b->id() == 0)) {
```
MightyPlaza, line 334: what the order we are trying to go for here?

src/modules/hyprland/workspaces.cpp
```diff
@@ -255,7 +381,7 @@ auto Workspace::handle_clicked(GdkEventButton *bt) -> bool {
   try {
     if (id() > 0) {  // normal
       gIPC->getSocket1Reply("dispatch workspace " + std::to_string(id()));
-    } else if (!is_special()) {  // named normal
+    } else if (!is_special()) {  // named normal or persistent workspace
```
MightyPlaza, line 384: persistent workspaces will also have id > 0

MightyPlaza (changes requested):

src/modules/hyprland/workspaces.cpp
zjeffer: Because I use this ID to calculate the IDs the persistent workspaces should get.

I [forked](https://github.com/zjeffer/split-monitor-workspaces) Duckonaut's `split-monitor-workspaces` Hyprland plugin to make it possible to automatically set up keyboard shortcuts to switch to the xth workspace on the current focused monitor. The plugin calculates the IDs the workspaces should get based on the monitor ID, and here I used the same formula (because the IDs have to match).

MightyPlaza: would make more sense to use ```workspace name:<monitor-desc-here>_<workspace-number-here>```, monitor id can change on restart but monitor desc is persistent.
Also it's more consistent with the rest of hyprland

src/modules/hyprland/workspaces.cpp
zjeffer: Ok, I implemented the first option.

src/modules/hyprland/workspaces.cpp
zjeffer: But how can I then ensure both waybar and the plugin give the same name to the workspaces?

MightyPlaza: waybar should only use the name you set in the config file

src/modules/hyprland/workspaces.cpp
```diff
@@ -202,28 +321,45 @@ void add_or_remove_class(const Glib::RefPtr<Gtk::StyleContext> &context, bool co
...
+                return a->id() > 0;
+              }
+
+              // both special
+              if (a->is_special() && b->is_special()) {
+                // if one is -99 => put it last
+                if (a->id() == -99 || b->id() == -99) {
+                  return b->id() == -99;
+                }
+                // both are 0 (not yet named persistents) or both are named specials (id <= -1337)
```
MightyPlaza, line 357: id of named special are from -98 to -1

MightyPlaza (approved):
LGTM

src/modules/hyprland/workspaces.cpp
zjeffer: What if you want waybar to work with any monitor you connect, without having to specify the name in the config file?

For example, if I connect my laptop to a monitor it has never seen before, waybar and my fork of the plugin will work together to automatically show persistent workspaces and set up the correct keybinds.

If we did it with monitor descriptions, you would have to do all that stuff manually.

MightyPlaza: use the port name then

zjeffer: I don't understand why it would be any different with the port name instead of the monitor description. I would still have to manually specify the port name in the configs.

In [the plugin](https://github.com/zjeffer/split-monitor-workspaces/blob/main/src/main.cpp#L63), I use the formula `workspace_name = (monitor_ID * amount_of_workspaces) + index + 1`. Here, I use the same formula so that any monitor I connect, both the plugin and the monitor will assign the same index to the same workspace ID (with 5 workspaces per monitor, the 4th workspace on the monitor with ID 2 will have workspace ID = 2*5 + 3 + 1 = 15).

I don't see how I can implement the same functionality with port names or monitor desc. I'm open to suggestions.

MightyPlaza: simply use named workspaces, containing the monitor name and a number (```name:<monitor-desc-here>_<workspace-number-here>```)
this way you wouldn't rely on using initial id, which depends on the order you connect the monitors

if you want every monitor you connect on DP-1 (for example) to have the same configuration you would use DP-1 in the config
if you want each monitor you connect on DP-1 (for example) to have a different config you could use the monitor description

using id gives you much less control than either of these options

zjeffer: I'm starting to understand. But I think changing it to use monitor descriptions might be more appropriate for another PR, as a lot of things would need to be refactored for it to work. Not to mention the changes I would have to make to the plugin.

I think we can merge this if you agree? CC @Alexays

Alexays (approved):

src/modules/hyprland/workspaces.cpp
aaronshahriari: Did anyone create a PR for using the monitor description? My monitor IDs change quite frequently and I would love to be able to set specific monitor descriptions to the output for my waybar as I do in my hyprland configuration.

Thanks!

Issue #2344, opened 2023-07-24
Seconds not displaying correctly


casey-miller:
Seeing an issue with seconds being displayed with a long decimal.
Ran a bisect and it looks like the issue is with commit: 4d9e0ea8029f769c8edb65007150de376b71a6f4.
Happy to provide any other logs/config information.

![swappy-20230724-133103](https://github.com/Alexays/Waybar/assets/22628838/e2075a4f-120d-4714-8aef-a65981c83902)

~/.config/waybar/config:
...
    "clock": {
        "interval": 1,
        "format": "{:%e %b %Y %T} ",
        "tooltip-format": "{:%e %B %Y}",
	    "on-click": "alacritty -e 'cal'"
    },
...

LukashonakV:
Seconds displays correct :) According [Chrono Format Specifications](https://fmt.dev/latest/syntax.html)

> 

'S' | Seconds as a decimal number. If the number of seconds is less than 10, the result is prefixed with 0. If the precision of the input cannot be exactly represented with seconds, then the format is a decimal floating-point number with a fixed format and a precision matching that of the precision of the input (or to a microseconds precision if the conversion to floating-point decimal seconds cannot be made within 18 fractional digits). The character for the decimal point is localized according to the locale. The modified command %OS produces the locale’s alternative representation.
-- | --


So mentioned commit brought more higher accuracy. )))
Anyway of course I'll return back seconds precision for the local time. See #2345

casey-miller:
9207fff627059b922fb790e30d68fea23f76146e fixed the issue for me.
Thank you!

AtomicFS:
Temporary fix is to use `%OS` instead of `%S` in the configuration file.

PR #2345, opened 2023-07-24
Clock. Narrow seconds precision

LukashonakV:
Hi @Alexays . While previous commits, I lost the part of code which was truncating of the system time to seconds. So ... Even precision now is more higher ... users complains about it. So this commit solves #2344


Alexays:
Thanks for the quick fix!

PR #2347, opened 2023-07-25
hyprland/workspaces fix typo

MightyPlaza:
fixes small typo

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -97,7 +97,7 @@ void Workspaces::onEvent(const std::string &ev) {
     for (Json::Value workspace_json : workspaces_json) {
       if (workspace_json["name"].asString() == payload &&
           (all_outputs() || bar_.output->name == workspace_json["monitor"].asString()) &&
-          (show_special() || workspace_json["name"].asString().find("special:") != 0)) {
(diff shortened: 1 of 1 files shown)


src/modules/hyprland/workspaces.cpp
```diff
@@ -97,7 +97,7 @@ void Workspaces::onEvent(const std::string &ev) {
     for (Json::Value workspace_json : workspaces_json) {
       if (workspace_json["name"].asString() == payload &&
           (all_outputs() || bar_.output->name == workspace_json["monitor"].asString()) &&
-          (show_special() || workspace_json["name"].asString().find("special:") != 0)) {
+          (show_special() || workspace_json["name"].asString().find("special") != 0)) {
```
zjeffer, line 100: You can change it to starts_with now that https://github.com/Alexays/Waybar/pull/2322 is merged. I'll also apply this fix in my PR.

MightyPlaza: I just want to get this fixed asap, since it's showing the special workspace if it's open before the bar starts which I don't like.
feel free to change all the occurences

MightyPlaza:
@Alexays can we get this merged

zjeffer (approved):

PR #2372, opened 2023-08-01
[Feature] Network module: Network connected but no internet format

### LINKED ISSUE #1487: network module: network connected but no internet
It would be quite handy to have indicator for netowork (wlan on eth) is connected but no internet.

I think, it can be achieved by running `ping`
1. For specified interval
2. Once network is connected, t

So, the config to be like,
"network": {
...
    "format-nointernet": {
                  "wifi": "No Internet",
                  "lan": "No ineternet",
},
    "ping-interval": "30",
..
}

Absobel:
Attempt to resolve #1487.

It seem to work with wifi but I couldn't test it with ethernet because I lack an ethernet port on my laptop.

The network module will try pinging the list of servers in the array "ping-servers" (the module will keep its default behaviour if it is empty or not initialised) every "interval" seconds. I also added the format for WiFi without internet "format-wifi-no-internet" and the format for ethernet without internet "format-ethernet-no-internet" if the pings fail


Absobel:
Hey, I saw that this pull request never had any comments or anything. Would it be possible to review my code and tell me if I need to modify it, or if it is an eldritch horror of an abomination, or anything in between ?

I guess my pull request just went unnoticed x) (tbh i forgot it too)

PR #2373, opened 2023-08-01
Fix logical errors, formatting error

marshallmoats:
PR #2341 resulted in my workspaces not showing up. It seems this is a result of a couple minor logical mistakes (due to switching from `find` to `starts_with`), along with `id` not being formatted in, which was throwing an "argument not found" exception. I don't know the reasoning behind not formatting `id`, if anyone knows why, let me know.


zjeffer:
@marshallmoats I can't find the review comment anymore but @MightyPlaza said the `id` label would not work anymore due to persistent workspaces. I can't remember why exactly, though.

Can you share the config you are using that caused your workspaces not showing up? I didn't have this problem.

marshallmoats:
@zjeffer Thanks for the clarification. I hadn't been using a config for that module. I now see that the issue is that the default format is `{id}`, apologies for the oversight. I suppose the default format should be changed from `{id}` to `{name}`, then?

zjeffer:
Yes

PR #2393, opened 2023-08-10
Hyprland/workspaces: use named icons instead of searching by id

zjeffer:
Bug report by @VincentBerthier: 

>  It seems the icons for the workspaces specified in the configuration only work if they’re specified for a workspace ID, not its name.

This PR should fix this. @VincentBerthier can you test this?


VincentBerthier:
Erm, I changed my config to use IDs now which wouldn’t be too much of a problem, but I’m also playing around with NixOS and I don’t think I have anything installed to test this sorry :o

zjeffer:
Alright, it seems to work on my system and it makes more sense to use name instead of id, so this can be merged @Alexays

Issue #1968, opened 2023-01-22
Clicks registered at wrong location


UltraBlackLinux:
Hey there,
I just noticed a weird issue with waybar. Whenever I click on a custom widget which launches fuzzel, the next few clicks, nomatter where on the bar they happen, will trigger that previous widget for many clicks.

What is causing this?

I'm running Waybar v0.9.16

Thanks!

PolGZ:
I have the "same" issue. In my case, clicking on a widget that launches rofi, makes rofi appear back again wherever on the bar y click after that first time.

Even if I make a killall rofi while the window is showing and it closes.

Waybar v0.9.17 installed through [waybar-hyprland-git](https://aur.archlinux.org/packages/waybar-hyprland-git).

Here is my config.jsonc extract:

```
    "battery": {
    	"bat": "BAT0",
	"format": "{icon} {capacity}%",
	"format-icons": ["", "", "", "", ""],
	"on-click": "~/.config/rofi/applets/bin/battery.sh"
    },
```

At first I thought it had to do with rofi location and anchor values, and their x-offset/y-offset values, that they would interfere with the bar position. But even if I set it to the center of the screen, clicking anywhere on the bar will remain launching that rofi script from anywhere else.

opekope2:
Same here with [SwayNC](https://github.com/ErikReider/SwayNotificationCenter) on Hyprland. If I use a different mouse button (right or middle), the bug disappears. I worked around it by adding a sleep in the on-click: `"on-click": "sleep 0.1; swaync-client -t -sw"`. The duration may need to be tweaked.
[waybar-hyprland-git](https://aur.archlinux.org/packages/waybar-hyprland-git) 0.9.17.r89.gb34d325a-1 (the package only patches the workspace module)

LDAP:
Same setup as @opekope2. I noticed that if a right-click twice, the left click on other modules work again.

niraami:
I've also noticed this with multiple modules, even the integrated `wlr/workspaces`, on which the `sleep 0.1` workaround doesn't work (or I don't know how to make it work).
The issue seems to occur when waybar is defocused, while the user still has the left-mouse button pressed down, so this might also be related to #1047. Even with the `sleep` added, if you hold down the mouse button, the issue will still occur. As @LDAP mentioned, right clicking does seem to solve this, if that action doesn't also defocus waybar.

Overall, pretty annoying, and I'd love for this to get fixed.

AnoRebel:
[...]

PR #2397, opened 2023-08-13
Add on-button release actions

KosmX:
Allow `on-click-release` actions for most buttons, this will exec the action when the button is released.  

This is a workaround for #1968 (I believe the bug is caused by an unexpected focus lost while holding the button)  
> at least, it works for me.

By doing `"on-click-release": "swaync-client -t -sw",` I could fix the wrong location click bug without a sleep longer than the click.

But I think, on-release will have other uses too!


KosmX:
My logic is incorrect, I'll open a new PR with a correct fix (sorry for the garbage PR)

PR #2399, opened 2023-08-13
Button release action

KosmX:
Allow `on-click-release` actions for most buttons, this will execute the action when the button is **released**.  

This is a workaround for #1968 (I believe the bug is caused by an unexpected focus loss while holding the button)  
> at least, it works for me.
By doing `"on-click-release": "swaync-client -t -sw",` I could fix the wrong location click bug without a sleep longer than the click.

And  I think, on-release will have other uses too!


Alexays (approved):

KosmX:
I've updated the configuration wiki page to include this action.

PR #2408, opened 2023-08-15
Implement 'active_only' option and 'visible' class in hyprland/workspaces

zjeffer:
With this PR, `active_only` can be set to true, which will only show the active workspace. 

If I understand the feature request correctly, each monitor should show the currently visible workspace, right? This PR currently only shows the active one as reported by Hyprland, so if a monitor is not focused, it will not show any workspaces. Is this supposed to happen? I would expect it to be better if each monitor shows one "active" workspace, i.e. the one currently visible.

Let me know if this can be improved.

CC @Anakael @MightyPlaza

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -38,6 +38,11 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value 
     show_special_ = config_show_special.asBool();
   }
 
+  auto config_active_only = config_["active-only"];
+  if (config_active_only.isBool()) {
+    active_only_ = config_active_only.asBool();
+  }
+
   box_.set_name("workspaces");
   if (!id.empty()) {
     box_.get_style_context()->add_class(id);
@@ -75,11 +80,29 @@ auto Workspaces::update() -> void {
 
   workspaces_to_create_.clear();
 
+  // get all active workspaces
+  auto monitors = gIPC->getSocket1JsonReply("monitors");
+  std::vector<std::string> visible_workspaces;
+  for (Json::Value &monitor : monitors) {
+    auto ws = monitor["activeWorkspace"];
+    if (ws.isObject() && (ws["name"].isString())) {
+      visible_workspaces.push_back(ws["name"].asString());
+    }
+  }
+
   for (auto &workspace : workspaces_) {
+    // active
     workspace->set_active(workspace->name() == active_workspace_name_);
-    if (workspace->name() == active_workspace_name_ && workspace.get()->is_urgent()) {
+    // disable urgency if workspace is active
+    if (workspace->name() == active_workspace_name_ && workspace->is_urgent()) {
       workspace->set_urgent(false);
     }
+
+    // visible
+    workspace->set_visible(std::find(visible_workspaces.begin(), visible_workspaces.end(),
+                                     workspace->name()) != visible_workspaces.end());
+
+    // set workspace icon
     std::string &workspace_icon = icons_map_[""];
     if (with_icon_) {
       workspace_icon = workspace->select_icon(icons_map_);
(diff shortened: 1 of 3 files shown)


src/modules/hyprland/workspaces.cpp
```diff
@@ -319,6 +330,12 @@ void add_or_remove_class(const Glib::RefPtr<Gtk::StyleContext> &context, bool co
 }
 
 void Workspace::update(const std::string &format, const std::string &icon) {
+  if (this->workspace_manager_.active_only() && !this->active()) {
```
zjeffer, line 333: Once `urgent` workspaces are implemented in this module, we can simply add a check for this->urgent() here. Then it would work the same as in the `wlr/workspaces` module.

MightyPlaza (approved):
the rest LGTM

MightyPlaza:
```
"hyprland/workspaces": {
    "persistent_workspaces": {
        //"1":[],"2":[],"3":[],"4":[],"5":[],"6":[],"7":[],"8":[],"9":[]
        "*": 10
    },
    "disable-scroll": false,
    "all-outputs": false,
    "warp-on-scroll": false,
    "on-click": "activate",
    "format": "{icon}",
    "format-icons": {
        "1":"一","2":"二","3":"三","4":"四","5":"五","6":"六","7":"七","8":"八","9":"九","10":"十"
    }
},
```
this config for example
image

was not working as intended, not sure if it has been fixed

zjeffer:
Ah, I thought you mean something like this, which does work:

```jsonc
    "format-icons": {
      "default": "O",
      "active": "a",
      "persistent": "p",
      "special": "s"
    },
```

Your example indeed doesn't work, I'll fix that.

zjeffer:
Sorry this took so long, I haven't had much time lately and I kept getting issues compiling due to the `fmt` & `spdlog` stuff. Should be ready to review now. 

Let me know if this is the correct behaviour or not:

* Setting both `active-only` and `all-outputs` to `true` => each bar will show all monitor's visible workspaces, and mark the globally active one as active, and the rest as visible.
* Setting `active-only` to `true` and `all-outputs` to `false` => each bar shows only its visible workspace. The focused monitor shows the visible workspace as active.
* Setting `active-only` to `false` and `all-outputs` to `true` => each bar shows all non-empty workspaces (so each bar shows the exact same thing)
* Persistent workspaces are not effected, those are always shown.

zjeffer:
The `visible` css class doesn't seem to be working yet for me, but the visible icon does work, investigating...

MightyPlaza (approved):
lgtm

khaneliman (approved):

khaneliman:
@zjeffer looks like this needs a rebase btw

zjeffer:
@Alexays Ready to merge.

khaneliman:
Closes https://github.com/Alexays/Waybar/issues/2465

Alexays (approved):

Alexays:
Thanks!

NotAShelf:
Is it intended behaviour that active-only displays both workspaces if two monitors are at the same time?

zjeffer:
@NotAShelf only if all-outputs is true. If it's false, it should only show the current monitor's active workspace.

NotAShelf:
Understood. Any chance we could get a rule to see one active workspace at a time, regardless of monitor? Similar to active-only + all-outputs behaviour on wlr/workspaces

zjeffer:
Do you mean you want to see the same active workspace on every monitor? Like if workspace 1 is active on monitor 1, you also want to see "1" in monitor 2?

NotAShelf:
More or less, yes.

Diaoul:
I'm looking for a way to have active workspace (e.g. 9) on a not focused monitor (e.g. B) to have a different style applied to it on the bar on monitor B compared to other workspaces there.

Reading this, I'm not sure it is possible nor which combination to try :confused:

Issue #2403, opened 2023-08-14
0.9.21: build fails


kloczek:
gcc 13.2 and build fails with
```console
[...]

PR #2412, opened 2023-08-16
Fix simpleclock module compilation with recent fmt

jpalus:
both `format_` and `tooltip_format` are dynamically provided formats so wrap them in `fmt::runtime()`


jpalus:
It attempts to fix another [issue](https://github.com/Alexays/Waybar/issues/2403#issuecomment-1680242361) raised in #2403 but note that while I verified the change fixes build, I didn't verify if it actually works. I would be surprised if it didn't though.

PR #2413, opened 2023-08-16
Add format-no-controller to bluetooth module and fix it still being visible if the format string is empty

### LINKED ISSUE #2163: bluetooth module visible if no bluetooth controller found
Issue #1707 still occurs in v0.9.17. I wasn't sure if I should tag you so I opened a new issue.

When the module has some padding or margin set, it is visible as an empty field.
screenshot_20230508_193951-726744043

When you click on it, it appears as if bluetooth was enabled (my computer doesn't have bluetooth):
screenshot_20230508_194032-378481588

PucklaJ:
Fixes #2163

This PR contains the following changes:
1. Adds the `format-no-controller` and `tooltip-format-no-controller` format strings that will be used if no bluetooth controller could be found.
2. Fixes the bluetooth module still being visible if the format string is empty.

The reason for why the bluetooth module was still being displayed is the return statement in the constructor of the bluetooth module:
https://github.com/Alexays/Waybar/blob/aecd80702e746e7cb9a78c70d0578a88547341b1/src/modules/bluetooth.cpp#L95-L106

This return statement caused `dp.emit()` not to be called and somehow this removed the ability for the module to hide.

I also think that it makes sense to have an additional format string for the case when the system does not have a bluetooth controller. If the user still wants to hide the bluetooth module he just needs to set `format-no-controller` to an empty string.

### DIFF
--- a/include/modules/bluetooth.hpp
+++ b/include/modules/bluetooth.hpp
@@ -59,7 +59,8 @@ class Bluetooth : public ALabel {
   auto getDeviceProperties(GDBusObject*, DeviceInfo&) -> bool;
   auto getControllerProperties(GDBusObject*, ControllerInfo&) -> bool;
 
-  auto findCurController(ControllerInfo&) -> bool;
+  // Returns std::nullopt if no controller could be found
+  auto findCurController() -> std::optional<ControllerInfo>;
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM thanks!

PR #2414, opened 2023-08-16
Release button event second attempt

KosmX:
Attempt to properly add button release events
#2399 (but I hope without bugs)  

I've created a new function in AModule to handle release events without invoking handleToggle methods (what is not prepared to handle release events)  

Also, I've moved the user input handling to a private function (simply to avoid code duplication).  

> I've no access to my workstation, I couldn't test this PR. Please verify it before merging.

### DIFF
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -36,8 +36,11 @@ class AModule : public IModule {
 
   virtual bool handleToggle(GdkEventButton *const &ev);
   virtual bool handleScroll(GdkEventScroll *);
+  virtual bool handleRelease(GdkEventButton *const &ev);
 
  private:
+  bool handleUserEvent(GdkEventButton *const &ev);
+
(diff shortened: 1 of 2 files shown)


KosmX:
I've tested #1968 , this PR fixes it. (when used `-release` event)  
I can't test every single module simply because I don't have the hardware, but I've implemented a new release event handler what should not interfere with any existing handler.

Alexays:
LGTM, @cinerea0 or @hekatonkheirex can you test just to be sure that a single click to trigger module work?

cinerea0:
I've been using these changes for a few days, I can confirm that the double-click issue I was encountering before is not present.

Alexays:
Thanks!

hekatonkheirex:
Sorry, been AFK for a while. Also can confirm there's no issue on my side. Thanks for the heads-up @Alexays

PR #2423, opened 2023-08-19
Hyprland hyprctl dispatch workspace compatibility

HanoJing:
This line of code enables better compatibility with the Hyprland window manager, specifically, when clicking on workspaces shown in Waybar.

I have verified that this does not break the activate-workspace-on-click function on sway for example. 
I didn't test other wlroots-based compositors though.

At https://wiki.hyprland.org/Useful-Utilities/Status-Bars/
see: `sed -i -e 's/zext_workspace_handle_v1_activate(workspace_handle_);/const std::string command = "hyprctl dispatch workspace " + name_;\n\tsystem(command.c_str());/g' src/modules/wlr/workspace_manager.cpp`

Thanks a lot!

### DIFF
--- a/src/modules/wlr/workspace_manager.cpp
+++ b/src/modules/wlr/workspace_manager.cpp
@@ -523,7 +523,8 @@ auto Workspace::handle_clicked(GdkEventButton *bt) -> bool {
   if (action.empty())
     return true;
   else if (action == "activate") {
-    zext_workspace_handle_v1_activate(workspace_handle_);
+    const std::string command = "hyprctl dispatch workspace " + name_;
+	system(command.c_str());
   } else if (action == "close") {
     zext_workspace_handle_v1_remove(workspace_handle_);
   } else {


zjeffer (changes requested):
This would cause other Wayland compositors to not be able to activate workspaces: it would only work with Hyprland.

HanoJing:
> This would cause other Wayland compositors to not be able to activate workspaces: it would only work with Hyprland.

I see. Thanks for the reply. 
This feature isn't that important to me atleast. Without this patch, Waybar sometimes segfaults when clicking on workspaces. 
I just disabled on-click in the config for now.

zjeffer:
As the wiki says, it is recommended to compile it manually with that sed command, or use a patched package (like Arch's `waybar-hyprland(-git)` package.

But currently, it is also possible to use the new `hyprland/workspaces` module, and in that case you don't even need that patch.

HanoJing:
> As the wiki says, it is recommended to compile it manually with that sed command, or use a patched package (like Arch's `waybar-hyprland(-git)` package.
> 
> But currently, it is also possible to use the new `hyprland/workspaces` module, and in that case you don't even need that patch.


Oh great, I didn't know about the hyprland/workspaces. Thanks a lot!

Issue #2415, opened 2023-08-16
The build fails after updating fmt to version 10.0


fakeowl1:
Meson version: 1.2.1
Ninja version: 1.11.1

```
FAILED: test/waybar_test.p/.._src_config.cpp.o 
[...]

PR #2431, opened 2023-08-24
hyprland/workspaces: Add "empty" icon and class

### LINKED ISSUE #2430: Difference in behaviour of persistent icon between wlr/workspaces and hyprland/workspaces
In wlr/workspaces, the "persistent" icon is only set if the workspace is not empty: https://github.com/Alexays/Waybar/blob/b7a527b122831235fb36f661d415f5f3ed378d92/src/modules/wlr/workspace_manager.cpp#L498-L503
But in hyprland/workspaces, the "persistent" icon is set if the workspace is a persistent one: https://github.com/Alexays/Waybar/blob/b7a527b122831235fb36f661d415f5f3ed378d92/src/modules/hyprland/workspaces.cpp#L390-L395

The behavior in wlr/workspaces allowed me to have this rule:
```
"format-icons": {
    "active": "",
    "default": "",
    "persistent": "",
},
```
where the active workspace gets the active icon, inactive workspaces that contain windows get the default icon (since they don't count as persistent, as they have a window in them), and inactive workspaces that don't contain windows get the persistent icon.

When I switched to using hyprland/workspaces, all inactive workspaces now use the persistent icon regardless of whether they have a window in them.

I like to have all my workspaces visible, and also an indication that a workspace contains a window... and I do think this way of doing it is pretty hacky, but I can't think of any other way I'm supposed to do this. It also seems this behaviour is intended considering on hyprland/workspaces, persistent workspaces do have the `.persistent` class taken away when they contain a window: https://github.com/Alexays/Waybar/blob/b7a527b122831235fb36f661d415f5f3ed378d92/src/modules/hyprland/workspaces.cpp#L325

so I guess this means I could technically work around it with some weird `.persistent::before` CSS rule, but it would be nice to use the `format-icon` config object to format the icons.

zjeffer:
Fixes #2430 

I haven't tested this yet because I'm still having trouble compiling due to fmt (#2415), but I'm pretty sure this will work fine. Maybe someone else can test it?

### DIFF
--- a/man/waybar-hyprland-workspaces.5.scd
+++ b/man/waybar-hyprland-workspaces.5.scd
@@ -46,7 +46,8 @@ Additional to workspace name matching, the following *format-icons* can be set.
 - *default*: Will be shown, when no string match is found.
 - *active*: Will be shown, when workspace is active
 - *special*: Will be shown on non-active special workspaces
-- *persistent*: Will be shown on non-active persistent workspaces
+- *empty*: Will be shown on empty persistent workspaces
+- *persistent*: Will be shown on non-empty persistent workspaces
 
 # EXAMPLES
 
@@ -74,6 +75,7 @@ Additional to workspace name matching, the following *format-icons* can be set.
 - *#workspaces*
 - *#workspaces button*
 - *#workspaces button.active*
+- *#workspaces button.empty*
 - *#workspaces button.persistent*
 - *#workspaces button.special*
 - *#workspaces button.urgent*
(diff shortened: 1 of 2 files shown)


khaneliman:
> I haven't tested this yet because I'm still having trouble compiling due to fmt (#2415), but I'm pretty sure this will work fine. Maybe someone else can test it?

If you have nix installed i've been able to build with .envrc and nix-build .

khaneliman:
image

khaneliman:
Just update the man document and should be good to go, then.

khaneliman (approved):

khaneliman:
@Alexays this should be good whenever you get a chance.

Alexays:
Thanks!

Issue #2271, opened 2023-07-02
Hyprland/workspaces improvements and further development


zjeffer:
I'm creating this issue so we can keep track of further development for the newly implemented hyprland/workspaces module (by @Anakael in https://github.com/Alexays/Waybar/pull/2264)

Here are some features I would personally like to see:

- [ ] Urgent class
- [ ] Focused class
- [ ] Option to only show active workspaces
- [ ] Persistent workspaces (+persistent css class)
- [ ] Multi-monitor: option to disable showing all workspaces on every output
- [ ] on-click action
- [ ] on-scroll action

There is also currently a segfault happening when rapidly switching workspaces: https://github.com/Alexays/Waybar/pull/2264#issuecomment-1615772176.

Anakael:
Active (focused) class is already implemented.

Anakael:
Wiki is updated:
https://github.com/Alexays/Waybar/wiki/Module:-Hyprland
(Also fixed wrong class name in https://github.com/Alexays/Waybar/wiki/Module:-Workspaces)

VincentBerthier:
Hi,
It seems the icons for the workspaces specified in the configuration only work if they’re specified for a workspace ID, not its name.

maxwelljens:
For me, multi-monitor displaying does not work. All the workspaces are shown on both bars in my dual monitor setup. Names also do not work, with Waybar claiming that the argument `{name}` does not exist.

I can open up a separate issue for this, if it sounds interesting.

VincentBerthier:
That’s not me for the unseen ^^
I changed my configuration to use IDs instead of names, it works fine now. I think.

maxwelljens:
@zjeffer I tried many different configurations and the one you presented is one of those I tried. None of the options seem to do anything, except `format`.

```
    "hyprland/workspaces": {
        "format": "{id}",
        "all-outputs": false,
        "persistent_workspaces": {
                "*": 5
            }
    },
```

The error in question is this, if you set `"format": "{name}"`:

```
[2023-08-11 08:54:32.006] [error] hyprland/workspaces: argument not found
[2023-08-11 08:54:32.121] [error] hyprland/workspaces: argument not found
```

It is definitely that element, because if I set `"format": "{id}"` or `"format": "{}"` then it does not show that error. No other errors are present either.

> What do you mean by "unseen workspace"?

I should have said "not visible" instead. If we have three workspaces 1, 2, and 3:

- 1 is visible on left monitor
[...]

PR #2438, opened 2023-08-29
hyprland persistent workspace support new config declaration

khaneliman:
Per discussion https://github.com/Alexays/Waybar/issues/2271#issuecomment-1692402703 with @zjeffer  about adding a new config style. You provide the monitors as keys and either an integer for generating that many workspaces or an array of workspaces to create on that monitor.


khaneliman:
@zjeffer what we were talking about

Alexays:
LGTM

PR #676, opened 2020-04-19
WIP: dynamic args

Alexays:

PR #2441, opened 2023-08-30
modules/custom: Allow json fields as format specifiers

### LINKED ISSUE #432: Custom format replacements in custom modules
I'm currently writing a custom module to monitor free space on my SSD. However, custom modules that return a JSON can only return a `text` field containing already formatted text. I would like to return different values (free space, used space, total space, usage) and let the format be specified in my waybar `config` file.

Here is an example of what I would like to return with my module:
```json
{"mount": "/home",
 "total": "871", 
 "used": "86",
 "free": "740",
 "usage": "9.8", 
 "tooltip": "", 
 "class": "ok"}
```

And I would like to define the format this way in my `config` file:
`"format": "{mount}: {used}G/{total}G"`

That would make custom modules better by dissociating data and format.

rad4day:
This PR changes the custom module in a (hopefully) backwards compatible way to allow using JSON fields as format specifiers.

As #404 and #676 have both been closed as non-working and I reeeaaally wanted this feature I spent a bit of time trying to figure out how to implement it. That's what I've come up with.
There are probably some weird bugs caused by this code. Disclaimer: I am not a C++ developer.

For example

```jsonc
    "custom/spotify": {
        "format-alt": "{status_icon}",
        "format-alt-click": "click-middle",
        "exec": "/usr/bin/python3 ~/.config/waybar/scripts/mediaplayer.py --player spotify",
        "format": "{status_icon} {artist} - {title}",
        "return-type": "json",
        "escape": true,
    },
```
which currently outputs the following JSON as a single line, where text is only provided as a fallback and not necessary.
```jsonc
{
  "player_name": "spotify",
  "player_icon": "\uf1bc",
  "artist": "Demoartist",
  "title": "totally a real song",
  "status": "Playing",
  "status_icon": "\uf144",
  "class": [
    "spotify",
    "playing"
  ],
  "text": "Demoartist - totally a real song"
}

```

What has been tested:
- classes still work
- alt format works
- tooltips work

Percentage Icons are currently untested as I don't have a use case for those. I'd enjoy if somebody could test if they still work or provide a working example.

resolves #432

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <fmt/format.h>
+#include <fmt/args.h>
 
 #include <csignal>
 #include <string>
@@ -34,6 +35,7 @@ class Custom : public ALabel {
   std::string alt_;
   std::string tooltip_;
   std::vector<std::string> class_;
+  fmt::dynamic_format_arg_store<fmt::format_context> fields_;
   int percentage_;
   FILE* fp_;
   int pid_;
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -122,14 +122,14 @@ auto waybar::modules::Custom::update() -> void {
       (output_.out.empty() || output_.exit_code != 0)) {
     event_box_.hide();
   } else {
+    fields_.clear();
     if (config_["return-type"].asString() == "json") {
       parseOutputJson();
     } else {
       parseOutputRaw();
     }
-    auto str = fmt::format(fmt::runtime(format_), text_, fmt::arg("alt", alt_),
-                           fmt::arg("icon", getIcon(percentage_, alt_)),
-                           fmt::arg("percentage", percentage_));
+    auto str = fmt::vformat(format_, fields_);
+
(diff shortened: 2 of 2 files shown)


alohl669:
I'm really interested, this feature could open the doors to printing different "format" outputs like format-tooltip or format-alt in the future, isn't it?

rad4day:
@alohl669 With a small change to [L140](https://github.com/rad4day/Waybar/blob/a194755b4f2e615afb7b109bda023de7e9ca70b9/src/modules/custom.cpp#L140) (adding the formatting engine there) this *should* be possible I think.

ShellCode33:
@Alexays please merge this :heart: many tools already support json output, it would allow us to have a much more cleaner Waybar config or avoid having to wrap everything in custom scripts to convert json fields

christos-chatzifountas-biotz-io:
What is the state of this future?
I am was trying to combine nvidia-smi with miller until i realized that the JSON has to have - very inconveniently - output to specific fields

cyb3rkun:
Any update on this?
It would be a huge benefit.

Issue #2443, opened 2023-08-30
persistent_workspaces should be persistent-workspaces


PucklaJ:
This key of the different `workspaces` modules should be named `persistent-workspaces` with a hyphen instead on an underscore to make it consistent with all other keys. This inconsistency actually led to an issue as I intuitively typed `persistent-workspaces`. I was wondering what I did wrong until I realized that I used a hyphen instead of an underscore. Therefore the naming should be consistent and use a hyphen instead of an underscore.

khaneliman:
This is something that threw me off, as well. I think it's worth updating to maintain consistency... @zjeffer  thoughts?

zjeffer:
I've been thinking about this as well, but I feel we should make both options possible, so no existing configs break, and maybe log a warning saying the old name is deprecated. After a while we can then remove the deprecated style.

khaneliman:
> I've been thinking about this as well, but I feel we should make both options possible, so no existing configs break, and maybe log a warning saying the old name is deprecated. After a while we can then remove the deprecated style.

Guess that's one approach. :P I was taking the break -git people's now and publish in release notes for others. But, we can be more graceful too. lol

PR #2468, opened 2023-09-05
deprecate persistent_workspaces in favor of persistent-workspaces

khaneliman:
Closes: https://github.com/Alexays/Waybar/issues/2443 

Persistent workspaces currently is the only option using this naming style, changing to match all other options. 

Prompt updating config from `persistent_workspaces` to `persistent-workspaces` but keep current implementation support until after next release to remove. 

Follow up: Update wiki entries to match new name.

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -231,8 +231,15 @@ void Workspaces::remove_workspace(std::string name) {
 }
 
 void Workspaces::fill_persistent_workspaces() {
-  if (config_["persistent-workspaces"].isObject()) {
-    const Json::Value persistent_workspaces = config_["persistent-workspaces"];
+  if (config_["persistent_workspaces"].isObject()) {
+    spdlog::warn(
+        "persistent_workspaces is deprecated. Please change config to use persistent-workspaces.");
+  }
+
+  if (config_["persistent-workspaces"].isObject() || config_["persistent_workspaces"].isObject()) {
+    const Json::Value persistent_workspaces = config_["persistent-workspaces"].isObject()
+                                                  ? config_["persistent-workspaces"]
+                                                  : config_["persistent_workspaces"];
     const std::vector<std::string> keys = persistent_workspaces.getMemberNames();
 
     for (const std::string &key : keys) {
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -231,8 +231,8 @@ void Workspaces::remove_workspace(std::string name) {
 }
 
 void Workspaces::fill_persistent_workspaces() {
-  if (config_["persistent_workspaces"].isObject()) {
-    const Json::Value persistent_workspaces = config_["persistent_workspaces"];
+  if (config_["persistent-workspaces"].isObject()) {
+    const Json::Value persistent_workspaces = config_["persistent-workspaces"];
     const std::vector<std::string> keys = persistent_workspaces.getMemberNames();
 
     for (const std::string &key : keys) {
--- a/man/waybar-hyprland-workspaces.5.scd
+++ b/man/waybar-hyprland-workspaces.5.scd
@@ -69,7 +69,7 @@ Additional to workspace name matching, the following *format-icons* can be set.
 		"active": "",
 		"default": ""
 	},
-	"persistent_workspaces": {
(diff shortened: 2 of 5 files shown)


zjeffer (changes requested):
As I said in the issue, I think we should:

* Make both options parseable
* Log a warning if the user uses the old style
* After a couple of Waybar releases, we should remove the old style.

This way, no existing config would break. If we merge this in the current state, we'd have tons of people opening issues because their configs would break.

khaneliman:
> As I said in the issue, I think we should:
> 
> * Make both options parseable
> * Log a warning if the user uses the old style
> * After a couple of Waybar releases, we should remove the old style.
> 
> This way, no existing config would break. If we merge this in the current state, we'd have tons of people opening issues because their configs would break.

This new commit look better?

src/modules/hyprland/workspaces.cpp
```diff
@@ -207,8 +207,17 @@ void Workspaces::remove_workspace(std::string name) {
...
-  if (config_["persistent_workspaces"].isObject() && !all_outputs()) {
-    const Json::Value persistent_workspaces = config_["persistent_workspaces"];
+  if (config_["persistent_workspaces"].isObject()) {
+    spdlog::warn(
+        "persistent_workspaces is deprecated. Please change config to use persistent-workspaces.");
+  }
+
+  if ((config_["persistent-workspaces"].isObject() ||
+       config_["persistent_workspaces"].isObject()) &&
+      !all_outputs()) {
```
zjeffer, line 217: I wonder why I added `!all_outputs()` to here. I copied this from the `wlr/workspaces` module but I don't remember why I did that. Does it make sense to only allow persistent-workspaces when all_outputs is false?

zjeffer (approved):
Looks much better! I can't test right now but I'm sure it works fine.

src/modules/hyprland/workspaces.cpp
khaneliman: I personally think we should remove it, if you're okay with it. It feels counterintuitive... I am pinning a workspace to be displayed and if I want everything on all monitors I dont get to pin them?

khaneliman: The only non-intuitive thing is that when it's removed the persistent workspace shows on the output you have pinned it to, but not the other outputs that all_outputs implies (until it isn't empty). But with the setting, you lose the persistent workspace and all outputs match in appearance.

khaneliman: @zjeffer  Looks like you removed it already in that other PR when I rebased. :P

khaneliman:
Rebased to resolve merge conflicts. Still works fine, like before.

khaneliman:
@Alexays should be good now

src/modules/hyprland/workspaces.cpp
zjeffer: >  Looks like you removed it already in that other PR when I rebased. :P

Lol I don't even remember doing that haha

zjeffer: IIRC, I think the reason it's like that in `wlr/workspaces` was because some kind of limitation when persistent-workspaces was implemented in that module. 

@MonstrousOgre Do you remember why you did it this way in `wlr/workspaces`?

khaneliman: What I saw when testing locally, was that you'll see the persistent workspace in the output it's mapped to but not the other outputs, as you might expect from enabling all_outputs. So, the logic doesn't account for empty persistent workspaces to be shown on other outputs. Might just be confusing until reworked a bit?

MonstrousOgre: > IIRC, I think the reason it's like that in `wlr/workspaces` was because some kind of limitation when persistent-workspaces was implemented in that module.
> 
> @MonstrousOgre Do you remember why you did it this way in `wlr/workspaces`?

So when all-outputs is true, the number of `WorkspaceManager` instances for each bar is equal to the number of monitors. In my case, this would be 2 for every bar. But since there are two bars (one for each monitor), there are totally 4 instances of `WorkspaceManager` per application.

So if I had `"5": ["eDP-1", "HDMI-A-1"],` in my config, workspace 5 would show up twice on every monitor, or 4 times totally. I figured I'd just prevent the all-outputs and persistent workspaces being used together to avoid this issue with duplicate workspaces

PR #2472, opened 2023-09-06
fix hyprland module's workspace sort

mbpowers:
changed sort from lexicographic to numeric


zjeffer:
I don't understand how it used to be lexicographic? The workspace id is an integer, so we're already sorting numerically, no?

Issue #2481, opened 2023-09-08
hyprland/workspaces: Auto rename lost active information


foonathan:
I'm on the latest git commit of both waybar and Hyprland.

When I rename the active workspace, waybar correctly updates the name/icon, but while the workspace was correctly styled as active before, it no longer is after the rename. I have to switch to another workspace and back to force an update.

khaneliman:
@foonathan Can you try with the PR https://github.com/Alexays/Waybar/pull/2485 ?

foonathan:
I can confirm it's fixed. Thank you.

PR #2485, opened 2023-09-08
fix: rename workspace active fix

khaneliman:
Closes: https://github.com/Alexays/Waybar/issues/2481

Verified locally. With a `hyprctl dispatch renameworkspace 3 "4"` and saw old behavior (top) and new behavior (bottom) of it keeping the active state. My config visible = blue and active = green. 

image


khaneliman:
@Alexays bug fix for active status with a workspace rename whenever you get a chance.

Issue #2474, opened 2023-09-07
hyprland workspaces sorted differently


alexmo1997:
I have a "zeroth" workspace in hyprland that is `name:0`. When using wlr/workspaces I could use sort-by-number so it was sorted before the first numbered workspace. Also, hyprctl sorts them like this internally (as can be seen when I use `hyprctl dispatch workspace e+1`). 

So it should either be possible to use something like wlr/workspaces sort-by-number or waybar should use the hyprland internal sorting, otherwise it leads to weird behaviour (like a hyprland bind doing `hyprctl dispatch workspace e+1` jumping around in the bar)

PR #2486, opened 2023-09-09
feat: hyprland workspaces add sort-by

khaneliman:
Closes: https://github.com/Alexays/Waybar/issues/2474

Adds a `sort-by` option that takes in a string to choose how to sort workspaces. This allows you to manually specify the sorting method and future sorting methods can just utilize the same config option. Fallsback to the original sorting.

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -86,6 +86,7 @@ class Workspaces : public AModule, public EventHandler {
   bool all_outputs_ = false;
   bool show_special_ = false;
   bool active_only_ = false;
+  std::string sort_by = "default";
 
   void fill_persistent_workspaces();
   void create_persistent_workspaces();
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -11,6 +11,7 @@
 #include "AModule.hpp"
 #include "bar.hpp"
 #include "modules/hyprland/backend.hpp"
+#include "util/enum.hpp"
 
 namespace waybar::modules::hyprland {
 
@@ -86,7 +87,8 @@ class Workspaces : public AModule, public EventHandler {
   bool all_outputs_ = false;
   bool show_special_ = false;
   bool active_only_ = false;
-  std::string sort_by = "default";
+  util::EnumParser enum_parser_;
+  util::EnumParser::SORT_METHOD sort_by_ = util::EnumParser::SORT_METHOD::DEFAULT;
 
   void fill_persistent_workspaces();
   void create_persistent_workspaces();
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -80,6 +80,8 @@ class Workspaces : public AModule, public EventHandler {
   void create_workspace(Json::Value& value);
   void remove_workspace(std::string name);
   void set_urgent_workspace(std::string windowaddress);
+  void parse_config(const Json::Value& config);
+  void register_ipc();
 
   bool all_outputs_ = false;
   bool show_special_ = false;
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -87,8 +87,14 @@ class Workspaces : public AModule, public EventHandler {
   bool all_outputs_ = false;
   bool show_special_ = false;
   bool active_only_ = false;
-  util::EnumParser enum_parser_;
-  util::EnumParser::SORT_METHOD sort_by_ = util::EnumParser::SORT_METHOD::DEFAULT;
+
+  enum SORT_METHOD { ID, NAME, NUMBER, DEFAULT };
+  util::EnumParser<SORT_METHOD> enum_parser_;
+  SORT_METHOD sort_by_ = SORT_METHOD::DEFAULT;
+  std::map<std::string, SORT_METHOD> sort_map_ = {{"ID", SORT_METHOD::ID},
+                                                  {"NAME", SORT_METHOD::NAME},
+                                                  {"NUMBER", SORT_METHOD::NUMBER},
+                                                  {"DEFAULT", SORT_METHOD::DEFAULT}};
 
   void fill_persistent_workspaces();
   void create_persistent_workspaces();
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -88,7 +88,7 @@ class Workspaces : public AModule, public EventHandler {
   bool show_special_ = false;
   bool active_only_ = false;
 
-  enum SORT_METHOD { ID, NAME, NUMBER, DEFAULT };
+  enum class SORT_METHOD { ID, NAME, NUMBER, DEFAULT };
   util::EnumParser<SORT_METHOD> enum_parser_;
   SORT_METHOD sort_by_ = SORT_METHOD::DEFAULT;
   std::map<std::string, SORT_METHOD> sort_map_ = {{"ID", SORT_METHOD::ID},
--- a/include/util/enum.hpp
+++ b/include/util/enum.hpp
@@ -6,32 +6,24 @@
 #include <stdexcept>
 #include <string>
 
+#include "util/string.hpp"
+
 namespace waybar::util {
 
 template <typename EnumType>
 struct EnumParser {
   EnumParser() {}
 
-  // Helper function to capitalize a string
-  std::string capitalizeString(const std::string& str) {
-    std::string result = str;
-    std::transform(result.begin(), result.end(), result.begin(),
-                   [](unsigned char c) { return std::toupper(c); });
-    return result;
-  }
-
   EnumType parseStringToEnum(const std::string& str,
                              const std::map<std::string, EnumType>& enumMap) {
     // Convert the input string to uppercase
-    std::string uppercaseStr = capitalizeString(str);
+    std::string uppercaseStr = capitalize(str);
 
     // Capitalize the map keys before searching
     std::map<std::string, EnumType> capitalizedEnumMap;
-    std::transform(enumMap.begin(), enumMap.end(),
-                   std::inserter(capitalizedEnumMap, capitalizedEnumMap.end()),
-                   [this](const auto& pair) {
-                     return std::make_pair(capitalizeString(pair.first), pair.second);
-                   });
+    std::transform(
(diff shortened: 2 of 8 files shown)


include/modules/hyprland/workspaces.hpp
```diff
@@ -80,10 +80,13 @@ class Workspaces : public AModule, public EventHandler {
   void create_workspace(Json::Value& value);
   void remove_workspace(std::string name);
   void set_urgent_workspace(std::string windowaddress);
+  void parse_config(const Json::Value& config);
+  void register_ipc();
 
   bool all_outputs_ = false;
   bool show_special_ = false;
   bool active_only_ = false;
+  std::string sort_by = "default";
```
zjeffer, line 89: We could instead use an enum and then parse the string into the enum in the constructor.

src/modules/hyprland/workspaces.cpp
```diff
@@ -43,18 +57,19 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value
     active_only_ = config_active_only.asBool();
   }
 
-  box_.set_name("workspaces");
-  if (!id.empty()) {
-    box_.get_style_context()->add_class(id);
+  auto config_sort_by = config_["sort-by"];
+  if (config_sort_by.isString()) {
+    sort_by = config_sort_by.asString();
```
zjeffer, line 62: Convert the string argument to an enum here. Fallback to DEFAULT if given argument is misspelled (with a warning)

src/modules/hyprland/workspaces.cpp
```diff
@@ -14,6 +14,20 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value
     : AModule(config, "workspaces", id, false, false),
       bar_(bar),
       box_(bar.vertical ? Gtk::ORIENTATION_VERTICAL : Gtk::ORIENTATION_HORIZONTAL, 0) {
+  parse_config(config);
```
zjeffer, line 17: Nice cleanup!

src/modules/hyprland/workspaces.cpp
```diff
@@ -406,36 +421,53 @@ void Workspace::update(const std::string &format, const std::string &icon) {
...
-              if (a->id() > 0 && b->id() > 0) {
-                return a->id() < b->id();
-              }
+            [&](std::unique_ptr<Workspace> &a, std::unique_ptr<Workspace> &b) {
+              // Helper comparisons
+              auto is_id_less = a->id() < b->id();
+              auto is_name_less = a->name() < b->name();
+              auto is_number_less = std::stoi(a->name()) < std::stoi(b->name());
+
+              if (sort_by == "number") {
```
zjeffer, line 430: And finally here you can replace the ifs with a switch statement with enums.

zjeffer (changes requested):
Looks great! Just suggesting one minor change.

I don't really have time to test this thoroughly right now, so make sure you test:

* switching between persistent workspaces
* adding new workspaces by name
* switching to special workspaces
* switching to named special workspaces
* making a (named) special workspace persistent, adding and removing the windows from it, and switching between this workspace and another: the special workspace shouldn't change position if the workspace is empty or not.

include/modules/hyprland/workspaces.hpp
khaneliman: Is there another example of a config value using an enum? From what I can see you have to implement mapping logic to convert a string to an enum in C++ since it's not an actual object.

khaneliman:
@zjeffer Updated to use an enum based approach. Sort-by overrides the sorting mechanism and falls back to previous sorting method with a warning if an incorrect value is supplied.

include/modules/hyprland/workspaces.hpp
zjeffer: Yes, you write a function to map a string to the correct enum value. Implementing ostreams is also useful.

include/util/enum.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <iostream>
+#include <map>
+#include <string>
+
+namespace waybar::util {
+
+struct EnumParser {
```
zjeffer, line 9: This is maybe kind of overkill :D

A simple set of if statements in a function in workspaces.cpp is fine, then everywhere else you can use the enums in switch statements.

I'll write an example when I have some more time.

khaneliman: I didn't want to bloat workspaces.cpp with details of how to parse an enum.

include/util/enum.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+  SORT_METHOD sortStringToEnum(const std::string& str) {
+    // Convert the input string to uppercase (make it lenient on config input)
+    std::string uppercaseStr;
+    for (char c : str) {
+        uppercaseStr += std::toupper(c);
+    }
+
+    static const std::map<std::string, SORT_METHOD> enumMap = {
+        {"ID", ID}, {"NAME", NAME}, {"NUMBER", NUMBER}, {"DEFAULT", DEFAULT}};
```
zjeffer, line 22: The problem is also that you're defining a fixed map here: if you want to add a way for other modules to parse enums, you'd have to pass the enum map to the EnumParser constructor.

khaneliman: I mean, this was specific to sort-by. If you want to take a stab at creating a more robust enum system I can just back out those changes and you can have at it.

zjeffer: Here's what I propose: https://github.com/Alexays/Waybar/commit/c628bc780c53c6e5c90cefcaf427cdcd4ab438b9

I'd love to look into a more robust enum parsing system, but indeed better in a different PR :)

khaneliman: I might have something that works like you are mentioning, though. I'll take one last stab at it and if we want to change it we can refine it :P

khaneliman:
@zjeffer new commit with changes  to make it easier for other classes to use

include/util/enum.hpp
```diff
@@ -1,26 +1,25 @@
...
 
+template <typename EnumType>
 struct EnumParser {
   EnumParser() {}
 
-  enum SORT_METHOD { ID, NAME, NUMBER, DEFAULT };
-
-  SORT_METHOD sortStringToEnum(const std::string& str) {
-    // Convert the input string to uppercase (make it lenient on config input)
+  EnumType sortStringToEnum(const std::string& str,
```
zjeffer, line 15: sortStringToEnum -> parseStringToEnum

include/util/enum.hpp
```diff
@@ -1,26 +1,25 @@
...
+                            const std::map<std::string, EnumType>& enumMap) {
+    // Convert the input string to uppercase
     std::string uppercaseStr;
     for (char c : str) {
-        uppercaseStr += std::toupper(c);
+      uppercaseStr += std::toupper(c);
```
zjeffer, line 20: You can use std::transform for this instead of a for loop

include/util/enum.hpp
```diff
@@ -1,26 +1,25 @@
...
     }
 
-    static const std::map<std::string, SORT_METHOD> enumMap = {
-        {"ID", ID}, {"NAME", NAME}, {"NUMBER", NUMBER}, {"DEFAULT", DEFAULT}};
-
     auto it = enumMap.find(uppercaseStr);
     if (it != enumMap.end()) {
       return it->second;
```
zjeffer, line 25: on the next line, you can remove the `else {` because you're returning here.

include/modules/hyprland/workspaces.hpp
```diff
@@ -87,8 +87,14 @@ class Workspaces : public AModule, public EventHandler {
   bool all_outputs_ = false;
   bool show_special_ = false;
   bool active_only_ = false;
-  util::EnumParser enum_parser_;
-  util::EnumParser::SORT_METHOD sort_by_ = util::EnumParser::SORT_METHOD::DEFAULT;
+
+  enum SORT_METHOD { ID, NAME, NUMBER, DEFAULT };
```
zjeffer, line 91: use enum class: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Renum-class

zjeffer (changes requested):
Great, just some small suggestions left:

include/util/enum.hpp
zjeffer: and then you can also remove the {} of the else

khaneliman:
> Great, just some small suggestions left:

Sounds good, I also realized I now need to make sure the keys are uppercase so i'll add a commit to make that work since people can provide their own keys. Just trying to make it so you dont have to worry about case when a person is writing their config.

include/util/enum.hpp
khaneliman: not sure I understand the ask here. The if else is just for determining if they passed in a valid setting and whether to throw an invalid argument exception to bubble up to the caller to determine how to handle

khaneliman: nvm... you mean if it never hits the return it will throw the error.

zjeffer (approved):

khaneliman:
@zjeffer appreciate the thorough review... learned a few things :)

include/util/enum.hpp
```diff
@@ -12,12 +12,26 @@ template <typename EnumType>
 struct EnumParser {
   EnumParser() {}
 
+  // Helper function to capitalize a string
```
zjeffer, line 15: Two more things:

* I would put this function in a separate file "utils/strings.hpp"
* It's better to declare the functions in the hpp, but implement them in a cpp file, otherwise compilation might take longer: https://stackoverflow.com/a/1809753/10180569. It's not really a problem right now (because we're only including it in one place), but it's good practice to change it.

zjeffer (changes requested):

include/util/enum.hpp
khaneliman: That makes me kinda want to go through and clean up all the other poor separations if it can affect that much during compilation.

khaneliman:
@zjeffer Alright, handled those last couple suggestions.

include/util/string.hpp
zjeffer, line 1: You can also split up string.hpp into a separate cpp file :)

khaneliman: yeah, that can be a separate refactor pr lol

khaneliman: i saw a ton of headers that were the implementations that need to be refactored

zjeffer (approved):

khaneliman:
@Alexays This one should be ready, now.

Alexays:
LGTM, thx to both of you!

Issue #2475, opened 2023-09-07
hyprland/workspaces do not work


Nyayurn:
![image](https://github.com/Alexays/Waybar/assets/132195516/391ff21a-13da-461e-b713-c03670154853)
![image](https://github.com/Alexays/Waybar/assets/132195516/d9f37668-3585-49b0-8076-6e000a05caa3)

Dungeonseeker:
Add the following to the bottom of ~/.config/waybar/config....

```
    "hyprland/workspaces": {
        "disable-scroll": true,
		"format": "{name} {icon}",
		"format-icons": {
			"active": " ",
			"default": " "
		},
        "persistent_workspaces": {
            "1": [],
            "2": [],
            "3": [],
            "4": [],
        },
    },
```

zjeffer:
@Nyayurn What's your actual hyprland/workspaces config? We can only see you added the "modules/left": ["hyprland/workspaces"] module, but not the config itself. Also please copy paste instead of screenshotting, makes it easier for other people to find similar issues.

Nyayurn:
> @Nyayurn What's your actual hyprland/workspaces config? We can only see you added the `"modules-center": ["hyprland/workspaces"]` module, but not the config itself. Also please copy paste instead of screenshotting, makes it easier for other people to find similar issues.

none or
```
"hyprland/workspaces": {}
```
or


> Add the following to the bottom of ~/.config/waybar/config....
> 
> ```
>     "hyprland/workspaces": {
>         "disable-scroll": true,
> 		"format": "{name} {icon}",
> 		"format-icons": {
> 			"active": " ",
> 			"default": " "
> 		},
>         "persistent_workspaces": {
>             "1": [],
>             "2": [],
>             "3": [],
>             "4": [],
>         },
>     },
> ```

as the same error

zjeffer:
Which version of Waybar are you using?

celepharn:
Changing wlr/workspaces to hyprland/workspaces fixed it for me.

Nyayurn:
> Which version of Waybar are you using?

extra/waybar 0.9.22-1

zjeffer:
That's so strange. Can you copy paste your entire config? Both config.json and style.css.

Nyayurn:
[config](https://github.com/Alexays/Waybar/files/12565287/config.txt)
[style.css](https://github.com/Alexays/Waybar/files/12565288/style.css.txt)

zjeffer:
[...]

PR #2496, opened 2023-09-10
Fix JSON parsing with hexadecimal characters

### LINKED ISSUE #2495: The jsonparser doesn't work with escaped unicode characters
I wrote a test for this:

```cpp
TEST_CASE("Json with unicode", "[json]") {
  SECTION("Parse json with unicode") {
	std::string stringToTest = R"({"test": "\xab"})";
	waybar::util::JsonParser parser;
	Json::Value jsonValue = parser.parse(stringToTest);
	REQUIRE(jsonValue["test"].asString() == "\xab");
  }
}
```

```
../test/JsonParser.cpp:20: FAILED:
due to unexpected exception with message:
  * Line 1, Column 10
    Bad escape sequence in string
  See Line 1, Column 13 for detail.
```

causes problems with hyprland IPC if for example the monitor description/model contains unicode characters: #2475 

I'm not sure if this is a problem with our parsing code (include/util/json.hpp), or a problem with the upstream jsoncpp library we're using?

zjeffer:
Fixes #2495 and #2475

The JSON spec doesn't allow characters like `\xab`. Instead, it should be converted to unicode: `\u00ab`. 

It's my first time writing catch2 tests, let me know if I did something wrong.

### DIFF
--- a/include/util/json.hpp
+++ b/include/util/json.hpp
@@ -3,6 +3,12 @@
 #include <fmt/ostream.h>
 #include <json/json.h>
 
+#include <algorithm>
+#include <codecvt>
+#include <iostream>
+#include <locale>
+#include <regex>
+
 #if (FMT_VERSION >= 90000)
 
 template <>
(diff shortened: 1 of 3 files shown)


zjeffer:
@Nyayurn can you test if you still have your bug with this new parser?

Nyayurn:
> @Nyayurn can you test if you still have your bug with this new parser?

It's fine

Nyayurn:
image

But it's a new bug with clock, in screenshot's right top the clock compoment do not work

zjeffer:
Can you show the console output, I'd like to see if there are any errors from the clock component.

Nyayurn:
image
and the config
image

zjeffer:
@Nyayurn with your config, I get the same error on latest master. 
Something seems to be wrong with your format string..

Nyayurn:
> @Nyayurn with your config, I get the same error on latest master. Something seems to be wrong with your format string..
> 
> I think the clock icon needs to be outside of the {} brackets, like this: `"<icon>{:%I:%M %p %d/%m/%Y}"`
> 
> It works for me on both latest master and this PR:
> 
> image

Thanks, It's work now

zjeffer:
@Alexays This is ready to be reviewed.

Alexays:
LGTM, thx for this fix!

PR #2507, opened 2023-09-13
Update workspaces.cpp

### LINKED ISSUE #2501: hyprland/workspaces: module gets disabled.
Waybar v0.9.22-63-ga90e275d (branch 'master')

when launching hyprland from the tty, waybar isn't even displayed. when inside hyprland this is the message i was able to retrieve (ofc hyprland/workspaces doesn't work)

```
[2023-09-12 20:29:32.248] [info] Using configuration file /home/pbo/.config//waybar/config
[2023-09-12 20:29:32.250] [info] Unable to receive desktop appearance: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop
[2023-09-12 20:29:32.250] [info] Using CSS file /home/pbo/.config//waybar/style.css
[2023-09-12 20:29:32.255] [info] Hyprland IPC starting
[2023-09-12 20:29:32.257] [warning] module hyprland/workspaces: Disabling module "hyprland/workspaces", stoi

(waybar:20019): GLib-GObject-CRITICAL **: 20:29:32.747: ../glib/gobject/gsignal.c:2088: type 'GtkWindow' is already overridden for signal id '73'

(waybar:20019): GLib-GObject-CRITICAL **: 20:29:32.747: ../glib/gobject/gsignal.c:2088: type 'GtkWindow' is already overridden for signal id '72'
[2023-09-12 20:29:32.773] [warning] module hyprland/workspaces: Disabling module "hyprland/workspaces", stoi
[2023-09-12 20:29:33.192] [info] Bar configured (width: 2560, height: 17) for output: HDMI-A-1
[2023-09-12 20:29:33.192] [info] Bar configured (width: 1920, height: 17) for output: DVI-D-1
```

nothing changed inside my config since the last use, but could post config if needed.

hariseldon78:
Fix unchecked string to int conversion of workspace name (which can be a string) 
Closes #2501

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -439,7 +439,6 @@ void Workspaces::sort_workspaces() {
               // Helper comparisons
               auto is_id_less = a->id() < b->id();
               auto is_name_less = a->name() < b->name();
-              auto is_number_less = std::stoi(a->name()) < std::stoi(b->name());
 
               switch (sort_by_) {
                 case SORT_METHOD::ID:
@@ -448,7 +447,7 @@ void Workspaces::sort_workspaces() {
                   return is_name_less;
                 case SORT_METHOD::NUMBER:
                   try {
-                    return is_number_less;
+                    return std::stoi(a->name()) < std::stoi(b->name());
                   } catch (const std::invalid_argument &) {
                     // Handle the exception if necessary.
                     break;


Alexays:
Thanks!

src/modules/hyprland/workspaces.cpp
```diff
@@ -448,7 +447,7 @@ void Workspaces::sort_workspaces() {
                   return is_name_less;
                 case SORT_METHOD::NUMBER:
                   try {
-                    return is_number_less;
+                    return std::stoi(a->name()) < std::stoi(b->name());
```
zjeffer, line 450: I think this can be improved more by changing the 3 helper comparisons to be lambdas instead of boolean variables. 

Like this:

```cpp
auto is_id_less = [a, b]() { a->id() < b->id(); };
...
```

We can then easilly use: 

`return is_number_less()` (with the brackets, because it's a function)

hariseldon78: sure. I'm a bit rusty on c++ so i'm quite slow, but i want to make a PR to the Hyprland project enabling persistent workspaces from the compositor itself (if somebody don't beat me to it, it's all very active!), so i'll probably also take a better look at this part of waybar if there is space for improvement. 
That fix I made was done in a real rush!

PR #2517, opened 2023-09-18
Lighter Weight Signal-Based Custom Modules

an-prata:
These changes are intended to fix a performance problem I had on my laptop where using signals on my custom modules would cause excess CPU usage and battery drainage.

This PR makes it so that any custom module with an undefined `interval` but a defined `signal` will start up a thread with the new `Custom::waitingWorker()` function. This is identical to an `interval` based update with signals but will instead sleep indefinitely using a new `SleeperThread::sleep()` function.

Aside from making this a usable combination of config options this is higher performance than both the `"once"` option and any given integer interval. Full disclaimer, I am not an optimizations expert but waybar's CPU usage went from between 0.4% and 0.6% (sometimes getting much higher but I haven't recorded that and it might not actually be waybar's fault) to between 0.0% and 0.1% with these additions (CPU usage recorded with btop++).

Please tell me if anything should be changed, thanks! :D

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -22,6 +22,7 @@ class Custom : public ALabel {
  private:
   void delayWorker();
   void continuousWorker();
+  void waitingWorker();
   void parseOutputRaw();
   void parseOutputJson();
   void handleEvent();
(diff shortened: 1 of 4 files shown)


Alexays:
Lgtm, can you add some info about this change in man and wiki?

an-prata:
I added some info to the man page in the PR, since github wikis do not (that I know of) have a way to pull request here is a gist with the markdown for the custom module page: https://gist.github.com/an-prata/4395349af6913717876a16eca26176f2

Alexays:
You can directly edit the wiki :)

PR #2518, opened 2023-09-19
fix(backlight): wrong percentage numbers for device apple-panel-bl

idm1try:
fixes backlight module showings incorrect percentage(eg 101% instead of 100% or 47% instead of 45%) on asahi linux
we need use `brightness` not `actual_brightness`

### DIFF
--- a/src/modules/backlight.cpp
+++ b/src/modules/backlight.cpp
@@ -244,7 +244,7 @@ void waybar::modules::Backlight::upsert_device(ForwardIt first, ForwardIt last, 
   check_nn(name);
 
   const char *actual_brightness_attr =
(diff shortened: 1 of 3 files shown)


src/modules/backlight.cpp
```diff
@@ -244,7 +244,7 @@ void waybar::modules::Backlight::upsert_device(ForwardIt first, ForwardIt last,
   check_nn(name);
 
   const char *actual_brightness_attr =
-      strncmp(name, "amdgpu_bl", 9) == 0 ? "brightness" : "actual_brightness";
+      strncmp(name, "amdgpu_bl", 9) == 0 ? "brightness" : strcmp(name, "apple-panel-bl") == 0 ? "brightness" : "actual_brightness";
```
Alexays, line 247:
```suggestion
      strncmp(name, "amdgpu_bl", 9) == 0 || strcmp(name, "apple-panel-bl") == 0  ? "brightness" : "actual_brightness";
```

Issue #2533, opened 2023-09-23
Hyprland workspaces how to format with icon and set id as fallback


woojiq:
Config: https://github.com/woojiq/nixos-config/blob/main/modules/programs/waybar.nix#L41-L58
```nix
# NIX
      "wlr/workspaces" = {
        format = "<span font='13'>{icon}</span>";
        format-icons = {
          "1" = "";
          "2" = "";
        };
        active-only = false;
        on-click = "activate";
      };
```
When I was using wlr/workspaces I had this behavior where I only defined the first 2-3 icons and all other workspaces were displayed with a number (id). When I updated my system and probably updated hyprland, waybar, etc, my workspaces were no longer showing, so I switched to hyprland/workspaces module and now I don't have the desired behavior. With the same setup I've shown, all my subsequent workspaces, starting with 3 and inclusive, are marked with the same icon as the last icon defined (in my case the second one). Something like `format="{id}{icon}"` is really ugly. I want to set the `{id}` format as a fallback if there is no icon defined. Is it possible with hyprland module? It was possible with wlr module.

PR #2534, opened 2023-09-23
hyprland/workspaces: use `name` as fallback icon

woojiq:
Wlr and Sway modules use the workspace `name` as the default icon if no icon is provided. This adds the same behavior for the `hyprland/workspace` module. 

Closes https://github.com/Alexays/Waybar/issues/2533

Before this fix, it defaulted to the last defined icon, which is silly (see above issue).

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -549,7 +549,8 @@ std::string &Workspace::select_icon(std::map<std::string, std::string> &icons_ma
   if (default_icon_it != icons_map.end()) {
     return default_icon_it->second;
   }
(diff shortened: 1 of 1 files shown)


woojiq:
Wlr impl: https://github.com/Alexays/Waybar/blob/6997b34a816afea1e429e037249a9d67c38d35f7/src/modules/wlr/workspace_manager.cpp#L519
Sway impl: https://github.com/Alexays/Waybar/blob/6997b34a816afea1e429e037249a9d67c38d35f7/src/modules/sway/workspaces.cpp#L326

zjeffer (approved):

PR #2560, opened 2023-10-09
fix(modules/mpris): fix on-*-click typos

akshettrj:
In all other places, the norm is to use `on-click-(middle|right)` but in the mpris module, `on-(middle|right)-click` was being used which caused clicks to malfunction if set to some custom commands


khaneliman (approved):

khaneliman:
Always good to have the consistency. Thanks.

PR #2529, opened 2023-09-23
Feature: Hyprland dynamic window names on workspaces

Syndelis:
## About this PR

This PR introduces 3 new configurations for `hyprland/workspaces`, as well as a new formatter for `format`. These can be used to display all of a workspace's currently opened windows.

## Motivation

Unlike actual disciplined people, I can't seem to "assign responsabilities" to my workspaces, and tend to use them chaotically. This means that I frequently forget where I've left my browser or my VS Code windows and have to search for them.

As I'm aware I'm not alone in this, I ported over a functionality I had in my [Eww](https://github.com/elkowar/eww) setup that allows you to display a workspace's content dynamically.

## About the implementation

We've added the `windows_map_` attribute to the `Workspace` class, mapping a window's address (currently represented by a string) to its user-facing representation. The representation is computed by matching the window's class to any of the rules in `window-rewrite`, and defaults to `window-rewrite-default` if there are no matches. Once a given class has had its representation computed once, it will be stored in `Workspaces::regex_cache_`, meaning that new windows of the same class won't need to go through all of the rules again.

Windows are inserted and removed from `Workspace::windows_map_` depending on the event received in `Workspaces::onEvent`. The following events can trigger changes in this mapping (as per [Hyprland's IPC documentation](https://wiki.hyprland.org/hyprland-wiki/pages/IPC/)):

- `openwindow`: This event signals that a new windows has been opened, and accompanies data detailing the window's address, workspace, class and title;
- `closewindow`: This event signals that a window has been closed, and accompanies the window's address;
- `movewindow`: This event signals that a window has been moved to another workspace, and accompanies the window's address and its new workspace;

Additionally, on startup, a call to Hyprland's `clients` JSON is made to populate the initial state of windows.

## Examples 

<details>

<summary><b>Check out an example</b></summary>

image

</details>

<details>

<summary><b>Config options</b></summary>

```json
"hyprland/workspaces": {
  "format": "<sub>{icon}</sub>\n{windows}",
  "format-window-separator": "\n",
  "window-rewrite-default": "",
  "window-rewrite": {
    "firefox": "",
    "foot": "",
    "code": "󰨞",
    ...
  },
  ...
}
```

</details>

PR #2563, opened 2023-10-09
Feat/Hyprland: Allow matching against Window Titles for Rewrite

### LINKED ISSUE #2561: Feature request: in hyprland/workspaces allow also for window title and regex
I think that the recent commit on Hyprland dynamic window names on workspaces (58e506a6754fae213e409bcd079d38d03f70b958) by @Syndelis was very cool. It works very well in my setup.  If I understood correctly, at the moment it is only possible to use the window class to set the icon. Would it be possible/easy to add also the window title and/or allow for regex strings? I am sorry that I cannot help with this as I am not really proficient in coding.

Syndelis:
Resolves #2561 

## About this PR

This PR builds upon #2529 to introduce matching against a window's title in order to get its icon. In order to do that, the function that matches against regexes will now be supplied a string in this format: `"class<WINDOW_CLASS> title<WINDOW_TITLE>"`, as opposed to just the class. This way, users may provide rules in the following forms:

```json
"window-rewrite": {
  "title<.*youtube.*>": "", // Windows whose titles contain "youtube"
  "class<firefox>": "", // Windows whose classes are "firefox"
  "class<firefox> title<.*github.*>": "", // Windows whose class is "firefox" and title contains "github". Note that "class" always comes first.
  "foot": "", // Windows that contain "foot" in either class or title. For optimization reasons, it will only match against a title if at least one other window explicitly matches against a title.
}
```

Rules are now also prioritized, from more specilized to least. In the previous example, the priority order would be:
1. GitHub — Contains both class and title matching;
2. YouTube — Matches against a title, which is preferred over simply a class;
3. FireFox — Explicitly matches against a class;
4. Foot — Doesn't explicitly match against anything and, thus, is considered low priority. This prevents having your browser be displayed as a terminal if you search for "foot" in FireFox, for example.

## Backwards Compatibility

Because of the possible problem pointed out in **4.**, matching against titles is disabled **unless** at least one rule explicitly matches against titles. If no rules do so, titles are not sent to the aforementioned function, and, instead, the string `"class<WINDOW_CLASS>"` is used.

This does mean, however, that rules written as `"^something_here$"` will not work anymore, and should be replace by either `"somethind_here"`, `"class<something_here>"` or `"^class<something_here>$"`. While I don't believe anyone might've written such a specific rule, I'm open to hear the reasoning for that and work out something that doesn't break this niche use case.

## Implementation Details

A refactor has been made to separate the logic behind the regex map, its cache and default values. This whole logic is now available as the `waybar::util::RegexCollection` class, which can be constructed from a `Json::Value`. This class also provides a way to sort rules, giving priority to whichever rule scores higher in the parameter function.

When titles are used in any rules, the workspace module will be subscribed to an additional event called `windowtitle`, which, [according to Hyprland's IPC documentation](https://wiki.hyprland.org/IPC/), only features the window's address as a parameter. This means that an additional call to Hyprland's `"clients"` socket1 must be made in order to retrieve the window's latest name.

The module is not subscribed to this event when titles aren't explicitly used simply because it would be of no use and could break compatibility as mentioned above.

### DIFF
--- a/src/modules/disk.cpp
+++ b/src/modules/disk.cpp
@@ -49,9 +49,9 @@ auto waybar::modules::Disk::update() -> void {
   float specific_free, specific_used, specific_total, divisor;
 
   divisor = calc_specific_divisor(unit_);
-  specific_free = (stats.f_bavail * stats.f_frsize)/divisor;
-  specific_used = ((stats.f_blocks - stats.f_bfree) * stats.f_frsize)/divisor;
-  specific_total = (stats.f_blocks * stats.f_frsize)/divisor;
+  specific_free = (stats.f_bavail * stats.f_frsize) / divisor;
+  specific_used = ((stats.f_blocks - stats.f_bfree) * stats.f_frsize) / divisor;
+  specific_total = (stats.f_blocks * stats.f_frsize) / divisor;
 
   auto free = pow_format(stats.f_bavail * stats.f_frsize, "B", true);
   auto used = pow_format((stats.f_blocks - stats.f_bfree) * stats.f_frsize, "B", true);
@@ -72,7 +72,7 @@ auto waybar::modules::Disk::update() -> void {
         fmt::runtime(format), stats.f_bavail * 100 / stats.f_blocks, fmt::arg("free", free),
         fmt::arg("percentage_free", stats.f_bavail * 100 / stats.f_blocks), fmt::arg("used", used),
         fmt::arg("percentage_used", percentage_used), fmt::arg("total", total),
-        fmt::arg("path", path_), fmt::arg("specific_free", specific_free), 
+        fmt::arg("path", path_), fmt::arg("specific_free", specific_free),
         fmt::arg("specific_used", specific_used), fmt::arg("specific_total", specific_total)));
   }
 
@@ -85,7 +85,7 @@ auto waybar::modules::Disk::update() -> void {
         fmt::runtime(tooltip_format), stats.f_bavail * 100 / stats.f_blocks, fmt::arg("free", free),
         fmt::arg("percentage_free", stats.f_bavail * 100 / stats.f_blocks), fmt::arg("used", used),
         fmt::arg("percentage_used", percentage_used), fmt::arg("total", total),
-        fmt::arg("path", path_), fmt::arg("specific_free", specific_free), 
+        fmt::arg("path", path_), fmt::arg("specific_free", specific_free),
         fmt::arg("specific_used", specific_used), fmt::arg("specific_total", specific_total)));
   }
   // Call parent update
@@ -109,7 +109,7 @@ float waybar::modules::Disk::calc_specific_divisor(std::string divisor) {
     return 1000.0 * 1000.0 * 1000.0 * 1000.0;
   } else if (divisor == "TiB") {
     return 1024.0 * 1024.0 * 1024.0 * 1024.0;
-  } else { //default to Bytes if it is anything that we don't recongnise
+  } else {  // default to Bytes if it is anything that we don't recongnise
     return 1.0;
   }
 }
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -48,7 +48,7 @@ class Workspace {
   void set_visible(bool value = true) { is_visible_ = value; };
   void set_windows(uint value) { windows_ = value; };
   void set_name(std::string value) { name_ = value; };
-  bool contains_window(WindowAddress addr) { return window_map_.contains(addr); }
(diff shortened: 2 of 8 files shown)


src/modules/disk.cpp
Syndelis, line 1: This file is unrelated to this PR and was simply linted it in f8340d8

khaneliman:
Looks like freebsd check failed to build properly.

Syndelis:
> Looks like freebsd check failed to build properly.

Yeah, my last PR #2568 did too. Unfortunately I can't do much to help it; the logs don't mention the reason it failed, it just goes from "Copying files" to "/bin/bash failed". Would need to re-run it with debug logging enable, but I don't have permissions for that.

khaneliman:
In file included from /usr/include/c++/v1/__functional/function.h:20:
  /usr/include/c++/v1/__memory/allocator_traits.h:298:9: error: no matching function for call to 'construct_at'
          _VSTD::construct_at(__p, _VSTD::forward<_Args>(__args)...);
          ^~~~~~~~~~~~~~~~~~~
  /usr/include/c++/v1/__config:826:15: note: expanded from macro '_VSTD'
  #define _VSTD std
                ^
  /usr/include/c++/v1/vector:778:21: note: in instantiation of function template specialization 'std::allocator_traits<std::allocator<waybar::util::Rule>>::construct<waybar::util::Rule, const std::regex &, std::string, int &, void, void>' requested here
      __alloc_traits::construct(this->__alloc(), _VSTD::__to_address(__tx.__pos_),
                      ^
  /usr/include/c++/v1/vector:1577:9: note: in instantiation of function template specialization 'std::vector<waybar::util::Rule>::__construct_one_at_end<const std::regex &, std::string, int &>' requested here
          __construct_one_at_end(_VSTD::forward<_Args>(__args)...);
          ^
  ../src/util/regex_collection.cpp:24:15: note: in instantiation of function template specialization 'std::vector<waybar::util::Rule>::emplace_back<const std::regex &, std::string, int &>' requested here
          rules.emplace_back(rule, it->asString(), priority);
                ^
  /usr/include/c++/v1/__memory/construct_at.h:36:16: note: candidate template ignored: substitution failure [with _Tp = waybar::util::Rule, _Args = <const std::regex &, std::string, int &>]: no matching constructor for initialization of 'waybar::util::Rule'
  constexpr _Tp* construct_at(_Tp* __location, _Args&& ...__args) {
                 ^
  1 error generated.

Syndelis:
> This is what I see looking at the checks logs under` Run 'run' in VM`
> 
> ```
>   In file included from /usr/include/c++/v1/__functional/function.h:20:
>   /usr/include/c++/v1/__memory/allocator_traits.h:298:9: error: no matching function for call to 'construct_at'
>           _VSTD::construct_at(__p, _VSTD::forward<_Args>(__args)...);
>           ^~~~~~~~~~~~~~~~~~~
>   /usr/include/c++/v1/__config:826:15: note: expanded from macro '_VSTD'
>   #define _VSTD std
>                 ^
>   /usr/include/c++/v1/vector:778:21: note: in instantiation of function template specialization 'std::allocator_traits<std::allocator<waybar::util::Rule>>::construct<waybar::util::Rule, const std::regex &, std::string, int &, void, void>' requested here
>       __alloc_traits::construct(this->__alloc(), _VSTD::__to_address(__tx.__pos_),
>                       ^
>   /usr/include/c++/v1/vector:1577:9: note: in instantiation of function template specialization 'std::vector<waybar::util::Rule>::__construct_one_at_end<const std::regex &, std::string, int &>' requested here
>           __construct_one_at_end(_VSTD::forward<_Args>(__args)...);
>           ^
>   ../src/util/regex_collection.cpp:24:15: note: in instantiation of function template specialization 'std::vector<waybar::util::Rule>::emplace_back<const std::regex &, std::string, int &>' requested here
>           rules.emplace_back(rule, it->asString(), priority);
>                 ^
>   /usr/include/c++/v1/__memory/construct_at.h:36:16: note: candidate template ignored: substitution failure [with _Tp = waybar::util::Rule, _Args = <const std::regex &, std::string, int &>]: no matching constructor for initialization of 'waybar::util::Rule'
>   constexpr _Tp* construct_at(_Tp* __location, _Args&& ...__args) {
>                  ^
>   1 error generated.
> ```

My bad, I was looking at the wrong step. I'll take a look at it

Syndelis:
Apparently the pipeline runs with Clang 14, which still didn't support [direct initialization](https://en.cppreference.com/w/cpp/feature_test#cpp_aggregate_paren_init)

khaneliman:
Tested locally seems to work, as intended. @Alexays should be good to go

khaneliman (approved):

Alexays:
Thanks!

PR #2568, opened 2023-10-11
Feat/Hyprland/Workspaces: Ignore workspaces

Syndelis:
## About this PR

This PR introduces a new option to `hyprland/workspaces` called `ignore-workspaces`. This option is an array of regexes that will be matched against workspace names. If there's a match, the workspace being created will not be shown on the user's bar.

## Motivation

Chrome is notorious for launching a "sharing indicator" when your screen is being shared. This is specially frustrating when using tiling managers, such as Hyprland. 

image

My current solution for not having to see it is by forcing it to move to a workspace named `special:hell`. This PR, then, lets me add it to a ignored list of workspaces, with the following rule:

```json
"hyprland/workspaces": {
        "ignore-workspaces": [
            "(special:)?hell"
        ]
}
```

*(note that I'm optionally matching against "special:" because Hyprland is not very consistent when it comes to naming special workspaces)*

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -9,6 +9,7 @@
 #include <map>
 #include <memory>
 #include <optional>
+#include <regex>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM, can you also update the GitHub wiki?
Thanks!

Syndelis:
Done!

Readf0x:
I'm actually lmao right now. Naming the special workspace hell is absolute comedy gold. I am stealing that immediately lol. Thank you for this.

PR #2573, opened 2023-10-14
C ABI plugin system

CromFr:
Hi !
I played around with waybar a bit, and developed a quick proof of concept for a C ABI plugin system, that could allow people to develop more complex waybar modules, without modifying waybar itself.

The idea is to have waybar allocate a GTK container for the cabi plugin, and send its pointer to the plugin, that takes care of populating the widget as needed. The C ABI is very handy because GTK is in C and almost most languages can generate a C ABI (C++, rust, python, ...) and call GTK functions.

My end goal is to develop a custom system monitor module, that displays CPU, IO and network activity in a very compact manner (eventually with graphs) !

As I said, this currently a PoC and needs to be worked on to improve the ABI, add more features and config options, ... 
I would like to get your opinion on the C ABI idea, and if it is worth spending more time on this PR

### To try it out:
1. Compile the plugin dynamic library: `cd cabi_example && meson build && meson compile -C build`
2. Copy this waybar config:
```json
{
    "modules-center": ["cabi/c_example"],
    "cabi/c_example": {"path": "cabi_example/build/libwaybar_cabi_c_example.so"}
}
```
3. Run waybar `./build/waybar -c cabi_config`


Alexays:
This could be a really nice addition!
Would love to see how far we can get in this direction 😃

CromFr:
Sorry for the delay, I got carried away by other projects

I think this PR is ready to be reviewed / merged !

The current ABI may not be perfect yet, but it's definitely usable and can always be improved later using feedback from custom module devs.
 
There's an example module located in `resources/custom_modules/cffi_example/`, written in C. I also started writing my own plugin in Rust for displaying CPU graphs in a GtkDrawingArea, but it's nowhere near ready. 

Some key points for maintaining and upgrading the CFFI ABI:
- Waybar detects the ABI version used by the plugin with the `wbcffi_version` constant. Waybar should be able to handle different ABI versions by wrapping the raw function pointers with std::function (or do some later logic using the version number) 
- The ABI version should only be incremented with breaking changes, i.e. when modifying existing symbols or requiring new symbols. You can add additional optional hooks or append new members to wbcffi_init_info without incrementing the ABI version number.
- Plugins can require specific waybar versions (e.g. when a new optional hook has been added without incrementing the ABI version) by checking the waybar version during plugin initialization

Alexays:
Nice, look pretty simple for a first iteration, thx!
Can you also add a man to redirect the user to the example files and possible module conf?
And aswell for the Github wiki :)

CromFr:
Done !

The file resources/custom_modules/cffi_example/waybar_cffi_module.h documents the symbols to define for a module to be loaded. I'm not sure it should be stored in the example directory, but I'm not sure where it belongs to

PR #2578, opened 2023-10-16
Fix: Hyprland/Workspaces' windows disappearing when moving around

### LINKED ISSUE #2557: Hyprland/workspaces: window rewrite issues
Thought https://github.com/Alexays/Waybar/pull/2529 was a cool feature introduced so tried setting it up like this https://github.com/khaneliman/khanelinix/commit/9ab7e19c3a5dd4026047045a9837c9bd53780477 but ran into issues with window-rewrite not taking effect on apps launched from hyprland startup and secondary bar not getting updated.

Another thing that might be nice is implementing a way to declare window-rewrite rule default map or a way to look up icons that makes it more automated to setup without having to declare every icon in your personal config and using the personal config as an override.

Syndelis:
Fixes #2557

## About this PR

This PR fixes an issue that caused windows to disappear. This was caused because, sometimes, Hyprland's IPC decides to report the window-related event (`openwindow`, `movewindow`) before the `createworkspace` event. This, in turn, caused Waybar to have nowhere to put the moved/opened window, causing it to be discarted.

grim_2023-10-16T16:27:06-03:00

## About the solution

In order to solve this problem, the window creation has been delayed, much akin to how workspaces are created; they are pushed into a vector and are processed during `Workspaces::update()`. There will be 3 attempts to create each window, accounting for the possibility that the `createworkspace` message got *just* late enough that it wasn't processed in the same `update()` call.

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -11,6 +11,7 @@
 #include <optional>
 #include <regex>
 #include <string>
+#include <variant>
 #include <vector>
 
 #include "AModule.hpp"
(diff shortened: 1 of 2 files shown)


khaneliman (approved):
Awesome fixes, appreciate all your effort! Feel like this is in a much more stable and functional state for persistent workspace users. Only 1 remaining bug that is more niche.

Alexays:
Thanks a lot again!

Issue #2167, opened 2023-05-10
request for waybar-network interface priority and exclusion


jigish:
I use Wi-Fi, four different ethernet interfaces, and VPN. Here are a few issues I've found:

1. When connected via WiFi and Ethernet, I would like to prioritize showing the Ethernet interface over WiFi without needing to display two network modules. Currently if I'm connected to WiFi when I plug in Ethernet, the network module does not change interface to Ethernet (even though traffic will be routed through Ethernet)
2. If I split network into `network#enp` and `network#wlp` with wildcard matching for the interface (`enp*` and `wlp*`), the wrong Ethernet interface is chosen. It should automatically select a connected one, not a disconnected one that is alphanumerically first.
3. If I try to use one network module and am on VPN, `tun0` will be shown. I don't care about `tun0` except for the fact that it exists. I can solve this by specifying interface but there is no way currently to _exclude_ an interface which would be useful to exclude `tun0` and create a new `network#tun0` to display a lock if I'm on VPN.

I think this can be solved by either adding an interface priority configuration or changing how interfaces are chosen (always choose connected over disconnected) or preferably both.

31PNMR:
I just noticed this as well when I switched to Hyprland on my laptop last night. I use `NetworkManager` which sets the priority of connections. I wonder if it would be easy enough to simply look at what device it has set as priority.

jinnko:
I have similar issues and was wondering how the interface is chosen.  I was expecting that the interface being used as the primary route should be displayed, and I found what [seems like this approach in the code](https://github.com/Alexays/Waybar/blob/master/src/modules/network.cpp#L659) (or am I reading it wrong?).

As an example, I have wifi & ethernet both connected, and each has a default route, with the ethernet being the preferred default route (specifically a lower metric value).  In this situation the module is displaying the wifi interface, however I would expect the ethernet interface to be shown.

Here's the relevant parts of the routing table.

```
$ ip route | grep default
default via 192.168.1.1 dev enp6s0f4u2 proto dhcp src 192.168.1.64 metric 100 
default via 192.168.1.1 dev wlp3s0 proto dhcp src 192.168.1.81 metric 600

$ ip -6 route | grep default
default via fe80::2211:33ff:fe11:ec2e dev enp6s0f4u2 proto ra metric 100 pref medium
default via fe80::2211:33ff:fe11:ec2e dev wlp3s0 proto ra metric 20600 pref medium
```

[...]

PR #2585, opened 2023-10-20
Show the network interface with highest priority (aka lower metric)

davxy:
On new interface discovery a new index is (may be) assigned to the `temp_idx`.

The new interface should be assigned to `net` if is the first one (`net->ifid_ == -1`) or if has higher priority (`priority < net->route_priority`... lower is higher according to metric).

This PR removes the additional condition `net->ifid_ != temp_idx` which is always be true and incorrectly ends up showing the last found route instead of the one with higher priority

Closes https://github.com/Alexays/Waybar/issues/2167

### DIFF
--- a/src/modules/network.cpp
+++ b/src/modules/network.cpp
@@ -656,7 +656,7 @@ int waybar::modules::Network::handleEvents(struct nl_msg *msg, void *data) {
         checking route id
         **/
         if (!is_del_event &&
-            ((net->ifid_ == -1) || (priority < net->route_priority) || (net->ifid_ != temp_idx))) {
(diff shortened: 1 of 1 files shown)


Alexays:
We should check why we introduced that condition with this commit.
https://github.com/Alexays/Waybar/commit/a4d27ea8067a1767dfe1cdc526f5170d515fe226

davxy:
If I'm not wrong like you are the author of that modification right 😃.

However at the moment is broken as it doesn't show the interface which is effectively used

Alexays:
Yep, but i don't remember why :D
Thanks, let's merge it!

khaled4vokalz:
did we release this fix?

PR #2586, opened 2023-10-20
Fix potential memory leaks

taminob:
While looking through the source code, I found a few potential memory leaks.

Most of them were caused by early returns (e.g. because of an exception) without freeing a resource first.
To prevent those and make the code more robust, I added a `scope_guard` which will ensure via RAII that the resources are freed.
Additionally, I added them in a few places where `GError`s are in use and where I think there might be a risk in the future.

Instead of this custom `scope_guard` it could also be considered to use a third-party library but I think it's easier to provide this simple functionality in `util` instead of integrating yet another dependency.

I tried to group the different kinds of fixes I did in different commits, I'll squash them in the end to make history more readable.
Also, I made the fixes as small as possible to avoid regressions or break something by accident.

### DIFF
--- a/include/modules/upower/upower.hpp
+++ b/include/modules/upower/upower.hpp
@@ -69,7 +69,7 @@ class UPower : public AModule {
   UpDevice *displayDevice;
   guint login1_id;
   GDBusConnection *login1_connection;
-  UPowerTooltip *upower_tooltip;
+  std::unique_ptr<UPowerTooltip> upower_tooltip;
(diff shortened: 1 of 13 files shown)


taminob:
Sorry that I forgot to auto-format the commits. Should be fixed now.

I noticed that the CI doesn't seem to check the `hpp` files which also had a formatting inconsistency in the previous commits.

Alexays:
LGTM thanks!

PR #2592, opened 2023-10-21
General cleanup in hyprland/workspaces

zjeffer:
Did some cleanup in hyprland/workspaces:

* Renamed the CreateWindow class to WindowCreationPayload, to make it clearer it isn't a function but a class.
* Added const qualifiers and references to functions 
* Added std::move in initializer lists to avoid making copies
* ~~Put the WINDOW_CREATION_TIMEOUT variable in an anonymous namespace (it used to be statically defined inside the function it was used in, I don't understand why?)~~

On a related note, I also started a discussion here about adding a clang-tidy file to the repo: https://github.com/Alexays/Waybar/discussions/2591. I'd love to hear your thoughts on this. 

CC: @Syndelis @khaneliman @Anakael @MightyPlaza

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -88,7 +88,7 @@ class Workspace {
   void initialize_window_map(const Json::Value& clients_data);
 
   bool on_window_opened(WindowCreationPayload const& create_window_paylod);
-  std::optional<std::string> on_window_closed(WindowAddress const& addr);
+  std::optional<std::string> close_window(WindowAddress const& addr);
 
   void update(const std::string& format, const std::string& icon);
 
@@ -127,7 +127,7 @@ class Workspaces : public AModule, public EventHandler {
 
   std::string get_rewrite(std::string window_class, std::string window_title);
   std::string& get_window_separator() { return format_window_separator_; }
-  bool is_workspace_ignored(std::string& workspace_name);
+  bool is_workspace_ignored(std::string const& workspace_name);
 
   bool window_rewrite_config_uses_title() const { return any_window_rewrite_rule_uses_title_; }
 
@@ -142,9 +142,22 @@ class Workspaces : public AModule, public EventHandler {
   void parse_config(const Json::Value& config);
   void register_ipc();
 
+  // workspace events
+  void on_workspace_activated(std::string const& payload);
+  void on_workspace_destroyed(std::string const& payload);
+  void on_workspace_created(std::string const& payload);
+  void on_workspace_moved(std::string const& payload);
+  void on_workspace_renamed(std::string const& payload);
+
+  // monitor events
+  void on_monitor_focused(std::string const& payload);
+
+  // window events
   void on_window_opened(std::string const& payload);
   void on_window_closed(std::string const& payload);
   void on_window_moved(std::string const& payload);
+
+  void on_window_title_event(std::string const& payload);
 
   int window_rewrite_priority_function(std::string const& window_rule);
 
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -26,13 +26,13 @@ namespace waybar::modules::hyprland {
 
 class Workspaces;
 
-class WorkspaceWindow {
+class WindowCreationPayload {
  public:
-  WorkspaceWindow(std::string workspace_name, WindowAddress window_address,
-                  std::string window_repr);
-  WorkspaceWindow(std::string workspace_name, WindowAddress window_address,
-                  std::string window_class, std::string window_title);
-  WorkspaceWindow(Json::Value const& client_data);
+  WindowCreationPayload(std::string workspace_name, WindowAddress window_address,
+                        std::string window_repr);
+  WindowCreationPayload(std::string workspace_name, WindowAddress window_address,
+                        std::string window_class, std::string window_title);
+  WindowCreationPayload(Json::Value const& client_data);
 
   int increment_time_spent_uncreated();
   bool is_empty(Workspaces& workspace_manager);
@@ -83,11 +83,11 @@ class Workspace {
   void set_windows(uint value) { windows_ = value; };
   void set_name(std::string const& value) { name_ = value; };
   bool contains_window(WindowAddress const& addr) const { return window_map_.contains(addr); }
-  void insert_window(WorkspaceWindow create_window_paylod);
+  void insert_window(WindowCreationPayload create_window_paylod);
   std::string remove_window(WindowAddress const& addr);
   void initialize_window_map(const Json::Value& clients_data);
 
-  bool on_window_opened(WorkspaceWindow const& create_window_paylod);
+  bool on_window_opened(WindowCreationPayload const& create_window_paylod);
   std::optional<std::string> on_window_closed(WindowAddress const& addr);
 
   void update(const std::string& format, const std::string& icon);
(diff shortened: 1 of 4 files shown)


MightyPlaza:
https://github.com/hyprwm/Hyprland/pull/3530

zjeffer:
> [hyprwm/Hyprland#3530](https://github.com/hyprwm/Hyprland/pull/3530)

@MightyPlaza  Yep, that's the next step, for another PR :) 
I first wanted to do some general cleaning up before implementing that.

MightyPlaza:
just wanted to point it out since most code for persistent workspaces can be removed now

include/modules/hyprland/workspaces.hpp
```diff
@@ -26,12 +26,13 @@ namespace waybar::modules::hyprland {
 
 class Workspaces;
 
-class CreateWindow {
+class WorkspaceWindow {
```
Syndelis, line 29: > Renamed the CreateWindow class to WorkspaceWindow, to make it clearer it isn't a function but a class.

I agree `CreateWindow` could have a different name to differentiate it from a function, but `WorkspaceWindow` makes it seems like it's a window on a workspace, while in reality it's the payload that will be used to push a new window to a workspace. Maybe `WindowCreationPayload` would be a better compromise between the two?

src/modules/hyprland/workspaces.cpp
```diff
@@ -863,58 +864,62 @@ std::string Workspaces::get_rewrite(std::string window_class, std::string window
...
+  if (std::holds_alternative<ClassAndTitle>(window_)) {
     auto [window_class, window_title] = std::get<ClassAndTitle>(window_);
     return workspace_manager.get_rewrite(window_class, window_title);
-  } else {
-    // Unreachable
-    return "";
   }
+  // Unreachable
+  spdlog::error("WorkspaceWindow::repr: Unreachable");
+  throw std::runtime_error("WorkspaceWindow::repr: Unreachable");
```
zjeffer, line 903: One thing I'm unsure of: this is unreachable code so I'm throwing here instead of returning an empty string. 

It makes more sense to me that the program throws an exception if the code is truly never supposed to reach there.

include/modules/hyprland/workspaces.hpp
zjeffer: I see. What about `WindowData`? WindowCreationPayload seems a bit long to me ;)

Syndelis:
> Put the WINDOW_CREATION_TIMEOUT variable in an anonymous namespace (it used to be statically defined inside the function it was used in, I don't understand why?)

It's just a personal preference, but if a "global" will only be used in one scope, I find it better to define it as a `static const` inside that scope. The reasoning is:

- It's not slower in any shape or form. [See it for yourself](https://godbolt.org/z/M1W13o9zq);
- It avoids needing to jump files in order to find the constant's value (it's right there in the same scope);
- Avoids raising the question "is this used somewhere else?", which directly leads you to global searching for its name. You can be a testimony for that: you literally took a look and realized it was only used inside of that function;
- Doesn't litter global namespace, which in turn avoids the need to define an anonymous namespace, reducing bloat. 

Given the above reasons, I'd say I don't understand why one would prefer a global variable over a scoped `static const`. Although I can't and wouldn't enforce you to keep it as it was, I do ask you to take my points into consideration and reconsider this change.

include/modules/hyprland/workspaces.hpp
Syndelis: `WindowCreationData`/`WinCreateData`, then? Just `WindowData` doesn't seem to be quite enough to decipher the class' purpose

khaneliman: My only wrench to throw in is that everything related to the data arguments for it are using the 'payload' nomenclature.

src/modules/hyprland/workspaces.cpp
Syndelis: Good one! I definitely agree a throw is better than an empty return

include/modules/hyprland/workspaces.hpp
zjeffer: Valid points, changed it to WindowCreationPayload.

zjeffer:
Here are the reasons I thought it was strange:

* We might use the constant in another function in the future, so my change takes that into account.
* If all constants used in a file are defined up top, it makes changing them easy without having to find the function they are used in. Say we have 5 constants, tuning these constants is really quick, without having to go to the functions they are used in.
* Where I work, we always use an anonymous namespace at the top to define constants, so it was a force of habit ;)

But your reasons are more logical in this case as it's only used in one place. I changed it back.

zjeffer:
Just out of interest, could there be a (tiny) runtime performance difference when defining a static variable inside the function it is used in, if that function is called repeatedly? We're statically defining the WINDOW_CREATION_TIMEOUT variable in the update() function, which is called on every event. When running, does it check every iteration whether the static variable was defined yet? This obviously would make zero noticable difference in real-world usage, but I'm interested in how that works.

Syndelis:
> Just out of interest, could there be a (tiny) runtime performance difference when defining a static variable inside the function it is used in, if that function is called repeatedly? We're statically defining the WINDOW_CREATION_TIMEOUT variable in the update() function, which is called on every event. When running, does it check every iteration whether the static variable was defined yet? This obviously would make zero noticable difference in real-world usage, but I'm interested in how that works.

From the [compiler explorer link](https://godbolt.org/z/M1W13o9zq), they should be compiled down to the same thing even with `-O0`. From the manual:

> Most  optimizations  are completely disabled at -O0

```cpp
#include <cstdio>

const int A = 10;

void with_global() {
  printf("A=%d\n", A);
}

void with_scoped() {
  static const int B = 20;
  printf("B=%d\n", B);
}
```

```asm
.LC0:
        .string "A=%d\n"
with_global():
        push    rbp
        mov     rbp, rsp
        mov     esi, 10
        mov     edi, OFFSET FLAT:.LC0
        mov     eax, 0
        call    printf
        nop
        pop     rbp
        ret
.LC1:
        .string "B=%d\n"
with_scoped():
        push    rbp
        mov     rbp, rsp
        mov     esi, 20
        mov     edi, OFFSET FLAT:.LC1
        mov     eax, 0
        call    printf
        nop
        pop     rbp
        ret
```

The only real difference is the scopes in which the compiler will allow you to use the variable

Syndelis (approved):
Great PR, and I look forward to the clang-tidy PR, too!

zjeffer:
Just made some more changes, could you re-review? Thanks!

zjeffer:
@Alexays ready to merge :)

include/modules/hyprland/workspaces.hpp
```diff
@@ -142,10 +142,23 @@ class Workspaces : public AModule, public EventHandler {
   void parse_config(const Json::Value& config);
   void register_ipc();
 
+  // workspace events
+  void on_workspace_activated(std::string const& payload);
+  void on_workspace_destroyed(std::string const& payload);
+  void on_workspace_created(std::string const& payload);
+  void on_workspace_moved(std::string const& payload);
+  void on_workspace_renamed(std::string const& payload);
```
khaneliman, line 150: Thank you for doing this... I didn't like how busy that method was.

khaneliman (approved):

include/modules/hyprland/workspaces.hpp
zjeffer: This was also a hint by clang-tidy: it shows a warning if a function is too complex. Another reason to add it to the repo :)

Syndelis (approved):
Forgot to submit the review yesterday, my apologies!

Alexays:
Thanks!

PR #2595, opened 2023-10-21
Added initial .clang-tidy file

zjeffer:
More info here: https://github.com/Alexays/Waybar/discussions/2591

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,28 @@
+Checks: >
+  -*,
+  bugprone-*
+  misc-*,
+  modernize-*,
+  performance-*,
+  portability-*,
+  readability-*,
+  -fuchsia-trailing-return,
+  -readability-magic-numbers,
+  -modernize-use-nodiscard,
+  -modernize-use-trailing-return-type,
+  -readability-braces-around-statements,
+  -readability-redundant-access-specifiers,
+  -readability-redundant-member-init,
+  -readability-redundant-string-init,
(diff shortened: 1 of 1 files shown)


khaneliman:
Looks like it's already working in the linter :P

zjeffer:
> Looks like it's already working in the linter :P

No, that's clang-format detecting a file in the current master not being formatted correctly. I had the same linter error in my other PR.

If we would add a GitHub action that checks clang-tidy, we'd see *lots* more errors :D

Alexays:
I'm ok with making a big PR that fixes all the lint at once as well as adding other rules to improve code quality.

Alexays:
And updating Github Action to reflect that.

zjeffer:
@Alexays I added the casing hints to the file, to show warnings in the IDE when the casing of the name of a class, function, struct, member variable, ... is inconsistent with the clang-tidy file. I based my choices on the existing code, to keep the impact to the code the lowest. 

One thing I'm unsure of: I added a `m_` as a PrivateMemberPrefix, so member variables have to start with `m_`, which I think makes the code much clearer. I don't see it used very much, though, so let me know if you think this should be included.

Can you also check whether the other CheckOptions make sense to you? Let me know if you want to make any changes.

I'll start tidying-up the code (in a separate commit) once you're sure about all the options.

Alexays:
LGTM, we'll make small changes to the config if we realise it's too much.

PR #2596, opened 2023-10-21
modules/mpris: change default interval value to 0

### LINKED ISSUE #2392: MPRIS module takes too long to update metadata
`mpris` module displays previous title for about 2-3 seconds before displaying the correct one while `playerctl metadata` displays the most up to date info.

rehanzo:
This PR sets the interval value to 0. It seems this is what it was meant to be set at, but was changed to '5' in commit ba6faa7.

This closes #2392.


khaneliman (approved):
Good catch.

PR #2601, opened 2023-10-22
Fix sleep_for for max duration

taminob:
As mentioned in the issue, the bug is caused by the standard library expecting for `std::condition_variable::sleep_for` a duration which does not cause an overflow when added to the current time.

This PR fixes the issue by using `sleep_until` to emulate `sleep_for` since `sleep_for` is defined as identical to adding the current time and then using `sleep_until` by the standard.

Additionally, this PR re-applies #2586 which was reverted in #2599 because of this bug.

I also tried to write a test to catch this kind of error for `SleeperThread`, but I wasn't able to check for this kind of failure.
The standard does not specify the behavior if a too big duration is passed.
I ended up writing a few basic other tests for the class and restructured the build system to allow for easier addition to the tests by providing all sources of `waybar` as a static library which is now simply linked against `main.cpp` for the executable.

Resolves #2598

PR #2608, opened 2023-10-24
Add basic tests for SleeperThread

taminob:
I wrote these after failing to construct test cases for #2601.
Unfortunately, I couldn't think of tests without relying on some timing, but I hope that I chose the times so that the tests will work on all (relevant) systems.

For this, I restructured the build system to allow for easier addition to the tests by compiling all sources of waybar (except `main.cpp`) as a static library.
This library is then linked against `main.cpp` for the main executable and against the test files for the test executable.


taminob:
I don't think anymore that these `SleeperThread` tests are worth merging since they are too brittle and depend on specific timing to succeed - I tried to improve them, but wasn't really able to achieve guaranteed success without any timing dependency. Thus, I'll close the PR.

@Alexays if you think the other changes to allow linking against a static waybar lib in the tests are worth adding, feel free to ping me and I can open another PR for that.

PR #2609, opened 2023-10-24
Reduce freebsd ci timeout

taminob:
Since the freebsd job is pretty unreliable, I reduced the timeout to 30min instead of the default of 6h.

The issue is known: https://github.com/vmactions/freebsd-vm/issues/74

Alternatively, if there are no fixes for this github action, it could be considered to switch to a different action.
One project already moved away from the one that is currently used in waybar:
https://github.com/mpv-player/mpv/commit/71a497deee8f5cb47a9ea37bc14038ecdf238e21


Alexays:
We could try to switch to the same mpv action.

Issue #2607, opened 2023-10-24
[Bug]Several requests were sent at the same time in a short period of time, and some failed to respond


DreamMaoMao:
![image](https://github.com/Alexays/Waybar/assets/30348075/21fac9cf-e6a9-4c15-9b75-a0d09413a524)

https://github.com/Alexays/Waybar/assets/30348075/4dfc9d64-ff5e-4b96-9064-48c8137ebdbe

In the video, you can see that worksapce 3 sometimes does not show up, but from the log,waybar has successfully received the IPC request

Syndelis:
Could you describe your setup so I can try to reproduce the problem? Specifically, these questions came to mind:
1. What is it you're doing that seemingly spawns/destroys 5 workspaces at once?
2. Are these workspaces populated by any windows?
3. Are they set as "persistent" on your Waybar/Hyprland configuration?

Thanks in advance for your time opening this issue and recording a video!

DreamMaoMao:
> Could you describe your setup so I can try to reproduce the problem? Specifically, these questions came to mind:
> 
> 1. What is it you're doing that seemingly spawns/destroys 5 workspaces at once?
> 2. Are these workspaces populated by any windows?
> 3. Are they set as "persistent" on your Waybar/Hyprland configuration?
> 
> Thanks in advance for your time opening this issue and recording a video!

1.What is it you're doing that seemingly spawns/destroys 5 workspaces at once?
I am making a hyprland plugin[plugin](https://github.com/DreamMaoMao/hycov)

setup1 I moved the Windows of five workspaces to another workspace at the same time
setup2 I moved all the Windows I had previously moved back to the previous workspace

2.Are these workspaces populated by any windows?
the answer is no

3.Are they set as "persistent" on your Waybar/Hyprland configuration?
the answer is no 

logfile:

[waybar.log](https://github.com/Alexays/Waybar/files/13138352/waybar.log)

It is worth noting that when I have only three workspaces, this bug is very unlikely to occur, but when I have six workspaces with Windows, this operation is very likely to occur (repeat the operation once to five times).

This seems to be a software performance issue under high load.

Syndelis:
> 2.Are these workspaces populated by any windows?
> the answer is no

Correct me if I'm wrong, but Hyprland deletes empty workspaces, so that should be an expected behavior (?)

DreamMaoMao:
@Syndelis 
What I mean is that no window is occupied before the deletion request is sent

[...]

PR #2613, opened 2023-10-27
Fixed variable synchronization exception caused by the "hyprland/workspace" module

DreamMaoMao:
hey i want to fix:[#2607](https://github.com/Alexays/Waybar/issues/2607)

I've tested this commit several times.

https://github.com/Alexays/Waybar/assets/30348075/6bbe534a-435c-483d-9dfe-fc154e318142

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -153,17 +153,25 @@ auto Workspaces::register_ipc() -> void {
 }
 
 auto Workspaces::update() -> void {
+  //remove workspaces that wait to be removed
+  unsigned int current_remove_workspace_num = 0;
   for (const std::string &workspace_to_remove : workspaces_to_remove_) {
     remove_workspace(workspace_to_remove);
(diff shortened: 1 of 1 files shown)


Alexays:
Hi, thanks for the fix!
cc @Syndelis does this look good for you?

Syndelis (approved):
Good catch! @Alexays LGTM

zjeffer:
I feel like it would be cleaner with some kind of mutex locking solution, no? Then we make sure the vector is not accessed when we update(). I don't really have time to look into that, though.

DreamMaoMao:
> I feel like it would be cleaner with some kind of mutex locking solution, no? Then we make sure the vector is not accessed when we update(). I don't really have time to look into that, though.

If the vector is locked during the update, then if the event is triggered at this time, won’t the creation and removal requests be lost?

DreamMaoMao:
To ensure that requests are not lost, the vector used to receive buffering should not be locked from writing at any time. It's like a keyboard's input buffer.

PR #2620, opened 2023-10-30
Fixed typos

fritzrehde:
I fixed some typos that I found with [`typos`](https://github.com/crate-ci/typos).
I'm not sure about all of my fixes, so I have highlighted those places in some following messages.

### DIFF
--- a/include/modules/sway/language.hpp
+++ b/include/modules/sway/language.hpp
@@ -21,7 +21,7 @@ class Language : public ALabel, public sigc::trackable {
   auto update() -> void override;
 
  private:
-  enum class DispayedShortFlag { None = 0, ShortName = 1, ShortDescription = 1 << 1 };
+  enum class DisplayedShortFlag { None = 0, ShortName = 1, ShortDescription = 1 << 1 };
 
   struct Layout {
     std::string full_name;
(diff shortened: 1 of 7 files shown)


include/modules/sway/language.hpp
```diff
@@ -21,7 +21,7 @@ class Language : public ALabel, public sigc::trackable {
   auto update() -> void override;
 
  private:
-  enum class DispayedShortFlag { None = 0, ShortName = 1, ShortDescription = 1 << 1 };
+  enum class DisplayedShortFlag { None = 0, ShortName = 1, ShortDescription = 1 << 1 };
```
fritzrehde, line 24: You mean `Displayed` here, right?

protocol/dwl-ipc-unstable-v2.xml
```diff
@@ -36,8 +36,8 @@ I would probably just submit raphi's patchset but I don't think that would be po
     </request>
 
     <request name="get_output">
-      <description summary="get a dwl_ipc_outout for a wl_output">
```
fritzrehde, line 39: Is `outout` some technical term from the protocol, or do you mean `output`?

Alexays: I'll prefer to don't touch this file, as it's a protocol file.

include/modules/sway/language.hpp
Alexays: Yep.

PR #2626, opened 2023-11-01
tray: load_icon use request_size directly

akliuxingyuan:
I delete tmp_size part, when call `load_icon` function, use request_size `icon_size * image.get_scale_factor()` like wlr/taskbar
`get_icon_sizes` doesn't handle scaling, if icon isn't scalable, tmp_size will be smaller size
https://github.com/Alexays/Waybar/blob/1149e51f7271fbb81b610c223dccc8181926c7e2/src/modules/wlr/taskbar.cpp#L181-L184

for example: nm-applet

### DIFF
--- a/src/modules/sni/item.cpp
+++ b/src/modules/sni/item.cpp
@@ -356,32 +356,15 @@ Glib::RefPtr<Gdk::Pixbuf> Item::getIconPixbuf() {
 }
 
 Glib::RefPtr<Gdk::Pixbuf> Item::getIconByName(const std::string& name, int request_size) {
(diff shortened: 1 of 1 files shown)


archseer:
I think this ruined scaling on my bar for certain icons: They used to be the same size, but now the volume icons, the first workspace icon and the CPU icon are slightly smaller than the rest. I'm using Font Awesome for all of them.
icons

archseer:
Hmm maybe not since the icons are a part of the font, not a separate image. I need to do a bisect from the last release

Issue #2244, opened 2023-06-19
Turn off laptop's built-in screen Cause zombie process.


Bryan2333:
My WM is Hyprland 0.26.0, Waybar 0.9.18

I have two screens: one is my laptop built-in screen, the other one is my external monitor. Recently I find that there are zombie process of Waybar in Htop. And I found that zombie process is generated every time I turn off one screen when the other one is on

Here is the screen record of that situation.

I don't know if this problem is related to #1713 

https://github.com/Alexays/Waybar/assets/31512600/1ee78cfa-37a8-44f6-98a1-4d9e30938e3d

PR #2629, opened 2023-11-02
custom: reap zombie processes on termination

### LINKED ISSUE #1713: Some child processes are still left open when waybar terminates
To reproduce:

1. Setup a custom module (and add that module to the waybar):

```json
    "custom/process-broken-demo": {
        "return-type": "json",
        "escape": false,
        "exec": "$HOME/.config/waybar/process-test.sh"
    }
```

2. Put this into the `process-test.sh`:

```bash
#!/usr/bin/env bash

function_call() {
    sleep 1
}

while read line; do
        function_call
done < <(sleep 9999)
```

3. Run `ps -u | grep process-test`. Note that the only returned entry should be the grep process.
4. Start waybar.
5. Run step 3 again. Note that there are two extra processes now, started by waybar
6. Close waybar
7. Run step 3 again. Note that one extra process is still there, even though waybar has been closed

Probably related to #358?

andresilva:
When turning off/on the monitor it would lead to defunct processes started by custom modules. I don't know why this is needed since we are tracking the pid of forked processes and properly waiting them (either on `command::close` or on `Custom` destructor).
Fixes #1713 (and probably #2244).

### DIFF
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -112,6 +112,10 @@ inline FILE* open(const std::string& cmd, int& pid) {
     execlp("/bin/sh", "sh", "-c", cmd.c_str(), (char*)0);
     exit(0);
   } else {
+    reap_mtx.lock();
+    reap.push_back(child_pid);
+    reap_mtx.unlock();
+
     ::close(fd[1]);
   }
   pid = child_pid;


Alexays:
Thanks!

Issue #2615, opened 2023-10-29
Daylight saving hour breaks clock module


hrdl-github:
Waybar stopped updating the time displayed today at 01:59 AM CEST, outputting something like

```
[2023-10-29 02:11:00.000] [error] clock: 2023-10-29 02:11:00 is ambiguous.  It could be
2023-10-29 02:11:00 CEST == 2023-10-29 00:11:00 UTC or
2023-10-29 02:11:00 CET == 2023-10-29 01:11:00 UTC
```

every minute. Restarting waybar removes any displayed clock field. I have `/etc/localtime -> /usr/share/zoneinfo/Europe/Brussels` and `date` reporting `Sun Oct 29 02:11:26 AM CEST 2023`.

Version: `Waybar v0.9.22-200-ge46f66b4 (branch 'master')`

SuperTux88:
LOL, I also just wanted to report that, luckily I checked again before actually submitting :joy:

kescherCode:
Even putting `{ "timezone": "Europe/Brussels" }` into the clock config results in this. The log message indicates that somewhere, the already converted time is converted back to UTC.

SuperTux88:
It looks like it's getting the time in local time but without offset information. I don't know if it wants to convert back to UTC (or if it's just for the logging), it's just the error messages saying it could be either CEST or CET, and it doesn't know which one it is and that those would be 2 different times in UTC. And setting "timezone" doesn't contain the information if it's currently CEST or CET either.

So if it needs to know if it's CEST or CET it needs to get the time including the offset.

bitterhalt:
Same here.. I googled this like mad until I looked here :D

qbe:
this error seems to come from within std::chrono [see here](https://github.com/microsoft/STL/issues/3648)

LittleFox94:
setting `"timezone": "UTC"` gives at least some clock display again, in UTC though - at least for me it's better than no date/time at all ^^'

ezkha:
Well, same here. Regardless, happy winter time to y'all!

krei-se:
Glad i'm not the only nerd not sleeping at 2am :D

6b6279:
I tried creating a custom time module as a workaround (see code below), but it freezes 5 seconds after launching Waybar, no error messages in the logfile.
```
    "custom/time": {
        "interval": 1,
        "exec": "date '+%T %Z'",
        "format": "{}",
        "tooltip": false
    }
```

[...]

PR #2643, opened 2023-11-06
avoid converting to local_time when converting to zoned_time afterwards

qbe:
see [Issue 2615: Daylight saving hour breaks clock module](https://github.com/Alexays/Waybar/issues/2615)


qbe:
needs to be tested for compatibility with shifted day / time

Alexays:
Will revert if broking something, otherwise lgtm

PR #2648, opened 2023-11-06
Fix segmentation fault in WirePlumber module

Mrcarrot1:
The WirePlumber module assumes that either the node's name or description will not be null. This leads to a segmentation fault when both are.

The solution provided is to set self->node_name_ to a default value in this case.


Alexays:
LGTM!

Issue #921, opened 2020-11-24
wlr/taskbar: Missing icons for signal-desktop and gnome-terminal


cryptoluks:
Hi,

when running `wayland --log-level trace` it can be observed, that for Signal (`signal-desktop`) and the Gnome Terminal (`gnome-terminal` which spawns `gnome-terminal-server`) no icons can be found.

```
[2020-11-24 09:05:53.307] [debug] Couldn't find icon for Signal
```

For both applications, icons are defined in their respective `.desktop` file and they are also available on my system under `/usr/share/icons`.

```
❯ grep -i icon /usr/share/applications/signal-desktop.desktop 
Icon=signal-desktop
```

```
❯ find /usr/share/icons -iname "*"signal-desktop"*"
/usr/share/icons/hicolor/128x128/apps/signal-desktop.png
/usr/share/icons/hicolor/16x16/apps/signal-desktop.png
/usr/share/icons/hicolor/24x24/apps/signal-desktop.png
/usr/share/icons/hicolor/256x256/apps/signal-desktop.png
/usr/share/icons/hicolor/32x32/apps/signal-desktop.png
/usr/share/icons/hicolor/48x48/apps/signal-desktop.png
/usr/share/icons/hicolor/512x512/apps/signal-desktop.png
/usr/share/icons/hicolor/64x64/apps/signal-desktop.png
/usr/share/icons/hicolor/1024x1024/apps/signal-desktop.png
/usr/share/icons/Papirus/16x16/apps/signal-desktop-beta.svg
/usr/share/icons/Papirus/16x16/apps/signal-desktop.svg
/usr/share/icons/Papirus/22x22/apps/signal-desktop-beta.svg
/usr/share/icons/Papirus/22x22/apps/signal-desktop.svg
/usr/share/icons/Papirus/24x24/apps/signal-desktop-beta.svg
/usr/share/icons/Papirus/24x24/apps/signal-desktop.svg
/usr/share/icons/Papirus/32x32/apps/signal-desktop-beta.svg
/usr/share/icons/Papirus/32x32/apps/signal-desktop.svg
/usr/share/icons/Papirus/48x48/apps/signal-desktop-beta.svg
/usr/share/icons/Papirus/48x48/apps/signal-desktop.svg
/usr/share/icons/Papirus/64x64/apps/signal-desktop-beta.svg
/usr/share/icons/Papirus/64x64/apps/signal-desktop.svg
```

[...]

PR #2677, opened 2023-11-16
Using 'image-missing' fallback if no taskbar icon is found

sjoblomj:
Icons are sometimes not found in the taskbar, for two reasons:
1. The icon given in the .desktop file for that program cannot be found.
2. Waybar is not given an `app_id`, so lookup is impossible.

If the user has configured Waybar to only show applications as icons, this means that the program is now essentially "hiding":

Hidden program

With this patch, when an icon cannot be found, Waybar will now load the "image-missing" icon of Gtk:

Image missing as a fallback

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -22,6 +22,7 @@
 #include "util/format.hpp"
 #include "util/rewrite_string.hpp"
 #include "util/string.hpp"
+#include "util/gtk_icon.hpp"
 
 namespace waybar::modules::wlr {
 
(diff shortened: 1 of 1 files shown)


src/modules/wlr/taskbar.cpp
```diff
@@ -182,11 +183,21 @@ bool Task::image_load_icon(Gtk::Image &image, const Glib::RefPtr<Gtk::IconTheme>
...
-    if (Glib::file_test(ret_icon_name, Glib::FILE_TEST_EXISTS))
+    if (Glib::file_test(ret_icon_name, Glib::FILE_TEST_EXISTS)) {
       pixbuf = load_icon_from_file(ret_icon_name, scaled_icon_size);
-    else
-      pixbuf = {};
+      spdlog::debug("{} Loaded icon from file '{}'", repr(), ret_icon_name);
+    } else {
+      try {
+        pixbuf = DefaultGtkIconThemeWrapper::load_icon("image-missing", scaled_icon_size,
+            Gtk::IconLookupFlags::ICON_LOOKUP_FORCE_SIZE);
```
sjoblomj, line 194: The "image-missing" icon is maybe not the prettiest, something like [this](https://commons.wikimedia.org/wiki/File:Application-default-icon.svg) maybe looks prettier. A different approach is thus to bundle in a fallback icon in resources/icons (and update resources/icons/waybar_icons.gresource.xml).

This line is then replaced by `pixbuf = Gdk::Pixbuf::create_from_resource("/fr/arouillard/waybar/icons/scalable/apps/default-app-icon", scaled_icon_size, scaled_icon_size)`

If bundling an icon preferred, I am happy to use this approach instead and change the PR accordingly.

sjoblomj:
Fixes #754 and fixes #921

Alexays:
LGTM, thx!

PR #2679, opened 2023-11-18
hyprland module: added sort_workspaces after rename occured

aserowy:
added sorting of workspaces after a rename event got handled. The solution was mentioned in #2271

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -339,6 +339,7 @@ void Workspaces::on_workspace_renamed(std::string const &payload) {
       break;
     }
   }
+  sort_workspaces();
 }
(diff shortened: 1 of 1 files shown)


src/modules/hyprland/workspaces.cpp
```diff
@@ -339,6 +339,7 @@ void Workspaces::on_workspace_renamed(std::string const &payload) {
       break;
     }
   }
+  sort_workspaces();
```
zjeffer, line 342: So does this fix the issue? Did you test it?

aserowy: Heho,

worked on my machine. :D

zjeffer (approved):

PR #2686, opened 2023-11-22
privacy: fix visibility when a module is disabled

aruhier:
Only consider a privacy module as visible if it is enabled in the configuration. Otherwise, when screensharing or audio-in or audio-out is in use but the associated module is not enabled, the privacy widget is empty but still considered as visible.


Alexays:
Thanks!

PR #2701, opened 2023-11-28
Add button click check to hyprland workspaces module

thejch:
When clicking on the hyprland special workspace button twice in a row, the toggle doesn't behave correctly.
This fixes the issue that hyprland special workspaces are not toggled correctly when you click on the buttons quickly more than once. Also eliminates some duplicate calls made to hyprland for any workspace switching.


Alexays:
Thanks!

PR #2711, opened 2023-12-01
flake: update; simplify dev shell

chayleaf:
this takes the dependencies for the dev shell from the waybar package in nixpkgs, removes some unnecessary dependencies, and updates nixpkgs (this is important for those who don't use `inputs.waybar.inputs.nixpkgs.follows = "nixpkgs";` so their Waybar dependencies stay up to date)


khaneliman (approved):
Works for me and is much cleaner. Thank you.

PR #2712, opened 2023-12-01
fix: nix flake build

khaneliman:
Update nix flake to fix nix build errors.


chayleaf:
see https://github.com/Alexays/Waybar/pull/2711

khaneliman:
> #2711 should fix it

Better than lipstick on a pig, thank you for doing that proper.

PR #2720, opened 2023-12-04
Applied clang-tidy suggestions for hyprland workspaces

zjeffer:
First test after merging https://github.com/Alexays/Waybar/pull/2595

Let me know if anything needs to change.


Alexays:
We can add sudo to images, Dockerfile are also stored in the repo, once changed i'll trigger a rebuild :)

zjeffer:
I'm doing some manual tests by adding sudo and python3-pip (which also seems needed) in the yml file. I'll make a PR to change the Dockerfile once I've figured out the next errors I'm getting.

PR #2744, opened 2023-12-13
Specify config is JSONC

rpdelaney:
I always find pure JSON cumbersome for configuration files due to the lack of support for comments. Seeing up-front that the configuration file is actually JSONC would have aided my adoption of waybar.

Changes proposed in this pull request:

- Clarify that the configuration file is JSON*C* format
- The configuration file can also be named config.jsonc
- Remove 'config' from list of documented valid locations


Alexays:
Thanks!

PR #2760, opened 2023-12-22
Fix wireplumber reverse-scroll option

avannereau:
The reverse-scroll option does nothing currently, since the reverse logic is present in both **AModule::getScrollDir** and **Wireplumber::handleScroll** thus reversing it twice when enabled.

This PR remove the reverse logic in **Wireplumber::handleScroll** to be consistent with other modules.


Alexays:
Thanks!

PR #2764, opened 2023-12-23
fix:the workspace data is null in a small probability

DreamMaoMao:
Hey, i found a small probability bug.

the `workspaceToCreate` json object is ready in here:
image

but sometime,when the `workspaceToCreate` json data which get from m_workspacesToCreate in here will be `null`.
image

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -6,13 +6,17 @@
 #include <algorithm>
 #include <charconv>
 #include <memory>
+#include <shared_mutex>
 #include <string>
+#include <thread>
 #include <utility>
 #include <variant>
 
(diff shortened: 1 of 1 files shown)


DreamMaoMao:
waybar crash, coredump:

image

DreamMaoMao:
https://github.com/Alexays/Waybar/assets/30348075/7aa9d936-47e9-4f40-b143-aec2bdc40068

With the addition of read-write locks, it can now create 30 workspaces simultaneously and has run over 100 tests without any problems

@Syndelis 
What do you think about that?

DreamMaoMao:
Relate to [issue:#2769](https://github.com/Alexays/Waybar/issues/2769)

DreamMaoMao:
how aboout this? @Alexays

PR #2773, opened 2023-12-27
Implement windows formating in sway/workspaces

grimpy:
This implementation mimics to some extend the implementation of hyprland

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -12,6 +12,7 @@
 #include "client.hpp"
 #include "modules/sway/ipc/client.hpp"
 #include "util/json.hpp"
+#include "util/regex_collection.hpp"
 
 namespace waybar::modules::sway {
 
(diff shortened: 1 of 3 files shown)


grimpy:
@Alexays anything else you want me to address in this PR?

Alexays:
LGTM, can you also update the Github wiki?

Alexays:
Reverted, freebsd check failed.

alebastr:
```
../src/modules/sway/workspaces.cpp:250:20: error: no member named 'any_of' in namespace 'std::ranges'
  if (std::ranges::any_of(node["nodes"], [&](auto const &e) { return hasFlag(e, flag); })) {
      ~~~~~~~~~~~~~^
```

The CI runs FreeBSD 13.2 (still supported) with clang 14.x. It either has no `std::ranges` support or may need additional includes. Rewriting this line with `std::any_of` would be better for compatibility (and for the binary size).

Issue #2650, opened 2023-11-07
Custom module doesn't update in version 0.9.24


tamirzb:
I have a custom module which prints multiple outputs throughout its run, with the intention of each output updating what is displayed on waybar. It seems that in version 0.9.24 waybar keeps displaying the first output, not updating what is displayed even when other stuff are printed.

To narrow it down, if the script is:
```bash
    echo "{\"text\": \"1\", \"tooltip\": \"tooltip\"," "\"class\": \"class\" }"
    echo "{\"text\": \"2\", \"tooltip\": \"tooltip\"," "\"class\": \"class\" }"
```
I still see `1` printed, not `2`.

On version 0.9.22 this doesn't happend.

mxraw:
I have the same issue. I noticed the bug after updating from Fedora 38 -> 39, which bumped Waybar from 0.9.22 -> 0.9.24.

mxraw:
Bisected this.

```
> git bisect log
git bisect start
# status: waiting for both good and bad commits
# bad: [48ec834ec9c7cdf146169c8b0f669ce1e9a8849e] chore: 0.9.24
git bisect bad 48ec834ec9c7cdf146169c8b0f669ce1e9a8849e
# status: waiting for good commit(s), bad commit known
# good: [e30fba0b8f875c7f35e3173be2b9f6f3ffe3641e] chore: 0.9.22
git bisect good e30fba0b8f875c7f35e3173be2b9f6f3ffe3641e
# bad: [7645ae1907904d7283b1fac70b168e69e52a6f79] Merge pull request #2563 from Syndelis/feat/hyprland-window-rename-by-title
git bisect bad 7645ae1907904d7283b1fac70b168e69e52a6f79
# good: [05a2af2d7c57ce320053b73ed86a58449b5332f1] Merge pull request #2486 from khaneliman/workspace-sorting
git bisect good 05a2af2d7c57ce320053b73ed86a58449b5332f1
# bad: [4c0347d9f243322352a18bdf2c4d4bbcde37163e] Merge pull request #2540 from Syndelis/fix/hyprland-special-workspaces
git bisect bad 4c0347d9f243322352a18bdf2c4d4bbcde37163e
# bad: [f14fe96e199af4bd8050c304ae5ea6c300ba28df] add info to `interval` and `signal` in manpage for custom modules
git bisect bad f14fe96e199af4bd8050c304ae5ea6c300ba28df
[...]

PR #2778, opened 2023-12-29
fix `restart-interval` being ignored while `signal` is defined (#2650)

an-prata:
Bug popped up from my previous PR (#2517) where users with modules that had both a `restart-interval` and `signal` defined would end up with a module that behaves as though only a signal is defined.

This PR fixes this issue (#2650).


Alexays:
Thanks!

PR #2787, opened 2023-12-31
hyprland/workspaces: allow using the original window class/title

scgtrp:
This turns the values of window rewrite rules in hyprland/workspaces from static strings to format strings with the values {class} and {title} available.

For example, I'm now doing this to list windows in each workspace without having to add a rule for every application I use:
```
  "hyprland/workspaces": {
    "format": "{name}: {windows}",
    "window-rewrite-default": "{class}",
  },
```

image


Alexays:
Thanks!

Issue #2769, opened 2023-12-25
'hyprland/workspace' module data reads and writes do not always keep atomicity


DreamMaoMao:
![image](https://github.com/Alexays/Waybar/assets/30348075/60624f05-477c-4f04-9451-ed303f8a8567)

![image](https://github.com/Alexays/Waybar/assets/30348075/7be9699d-6357-4c15-8c94-c1c156b8135f)

When reading the data of `workspaceToCreate`, sometimes you will read the address of the data that has not been fully written.

![image](https://github.com/Alexays/Waybar/assets/30348075/d4e4fef2-8704-4af3-8a23-ace91d2be69c)

PR #2789, opened 2023-12-31
fix:#2769

DreamMaoMao:
fix:(#2769 )


DreamMaoMao:
yd I've been testing this patch for a long time and it's greatly improved stability of `hyprland/workspace` module.

Alexays:
Thanks!

Issue #2718, opened 2023-12-03
window-rewrite missing firefox icon upon boot (hyprland/workspaces module)


Masber:
Hi, I am using waybar with hyprland, my issue is waybar is inconsistent in rendering the workspaces icons upon login. I have to kill and start warbar again in order to restore the missing icons.

hyprland instructions to start the apps:

```
exec-once=waybar & hyprpaper
exec-once=[workspace 1 silent] alacritty
exec-once=[workspace 2 silent] thunderbird & slack
exec-once=[workspace 3 silent] firefox
```

waybar hyprland/workspaces config

```
    "hyprland/workspaces": {
	"format": " <b>{icon}</b> {windows} ",
	"window-rewrite-default": "",
	"window-rewrite": {
	  "firefox": "󰈹",
	  "Alacritty": "󰞷",
	  "slack": "󰒱",
	  "thunderbird": "󰇮",
	  "spotify": "󰓇"
	},
        "on-scroll-up": "hyprctl dispatch workspace e+1",
        "on-scroll-down": "hyprctl dispatch workspace e-1",
        "on-click": "activate"
    },
    "hyprland/window": {
        "format": "{}",
    	"rewrite": {
        	"(.*) — Mozilla Firefox": "󰈹 $1",
        	"(.*) — Mozilla Thunderbird": "󰇮 $1",
        	"Alacritty": "󰞷 Alacritty",
        	"(.*) - Slack": "󰒱 $1",
	}
    },
```

These are how the workspaces looks like after boot

![20231203_13h02m59s_grim](https://github.com/Alexays/Waybar/assets/722064/b66e0440-97e4-4152-9bcd-29c2d616aaac)

and this after restart waybar `$ killall waybar ; waybar &`

![20231203_13h06m09s_grim](https://github.com/Alexays/Waybar/assets/722064/45903f90-d458-49e7-94b0-5203280908ed)

As you can see both the firefox and the terminal icons are missing even tho both apps are running

Any idea of what could happen?

thank you

khaneliman:
Yeah, this has been an issue since the features creation and I keep wanting to take a stab at it when I have time... I basically just always do a systemctl --user restart waybar after login, for the time being. But, would rather it not be needed, obviously.

Masber:
no problem, I understand this is pending issue, I can try your workaround, that is not a problem.

thank you very much for your help and waybar is very cool :)

PR #2809, opened 2024-01-05
fix: hyprlands window-rewrite 

khaneliman:
Resolves: https://github.com/Alexays/Waybar/issues/2718 found during testing another feature in pipeline. The window rewrites work on boot for me and no issues dragging between monitors anymore. 

cc @Syndelis

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -50,8 +50,8 @@ Workspaces::Workspaces(const std::string &id, const Bar &bar, const Json::Value 
     gIPC = std::make_unique<IPC>();
   }
 
-  init();
   registerIpc();
+  init();
 }
 
 auto Workspaces::parseConfig(const Json::Value &config) -> void {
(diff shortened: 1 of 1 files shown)


src/modules/hyprland/workspaces.cpp
```diff
@@ -181,6 +181,9 @@ auto Workspaces::update() -> void {
   }
 
   for (auto &workspace : m_workspaces) {
+    auto clients_data = gIPC->getSocket1JsonReply("clients");
```
zjeffer, line 184: Shouldn't you get the clients_data outside of the for-loop? Right now you're getting the data every iteration of the loop, does the data change in between iterations?

khaneliman: Good catch, pasted it inside loop accidentally.

Syndelis:
Although this would solve the issue, I feel like getting the `clients` information for every update seems overkill. The problem seems to appear when a window is dragged to a workspace that doens't exist in tha same bar it originated from. A less performance-intensive solution would be to check if the icon is empty when the window is moved to another workspace [here](https://github.com/Syndelis/Waybar/blob/8b1c0f3f7aee14b01eca6c010c157dc13e0c5b81/src/modules/hyprland/workspaces.cpp#L401-L407). I'll see what I can hack together tomorrow and will get back here.

khaneliman:
Oh yeah, it's definitely not the most elegant... I was just excited to finally see the bar function the way I had hoped finally..

Syndelis:
Sorry for being late, @khaneliman ! I've opened #2817 which should hopefully fix this problem as well. Could you verify that works for your setup?

khaneliman:
closing with work being done in https://github.com/Alexays/Waybar/pull/2817

PR #2814, opened 2024-01-08
flake update and direnv fix

khaneliman:
Enables direnv loading flake, fix flake devshell output to be properly at root. 

I realized that catch2 dependency was updated and currently hasn't been pushed to nixos-unstable: https://nixpk.gs/pr-tracker.html?pr=278807

Temporarily added an override to fetch the 3.5.1 version from source until nixpkgs is updated.


Alexays:
Thanks!

Issue #1187, opened 2021-07-30
[Feature Request] systemd system/user unit control


somini:
It should be possible to start, stop and monitor the state of systemd units, both for system and user.

For system units, there's the issue of requesting credentials, but that is not a problem for user units.

It should be possible to use the systemd dbus interface to avoid costly polling, instead of this custom module does:

```
    "custom/systemd#information": {
	"format": "§ {}",
	"interval": 5,
	"exec": "systemctl-state-waybar --user $UNIT",
	"on-click": "systemctl-toggle -u $UNIT",
	"on-click-right": "systemctl --user stop $UNIT"
    }

```

Anakael:
Do you have idea how it should look like?
I think it can be useful only if several units can be attached to this module, but in that case I don't know how to find unit to handle clicks.

somini:
There's already support for multiple blocks with separate configuration, right? It's enough to have a single dbus connection across all systemd sections.

Or maybe you mean the configuration? This is what I would do, for `daemon1`:

```
systemd#user_daemon1: {
  "unit": "$UNIT",
  "user": true,
  "format": "{unit}",
  "on-click": "systemctl-toggle {manager} {unit}",
  "on-click-right": "systemctl {manager} stop {unit}",
}
```

- Not sure how the on-click templates is done on other plugins. IMHO, it would be fine if no templates were allowed and passed the info in env variables or something.
- No polling would be needed since it subscribes to dbus messages. The multiple states would be implemented as CSS classes, just like other plugins.

I can also share the bash scripts I'm using on this example:

## `systemd-state-waybar`

Uses the old i3blocks protocol:

```bash
#!/bin/sh
set -e
what='--system'
unit="$1"
if [ "$1" = '--user' ]; then
	what="$1"
	unit="$2"
fi
text="${3:-$unit}"

echo "$text"
echo "Unit: $unit"
if systemctl --quiet "$what" is-active "$unit"; then
	echo "good"
else
	echo "bad"
fi
exit 0
```

## `systemctl-toggle`

```bash
#!/bin/bash
extra_args=()
check_command='is-active'
start_command='start'
stop_command='stop'

usage() {
	# shellcheck disable=SC2016
	(
	echo "Usage: $0 [ -u ] [ -c CHECK_COMMAND ] UNIT"
	echo '  -u: Talk to the `--user` systemd daemon'
	)>&2
	exit 1
}

while getopts ':uc:' opt; do
	case "$opt" in
		u)
[...]

PR #2820, opened 2024-01-09
Add module systemd-failed-units to monitor failed systemd units

oxalica:
This module monitors the number of failed systemd units and can warn user if there is something going wrong. It's especially useful for periodic services.

Currently only the count is monitored and displayed, and there's no tooltip. It's nontrivial to get the actual failed unit names and it seems to require enumerating all units and then filter, which would be too CPU intensive to be done frequently.

Related but not equivalent to #1187.

### DIFF
--- a/include/factory.hpp
+++ b/include/factory.hpp
@@ -93,6 +93,9 @@
 #ifdef HAVE_LIBCAVA
 #include "modules/cava.hpp"
 #endif
+#ifdef HAVE_SYSTEMD_MONITOR
+#include "modules/systemd_failed_units.hpp"
+#endif
 #include "bar.hpp"
 #include "modules/cffi.hpp"
 #include "modules/custom.hpp"
--- a/include/modules/systemd_failed_units.hpp
+++ b/include/modules/systemd_failed_units.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+#include <string>
+#include <giomm/dbusproxy.h>
+
+#include "ALabel.hpp"
+
+namespace waybar::modules {
+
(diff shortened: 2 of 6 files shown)


Alexays:
LGTM!
Can you also add a new entry in the github wiki?

oxalica:
> Can you also add a new entry in the github wiki?

Added https://github.com/Alexays/Waybar/wiki/Module:-Systemd-failed-units

HilkopterBob:
ive tried for days to use it. Please edit the wiki so its visible its a draft and not published

oxalica:
@HilkopterBob
> Please edit the wiki so its visible its a draft and not published

I tried in a private browser session without signing in GitHub and the page *is* public visible. There's no "draft" or anything relevant state, button or toggles in the edit page. [The link above](https://github.com/Alexays/Waybar/wiki/Module:-Systemd-failed-units) is also working publicly. Maybe you need to refresh your browser cache and try again? Or could you elaborate why you think it's "draft state and not published" and how to fix it?

HilkopterBob:
Sorry, i meant that it isnt visible that the feature isnt published yet and i think _that_ should be visible.
Sorry for the miscommunication.

oxalica:
> Sorry, i meant that the feature isnt published yet and i think _that_ should be visible in the wiki article.

Ah, that makes sense. I added a warning to the wiki. Thanks.

andyrtr:
Is there a way to style the colour of format-ok/format in style.css to make the small but helpful notification more visible or would this require additional changes to the source?

PR #2831, opened 2024-01-14
fix(tray): completely hide tray when it is empty

JManch:
Given the following tray:
populated

Before when empty:
space

After when empty:
no-space


Alexays:
Nice catch! thx :)

Issue #2815, opened 2024-01-08
GTK4


LukashonakV:
Hi @Alexays , @alebastr  I've done some digging into GTK4 and opportunity to migrate Waybar to it. 
So. I have the first base example of the Waybar which can be build and run with gtkmm4 . 
There only base engine... without any modules and event handlers. But I think it's a good start to migrate Waybar. 
@Alexays  is it possible for you to create for example separate branch and mark it with labels: unstable, gtk4 and to do cloning from my branch : https://github.com/LukashonakV/Waybar/tree/gtk4
I think it will provide an opportunity to contributors to make contributions, fix errors, find more effective approaches, bring new ideas, speedup the process

About versioning: 4.1.0  - the first digit means (stable release version. Currently it points to GTK4 chain), the second digit experimental version. The third digit update of the stable release. So for example the first stable release of the Waybar GTK4 should be looked like: 4.0.1. - The same versioning system now for example is used by Wine project.
So current 4.1.0 - meanse - Waybar 4 experimental 1 release without updates of the stable release

Once this approach is approved and separate branch is created it's possible to keep TODO list in this issue.

Thank you

----
Migration tasks
- [x] [Backlight](https://github.com/Alexays/Waybar/wiki/Module:-Backlight)
- [x] [Backlight/Slider](https://github.com/Alexays/Waybar/wiki/Module:-Backlight-Slider)
- [x] [Battery](https://github.com/Alexays/Waybar/wiki/Module:-Battery)
- [x] [Bluetooth](https://github.com/Alexays/Waybar/wiki/Module:-Bluetooth)
- [x] [Cava](https://github.com/Alexays/Waybar/wiki/Module:-Cava)
- [x] [CFFI]()
- [x] [Clock](https://github.com/Alexays/Waybar/wiki/Module:-Clock)
- [x] [CPU](https://github.com/Alexays/Waybar/wiki/Module:-CPU)
- [x] [Custom](https://github.com/Alexays/Waybar/wiki/Module:-Custom)
- [x] [Disk](https://github.com/Alexays/Waybar/wiki/Module:-Disk)
- [x] [DWL](https://github.com/Alexays/Waybar/wiki/Module:-Dwl)
- [x] [Gamemode](https://github.com/Alexays/Waybar/wiki/Module:-Gamemode)
- [x] [Group](https://github.com/Alexays/Waybar/wiki/Module:-Group)
[...]

PR #2836, opened 2024-01-16
Require gtk-layer-shell

alebastr:
BREAKING CHANGE: gtk-layer-shell is now required and unconditionally used. The corresponding config option is removed.

As a part of preparation for future versions of GTK, remove an ability to use wlr-layer-shell directly. The APIs it required were dropped in GTK4, and with the menus/tooltips positioning issue being practically unsolvable it doesn't make sense to keep maintaining the code.

See: #2815


Alexays:
Merging it to be apart of the upcoming release of the next month.
Thanks Aleksei!

PR #2852, opened 2024-01-22
Adding css reloader

dpayne:
This adds a css file watcher that reloads the style if the css files change.

A common use case is changing the colors of waybar using pywal. I wanted to make the change seamless, a full reset causes hyprland to briefly change window sizes since the gtk window is destroyed and recreated. Reloading only the style avoids that issue.


alebastr:
The code changes the global GTK app state from a thread, which is not safe.
It also creates an inotify instance; that is a limited resource, and Waybar already creates way too many :(

Have you considered using [Gio::FileMonitor](https://developer-old.gnome.org/glibmm/stable/classGio_1_1FileMonitor.html) or the underlying [g_file_monitor](https://docs.gtk.org/gio/method.File.monitor_file.html) instead? That would address both points.

Also, please consider adding a flag to disable the file watcher or even make it opt-in.

dpayne:
Updated with the changes suggested above.

Alexays:
LGTM, thx to both of you!
Can you add this new option to the man and github wiki? 🙏

dpayne:
I added the man page entry and edited the wiki.

Alexays:
Can you check the CI, freebsd seems to fail @dpayne

dpayne:
I believe I fixed the compilation error in freebsd, it was an issue with a missing import that only errors out on clang. I do not know why it hung after. It did seem to compile, this is from the workflow logs 

``` [337/339] Compiling C++ object test/waybar_test.p/.._src_config.cpp.o
[338/339] Compiling C++ object test/waybar_test.p/.._src_util_css_reload_helper.cpp.o
[339/339] Linking target test/waybar_test
1/1 waybar / waybar OK              0.12s

Ok:                 1   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   ```

Alexays:
LGTM, thx!
CI will rerun on master and timeout should be good

PR #2855, opened 2024-01-22
Fix module UPower display device poniter

SquishyPandaDev:
Force displayDevice to be a nullptr on class setup

Should fix seg faults and device not displaying

Current code expects UPower::displayDevice to be initially null. If it is not, as was the case on my box, it can lead to seg faults or the device not displaying information


Alexays:
Thanks!

PR #2862, opened 2024-01-24
Move `.empty` CSS class from Img to Box in the Image module

Sryther:
Since the GTK Image is hidden, the CSS class wasn't used for every cases.


Sryther:
`clang-format / build` is failing on unchanged code.

Alexays:
Thanks, i'll re-lint the code.

PR #2865, opened 2024-01-25
Using steps() in animation to reduce cpu usage.

chntllh:
Changed animation-timing-function to reduce cpu usage.

Lowering steps, reduces cpu usage even lower, but will make the animations choppy.

In my testing linear will always be more taxing than using steps.

### DIFF
--- a/resources/style.css
+++ b/resources/style.css
@@ -128,12 +128,13 @@ button:hover {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks, i prefer a lower cpu usage :)

Issue #2619, opened 2023-10-30
pulseaudio module don't reconnect au pipewire-pulse restart


cyrinux:
Hi,

I feel like I have several times to restart waybar to get the audio module updating.
For sure I reproduce it now when restart pipewire-pulse.service.

I'm bad in C++, doesn't those audio module try to auto-reconnect on the socket?

Regards

kit-ty-kate:
I also hit the same issue. A simple `systemctl --user restart pipewire.service` makes the pulseaudio module show wrong values.
For example `{volume}` changes to `100` and stay this way even when I increase or decrease volume.

Restarting waybar fixes it but it's a bit annoying.

PR #2866, opened 2024-01-25
pulseaudio: reconnect context when pulseaudio server restarts

Jay-716:
For some reason, I sometimes restart the pulseaudio server (pipewire-pulse actually). But after restart, waybar's pulseaudio module just stop updating. The solution is to restart the whole waybar.
So I think it would be better if waybar itself could reconnect to pulseaudio server after it restarts.
The pulseaudio context connection would failed according to the [documentation](https://freedesktop.org/software/pulseaudio/doxygen/def_8h.html#a892684c03cf9edaed1a95e609ec7573c). So we can simply reconnect the context there.

Tested on waybar latest git version and pipewire-pulse 1.0.0

Fixes: #2619

PS: I am not so good at C++ actually. So I am sorry if I do something stupid. :)

### DIFF
--- a/src/util/audio_backend.cpp
+++ b/src/util/audio_backend.cpp
@@ -80,7 +80,22 @@ void AudioBackend::contextStateCb(pa_context *c, void *data) {
                            nullptr, nullptr);
       break;
     case PA_CONTEXT_FAILED:
-      backend->mainloop_api_->quit(backend->mainloop_api_, 1);
+      // When pulseaudio server restarts, the connection is "failed". Try to reconnect.
+      // pa_threaded_mainloop_lock is already acquired in callback threads.
+      // So there is no need to lock it again.
+      if (backend->context_ != nullptr) {
+        pa_context_disconnect(backend->context_);
+      }
+      backend->context_ = pa_context_new(backend->mainloop_api_, "waybar");
+      if (backend->context_ == nullptr) {
+        throw std::runtime_error("pa_context_new() failed.");
+      }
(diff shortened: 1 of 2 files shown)


src/util/audio_backend.cpp
```diff
@@ -80,7 +80,22 @@ void AudioBackend::contextStateCb(pa_context *c, void *data) {
...
+      }
+      backend->context_ = pa_context_new(backend->mainloop_api_, "waybar");
+      if (backend->context_ == nullptr) {
+        throw std::runtime_error("pa_context_new() failed.");
+      }
+      pa_context_set_state_callback(backend->context_, contextStateCb, data);
+      if (pa_context_connect(backend->context_, nullptr, PA_CONTEXT_NOFAIL, nullptr) < 0) {
+        auto err =
+            fmt::format("pa_context_connect() failed: {}", pa_strerror(pa_context_errno(backend->context_)));
+        throw std::runtime_error(err);
```
Alexays, line 97: Can you create a function to merge this code with one in constructor? thx :)

Jay-716:
I add a `connectContext()` function. I hope i do it right.
Btw, I think `pa_context_set_state_callback()` should be called before `pa_context_connect()`.(it's reversed in the original code) Otherwise there may be a race condition, i.e. the context may connect too fast that the callback is not yet registered and then no callbacks will be called.
And should I change the comment style from `//` to `/**/` to match the style of existing code?
Thanks for reviewing!

Alexays:
LGTM, thx!

Issue #2875, opened 2024-01-31
making a drawer open on the right


hfytr:
Hello, I am trying to construct a drawer that shows pulseaudio, and then shows mpris on hover like so:

 ```
   "modules-left": ["group/audio"],
    "modules-center": ["hyprland/workspaces"],
    "modules-right": ["network", "battery", "clock", "tray"],
    "group/audio": {
      "orientation": "inherit",
      "drawer": {
          "transition-duration": 500,
          "children-class": "not-power",
          "transition-left-to-right": true,
      },
      "modules": ["pulseaudio", "mpris"]
    },
    //configs for mpris and pulseaudio
```

I have 1 minor and a major problem.

1) since the module is on the left, the drawer needs to slide out on the left, but it comes out on the left and I can't find a fix.

2) my modules are round, so pulseaudio needs to have different css when the drawer is extended vs non-extended.

**Comment by hfytr** [comment]
clarification: when the drawer is closed it should look like this:

**Comment by hfytr** [comment]
accident sry

**Comment by jramrath** [comment]
I can recreate your problem. `transition-left-to-right` doesn't seem to change anything.

When looking at the Object Hierarchy with [this](https://github.com/Alexays/Waybar/wiki/Styling#interactive-styling), `Revealer` comes before `EventBox` (the leaderObject):
![20240202_16h43m46s_grim](https://github.com/Alexays/Waybar/assets/35172494/9787e841-9261-449f-a33b-dfd771c8dd01)

Since changing the `transition_type` of `revealer` doesn't change anything, I tried  moving `Revealer` after the `leaderObject`:
```diff
diff --git a/src/group.cpp b/src/group.cpp
index cad36e51..997ab628 100644
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -73,7 +73,7 @@ Group::Group(const std::string& name, const std::string& id, const Json::Value&
     revealer.get_style_context()->add_class("drawer");
 
     revealer.add(revealer_box);
-    box.pack_start(revealer);
+    box.pack_end(revealer);
 
     addHoverHandlerTo(revealer);
   }
```

It worked:
![20240202_17h00m51s_grim](https://github.com/Alexays/Waybar/assets/35172494/aec00d56-982f-427c-9d22-31e4c116a6db)

@Alexays did I miss something on how `revealer` works? 

I'll add an If statement to check which function to use based on `transition-left-to-right` and make a PR.

I might even be able to remove to whole `preferedTransitionType` thing.

[...]

PR #2881, opened 2024-02-02
Fix: drawer not appearing on configured side

jramrath:
See #2875


jramrath:
Turns out we can't remove the transition_type handling. I've added this comment:
```c++
/* The transition direction of a drawer is not actually determined by the transition type,
 * but rather by the order of 'box' and 'revealer_box':
 *   'REVEALER_TRANSITION_TYPE_SLIDE_LEFT' and 'REVEALER_TRANSITION_TYPE_SLIDE_RIGHT'
 *   will result in the same thing.
 * However: we still need to differentiate between vertical and horizontal transition types.
 */
```

Alexays:
LGTM thx.

PR #2886, opened 2024-02-04
Bluetooth module: fetch battery percentage from upower if not found from bluez

levnikmyskin:
# What is this PR about?
Current bluetooth module shows battery percentage from bluez experimental features. This has two drawbacks:

1.  Not super user friendly, as users need to activate these features;
2. Most importantly, a lot of devices (at least in my experience) do not report battery percentage anyway, despite having it in upower;

## Upower implementation
In this PR, the bluetooth module will first try to fetch battery percentage from bluez and, if not available, will try fetching it from upower.

## Other details
At the moment, the code is getting all upower devices, and then looking for the one with the correct serial number. I don't know if there's a better way of doing this

### DIFF
--- a/src/modules/bluetooth.cpp
+++ b/src/modules/bluetooth.cpp
@@ -340,13 +340,13 @@ auto waybar::modules::Bluetooth::getDeviceBatteryPercentage(GDBusObject* object)
     -> std::optional<unsigned char> {
   GDBusProxy* proxy_device_bat =
       G_DBUS_PROXY(g_dbus_object_get_interface(object, "org.bluez.Battery1"));
-  GDBusProxy* proxy_device = G_DBUS_PROXY(g_dbus_object_get_interface(object, "org.bluez.Device1"));
   if (proxy_device_bat != NULL) {
     unsigned char battery_percentage = getUcharProperty(proxy_device_bat, "Percentage");
     g_object_unref(proxy_device_bat);
 
     return battery_percentage;
   }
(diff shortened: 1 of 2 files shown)


src/modules/bluetooth.cpp
```diff
@@ -339,12 +340,36 @@ auto waybar::modules::Bluetooth::getDeviceBatteryPercentage(GDBusObject* object)
     -> std::optional<unsigned char> {
   GDBusProxy* proxy_device_bat =
       G_DBUS_PROXY(g_dbus_object_get_interface(object, "org.bluez.Battery1"));
+  GDBusProxy* proxy_device = G_DBUS_PROXY(g_dbus_object_get_interface(object, "org.bluez.Device1"));
   if (proxy_device_bat != NULL) {
     unsigned char battery_percentage = getUcharProperty(proxy_device_bat, "Percentage");
     g_object_unref(proxy_device_bat);
 
     return battery_percentage;
   }
```
Alexays, line 349:
```suggestion
  if (proxy_device_bat != NULL) {
    unsigned char battery_percentage = getUcharProperty(proxy_device_bat, "Percentage");
    g_object_unref(proxy_device_bat);

    return battery_percentage;
  }
  GDBusProxy* proxy_device = G_DBUS_PROXY(g_dbus_object_get_interface(object, "org.bluez.Device1"));
```

Alexays: Save the init overhead if proxy_device_bat is not null

levnikmyskin:
Thank you for your suggestions, implemented :)

levnikmyskin:
Hi @Alexays, is there anything missing here or could we merge this? thanks :)

Issue #366, opened 2019-06-03
Support CSS text-align


pedrocr:
For things like sway/window it would be nice to have CSS support for `text-align` to be able to adjust how the text is displayed in the bar.

Alexays:
Atm text-align is not supported on gtk label, we need to wait a gtk update

pedrocr:
Seems like gtk does that with something other than CSS:

https://developer.gnome.org/gtk3/stable/GtkLabel.html#gtk-label-set-justify

So maybe adding `text-align` is not in the plans.

Alexays:
But for your case, why not simply move module to the left/right block?

pedrocr:
The module is in the left block. But the actual text of the window title is centered and I wanted to make it left aligned. It's not too bad this way either, just a tweak to see if it felt better.

Alexays:
Can you share me a screenshot?
Window title module width must not be larger than the text size

pedrocr:
Sure, here it is:

![waybar](https://user-images.githubusercontent.com/246322/59287125-f2d45580-8c68-11e9-8ddc-127ad3d6fe1c.png)

Config is:

```
    "modules-left": ["sway/workspaces", "sway/window"],
    "modules-center": [],
    "modules-right": ["cpu", "memory", "temperature", "pulseaudio", "backlight", "network", "battery", "tray", 
```

I can add some CSS coloring on `sway/window` if that helps diagnose it.

Alexays:
Fixed with 3c9b533997ae21ebdd09dd40c7612616fbf0beb5

pedrocr:
Thanks. Just tested it. Now it's left aligned when put in `modules-left` and centered in the total bar when in `modules-center`, so seems to be working as intended. The previous behavior was actually not horrible though. It centered the window title in the space between left and right modules.

krnl0138:
@pedrocr @Alexays Really sorry, but is it possible to get back to the previous state with centered and filled module between two others? Or to toggle between those options?

MyNameIsCosmo:
Would using [GtkJustification](https://developer.gnome.org/gtk3/stable/gtk3-Standard-Enumerations.html#GtkJustification) to get `text-align: [left,center,right]` be a course of action?

I'd be happy to start a PR implementing this if it would be accepted.

Alexays:
@MyNameIsCosmo You can implement a justification for Label modules. it'll be a great improvement :)

MatthiasGrandl:
[...]

PR #2887, opened 2024-02-05
Add justify config option for Labels

dann-merlin:
This is PR adds the option to specify a "justify" field for aligning text modules. It should be either "left", "right" or "center".

This is especially useful for centering the text in labels on vertical bars. In situations like this:

`"format": "{icon}\n{value}%",`
This kind of fix was already mentioned in #366, however it was never implemented, so here's my take on it.

An example config for this:
```
    "memory": {
        "format": "\n{}%", // That's a character for a RAM module in the font I'm using.
        "justify": "center",
    },
```

### DIFF
--- a/man/waybar-backlight.5.scd
+++ b/man/waybar-backlight.5.scd
@@ -30,7 +30,11 @@ The *backlight* module displays the current backlight level.
 
 *align*: ++
 	typeof: float ++
-	The alignment of the text, where 0 is left-aligned and 1 is right-aligned. If the module is rotated, it will follow the flow of the text.
+	The alignment of the label within the module, where 0 is left-aligned and 1 is right-aligned. If the module is rotated, it will follow the flow of the text.
+
+*justify*: ++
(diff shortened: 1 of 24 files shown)


Alexays:
Nice, can you add this to the man and github wiki?

dann-merlin:
This would need to be added to basically all module documentations, right?

twohexed:
Would love to see this implemented, I myself have been wanting this!

Alexays:
Sorry for the delay, yep should be added to all modules that inherit label class :)

machitgarha:
So all we need is documentation for this to get merged?

dann-merlin:
Alright, I've added documentation for *justify* whereever I was able to find documentation for *align*. Which should be everything that inherits the label class.
I also slightly changed the documentation for *align*, in order to clarify the difference between these two options.

Let me know if there's anything else!

Alexays:
LGTM, thx!

PR #2888, opened 2024-02-06
Search for the first hwmon* directory

imran1008:
Background and Motivation
-------------------------
When the `hwmon-path-abs` and the `input-filename` fields are used for the temperature module, we evaluated the following path:

```
[hwmon-path-abs] / [gap] / [input-filename]
```

where `gap` is the first file or directory in the `hwmon-path-abs` directory. This usually works but it doesn't seem to work for NVME or WiFi temperature sensors. For those cases, there are a bunch of other files in the `hwmon-path-abs` directory. In the bad case, the first selected file is not the one with the prefix `hwmon` and we end up checking the wrong location for the `input-filename`.

Change description
------------------
We are simply going through the `hwmon-path-abs` directory and searching for the first file/directory that begins with `hwmon`.

Test case
---------
I tested this on a AMD based Framework 13 laptop.

### DIFF
--- a/src/modules/temperature.cpp
+++ b/src/modules/temperature.cpp
@@ -24,11 +24,15 @@ waybar::modules::Temperature::Temperature(const std::string& id, const Json::Val
       }
     }
   } else if (config_["hwmon-path-abs"].isString() && config_["input-filename"].isString()) {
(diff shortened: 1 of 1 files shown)


Alexays:
LGTM

PR #2891, opened 2024-02-06
net: add ethernet link speed 

### LINKED ISSUE #2636: Option to show the Link Speed of a wired network connection
I feel that it would be really useful to have the ability to show the current network Link Speed (simillar to the output of `ethtool enp3s0f1 | grep Speed` for example) as part of the network module.

awariac:
Resolves #2636
Link speed  in read from `/sys/net/<device>/speed`, works only for Ethernet interfaces.
Includes additional option for `pow_format` to skip decimal place, so 1000 will be printed as `1Gb/s`, instead of `1.0Gb/s`, but 2500 always as `2.5Gb/s`.

### DIFF
--- a/include/modules/network.hpp
+++ b/include/modules/network.hpp
@@ -54,6 +54,7 @@ class Network : public ALabel {
   const std::string getNetworkState() const;
   void clearIface();
   std::optional<std::pair<unsigned long long, unsigned long long>> readBandwidthUsage();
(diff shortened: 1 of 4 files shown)


b0ch3nski:
@Alexays :slightly_smiling_face:

abraxxa:
I'd love to see this getting merged! ❤️

PR #2894, opened 2024-02-07
Add css classes to hyprland special workspaces

jramrath:
This adds the following css classes to special workspaces in hyprland:
- active (+ correct behavior on monitor focus change)
- visible
- empty 

Styling only normal workspaces or only special workspaces is still possible. Example with `active`:

```css
#workspaces button.active:not(.special) {} /* only normal */
#workspaces button.active.special       {} /* only special */
```


Alexays:
LGTM, thx

PR #2899, opened 2024-02-08
Make battery estimate inaccessible in the "Full" state on default config

### LINKED ISSUE #2893: Battery time estimate is viewable while charging at 100%, leading to very large/ broken time estimates
The battery life estimate can't usually be viewed when the battery is charging, but when the battery reaches 100%, the battery doesn't seem to be considered to be charging anymore. This allows the time estimate for the battery life to be viewed, leading to comedically high battery life estimates being displayed:
image

This estimate also continues to rise extremely quickly; In the span of ~5 minutes, the estimate increases by over 250 hours.

A simple solution to this is likely to just disable viewing the battery estimate when the battery is charging at 100%, the same as any other scenario where the battery is charging.

I am running Waybar v0.9.24 on Fedora 39's Sway spin.

OpenSauce04:
Fixes #2893

This PR makes it so that, when the `battery` module is in the `Full` state, the battery life estimate is inaccessible similarly to how the module acts in the `Charging` and `Plugged` states.

This is done by adding a line in the default config which sets a format for the `Full` state. Setting a format for a given state disables access to the `format-alt` display while that state is active.

There are possibly better ways to approach this than a config change, but I couldn't think of any that wouldn't limit the functionality of other configs that may use `format-alt` here for things other than displaying a battery estimate, in which case the issue wouldn't occur.

If there is a better solution to this problem, please let me know and I'll look into that instead.

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -113,6 +113,7 @@
             "critical": 15
         },
         "format": "{capacity}% {icon}",
+        "format-full": "{capacity}% {icon}",
         "format-charging": "{capacity}% ",
         "format-plugged": "{capacity}% ",
         "format-alt": "{time} {icon}",


Alexays:
LGTM, thx

PR #2917, opened 2024-02-13
Add always on option for hyprland/submap

### LINKED ISSUE #2896: Feature Request: add option for default mode in sway/mode and hyprland/submap
Add an option to sway/mode and hyprland/submap to display the widget when no mode/submap is entered with a user-defined message.

The configuration could look something like this:
```
"sway/mode": {
    "format": " {}",
    "max-length": 50,
    "always-on": true,
    "no-mode-message": "No Mode",
}
```

"always-on" is a boolean, if false the widget behaves as it does now, if true, the widget is still rendered when no mode/submap is active using the "no-mode-message" string as a placeholder for the mode name.

alttabber:
Partially fixes #2896


Alexays:
Thanks!
Can you add it to the github wiki?

Issue #2914, opened 2024-02-13
How to change text of tooltips for custom modules?


bybrl:
I use this config, but want to use the supplied text instead of the icon while hovering the module.

```
{
    "layer": "top",
    "position": "top",
    "spacing": 0,
    "height": 34,
    "modules-left": [
        "custom/logo",
        "hyprland/workspaces","custom/launcher", 
        "custom/kitty",  "custom/firefox",
    ],
    "modules-center": [
        "clock",
        "custom/arch"
    ],
    "modules-right": [
        "tray",
        "memory",
        "network",
        "pulseaudio",
        "battery",
        "custom/power"
    ],
    "custom/launcher": {
            "format": "󱗼",
            "tooltip-format": "Wofi drun Menu",
    	    "on-click": "wofi --show drun -I -a -n -W 500 -H 376 -s ~/.config/wofi/themes/dracula.css ",
            "color": "#FF6347",
            "button": true,
        },
        "custom/logo":{
        	"format": "",
        	"tooltip-format": "Full system upgrade",
        	"on-click": "kitty --class float $HOME/.local/share/bin/syu"
        },
        "custom/firefox":{
        	"format": "",
            "on-click": "firefox",
            "tooltip-format": "Firefox web browser",
        },
        "custom/kitty":{
        	"format": "󰄛",
        	"tooltip-format": "Kitty terminal",
        	"on-click": "kitty $HOME"
        },
    "wlr/taskbar": {
        "format": "{icon}",
        "on-click": "activate",
        "on-click-right": "fullscreen",
        "icon-theme": "WhiteSur",
        "icon-size": 25,
        "tooltip-format": "{title}"
    },
    "hyprland/workspaces": {
        "on-click": "activate",
        "format": "{icon}",
        "format-icons": {
            "default": "",
            "1": "1",
            "2": "2",
            "3": "3",
            "4": "4",
            "5": "5",
            "6": "6",
            "7": "7",
            "8": "8",
            "9": "9",
            "active": "󱓻",
            "urgent": "󱓻"
        },
        "persistent_workspaces": {
            "1": [],
            "2": [],
            "3": [],
            "4": [],
            "5": []
        }
    },
    "memory": {
        "interval": 5,
        "format": "󰍛 {}%",
        "max-length": 10
    },
    "tray": {
        "spacing": 10
    },
    "clock": {
    	"timezone": "America/Chicago",
        "tooltip-format": "<tt>{calendar}</tt>",
[...]

PR #2923, opened 2024-02-16
Add check for tooltip-format for custom modules

jones-josh:
Issue https://github.com/Alexays/Waybar/issues/2914 tries to use `tooltip-format` on a custom module without a script. This change results in `tooltip-format` entries in config being used as the tooltip format, _**over**_ any tooltip-related return from the output of a custom script if there is one.

This makes the most sense to me as it is the more visible change. If I see a `tooltip-format` entry in the config file, I expect that this is truth, not the output of a script in a different place. While it wouldn't be best practice to setting the tooltip in both places, I'd argue that a user would expect that their change to the config would be reflected in the case of using a 3rd party script they don't fully understand.

### DIFF
--- a/man/waybar-custom.5.scd
+++ b/man/waybar-custom.5.scd
@@ -107,6 +107,11 @@ Addressed by *custom/<name>*
 	default: true ++
 	Option to disable tooltip on hover.
 
+*tooltip-format*: ++
+	typeof: string ++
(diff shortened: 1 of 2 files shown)


jones-josh:
Of course will add to man and wiki as appropriate if approved of.

Alexays (approved):

jones-josh:
Will add to wiki once merged

PR #2926, opened 2024-02-17
build: refactorings and deprecation fixes

alebastr:
See individual commits for details.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -22,8 +22,6 @@ endif
 
 if compiler.has_link_argument('-lc++fs')
     cpp_link_args += ['-lc++fs']
-elif compiler.has_link_argument('-lc++experimental')
-    cpp_link_args += ['-lc++experimental']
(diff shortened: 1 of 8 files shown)


meson.build
```diff
@@ -176,6 +162,7 @@ src_files = files(
     'src/ALabel.cpp',
     'src/AIconLabel.cpp',
     'src/AAppIconLabel.cpp',
+    'src/modules/bluetooth.cpp',
```
alebastr, line 165: This wasn't guarded by any OS-specific conditions (`-Dlogind` doesn't count as one), but somehow I doubt bluez is available on various BSD systems.
Maybe `src/modules/bluetooth.cpp` should go under `is_linux` to avoid confusion. @jbeich, what do you think?

Alexays: should definitely be under is_linux

Alexays:
Thanks!

PR #2930, opened 2024-02-18
fix clang-tidy errors in hyprland module

zjeffer:
CC: @Syndelis @khaneliman because you guys seem to be the most recently active in these files

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -22,7 +22,7 @@ CheckOptions:
   - { key: readability-identifier-naming.FunctionCase,           value: camelBack  }
   - { key: readability-identifier-naming.VariableCase,           value: camelBack  }
   - { key: readability-identifier-naming.PrivateMemberCase,      value: camelBack  }
-  - { key: readability-identifier-naming.PrivateMemberPrefix,    value: m_         }
+  - { key: readability-identifier-naming.PrivateMemberSuffix,    value: _          }
   - { key: readability-identifier-naming.EnumCase,               value: CamelCase  }
   - { key: readability-identifier-naming.EnumConstantCase,       value: UPPER_CASE }
   - { key: readability-identifier-naming.GlobalConstantCase,     value: UPPER_CASE }
--- a/include/modules/hyprland/backend.hpp
+++ b/include/modules/hyprland/backend.hpp
@@ -20,8 +20,8 @@ class IPC {
  public:
   IPC() { startIPC(); }
 
-  void registerForIPC(const std::string&, EventHandler*);
-  void unregisterForIPC(EventHandler*);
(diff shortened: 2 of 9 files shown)


alebastr:
`member_` comes from [Google code style](https://google.github.io/styleguide/cppguide.html#Variable_Names) recommended in the project README. The style isn't being followed strictly (esp. in regards to enum/constant names), but at least that part was consistent. Well, mostly.

Maybe modifying `.clang-tidy` to follow the project style would be better?

zjeffer:
I personally *really* dislike the trailing underscore style to mark member variables (which is why I initially set it to use the `m_` style [here](https://github.com/Alexays/Waybar/pull/2595#issuecomment-1791638922)). I don't think it's very clear at all which variables are member variables and which aren't. That style also often gets used for temporary variables in the code, so we're mixing styles (I think there's a couple ocurrences of this mixing happeing in the code I changed).

I'm obviously not the one to choose this, @Alexays what do you think?

alebastr:
I don't even mind the `m_` style as long as it is consistent across the codebase. I.e. someone does the complete conversion and adds [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).
Obviously, that would require manual merge and update for most of the pending PRs. That's the unavoidable price of consistency.

Currently, ~10 files of 200 are using the `m_` prefix style. Any change to the rest of the sources would make clang-tidy CI job fail, effectively rendering it useless. And all these `Invalid case style for...` editor warnings don't add to my happiness.

> That style also often gets used for temporary variables in the code, so we're mixing styles

That's fair. And I might even introduced some of those, just like inconsistent case for constants and enums. But fixing and enforcing style for that would be less disruptive.

zjeffer:
> I.e. someone does the complete conversion and adds [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

I would even volunteer to do this by myself for every single file in the repo, that's how much I hate the trailing underscore style ;)

That .git-blame-ignore-revs file is really interesting, I didn't know this feature existed.

Alexays:
I don't mind either between `m_` or `_`, I'd be more in favor of going back to the suffix to follow the readme guidelines and avoid too many conflicts over current PRs.
And it will surely be easier this way to get the green check on the clang-tidy side.

zjeffer:
>  And it will surely be easier this way to get the green check on the clang-tidy side.

This shouldn't really be an issue because clang-tidy only looks at the files changed in the PR, not all files. 

But I understand that current PRs would have lots of annoying conflicts. I'll look into changing the clang-tidy rule and making the necessary changes in the hyprland module.

PR #2934, opened 2024-02-19
Fixed window#waybar.swallowing for module hyprland/window

tomben13:
Due to [this change](https://github.com/hyprwm/Hyprland/pull/3186) in hyprland, the json value of the `"swallowing"` field is `"0x0"` when no window is swallowed, compared to `null` before.


tomben13:
Question for maintainers : should I create an issue that this PR resolves ? Or is it fine like this ?

Alexays:
Can you also keep the is null check to be backward compatible? :)

tomben13:
> Can you also keep the is null check to be backward compatible? :)
Yes, I will do tomorrow

PR #2939, opened 2024-02-20
Improve CI build times

alebastr:
Should be the last one for today :)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,9 +18,6 @@ jobs:
           - opensuse
           - gentoo
         cpp_std: [c++20]
-        include:
-          - distro: fedora
-            cpp_std: c++20
 
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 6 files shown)


alebastr:
@Alexays forgot to mention, can you please update the container images?

alebastr:
@Alexays I think you've accidentally pushed an arm64 image instead of amd64. That's the cause of CI breakage in #2944

Alexays:
Oups, should be good now, thx

alebastr:
Hmm, that accidentally broke clang-tidy action. Sorry :(

    /__t/Python/3.11.8/x64/bin/python: error while loading shared
libraries: libpython3.11.so.1.0: cannot open shared object file: No such
file or directory

I thought it won't be affected because they use actions/setup-python action
internally, but apparently they don't set the LD_LIBRARY_PATH, thus the
installation they got via actions/setup-python cannot find its own library
and relies on a system libpython.

Adding libpython3.11 to the container image could help. Or updating the
cpp-linter action to the latest release (v2.9.1) and adding libpython3.10

PR #2946, opened 2024-02-20
Fix #2945: Persistent workspace disappear after being active

### LINKED ISSUE #2945: hyprland/workspaces: Persistent workspace disappear after being active
Hi,
With commit 11310b89f063a305de0d23aa4dd21d6ef365a776, a workspace set as persistent is not shown anymore after being active and then destroyed by Hyprland.

Config:
```
    "hyprland/workspaces": {
        "all-outputs": true,
        "format": " {icon}",
        "persistent-workspaces": {
            "*": [1, 2, 3, 4, 5, 6, 7, 8]
        },
    },
```

How to reproduce:

1. Start waybar with persistent workspaces. All persistent workspaces are shown correctly at start. Using my config above, it'll show this: <br/>20240220_22h51m34s_grim
2. Switch to a an empty workspace in Hyprland, set as persistent in waybar. For example, switching to `i`, it works as intended: <br/>20240220_22h51m51s_grim
3. Switch to another workspace in Hyprland. Workspace `i` is destroyed in Hyprland, and not persistent anymore in Waybar (it is not listed anymore): <br/>20240220_22h52m04s_grim

Reverting commit 11310b89f063a305de0d23aa4dd21d6ef365a776 fixes this behavior.
CC @zjeffer for the #2603.

Edit: I forgot to specify that I don't set my workspaces as persistent in Hyperland, only in Waybar.

zjeffer:
If a user has a persistent workspace waybar config, and then switches to an empty workspace that is configured as persistent in the config, that empty workspace will be recreated with persistency disabled. 

This PR fixes this by not updating the persistency. In hindsight this line was never really needed in the first place.

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -599,8 +599,7 @@ void Workspaces::createWorkspace(Json::Value const &workspace_data,
       });
 
   if (workspace != m_workspaces.end()) {
-    // don't recreate workspace, but update persistency if necessary
-    (*workspace)->setPersistent(workspace_data["persistent"].asBool());
+    // don't recreate workspace if it already exists
     return;
   }
 


src/modules/hyprland/workspaces.cpp
```diff
@@ -599,8 +599,7 @@ void Workspaces::createWorkspace(Json::Value const &workspace_data,
       });
 
   if (workspace != m_workspaces.end()) {
-    // don't recreate workspace, but update persistency if necessary
-    (*workspace)->setPersistent(workspace_data["persistent"].asBool());
+    // don't recreate workspace if it already exists
     return;
```
aruhier, line 603: I think you might prefer this:

```
  if (workspace != m_workspaces.end()) {
    // don't recreate workspace, but update persistency if necessary
    //
    // An update is necessary only when the persistency was false and is now true. There is no way
    // in Hyprland to update the persistency of an existing workspace but to remove and recreate the
    // workspace. Therefore, going from persistency true to false cannot happen in this block.
    // Without this check, the persistency set in the waybar config will be overridden by the one
    // set in Hyprland when Hyprland creates the workspace.
    if (!(*workspace)->isPersistent()) {
      (*workspace)->setPersistent(workspace_data["persistent"].asBool());
    }
    return;
  }
```

aruhier (changes requested):

zjeffer:
Closing in favour of #2967

PR #2956, opened 2024-02-22
Gtk4. Migration. ALabel, AModule, factory, clock

LukashonakV:


LukashonakV:
Hi @Alexays , @alebastr 
Finally AModule, ALabel handle events correct

LukashonakV:
Partially covers #2815

LukashonakV:
+ backlight and harness

LukashonakV:
Hi @Alexays , @alebastr 

Actually the main work is almost done

Leftovers are:

1. ISSUES #2815 - not covered modules
2. Need to sync modules one by one with the master since c420b406
3. Need to satisfy linter
4. Need to rebuild docker containers
5. There is a bug with the Gtk::Label. When module sets the value to the set_tooltip_text/set_tooltip_markup Gtk::Label looks like try to refresh all tooltips for all opened Labels. And this leads to stealing of the focus. This point I'm trying to investigate now
Can you please take this PR for now.

Thanks

During migration I found out 

1. UPower module is rewritten and should be rewritten in gtk3 too
2. Client in GTK4 can be simplified due to no need to manage wayland logic more as Waybar relies on Gtk framework

LukashonakV:
Hi @alebastr , @Alexays  is it possible to have discussion somewhere regarding GTK4 and Waybar architecture ? 
The reason is I'm asking for : according documentation and the approach of GTK this framework since 3.36 release is no more thread safe ... and the approach is GTK API should be called from the main loop. Actually looking on the current Waybar architecture I'm inclined to think this architecture should be redesigned. Actually Waybar modules in their seoarate threads just need to do some logic and then signals main thread to start calling of the GTK API.
So I need someones opinion on this point of view... or to have discussion with some agreed approach...

LukashonakV:
> Hi @alebastr , @Alexays is it possible to have discussion somewhere regarding GTK4 and Waybar architecture ? The reason is I'm asking for : according documentation and the approach of GTK this framework since 3.36 release is no more thread safe ... and the approach is GTK API should be called from the main loop. Actually looking on the current Waybar architecture I'm inclined to think this architecture should be redesigned. Actually Waybar modules in their separate threads just need to do some logic and then signals main thread to start calling of the GTK API. So I need someones opinion on this point of view... or to have discussion with some agreed approach...

Need to wait for investigation of the core GTK team. To me it seems GTK issue which is under discussion here https://discourse.gnome.org/t/gtk4-migration-label-and-tooltip-from-gtk3/20710/2

LukashonakV:
> > Hi @alebastr , @Alexays is it possible to have discussion somewhere regarding GTK4 and Waybar architecture ? The reason is I'm asking for : according documentation and the approach of GTK this framework since 3.36 release is no more thread safe ... and the approach is GTK API should be called from the main loop. Actually looking on the current Waybar architecture I'm inclined to think this architecture should be redesigned. Actually Waybar modules in their separate threads just need to do some logic and then signals main thread to start calling of the GTK API. So I need someones opinion on this point of view... or to have discussion with some agreed approach...
> 
> Need to wait for investigation of the core GTK team. To me it seems GTK issue which is under discussion here [gtk4-migration-label-and-tooltip-from-gtk3](https://discourse.gnome.org/t/gtk4-migration-label-and-tooltip-from-gtk3/20710/8?u=viktar)

Finally no action from waybar side, https://gitlab.gnome.org/GNOME/gtk/-/commit/84fd420271928d171fd0d66a7f7d773fabb3ee2a
GTK fixed an issue with flickering tooltips in GTK4

PR #2969, opened 2024-02-25
Config tweaks

alebastr:
Small non-functional changes to the default configuration

### DIFF
--- a/resources/config
+++ b/resources/config
@@ -5,9 +5,31 @@
     // "width": 1280, // Waybar width
     "spacing": 4, // Gaps between modules (4px)
     // Choose the order of the modules
-    "modules-left": ["sway/workspaces", "sway/mode", "sway/scratchpad", "custom/media"],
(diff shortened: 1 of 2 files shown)


zjeffer:
Is there a reason the default config has no extension? Adding `.jsonc` would make it slightly easier for editors to deduce the file type.

Alexays:
Nope this can be renamed, waybar already lookup for config.jsonc

zjeffer:
No, I meant why are we putting this file in the repo without any extension?

alebastr:
> No, I meant why are we putting this file in the repo without any extension?

Because it was in the repo before `.jsonc` extension support. Next major release could be a good point to rename the config, both in the repo and in the installation.

For the record, that won't fix the file type detection for all editors. E.g. Emacs still needs `// -*- mode: jsonc; -*-` modeline.

alebastr:
#2972

Issue #1539, opened 2022-05-01
[Feature] Add module to configure PowerProfiles


columbarius:
[power-profiles-daemon](https://gitlab.freedesktop.org/hadess/power-profiles-daemon) is a daemon, which lets you change the Powerprofile, which is useful to prolong battery life on laptops. 

I've created a prototype in [python](https://gist.github.com/columbarius/88c753aa4f4582371243d9a98756184b).

Would you be interested if I would port this to a c++ module?

I'm not familiar with C++ so I probably need some help doing this.

colemickens:
Very disappointing this hasn't been added. I simply am not qualified to write C++ but this is a single executable or dbus call to get/set this value. Not sure how this doesn't have more attention from laptop users.

picnoir:
I can't close this issue by myself, but this has been implemented by https://github.com/Alexays/Waybar/pull/2971. 

See manual or the [module wiki page](https://github.com/Alexays/Waybar/wiki/Module:-PowerProfilesDaemon) for more information.

PR #2971, opened 2024-02-25
Introduce a power-profiles-daemon module

picnoir:
We introduce a module in charge to display and toggle on click the power profiles via power-profiles-daemon.

https://gitlab.freedesktop.org/upower/power-profiles-daemon

This daemon is pretty widespread. It's the component used by Gnome and KDE to manage the power profiles. The power management daemon is a pretty important software component for laptops and other battery-powered devices.

We're using the daemon DBus interface to:

- Fetch the available power profiles.
- Track the active power profile.
- Change the active power profile.

The original author recently gave up maintenance on the project. The Upower group took over the maintenance burden… …and created a new DBus name for the project. The old name is still advertised for now. We use the old name for compatibility sake: most distributions did not release 0.20, which introduces this new DBus name. We'll likely revisit
this in the future and point to the new bus name. See the inline comment for more details.

Given how widespread this daemon is, I activated the module in the default configuration. I'm not 100% sure about myself on that one though :)

Demo video: https://social.alternativebit.fr/media/514a598a09c6c9047e61af405d188aa7b99a5730015c95ad787f7c023ea1f461.mp4

### DIFF
--- a/include/modules/power_profiles_daemon.hpp
+++ b/include/modules/power_profiles_daemon.hpp
@@ -7,25 +7,25 @@
 
 namespace waybar::modules {
 
-typedef struct {
+struct Profile {
   std::string name;
   std::string driver;
-} Profile;
+};
 
 class PowerProfilesDaemon : public ALabel {
  public:
   PowerProfilesDaemon(const std::string&, const Json::Value&);
-  ~PowerProfilesDaemon();
+  ~PowerProfilesDaemon() override;
   auto update() -> void override;
-  void profileChanged_cb(const Gio::DBus::Proxy::MapChangedProperties&,
-                         const std::vector<Glib::ustring>&);
+  void profileChangedCb(const Gio::DBus::Proxy::MapChangedProperties&,
+                        const std::vector<Glib::ustring>&);
   void populateInitState();
-  virtual bool handleToggle(GdkEventButton* const& e);
+  bool handleToggle(GdkEventButton* const& e) override;
 
  private:
   // Look for a profile name in the list of available profiles and
   // switch activeProfile_ to it.
-  void switchToProfile_(std::string);
+  void switchToProfile(std::string const&);
   // Used to toggle/display the profiles
   std::vector<Profile> availableProfiles_;
   // Points to the active profile in the profiles list
@@ -33,7 +33,7 @@ class PowerProfilesDaemon : public ALabel {
   // Current CSS class applied to the label
   std::string currentStyle_;
   // DBus Proxy used to track the current active profile
-  Glib::RefPtr<Gio::DBus::Proxy> power_profiles_proxy_;
+  Glib::RefPtr<Gio::DBus::Proxy> powerProfilesProxy_;
   sigc::connection powerProfileChangeSignal_;
 };
 
--- a/include/modules/power_profiles_daemon.hpp
+++ b/include/modules/power_profiles_daemon.hpp
@@ -1,0 +1,38 @@
+# pragma once
+
+#include <fmt/format.h>
+
+#include "ALabel.hpp"
+#include "giomm/dbusproxy.h"
+
+namespace waybar::modules {
+
+typedef struct {
+  std::string name;
+  std::string driver;
+} Profile;
+
+class PowerProfilesDaemon : public ALabel  {
+ public:
+  PowerProfilesDaemon(const std::string&, const Json::Value&);
+  ~PowerProfilesDaemon();
+  auto update() -> void override;
+  void profileChanged_cb( const Gio::DBus::Proxy::MapChangedProperties&, const std::vector<Glib::ustring>&);
+  void populateInitState();
+  virtual bool handleToggle(GdkEventButton* const& e);
+ private:
+  // Look for a profile name in the list of available profiles and
+  // switch activeProfile_ to it.
+  void switchToProfile_(std::string);
+  // Used to toggle/display the profiles
+  std::vector<Profile> availableProfiles_;
+  // Points to the active profile in the profiles list
+  std::vector<Profile>::iterator activeProfile_;
+  // Current CSS class applied to the label
+  std::string currentStyle_;
+  // DBus Proxy used to track the current active profile
+  Glib::RefPtr<Gio::DBus::Proxy> power_profiles_proxy_;
+  sigc::connection powerProfileChangeSignal_;
+};
+
+}
--- a/include/modules/power_profiles_daemon.hpp
+++ b/include/modules/power_profiles_daemon.hpp
@@ -14,18 +14,24 @@ struct Profile {
 
 class PowerProfilesDaemon : public ALabel {
  public:
-  PowerProfilesDaemon(const std::string&, const Json::Value&);
+  PowerProfilesDaemon(const std::string &, const Json::Value &);
   ~PowerProfilesDaemon() override;
   auto update() -> void override;
-  void profileChangedCb(const Gio::DBus::Proxy::MapChangedProperties&,
-                        const std::vector<Glib::ustring>&);
+  void profileChangedCb(const Gio::DBus::Proxy::MapChangedProperties &,
+                        const std::vector<Glib::ustring> &);
+  void busConnectedCb(Glib::RefPtr<Gio::AsyncResult> &r);
+  void getAllPropsCb(Glib::RefPtr<Gio::AsyncResult> &r);
+  void setPropCb(Glib::RefPtr<Gio::AsyncResult> &r);
   void populateInitState();
-  bool handleToggle(GdkEventButton* const& e) override;
+  bool handleToggle(GdkEventButton *const &e) override;
 
  private:
+  // True if we're connected to the dbug interface. False if we're
+  // not.
+  bool connected_;
   // Look for a profile name in the list of available profiles and
   // switch activeProfile_ to it.
-  void switchToProfile(std::string const&);
+  void switchToProfile(std::string const &);
   // Used to toggle/display the profiles
   std::vector<Profile> availableProfiles_;
   // Points to the active profile in the profiles list
--- a/include/modules/power_profiles_daemon.hpp
+++ b/include/modules/power_profiles_daemon.hpp
@@ -15,7 +15,6 @@ struct Profile {
 class PowerProfilesDaemon : public ALabel {
  public:
   PowerProfilesDaemon(const std::string &, const Json::Value &);
-  ~PowerProfilesDaemon() override;
   auto update() -> void override;
   void profileChangedCb(const Gio::DBus::Proxy::MapChangedProperties &,
                         const std::vector<Glib::ustring> &);
@@ -38,12 +37,10 @@ class PowerProfilesDaemon : public ALabel {
   std::vector<Profile>::iterator activeProfile_;
   // Current CSS class applied to the label
   std::string currentStyle_;
-  // Format strings
-  std::string labelFormat_;
+  // Format string
   std::string tooltipFormat_;
   // DBus Proxy used to track the current active profile
   Glib::RefPtr<Gio::DBus::Proxy> powerProfilesProxy_;
-  sigc::connection powerProfileChangeSignal_;
 };
 
 }  // namespace waybar::modules
--- a/include/modules/power_profiles_daemon.hpp
+++ b/include/modules/power_profiles_daemon.hpp
@@ -32,6 +32,9 @@ class PowerProfilesDaemon : public ALabel {
   std::vector<Profile>::iterator activeProfile_;
   // Current CSS class applied to the label
   std::string currentStyle_;
+  // Format strings
+  std::string labelFormat_;
+  std::string tooltipFormat_;
   // DBus Proxy used to track the current active profile
   Glib::RefPtr<Gio::DBus::Proxy> powerProfilesProxy_;
   sigc::connection powerProfileChangeSignal_;
--- a/include/modules/power_profiles_daemon.hpp
+++ b/include/modules/power_profiles_daemon.hpp
@@ -1,4 +1,4 @@
-# pragma once
+#pragma once
 
 #include <fmt/format.h>
 
@@ -12,14 +12,16 @@ typedef struct {
   std::string driver;
 } Profile;
 
-class PowerProfilesDaemon : public ALabel  {
+class PowerProfilesDaemon : public ALabel {
  public:
   PowerProfilesDaemon(const std::string&, const Json::Value&);
   ~PowerProfilesDaemon();
   auto update() -> void override;
-  void profileChanged_cb( const Gio::DBus::Proxy::MapChangedProperties&, const std::vector<Glib::ustring>&);
+  void profileChanged_cb(const Gio::DBus::Proxy::MapChangedProperties&,
+                         const std::vector<Glib::ustring>&);
   void populateInitState();
   virtual bool handleToggle(GdkEventButton* const& e);
+
  private:
   // Look for a profile name in the list of available profiles and
   // switch activeProfile_ to it.
@@ -35,4 +37,4 @@ class PowerProfilesDaemon : public ALabel  {
   sigc::connection powerProfileChangeSignal_;
 };
 
-}
+}  // namespace waybar::modules
--- a/src/modules/power_profiles_daemon.cpp
+++ b/src/modules/power_profiles_daemon.cpp
@@ -17,6 +17,18 @@ namespace waybar::modules {
 
 PowerProfilesDaemon::PowerProfilesDaemon(const std::string& id, const Json::Value& config)
     : ALabel(config, "power-profiles-daemon", id, "{profile}", 0, false, true) {
+  if (config_["format"].isString()) {
+    format_ = config_["format"].asString();
+  } else {
+    format_ = "{icon}";
+  }
+
+  if (config_["tooltip-format"].isString()) {
+    tooltipFormat_ = config_["tooltip-format"].asString();
+  } else {
+    tooltipFormat_ = "Power profile: {profile}\nDriver: {driver}";
+  }
+
   // NOTE: the DBus adresses are under migration. They should be
   // changed to org.freedesktop.UPower.PowerProfiles at some point.
   //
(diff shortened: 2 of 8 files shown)


picnoir:
Fixes https://github.com/Alexays/Waybar/issues/1539

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,147 @@
...
+// Look for the profile str in our internal profiles list. Using a
+// vector to store the profiles ain't the smartest move
+// complexity-wise, but it makes toggling between the mode easy. This
+// vector is 3 elements max, we'll be fine :P
+void PowerProfilesDaemon::switchToProfile_(std::string str) {
+  auto pred = [str](Profile p) { return p.name == str; };
+  activeProfile_ = std::find_if(availableProfiles_.begin(), availableProfiles_.end(), pred);
+  if (activeProfile_ == availableProfiles_.end()) {
+    spdlog::error("PowerProfilesDaemon: FATAL, can't find the active profile in the available profiles list");
+    exit(-1);
```
Alexays, line 56: why an exit?

picnoir: Erf! Two reasons 😅

1. When I wrote this, I did not know the best way to crash the module to prevent it to do some dangerous dereferences. I made a mental note about that and put this exit in the meantime.
2. My memory is not so good, I tend to forget my mental notes very fast 😀

Anyway, after looking the current codebase, it seems like throwing a runtime error does the trick. Don't hesitate to point me to another direction if this is not the best way to go.

picnoir:
Documentation-wise, I could use some help.

I'm not sure what to do for now. Is the wiki the canonical source for it? How are you dealing with it when adding a new feature like this?

There's not much to document for now, the label ain't configurable: I personally don't really need that. I planned to implement that iteratively in a subsequent PR. Unless you'd prefer this to be done all at the same time in this PR.

Alexays:
When adding new feature like this, a new entry in the man and wiki with module settings like others is ok :)

picnoir:
Thanks for the directions. I'll write the manpage section by the end of the week. (and fix the clang-tidy issues)

I'm a bit puzzled by the freebsd CI failure. Is it expected?

lheckemann:
It would be nice to have the format string be configurable; the full profile names are a bit wide to fit neatly in my already crowded status bar, so I'd like to be able to use just :seedling: / :balance_scale: / :zap: for example.

alebastr:
> I'm a bit puzzled by the freebsd CI failure. Is it expected?

FreeBSD CI is flaky. It works via qemu vm in a macOS virtual machine, and there are random connectivity problems or timeouts. The actual status in the log is fine, so please ignore the failure.

> It would be nice to have the format string be configurable; the full profile names are a bit wide to fit neatly in my already crowded status bar, so I'd like to be able to use just 🌱 / ⚖️ / ⚡ for example.

I'd agree with that and even suggest to use the icons by default. For a module enabled in the default config, it takes too much screen space.
Although fontawesome icons would be a better choice due to an ability to set the color with CSS.

picnoir:
Thanks for the detailed feedback @alebastr, it was super useful!

From a sub-average C++ developper and Glib newbie perspective, this DBus API is pretty confusing and painful to use. :/

I adressed @lheckemann and @alebastr feedback wrt. the label formatting in https://github.com/Alexays/Waybar/pull/2971/commits/61fed6a21474752e382f17724091b91dee273e9b. We're now only displaying a small icon in the default config.

I addressed https://github.com/Alexays/Waybar/pull/2971#issuecomment-1972364896 in https://github.com/Alexays/Waybar/pull/2971/commits/901314a351b9ac1f8fb09a88cb635c4d17bc822e.

I'll write the man page this EU afternoon.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+  // Revisit this in 2026, systems should be updated by then.
+  Gio::DBus::Proxy::create_for_bus(Gio::DBus::BusType::BUS_TYPE_SYSTEM, "net.hadess.PowerProfiles",
+                                   "/net/hadess/PowerProfiles", "net.hadess.PowerProfiles",
+                                   sigc::mem_fun(*this, &PowerProfilesDaemon::busConnectedCb));
+}
+
+PowerProfilesDaemon::~PowerProfilesDaemon() {
+  if (powerProfileChangeSignal_.connected()) {
+    powerProfileChangeSignal_.disconnect();
+  }
```
alebastr, line 64: For any connections made with `sigc::mem_fun` where the object inherits `sigc::trackable` (base class for the modules does), this will happen automatically. You don't need to track or disconnect the connection.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+  if (powerProfilesProxy_) {
+    powerProfilesProxy_.reset();
```
alebastr, line 66: `RefPtr` already does that during destruction.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+  try {
+    powerProfilesProxy_ = Gio::DBus::Proxy::create_for_bus_finish(r);
+    using GetAllProfilesVar = Glib::Variant<std::tuple<Glib::ustring>>;
+    auto callArgs = GetAllProfilesVar::create(std::make_tuple("net.hadess.PowerProfiles"));
+
+    auto container = Glib::VariantBase::cast_dynamic<Glib::VariantContainerBase>(callArgs);
+    powerProfilesProxy_->call("org.freedesktop.DBus.Properties.GetAll",
+                              sigc::mem_fun(*this, &PowerProfilesDaemon::getAllPropsCb), container);
+    // Connect active profile callback
+  } catch (const std::exception& e) {
```
alebastr, line 80: `Glib::Error` does not inherit from `std::exception`, so you'll have to handle it in a separate `catch` branch for all DBus callbacks.

`util/format.hpp` has a formatter required for `Glib::Error::what()` result.

picnoir:
The manpage is ready.

PTAL

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+void PowerProfilesDaemon::busConnectedCb(Glib::RefPtr<Gio::AsyncResult>& r) {
+  try {
+    powerProfilesProxy_ = Gio::DBus::Proxy::create_for_bus_finish(r);
+    using GetAllProfilesVar = Glib::Variant<std::tuple<Glib::ustring>>;
+    auto callArgs = GetAllProfilesVar::create(std::make_tuple("net.hadess.PowerProfiles"));
+
+    auto container = Glib::VariantBase::cast_dynamic<Glib::VariantContainerBase>(callArgs);
```
alebastr, line 76: Is the manual conversion here necessary? You should be able to pass callArgs to call directly.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+  auto pred = [str](Profile const& p) { return p.name == str; };
+  this->activeProfile_ = std::find_if(availableProfiles_.begin(), availableProfiles_.end(), pred);
+  if (activeProfile_ == availableProfiles_.end()) {
+    throw std::runtime_error("FATAL, can't find the active profile in the available profiles list");
+  }
+}
+
+auto PowerProfilesDaemon::update() -> void {
+  if (connected_) {
+    auto profile = (*activeProfile_);
```
alebastr, line 166: An error in `switchToProfile` can potentially leave the iterator in a bad state. You'll need to check it here before dereferencing.

Consider also setting event_box_ visibility to the check result.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+      auto container = Glib::VariantBase::cast_dynamic<Glib::VariantContainerBase>(callArgs);
+      powerProfilesProxy_->call("org.freedesktop.DBus.Properties.Set",
+                                sigc::mem_fun(*this, &PowerProfilesDaemon::setPropCb), container);
+    }
+  }
+  return true;
+}
+
+void PowerProfilesDaemon::setPropCb(Glib::RefPtr<Gio::AsyncResult>& r) {
+  auto _ = powerProfilesProxy_->call_finish(r);
```
alebastr, line 210: Please, check for exceptions here.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+    label_.get_style_context()->add_class(profile.name);
+    currentStyle_ = profile.name;
+
+    ALabel::update();
+  }
+}
+
+bool PowerProfilesDaemon::handleToggle(GdkEventButton* const& e) {
+  if (connected_) {
+    if (e->type == GdkEventType::GDK_BUTTON_PRESS && powerProfilesProxy_) {
```
alebastr, line 190: The check for `powerProfilesProxy_` could be redundant if you already check `connected_`

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+// Look for the profile str in our internal profiles list. Using a
+// vector to store the profiles ain't the smartest move
+// complexity-wise, but it makes toggling between the mode easy. This
+// vector is 3 elements max, we'll be fine :P
+void PowerProfilesDaemon::switchToProfile(std::string const& str) {
+  auto pred = [str](Profile const& p) { return p.name == str; };
+  this->activeProfile_ = std::find_if(availableProfiles_.begin(), availableProfiles_.end(), pred);
+  if (activeProfile_ == availableProfiles_.end()) {
+    throw std::runtime_error("FATAL, can't find the active profile in the available profiles list");
```
alebastr, line 160: I'm not sure throwing exception is necessary here. It will be caught by Glibmm exception handler anyways, with no consequences.
You can just log the error and hide the module on the next update.

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+      if (activeProfile_ == availableProfiles_.end()) {
+        activeProfile_ = availableProfiles_.begin();
+      }
+
+      using VarStr = Glib::Variant<Glib::ustring>;
+      using SetPowerProfileVar = Glib::Variant<std::tuple<Glib::ustring, Glib::ustring, VarStr>>;
+      VarStr activeProfileVariant = VarStr::create(activeProfile_->name);
+      auto callArgs = SetPowerProfileVar::create(
+          std::make_tuple("net.hadess.PowerProfiles", "ActiveProfile", activeProfileVariant));
+      auto container = Glib::VariantBase::cast_dynamic<Glib::VariantContainerBase>(callArgs);
```
alebastr, line 201: The manual cast might be unnecessary

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,214 @@
...
+  Profile profile;
+  for (auto& variantDict : profilesVariant.get()) {
+    if (auto p = variantDict.find("Profile"); p != variantDict.end()) {
+      name = p->second.get();
+    }
+    if (auto d = variantDict.find("Driver"); d != variantDict.end()) {
+      driver = d->second.get();
+    }
+    profile = {name, driver};
+    availableProfiles_.push_back(profile);
```
alebastr, line 125: `availableProfiles_.emplace_back(name, driver);` would avoid one temporary object.

And if you move name/driver declaration into the loop (which you might want to do anyways to avoid accidentally reusing values from previous iterations), it'll be safe to write `if (!name.empty()) { availableProfiles_.emplace_back(std::move(name), std::move(driver)); }`

alebastr (commented):
Thanks for addressing the DBus safety comments! Looks correct and works for me both with and without daemon.

I left a few more comments, mostly style nits and redundant code.

src/modules/power_profiles_daemon.cpp
picnoir: Ah, yes, indeed, my bad. I was sure I tried and it error-ed out. It definitely does not :confused:

picnoir:
Thanks for having taken the time to write these detailed explanations. Learning a lot here :)

Addressed these "nits" in the above commit.

include/modules/power_profiles_daemon.hpp
```diff
@@ -0,0 +1,47 @@
...
+  std::vector<Profile> availableProfiles_;
+  // Points to the active profile in the profiles list
+  std::vector<Profile>::iterator activeProfile_;
+  // Current CSS class applied to the label
+  std::string currentStyle_;
+  // Format string
+  std::string tooltipFormat_;
+  // DBus Proxy used to track the current active profile
+  Glib::RefPtr<Gio::DBus::Proxy> powerProfilesProxy_;
+  sigc::connection powerProfileChangeSignal_;
```
alebastr, line 44: `powerProfileChangeSignal_` is not used and can be removed (here and in `getAllPropsCb`).

src/modules/power_profiles_daemon.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    // Set CSS class
+    if (!currentStyle_.empty()) {
+      label_.get_style_context()->remove_class(currentStyle_);
+    }
+    label_.get_style_context()->add_class(profile.name);
+    currentStyle_ = profile.name;
+  } else {
+    event_box_.set_visible(false);
```
alebastr, line 170: Just in case, please set visibility in both branches accordingly.

alebastr (approved):
Looks good. Thanks for implementing this feature!

picnoir:
Loosely related note: there's a quite a few synchronous dbus calls throughout the codebase:

```
~/code-root/github.com/Alexays/Waybar (pic/power-profiles-daemon) » ag call_sync
src/util/portal.cpp
50:    response = call_sync(std::string(PORTAL_INTERFACE) + ".Read", params);

src/util/backlight_backend.cpp
223:  login_proxy_->call_sync("SetBrightness", call_args);

src/modules/systemd_failed_units.cpp
77:      Glib::VariantContainerBase data = proxy->call_sync("Get", parameters);

src/modules/gamemode.cpp
130:      Glib::VariantContainerBase data = gamemode_proxy->call_sync("Get", parameters);

~/code-root/github.com/Alexays/Waybar (pic/power-profiles-daemon) » ag bus_sync
src/util/backlight_backend.cpp
105:  login_proxy_ = Gio::DBus::Proxy::create_for_bus_sync(

src/modules/systemd_failed_units.cpp
31:    system_proxy = Gio::DBus::Proxy::create_for_bus_sync(
40:    user_proxy = Gio::DBus::Proxy::create_for_bus_sync(

src/modules/bluetooth.cpp
22:  GDBusObjectManager* manager = g_dbus_object_manager_client_new_for_bus_sync(
28:    spdlog::error("g_dbus_object_manager_client_new_for_bus_sync() failed: {}", error->message);

src/modules/gamemode.cpp
90:  gamemode_proxy = Gio::DBus::Proxy::create_for_bus_sync(Gio::DBus::BusType::BUS_TYPE_SESSION,

```

picnoir:
I just push a small visual fix. The fontawesome icons were not really centered in the box in the default config. I feel like there's something wrong with the calculation of the icon width in the typesetter code. 

Added a bit of padding-left to center everything.

alebastr:
You also accidentally reverted the last set of the review fixes :)

Would it make sense to use `power-profiles-daemon` (kebab-case) in the configuration? It's a bit awkward that the CSS element name differs from the config name.

alebastr:
> Loosely related note: there's a quite a few synchronous dbus calls throughout the codebase:

Yep, I did review all of those.
 - Systemd interfaces aren't affected by DBus activation issues and will not block.
 - Portal interface would break GTK if the DBus service is not available (all those "Waybar takes 25 seconds to start" issues in the tracker). So we already require a correct configuration for this one.
 - `bluetooth` and `gamemode` are off by default.

But here's a snippet that simulates a problem with gamemode service activation, blocks startup of waybar for the duration of a DBus timeout and then causes crash on invalid memory access. Fun.
```sh
% mkdir -p ~/.local/share/dbus-1/services
% cat >~/.local/share/dbus-1/services/com.feralinteractive.GameMode.service <<EOF
[D-BUS Service]
Name=com.feralinteractive.GameMode
Exec=/usr/bin/sleep 30
EOF
```

picnoir:
> You also accidentally reverted the last set of the review fixes :)

:facepalm: fixed

> Would it make sense to use power-profiles-daemon (kebab-case) in the configuration? It's a bit awkward that the CSS element name differs from the config name.

Yes, I also prefer that. The daemon itself is spelled using kebab cas.

I originally used snake-case seeing it used for `idle_inhibitor`, thinking it was the local style. But we actually already have some kebab-case names for some other modules (keyboard-state, systemd-failed-units).

Alexays:
LGTM, can you also update the github wiki? :)
Thanks for this nice module.

picnoir:
:tada: 

I created the module wiki page there: https://github.com/Alexays/Waybar/wiki/Module:-PowerProfilesDaemon

leiserfg:
@picnoir  thanks a lot for this feature! Would it make sense to make it so scrolling up and down cycles up and down? Or right-click cycles down (to keep the symmetry with the current behavior)?

picnoir:
Hey!

There's only 3 profiles available ATM. For my personal use case,
toggling through them is good enough. 

If you feel like implementing this, you'll likely have to replace the
availableProfiles_ vector with something that can be iterated in both
directions (std::list?)

See: https://github.com/Alexays/Waybar/blob/2c927de4c624d0c560f73ad78b4abc85fb703562/include/modules/power_profiles_daemon.hpp#L35

Once that done, you'll need to retrieve which button has been pressed in
the handleToggle
(https://github.com/Alexays/Waybar/blob/2c927de4c624d0c560f73ad78b4abc85fb703562/src/modules/power_profiles_daemon.cpp#L177)
function. See the GdkEventButton documentation for more details
https://api.gtkd.org/gdk.c.types.GdkEventButton.html

Don't hesitate to ping me on Matrix/IRC or here if you're stuck and need
a helping hand!

leiserfg:
Changing the container is not required, std::vector has a bidirectional iterator as well, I will make a pr for it.

leiserfg:
https://github.com/Alexays/Waybar/pull/3032

Issue #2966, opened 2024-02-24
sway/workspaces displays __i3_scratch as a workspace if all-outputs is true


RobertMueller2:
This seems to be a consequence of bb843e0 / PR #2944.  Specifically, the commit replaced `get_workspaces` with `get_tree`. `get_tree` includes __i3_scratch as a workspace object on the output "__i3", while `get_workspaces` does not.

![satty-20240224-15:05:11](https://github.com/Alexays/Waybar/assets/29629287/23f8eb33-6442-42ce-9946-daceac6da816)

The upper waybar includes the commit, the bottom does not.

@grimpy 
Is this intended? I suppose it could be useful, even more so if I could specifically rename it or display a counter. Clicking it creates a new workspace "__i3_scratch" beneath the active output, not the invisible "__i3" output. That is a bit weird. While this is something that can be done through IPC without waybar altogether, and there's nothing we can do about it, I do feel if this were to be a feature in the module, it should get some special handling. The identification for this could be "workspace __i3_scratch on output __i3".

grimpy:
Unfortunately that was noty intended behaviour, i never use the scratch
workspace.

I'm unsure what would be the best behaviour here, my gut feeling is to
filter out this special workspace, is it normally visible when its in use
or not even then?

RobertMueller2:
Thanks for the reply. The scratchpad is just a place to store windows in, but it is never directly visible.

So far, sway/workspaces has not displayed it. That's not to say it couldn't be useful at all, but as mentioned above, not really without special handling. I think your intuition is correct here. We should filter it for now.

If somebody feels like including it in a more sophisticated fashion, that's always possible as a new feature down the road. I believe that might be better suited for a separate scratchpad module, though. (In fact I got that one on my list for trying out a Rust cffi waybar module, but who knows if I can pull that off within the next 3 or 4 years. :stuck_out_tongue: )

PR #2979, opened 2024-02-27
Filter out special output __i3 which contains scratchpad

grimpy:
Fixes: #2966

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -107,11 +107,16 @@ void Workspaces::onCmd(const struct Ipc::ipc_response &res) {
         auto payload = parser_.parse(res.payload);
         workspaces_.clear();
         std::vector<Json::Value> outputs;
+        bool alloutputs = config_["all-outputs"].asBool();
         std::copy_if(payload["nodes"].begin(), payload["nodes"].end(), std::back_inserter(outputs),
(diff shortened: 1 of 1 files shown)


grimpy:
I'm not 100% sure what's wrong in the clang-tidy workflow, did I tip over the scale of complexity?

src/modules/sway/workspaces.cpp
```diff
@@ -107,11 +107,16 @@ void Workspaces::onCmd(const struct Ipc::ipc_response &res) {
...
         std::copy_if(payload["nodes"].begin(), payload["nodes"].end(), std::back_inserter(outputs),
-                     [&](const auto &workspace) {
-                       return !config_["all-outputs"].asBool()
-                                  ? workspace["name"].asString() == bar_.output->name
-                                  : true;
+                     [&](const auto &output) {
+                       if (alloutputs && output["name"].asString() != "__i3") {
+                         return true;
+                       }
+                       if (output["name"].asString() == bar_.output->name) {
```
Alexays, line 116: Mhmm we should check !alloutputs here otherwise the logic will be the same as before.

grimpy: the only way for it to reach here when `!allouput` is  `true` is when `output["name"]` is `__i3` which will not match `bar_.output->name`

PR #2983, opened 2024-02-29
ci: move FreeBSD to ubuntu runners

alebastr:
With the recent runner hardware upgrade[1] and support in the cross-platform-actions[2] it became possible to use a Linux runner for this workflow.
Linux-based configuration appears to be faster and stabler than macOS, so it's now recommended for use.

[1]: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/
[2]: https://github.com/cross-platform-actions/action/releases/tag/v0.23.0

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -8,19 +8,22 @@ concurrency:
 
 jobs:
   clang:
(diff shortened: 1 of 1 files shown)


Alexays:
Ohh, that's nice!

PR #2984, opened 2024-03-01
mpd: less aggressive logging and reconnections

alebastr:
mpd logs are on the second place in the list of complaints I'm getting from Fedora users, right after the missing font icons :)

So this change attempts to push all the 'no server available, business as usual' messages to debug, and add some powersaving with second precision timers (yes, these are promised to be more power-efficient due to being grouped and fired all at once internally).


Alexays:
LGTM, thx :)

PR #2987, opened 2024-03-04
[hyprland/workspaces] New options to change on click behaviour and active workspace status

tomben13:
Option `move-to-monitor`:
Use `hyprctl dispatch focusworkspaceoncurrentmonitor` when clicking on workspace button

Option `active-per-monitor`:
Each bar on each monitor has its own active workspace. So instead of all the bars showing the same workspace which is currently active, each show the currently active workspace of the bar's monitor.

This may require additional testing. Also I hope this doesn't break backwards compatibility, especially with special workspaces. I didn't test, but it should be fine as I didn't change anything special workspace related.

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -126,6 +126,8 @@ class Workspaces : public AModule, public EventHandler {
   auto allOutputs() const -> bool { return m_allOutputs; }
   auto showSpecial() const -> bool { return m_showSpecial; }
   auto activeOnly() const -> bool { return m_activeOnly; }
+  auto moveToMonitor() const -> bool { return m_moveToMonitor; }
+  auto activePerMonitor() const -> bool { return m_activePerMonitor; }
 
   auto getBarOutput() const -> std::string { return m_bar.output->name; }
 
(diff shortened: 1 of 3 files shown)


tomben13:
@Alexays I think this can be merged as is. We could implement a similar option for special workspaces but I tested and `focusworkspaceoncurrentmonitor` doesn't work correctly for special workspaces currently in Hyprland.

Also, this PR doesn't break backwards compatibilty, it's an opt-in option.

Alexays:
Thanks, can you also update the Github wiki?

tom-quadrille:
> Thanks, can you also update the Github wiki?

Done

tomben13:
> Thanks, can you also update the Github wiki?

Done

simonwiles:
> I will remove option active-per-monitor from this PR as it is being handled by https://github.com/Alexays/Waybar/pull/2968 already. And https://github.com/Alexays/Waybar/pull/2968 is a way better solution.

I don't think #2968 covers the same use-case, fwiw.  I was hoping to be able to target only the currently visible workspace on each monitor; with `.hosting-monitor` from #2968 target *every* workspace on the given monitor, and `.active` is only applied on the focussed monitor (so I can't use, e.g., `.hosting-monitor.active`).

Issue #2998, opened 2024-03-10
sway/workspaces `persistent-workspaces` throws error and does not show up


italovieira:
I tried to set up persistent workspaces but following the examples in the [wiki](https://github.com/Alexays/Waybar/wiki/Module:-Sway#persistent-workspaces):

```
    "sway/workspaces": {
        "persistent-workspaces": {
            "1": [],
            "2": [],
            "3": []
        }
    },
```
But the result is no workspaces at all and a error:
```
 [error] Workspaces: in Json::Value::find(begin, end): requires objectValue or nullValue
```

nut-3:
I'm experiencing same bug.
To add, it is not present on 0.9.24 tag, can reproduce only on build from master tree.
Could trace the error till `void Workspaces::onCmd(const struct Ipc::ipc_response &res)` in `src/modules/sway/workspaces.cpp`

grimpy:
@nut-3 @italovieira I've created a merge request to fix this, could you verify this returns the expected behavior?

italovieira:
> @nut-3 @italovieira I've created a merge request to fix this, could you verify this returns the expected behavior?

With your PR persistent workspaces shows up, but now there's another problem, the active workspace is not highlighted:

![image](https://github.com/Alexays/Waybar/assets/8337230/2c8a9f15-871f-4ae7-b098-cfd169047d03)

For comparison, in the `v0.9.24` the highlighting was as expected:
![image](https://github.com/Alexays/Waybar/assets/8337230/47a91736-2bca-4e38-b2ce-4c6e97f89db5)

grimpy:
I updated the merge request that behaviour should be fixed aswell now

PR #3001, opened 2024-03-11
Fix peristant workspaces for sway

grimpy:
Fixes: #2998


grimpy:
I'm not 100% sure this restores the original behavior as I've never used this feature before.

grimpy:
> I'm not 100% sure this restores the original behavior as I've never used this feature before.

Seems the people in created the issues have confirmed the fix

Alexays:
Top, thanks!

PR #3010, opened 2024-03-15
README: tweak repology badge

luzpaz:
* Change repology badge header  
* Use 3 columns instead of 1 to display badge

Preview: https://github.com/luzpaz/Waybar/blob/repology-badge/README.md#installation


Alexays:
Thanks!

PR #3011, opened 2024-03-15
Use the correct thermal zone in FreeBSD

nicolarevelant:


nicolarevelant:
I just fixed the missing argument to fmt::format

nicolarevelant:
I just fixed the missing argument to fmt::format and the explicit conversion from std::string to const char*

Alexays:
Thanks!

PR #3023, opened 2024-03-17
Fixed segfault

tomben13:
Fixes segfault introduced by #2968 .

I commented about this segfault in the original PR but it wasn't addressed.


Alexays:
Thanks!

PR #3034, opened 2024-03-18
Add empty workspace style for Sway

wispl:
Allows styling empty workspaces in Sway. While it is possible to do this using persistent tags, I feel like it may be a little confusing as persistent workspaces don't have to be empty. Thanks for reviewing!


Alexays:
Can you update the man and github wiki? 🙏

wispl:
Done, sorry for the wait!

Issue #3009, opened 2024-03-15
#workspaces button.persistent not working in 0.10.0


kingprawn22:
I've updated to 0.10.0 since it includes an update that hides the tray when there are no icons, which works well, but some functionality of my style.css looks to have changed. I have 4 persistent workspaces, and I want to apply a different style to those that are 'active' (have open windows), and those that are not. I have also tried the default style.css (renaming 'persistent-workspaces' to 'persistent_workspaces' to make it work), and can see the same issue there also, comparing between waybar 0.10.0 and 0.9.17 (Debian).

```
/* remove inheritance from GTK */
button:hover {
	box-shadow: none; /* Remove predefined box-shadow */
	text-shadow: none; /* Remove predefined text-shadow */
	background: none; /* Remove predefined background color (white) */
	transition: none; /* Disable predefined animations */
}

* {
	border: 0;
	border-radius: 0;
	font-family: "JetBrains Mono";
	font-size: 13px;
	font-weight: bold;
}

/* main waybar colors */
window#waybar {
	background-color: rgba(0, 0, 0, 0.3);
	color: #dedede;
}

/* gap on left side */
.modules-left {
	margin-left: 15px;
}

/* gap on right side */
.modules-right {
	margin-right: 15px;
}

#workspaces button {
	min-width: 20px;
	padding-left: 5px;
	padding-right: 5px;
	/* applies to 'active' workspaces */
	color: #dedede;
}

/* applies to all workspaces set as persistent */                 
#workspaces button.persistent {
        color: #555;
}

/* currently selected workspace */
#workspaces button.focused {
        background: rgba (0, 0, 0, 0.4);
        border-bottom: 2px solid rgba(255, 255, 255, 0.2);
}

/* when hovering over workspace button */
#workspaces button:hover {
	background: rgba (0, 0, 0, 0.2);
}

/* use a dash for module separated by slash in config
#custom-search {
	margin-right: 20px;
	border-bottom: 2px solid rgba(255, 255, 255, 0.2);
}*/

#pulseaudio {
	margin-right: 20px;
	border-bottom: 2px solid rgba(255, 255, 255, 0.2);
}

#cpu {
	margin-right: 20px;
	border-bottom: 2px solid rgba(255, 255, 255, 0.2);
}

#memory {
	margin-right: 20px;
	border-bottom: 2px solid rgba(255, 255, 255, 0.2);
}
```

[...]

PR #3037, opened 2024-03-19
Fix peristent class on buttons

grimpy:
Fixes: #3009


Alexays:
Thanks for the fix!

GaetanLepage:
Thanks for the fix.
Have you planned to ship it in a minor release ?

PR #3038, opened 2024-03-19
Add a way to hide keyboard state module when unlocked

### LINKED ISSUE #1299: Hide an element with CSS
I would like to have the keyboard state hidden when the caps lock is off.
However, it seems impossible as of now. `display` and `visibility` are reported as invalid property names.

milouse:
Hi,

This MR adds new optional format options for the keyboard-state module when any of its element is locked. The main idea is to allow one to build invisible labels when unlocked, and only display something when locked (fix #1299). The initial problem being that GTK CSS does not offer any usable property in the matter (no support for `visibility`, `display` or even `width`).

The MR "works for me". However I’ve no strong opinion on how the setting should be done. Currently my implementation adds 3 new property in the `format` object, respectively `numlock-locked`, `capslock-locked`, `scrolllock-locked` to do something like:

```json
{
    …
    "keyboard-state": {
        "capslock": true,
        "format": {
            "capslock": "",
            "capslock-locked": " CAPS "
        }
    },
    …
}
```

But I’m not sure what would be best between that, or a whole new property `format-locked`, which might be just a string, as the `format` property:

```json
{
    …
    "keyboard-state": {
        "capslock": true,
        "format": "",
        "format-locked": " {name} "
    },
    …
}
```

or 

```json
{
    …
    "keyboard-state": {
        "capslock": true,
        "numlock": true,
        "format": "",
         "format-locked": {
            "capslock": " CAPS ",
            "numlock": " NUM "
        }
    },
    …
}
```

What do you think?

### DIFF
--- a/include/modules/keyboard_state.hpp
+++ b/include/modules/keyboard_state.hpp
@@ -32,12 +32,15 @@ class KeyboardState : public AModule {
   Gtk::Label scrolllock_label_;
 
   std::string numlock_format_;
+  std::string numlock_locked_format_;
   std::string capslock_format_;
+  std::string capslock_locked_format_;
   std::string scrolllock_format_;
-  const std::chrono::seconds interval_;
+  std::string scrolllock_locked_format_;
   std::string icon_locked_;
   std::string icon_unlocked_;
   std::string devices_path_;
+  const std::chrono::seconds interval_;
 
   struct libinput* libinput_;
   std::unordered_map<std::string, struct libinput_device*> libinput_devices_;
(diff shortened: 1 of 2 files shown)


milouse:
Once we agree on a setting format, I’ll also add a commit to update the documentation :)

joesri:
Hi.
You could get the same effect with two instances of the module and the "format-icons" option:
```
    "keyboard-state": {
        "capslock": true,
        "format": "{icon}",
        "format-icons": {
            "locked": " CAPS ",
            "unlocked": ""
        }
    },
    "keyboard-state#num": {
        "numlock": true,
        "format": "{icon}",
        "format-icons": {
            "locked": " NUM ",
            "unlocked": ""
        }
    }

``` 
Cheers!

milouse:
You are absolutely right. We are often too much focused on implementation details, instead of thinking twice about the issue. Thank you very much for the heads up!

PR #3039, opened 2024-03-19
Also consider floating nodes when checking for flags

### LINKED ISSUE #3030: Lost focused when window float only
Hi, (sorry for my english)
My button at workspace, lose .focused when i have only window floating

hrdl-github:
Fixes #3030


Alexays:
Thanks!

PR #2919, opened 2024-02-14
Update Wireplumber API to 0.5

tokyovigilante:
Wireplumber 0.5 has an incompatible API. This is a first WIP crack at updating. Would fix #2873 (but should not be merged currently).

Two current problems:
- Ignores the results of the two setup calls changed which are now async.
- Doesn't actually seem to work, adding extra logging shows that volume is not actually returned.

However it does build at least... I don't know the WP API well enough to add much more.

[EDIT]: As below now updated to use the new async API and working, ready for review.

PR #3043, opened 2024-03-19
Update meson.build

RoccoRakete:
This should fix the dependency issue with wireplumber


ptrcnull:
it actually doesn't - see #2919

PR #3064, opened 2024-03-23
Hyprland/Workspaces: Added option to hide non-visible special workspaces

alttabber:
If `special-visible-only` is set, special workspaces will be shown only if visible, they'll be hidden if they're not visible.

### DIFF
--- a/src/modules/hyprland/workspace.cpp
+++ b/src/modules/hyprland/workspace.cpp
@@ -182,6 +182,10 @@ void Workspace::update(const std::string &format, const std::string &icon) {
     m_button.hide();
     return;
   }
+  if (this->m_workspaceManager.specialVisibleOnly() && this->isSpecial() && !this->isVisible()) {
+    m_button.hide();
+    return;
+  }
   m_button.show();
 
   auto styleContext = m_button.get_style_context();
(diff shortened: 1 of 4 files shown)


alttabber:
Why did the entire thing have to change all the sudden!? I have to reimplement this from scratch.

alttabber:
And it also broke active-only.

alttabber:
And it broke special workspace classes too!?

alttabber:
This should be fixed.

khaneliman (changes requested):
Just run a quick clang-format on it and it LGTM

src/modules/hyprland/workspace.cpp
```diff
@@ -182,6 +182,11 @@ void Workspace::update(const std::string &format, const std::string &icon) {
     m_button.hide();
     return;
   }
+  if (this->m_workspaceManager.specialVisibleOnly() && \
+      this->isSpecial() && ! this->isVisible()){
```
khaneliman: Sorry, suck at these code suggestion blocks... was just pointing out which file clang-format was complaining about.

```suggestion
  if (this->m_workspaceManager.specialVisibleOnly() && this->isSpecial() && !this->isVisible()) {
```

khaneliman:
@Alexays this one should be good.

PR #2944, opened 2024-02-20
Implement windows formating in sway/workspaces

grimpy:
This implementation mimics to some extend the implementation of hyprland

PR #3066, opened 2024-03-23
fix(sway/workspaces): Visible class doesn't work on visible and empty workspaces

MauroGuida:
This fix is necessary to correctly apply the `#workspaces button.visible` style to all visible workspaces. This includes cases where workspaces are empty but still visible on the screen. This issue was introduced in #2944, which changes `ipc_.sendCmd(IPC_GET_WORKSPACES)` to `ipc_.sendCmd(IPC_GET_TREE)`. Therefore, the logic should be adjusted to handle the new IPC response correctly.


Alexays:
Thanks!

PR #3067, opened 2024-03-23
Fix clock on-scroll value not being used for calendar

encbar5:
The functionality of being able to invert the month scrolling in the calendar by setting "on-scroll" to -1 appears to  have been broken in 86a3898.

This restores the ability to set the number of months that the scroll wheel scrolls, including negative numbers, as documented in the wiki https://github.com/Alexays/Waybar/wiki/Module:-Clock

### DIFF
--- a/include/modules/clock.hpp
+++ b/include/modules/clock.hpp
@@ -41,6 +41,7 @@ class Clock final : public ALabel {
   const int cldMonColLen_{20};   // calendar month column length
   WS cldWPos_{WS::HIDDEN};       // calendar week side to print
   months cldCurrShift_{0};       // calendar months shift
+  int cldShift_{1};              // calendar months shift factor
   year_month_day cldYearShift_;  // calendar Year mode. Cached ymd
   std::string cldYearCached_;    // calendar Year mode. Cached calendar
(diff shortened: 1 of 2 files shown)


include/modules/clock.hpp
```diff
@@ -41,6 +41,7 @@ class Clock final : public ALabel {
   const int cldMonColLen_{20};   // calendar month column length
   WS cldWPos_{WS::HIDDEN};       // calendar week side to print
   months cldCurrShift_{0};       // calendar months shift
+  int cldShift_{0};              // calendar months shift factor
```
Alexays, line 44: Should not be set to 1 by default?

encbar5: If you look at the commit I listed, the code used to initialize this to 0, so I guess I just copied that.
The Wiki does say the default is 1, and that makes sense, so I can update that.

encbar5: Actually I don't think this matters much, because given the structure of the code right now, this functionality isn't enabled unless the user explicitly puts a value into their config file. So really this "default" is never used.

Alexays:
LGTM, thx

Issue #2685, opened 2023-11-22
Losing alternative format on Group module


ordy:
When I move a module like pulseaudio to a group, I can no longer left click on the icon to display the `format-alt` variant (right click still works as expected). However, I get it back if I set another copy anywhere on the waybar modules -left -center -right arrays.

With pusleaudio only part of the group:
![vid01](https://github.com/Alexays/Waybar/assets/22659443/ed1cee47-38c8-4c09-9cbf-3875cd6411e2)

And here clicks work when I have an extra instance of pusleaudio outside of the group module:
![vid02](https://github.com/Alexays/Waybar/assets/22659443/ea2f230d-c387-48d2-b7a9-019b4fc30575)

**Code:**

> pulseaudio

  ```json
      "pulseaudio": {
        "format": "{icon}",
        "format-alt": "{format_source}   {volume}%  {icon}", 
        "format-bluetooth": "<span font='11' rise='-800'>󰂰</span>", 
        "format-bluetooth-muted": "<span font='11' rise='-800'>󰂰</span>", 
        "format-muted": " {format_source}",
        "format-source": "{volume}%  <span font='11'>󰍰</span>",
        "format-source-muted": "<span font='11'>󰍱</span>",
        "format-icons": {
          "headphone": "<span font='11'>󰋋 </span>",
          "hands-free": "",
          "headset": "",
          "phone": "",
          "portable": "",
          "car": "",
          "default": [
            "<span font='12'>󰜟</span>"
          ]
        },
  ```

> group

  ```json
  "group/pulse-group": {
    "orientation": "inherit",
    "drawer": {
      "transition-duration": 300,
      "children-class": "drawer-child",
      "transition-left-to-right": false
    },
    "modules": [
      "pulseaudio",
      "pulseaudio/slider"
    ]
  },
  ```

Same results after testing with other modules that support `format-alt`.

TheRialNiels:
I was going to ask if the format-alt was removed but you're right, when u added to a group module stop working.

I hope it can be fixed

PR #3068, opened 2024-03-23
(Fix) Alt format on group modules

ordy:
fix for #2685 

During the strings conversion in `setupAltFormatKeyForModuleList` for the modules left/center/right, we need to check for group modules and iterate their array to add alt-format support to their modules as well.


Alexays:
Thanks!

PR #3074, opened 2024-03-24
[hyprland/workspaces] Make focusworkspaceoncurrentmonitor update focused workspace

Davido264:
> [!NOTE]  
> As a disclaimer, I don't know C++, so the code I implemented might not have good practices.

This pull request is solving an issue that I had in my multimonitor setup, in which I have the bar duplicated in my 2 monitors.

The Issue happens when i use `focusworkspaceoncurrentmonitor` from hyprland (or `move-to-monitor` option), It doesn't update the focused monitor.

For testing, I used a configured bar in ags that works the way I spected, along side with my waybar. 

[Demo before](https://youtu.be/QnX64P2GCFU?t=9).

In the video, you can see that when I switch workspaces (bringing it from another monitor), the bar configured with ags gets updated, while the bar configured with waybar don't.

[Demo after](https://youtu.be/N4G19Zo9CgM?t=7).

After the changes, both bars get updated

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -318,7 +318,7 @@ void Workspaces::onEvent(const std::string &ev) {
     onWorkspaceCreated(payload);
   } else if (eventName == "focusedmon") {
     onMonitorFocused(payload);
-  } else if (eventName == "moveworkspace" && !allOutputs()) {
+  } else if (eventName == "moveworkspace") {
     onWorkspaceMoved(payload);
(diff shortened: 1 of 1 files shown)


tomben13:
I had the same problem but without using `all-outputs`. My solution was a bit simpler.
Could you try if this works ?
[patch.txt](https://github.com/Alexays/Waybar/files/14956839/patch.txt)

Davido264:
sure

Alexays:
Closing in favor of https://github.com/Alexays/Waybar/pull/3143

PR #3076, opened 2024-03-24
Fix reloading config

cptpcrd:
Fully clear the configuration before reloading, so that when the config is read and merged in there are no existing values which mergeConfig refuses to overwrite.


Alexays:
Thanks!

PR #3088, opened 2024-03-29
hyprland/workspaces: Add `enable-bar-scroll` option

VAWVAW:
This PR adds the option to capture scroll events for hyprland workspaces on the whole bar like it works with sway.

When using this, I had a bug that if I enable bar scroll on both hyprland and sway and put `sway/workspaces` after `hyprland/workspaces` in `modules-left` the bar scroll would no longer work on sway. When putting `sway/workspaces` first, it works on hyprland and sway. I don't know enough to track this down though.


zjeffer:
> When using this, I had a bug that if I enable bar scroll on both hyprland and sway and put sway/workspaces after hyprland/workspaces in modules-left the bar scroll would no longer work on sway. When putting sway/workspaces first, it works on hyprland and sway. I don't know enough to track this down though.

But why would you use both at the same time? :smile:

metakunt:
Hey guys, can we please have this merged, I tested it and I need the functionality. I want to run the upstream version and not my locally built version, so that I get all updates.

include/modules/hyprland/workspaces.hpp
```diff
@@ -181,6 +182,9 @@ class Workspaces : public AModule, public EventHandler {
   void loadPersistentWorkspacesFromConfig(Json::Value const& clientsJson);
   void loadPersistentWorkspacesFromWorkspaceRules(const Json::Value& clientsJson);
 
+  bool handleScroll(GdkEventScroll*) override;
```
zjeffer, line 185:
```suggestion
  bool handleScroll(GdkEventScroll* e) override;
```

src/modules/hyprland/workspaces.cpp
```diff
@@ -1196,4 +1207,35 @@ void WindowCreationPayload::moveToWorksace(std::string &new_workspace_name) {
   m_workspaceName = new_workspace_name;
 }
 
+bool Workspaces::handleScroll(GdkEventScroll *e) {
+  if (gdk_event_get_pointer_emulated((GdkEvent *)e)) {
+    /**
+     * Ignore emulated scroll events on window
+     */
```
zjeffer, line 1214:
```suggestion
    // Ignore emulated scroll events on window
```
I don't see why this should be a multiline comment

src/modules/hyprland/workspaces.cpp
```diff
@@ -1196,4 +1207,35 @@ void WindowCreationPayload::moveToWorksace(std::string &new_workspace_name) {
...
+    return false;
+  }
+  auto dir = AModule::getScrollDir(e);
+  if (dir == SCROLL_DIR::NONE) {
+    return true;
+  }
+
+  bool increase;
```
zjeffer, line 1222: Unused variable?

zjeffer (approved):

metakunt:
MUCHAS GRACIAS @zjeffer

zjeffer:
I'd like to test this out but there are way too many conflicts since the workspaces.cpp file was split up into two files. Can you fix the conflicts?

metakunt:
Is there any issue why this is still not merged?

JohnMH:
I wish I'd found this before writing #3401. There's a small problem with this, in that events will get handled twice.

For example, scrolling on volume modules will both change the volume and workspace. This is probably not ideal.

Craeckie:
I rebased the changes (once more) and adapted them to the new way of loading options:
https://github.com/Craeckie/Waybar/tree/hyprland-bar-scroll

JohnRTitor:
@Alexays could you look at this please?

paperbenni:
Is there a reason this didn't get merged? I want this functionality

JohnRTitor:
CC @Alexays

schitthub685:
It works great, thank you.

Personally I prefer including empty workspaces with scrolling using `hyprctl dispatch workspace r+1`.
Would it be possible to make it so scrolling uses the values from the on-scroll-up and on-scroll-down options? This would allow for more flexibility.

Issue #1610, opened 2022-07-04
Change cursor on hover of clickable module?


ton:
Would it be possible to change the cursor shape when hovering a module? Specifying the CSS `cursor` property makes Waybar error out saying the `cursor` CSS property is not supported.

It would be nice to get some visual feedback that a module is clickable when hovering over it with the mouse.

slithery0:
I was testing with css to change cursor on hover, but the waybar do not showup after setting any css with cursor property.

Sophed:
Still waiting on this functionality.

Prayag2:
Any updates on this?

lilydjwg:
I don't like this idea. For native UI widgets, the clickability is often indicated by a slight background change (e.g. buttons, menu items, tabs), or not indicated at all (e.g. menubar, item list).

The hand cursor is an invention of the Web, and not used on native interfaces (except when a hyperlink is indicated). You don't get the hand cursor on your browser's native UI widgets.

Sophed:
This might be a crazy take but perhaps not everything web and native has to stay separate? It's not like the two haven't massively influenced each others design principles.

If it really bothers you, maybe it could be made a per-widget option like polybar.

lilydjwg:
Yes, it would be great for it to be configurable (either per widget or global). The main issue it brings to me is that the cursor change draws attention when my cursor wanders around. (That's perhaps why I first noticed it on my laptop but not desktop.)

SeanHub:
Been investigating the cursor option and come to this issue. I have noticed that the waybar hyprland workspaces module uses both the arrow and hand in odd ways. If you hover on the very edge of the workspace item you will see a hand cursor, but when you hover over the workspace item you will see an arrow cursor. If anything it should be the other way round (ideally configurable).

PR #3108, opened 2024-04-03
Cursor change to indicate module clickability

bananaoomarang:
Potential fix for: https://github.com/Alexays/Waybar/issues/1610

For clickable modules change the cursor type when entering/leaving listening on the `AModule::event_box_`.

This seems to work for me but I'm not sure if there are edge cases or a better way to achieve this with GTK so open to suggestions.


haug1:
Sorry for unnecessary comment, but this is kind of funny. You already added the hover events and fixed the propagation bug prior to my: https://github.com/Alexays/Waybar/pull/3165/files and now I fix a bug in your code: https://github.com/Alexays/Waybar/pull/3195.

PR #3127, opened 2024-04-09
Add waybar-styles(5) manual page

dkg:
waybar(5) describes the configuration syntax but doesn't mention how the stylesheets are handled.

This documentation would have been helpful to me as i figured out how to configure waybar.


Alexays:
Thanks!

PR #3128, opened 2024-04-09
Fix #3125

### LINKED ISSUE #3125: clock module ignores TZ environment variable
On glibc and musl the `TZ` environment variable overrides the system-wide timezone configured in `/etc/localtime`. The default behavior of the clock module when a timezone isn't set should probably try to match that.

hvenev:
Read `$TZ` because libstdc++ doesn't.


LukashonakV:
Hi @hvenev , can you please fix clang format? Thanks

Alexays:
Thanks!

PR #3130, opened 2024-04-10
battery: Add {cycles}, {health} format replacements

kiriDevs:
This PR adds the following format replacements to the `battery` module (currently for Linux only):

- [x] `{cycles}` - Displays the amount of charge cycles the selected<sup>1</sup> battery has seen.
- [x] `{health}` Displays a percentage representing the selected<sup>1</sup> batteries max charge, relative to it's `design` (original) max charge.

## Battery selection<sup>1</sup>

Currently, both of the new format replacements select the battery candidate with the highest `charge_full_design`. This seemed the most intuitive for `{cycles}`, and I have applied the same logic to `{health}` to stay consistent with this. Should multiple batteries have the same `charge_full_design`, the "worse" value (i.e. higher cycle count, lower health percentage) is displayed.

However, I could also imagine using a `charge_full_design` weighted average of all eligible batteries' percentages for `{health}`, so it could represent the health of a full battery array.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -32,7 +32,7 @@ class Battery : public ALabel {
   void refreshBatteries();
   void worker();
   const std::string getAdapterStatus(uint8_t capacity) const;
-  const std::tuple<uint8_t, float, std::string, float, uint16_t, float> getInfos();
+  std::tuple<uint8_t, float, std::string, float, uint16_t, float> getInfos();
   const std::string formatTimeRemaining(float hoursRemaining);
   void setBarClass(std::string&);
 
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -32,7 +32,7 @@ class Battery : public ALabel {
   void refreshBatteries();
   void worker();
   const std::string getAdapterStatus(uint8_t capacity) const;
-  const std::tuple<uint8_t, float, std::string, float> getInfos();
+  const std::tuple<uint8_t, float, std::string, float, uint16_t> getInfos();
   const std::string formatTimeRemaining(float hoursRemaining);
   void setBarClass(std::string&);
 
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -32,7 +32,7 @@ class Battery : public ALabel {
   void refreshBatteries();
   void worker();
   const std::string getAdapterStatus(uint8_t capacity) const;
-  const std::tuple<uint8_t, float, std::string, float, uint16_t> getInfos();
+  const std::tuple<uint8_t, float, std::string, float, uint16_t, float> getInfos();
   const std::string formatTimeRemaining(float hoursRemaining);
   void setBarClass(std::string&);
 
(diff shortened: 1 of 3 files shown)


kiriDevs:
> floating point literal has suffix 'f', which is not uppercase

These warnings are (as far as I can tell) all caused by pre-existing code (the one float this PR introduces uses an uppercase `F`). Should I change them anyways?

> function 'getInfos' has cognitive complexity of 224 (threshold 25)

This was far over threshold before, and I would prefer to not refactor (rewrite) that thing. Would it be acceptable to just add to it in this case?

> function 'update' has cognitive complexity of 26 (threshold 25)

That one I could probably manage, if required.

Alexays:
Don't worry about clang-tidy, we need to adjust it :)
Thanks!

kiriDevs:
Thank you for merging, and maintaining this project in general \:)

> Can you also update the Github wiki?

Done!

nebulosa2007:
I've updated waybar to 0.10.1, updated config, reloaded `waybar`:

```console
...
"format-alt": "{time} {health} {cycles} {icon}",
... 
```
but got this error:

`[2024-04-23 11:48:31.588] [error] battery: argument not found`

with just `{health}` or `{cycles}` either. 
What should I check on my Arch Linux?

kiriDevs:
> with just `{health}` or `{cycles}` either.

Did you try each one independently, or both at once? 

After looking at the code again, it does appear like I did, in fact, never add `{health}` for the regular formatting, only the hover-tooltip (oops). However, `{cycles}` should already be registered for the "main" label.

kiriDevs:
I've opened #3167 to address `{health}` not being registered, along with any other issues that could be related here.

nebulosa2007:
> Did you try each one independently, or both at once?

I tried them independently and both:
```json
..
"format-alt": "{time} {health} {icon}",
..
"format-alt": "{time} {cycles} {icon}",
..
"format-alt": "{time} {health} {cycles} {icon}",
..
```
>only the hover-tooltip

Ok then, I've updated config to:

```json
"tooltip-format": "{health} {cycles} {capacity}%",
```
There are no errors, but tooltip gives me only `0 0 95%`

Also, I compiled 1.10.2 + fix above. Now it works as `format-alt` and `tooltip-format` both, but shows `0 0` only.

I should I do next?

nebulosa2007:
I have this files, not `capacity` but `energy`:
```console
$ ls -la /sys/class/power_supply/*/*_full{,_design}
-r--r--r-- 1 root root 4096 Apr 24 15:10 /sys/class/power_supply/BAT1/energy_full
-r--r--r-- 1 root root 4096 Apr 24 15:10 /sys/class/power_supply/BAT1/energy_full_design

$ \ls -la /sys/class/power_supply/*/cycle_count
-r--r--r-- 1 root root 4096 Apr 24 15:10 /sys/class/power_supply/BAT1/cycle_count

$ cat /sys/class/power_supply/*/*_full{,_design}
26000000
47520000

$ cat /sys/class/power_supply/*/cycle_count
0
```

kiriDevs:
Thank you. I will implement checking for `energy_full` and `energy_full_design` in the new PR, so that `{health}` should work. I don't think there's anything we can do about `{cycles}` though, unfortunately :(

nebulosa2007:
Compiled again with fixes
image
Now show in `text-alt` and in `tooltip` too.

As for `{cycles}`- file exists and contain a number, I think it's a battery issue, not the `waybar`.

Thanks a lot!

PR #3136, opened 2024-04-13
Pipewire backend: use pipewire thread lock

### LINKED ISSUE #3047: segfault when powering other monitor off
*** impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted

I'm getting this error most of the time when I turn my screen off, and waybar crashes

https://gitlab.com/that1communist/dotfiles/-/blob/master/.config/waybar/config

these are my dotfiles if it helps

hrdl-github:
Fixes #3047.

I haven't encountered any issues since adopting these locks. Calling https://github.com/hrdl-github/Waybar/blob/74773885c6166d190a294ad945b1f1d4d5f007ec/src/util/pipewire/pipewire_backend.cpp#L81 without locking causes the warning described in https://github.com/Alexays/Waybar/issues/3047#issue-2196726500. I am less certain whether the lock in the constructor is required.

### DIFF
--- a/src/util/pipewire/pipewire_backend.cpp
+++ b/src/util/pipewire/pipewire_backend.cpp
@@ -48,12 +48,17 @@ PipewireBackend::PipewireBackend(PrivateConstructorTag tag)
   if (mainloop_ == nullptr) {
     throw std::runtime_error("pw_thread_loop_new() failed.");
   }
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

Issue #2989, opened 2024-03-06
Feature Request: Hover for all Modules


5p4r74cu5:
Hover is only supported by workspaces modules and there has been alot of interest it being available to all modules, as mentioned in #282, #1838, #2682, #2691. This feature is unavailable with the revert of #1120 discussed in #1792, and it's been suggested in #2491 and #2815 that it will only be possible with a migration to GTK4, which I can't speak to as I am not familiar with the technical aspects.

samisul:
I would love to have this

EaGLexEyE:
Same

RafayAhmad7548:
yes hover effects yay yes please

Kaaswastablet:
Hype

roycrippen4:
+1 just spent like 40 minutes trying to debug my hover effect not working before getting here lol

MrAventador:
Such a bummer this is not a thing yet

JulesDmn:
> Such a bummer this is not a thing yet

this

haug1:
Yes, please!

Kommynct:
Please stop spamming this, if you aren't contributing to helping the issue it's really annoying to notify everybody with things like "this" and "yes, please" nobody cares, nobody is going to start working on this feature just because you post in this thread, +1's and everything are situations where you should use the reaction button. Thank you.

haug1:
I feel confident in my latest PR, we'll see what happens.

Alexays:
Just merged https://github.com/Alexays/Waybar/pull/3145

Issue #1792, opened 2022-11-15
⚠️ 0.9.15 regressions


Alexays:
Hi folks,
I'm creating a new issue to have all the regressions related to 0.9.15 in one place and be able to fix them more easily.
I also thought of revert https://github.com/Alexays/Waybar/pull/1120
What do you think?

https://github.com/Alexays/Waybar/issues/1781
https://github.com/Alexays/Waybar/issues/1786
https://github.com/Alexays/Waybar/issues/1763
https://github.com/Alexays/Waybar/issues/1775
https://github.com/Alexays/Waybar/issues/1778
https://github.com/Alexays/Waybar/issues/1743
https://github.com/Alexays/Waybar/issues/1741
https://github.com/Alexays/Waybar/issues/952
https://github.com/Alexays/Waybar/issues/1761
https://github.com/Alexays/Waybar/issues/1752
https://github.com/Alexays/Waybar/issues/1791
https://github.com/Alexays/Waybar/issues/1765

Estebiu:
I would also like to add that when I upgraded waybar to 0.9.15 the battery indicator is "strange". It shows me the exact half of what my battery capacity is. So if htop says I'm at 100%, waybar will say that I'm at 50%. If I'm at 34%, waybar will say that I'm at 17% and so on.

grfrederic:
@Estebiu I made changes to the battery calculation recently, could you give more info? What is your exact version? Do you have more than one battery (say, in `/sys/class/power_supply/`)?

PowerMan2206:
I think you should revert #1120.

Estebiu:
> @Estebiu I made changes to the battery calculation recently, could you give more info? What is your exact version? Do you have more than one battery (say, in `/sys/class/power_supply/`)?

I'm using the latest waybar-hyprland-git from the AUR as I'm on Hyprland. And no, I don't have another battery; my laptop doesn't support it (asus zenbook ux363ea). Just now I tried the standard waybar package and the battery monitor works normally. Very strange.

grfrederic:
@Estebiu Right, that does sound like #1788 might have broken it for you. The tricky part is I only have a laptop with TWO batteries. I can remove one of them once I'm home and see if I can replicate the issue.

roland-rollo:
[...]

PR #3139, opened 2024-04-14
feat(#2989)!: hover for all modules

haug1:
I recognize that this change may break or change many existing `style.css` configurations due to the new buttons and the default configuration containing a general css rule for buttons.
I've just done the minimal job with this approach of implementing the feature. If this approach is acceptable even with breaking changes to existing configurations, I can continue finishing this up with some guiding feedback.

If this is not an acceptable approach, I've been thinking about a different approach of just adding a ".hover" class to the GTK label instead. I haven't really explored that possibility yet and I'm very new to GTK and C++, so I'm not sure about the details in that approach, but it sounds very doable and probably not hard to figure out how to implement. This approach would likely not breaking any existing configurations.

### DIFF
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -6,6 +6,7 @@
 #include <json/json.h>
 
 #include "IModule.hpp"
+#include "gtkmm/button.h"
 
 namespace waybar {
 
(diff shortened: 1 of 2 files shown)


haug1:
https://github.com/Alexays/Waybar/issues/1792

haug1:
@Alexays 
Is it interesting? Could you please take some time to give feedback on this? How can we got some feature like this merged?
I'm determined to add this hover feature, but maybe this is not the right way because backwards-compatibility. Do you have any thoughts about the alternate solution hooking up GTK mouse enter/leave events and adding/removing the `.hover` class?

PR #3142, opened 2024-04-15
calendar: add shift_reset action

janpeterd:
Added `shift_reset` action to the calendar in the clock module. This is handy, when you go back a couple of months and want to jump back to current date.


Alexays:
Thanks!

PR #3143, opened 2024-04-16
Fixes for hyprland/workspaces

### LINKED ISSUE #3121: [Bug] hyprland/workspaces: Active workspace not updated when a screen is disconnected
# Steps to reproduce

1. Log into Hyprland
2. Attach 2 screens to a machine and make sure both are showing the desktop.
3. Start waybar with this config:

config.json
```json
[{
  "modules-left": [ "hyprland/workspaces" ],
  "hyprland/workspaces": { "move-to-monitor": true }
}]
```

style.css
```css
#workspaces button {
  padding: 0 0.5em;
  margin: 0 0.25em;
  border-radius: 3px;

  border: none;
}

#workspaces button.active {
  border: 3px solid red;
}
```
4. Focus the workspace on one of the monitors
5. Disable some other monitor (bind lid switch works too)
```shell
hyprctl keyword monitor 'eDP-1,preferred,0x0,1'
```

# Expected result

After the monitor is disabled the bar shows the selected workspace with a red outline.

# Actual result

No workspace is highlighted as active until a workspace is switched.

tomben13:
- Update active workspace on worskpace moved. Fixes #3121 
  - It should also cover #3074 . This PR solution is simpler than #3074 

- Don't update window count and sort workspace during init but AFTER their creation. It did nothing during init because the objects did not exist yet. It fixes a problem I had when configuring waybar where I launched it with already opened windows. This bug was not usually encountered because waybar usually starts with hyprland when there are no windows yet.

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -194,6 +194,10 @@ void Workspaces::doUpdate() {
   for (auto &[workspaceData, clientsData] : m_workspacesToCreate) {
     createWorkspace(workspaceData, clientsData);
   }
+  if (!m_workspacesToCreate.empty()) {
+    updateWindowCount();
+    sortWorkspaces();
+  }
   m_workspacesToCreate.clear();
 
   // get all active workspaces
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #3145, opened 2024-04-17
feat(#2989): (optional) hover for all modules

haug1:
A second attempt at implementing hover for all modules in a much less intrusive way.

See https://github.com/Alexays/Waybar/pull/3139 also (the more naive implementation which I found a similar approach had been tried and reverted).

Let me know if any additional documentation is required to get this merged. Alternatively, I think this could just be merged directly exposing the functionality for the super users immediately and then improved upon later on.

### DIFF
--- a/resources/style.css
+++ b/resources/style.css
@@ -48,6 +48,11 @@ button:hover {
     box-shadow: inset 0 -3px #ffffff;
 }
 
+/* you can set a style on hover for any module like this */
+#pulseaudio:hover {
+    background-color: #a37800;
+}
+
 #workspaces button {
     padding: 0 5px;
     background-color: transparent;
--- a/src/AModule.cpp
+++ b/src/AModule.cpp
@@ -27,6 +27,9 @@ AModule::AModule(const Json::Value& config, const std::string& name, const std::
       spdlog::warn("Wrong actions section configuration. See config by index: {}", it.index());
   }
 
+  event_box_.signal_enter_notify_event().connect(sigc::mem_fun(*this, &AModule::handleMouseEnter));
+  event_box_.signal_leave_notify_event().connect(sigc::mem_fun(*this, &AModule::handleMouseLeave));
+
   // configure events' user commands
   // hasUserEvent is true if any element from eventMap_ is satisfying the condition in the lambda
   bool hasUserEvent =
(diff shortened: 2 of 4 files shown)


haug1:
(Force-push removed debug log statements)

src/AModule.cpp
```diff
@@ -3,6 +3,7 @@
 #include <fmt/format.h>
 
 #include <util/command.hpp>
+#include "sigc++/functors/mem_fun.h"
```
haug1, line 6: That's weird, my editor did this automatically, but `mem_fun` was already called in this file and the include was not needed. Can revert if needed? I'm not that familiar with C++

resources/style.css
```diff
@@ -48,6 +48,11 @@ button:hover {
     box-shadow: inset 0 -3px #ffffff;
 }
 
+/* combined with the configuration option hover=true, you can set a style for any module like this */
+widget.hover #pulseaudio {
+    background-color: #a37800;
```
haug1, line 53: I specifically chose this color as it is slightly darker than the already set default color for the pulseaudio module.

src/AModule.cpp
Alexays: Yep remove this one :)

Alexays:
Nice approach, it might be the perfect workaround to have hover effect.
I would even remove the "hover" config and enable it for everyone, as it's just a class.

Alexays:
Can you also add it to the man?

haug1:
> Can you also add it to the man?

Thanks, I'll handle your feedback tonight when I have time!

haug1:
Made the following changes:
- removed unnecessary include
- removed configuration option, hover signals are attached by default
- added documentation in waybar-styles man-page

alebastr:
Might be useful: after a short dive into the GTK internals, I figured out how to make the `:hover` pseudo-class work:
```
diff --git a/src/AModule.cpp b/src/AModule.cpp
index f941ab97..a285da77 100644
--- a/src/AModule.cpp
+++ b/src/AModule.cpp
@@ -87,12 +87,16 @@ auto AModule::doAction(const std::string& name) -> void {
 }
 
 bool AModule::handleMouseEnter(GdkEventCrossing* const& e) {
-  event_box_.get_style_context()->add_class("hover");
+  if (auto* module = event_box_.get_child(); module != nullptr) {
+    module->set_state_flags(Gtk::StateFlags::STATE_FLAG_PRELIGHT);
+  }
   return true;
 }
 
 bool AModule::handleMouseLeave(GdkEventCrossing* const& e) {
-  event_box_.get_style_context()->remove_class("hover");
+  if (auto* module = event_box_.get_child(); module != nullptr) {
+    module->unset_state_flags(Gtk::StateFlags::STATE_FLAG_PRELIGHT);
+  }
   return true;
 }
 
```
Important part of the trick is that GtkEventBox does not propagate the state flags to the child widgets, so you have to apply the `PRELIGHT` (`:hover`) state flag directly to the module widget.

haug1:
I could attempt @ale

haug1:
I could attempt @alebastr's solution if that is preferrable. It looks good to me, I don't know.

haug1:
I jumped the gun and applied those changes. I think it's much more elegantly done like this. We get the normal CSS `:hover` and applied on the correct GTK component. Kudos to @alebastr for digging into GTK a lot better than me! :+1:

Alexays:
🔥

PR #3148, opened 2024-04-18
sway/workspaces: Fix scroll on unfocused monitor

### LINKED ISSUE #1496: sway-workspaces: can't scroll on inactive monitor when "all-outputs" = false
If you have multiple monitors and only have the workspaces shown on their respective waybars, then you can't use the scroll wheel to change workspace on the inactive monitor(s). 

Setting `all-outputs` to true allows scrolling on all monitors as expected. 

sway version 1.8-dev-ca016689 (Mar 22 2022, branch 'master')
Waybar v0.9.12

VAWVAW:
Fixes #1496.

Scrolling works in `sway/workspaces` only if the focused monitor is shown in the bar. This PR uses the visible monitor if the focused monitor is not in the bar.

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -434,9 +434,16 @@ bool Workspaces::handleScroll(GdkEventScroll *e) {
   }
   std::string name;
   {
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #3151, opened 2024-04-19
feat: add new module 'images' to render multiple images under one gtk container

### LINKED ISSUE #944: [Feature Request] Add support for custom Icons in custom modules

### LINKED ISSUE #2734: Make the images module behave like custom module
I made a custom script that prints the album art paths as soon as they change. The program runs continously in the background. Unfortunately with this approach waybar refuses to start. It works fine if I run the program like a custom module. 

(Example project: https://github.com/xrtxn/waybar_music_art)
Config:     
```
"image#album_art":{
        "exec":"$HOME/.config/waybar/modules/rs_album_art",
	"size": 32,
    },
```

VPavliashvili:
This module is inspired by already existing [image](https://github.com/Alexays/Waybar/wiki/Module:-Image) module. The difference is this module will give user capability to render multiple images with support of custom styles to use per image and each image will be able to run distinct `on-click` custom commands.

Resolves #944, resolves #2734


Alexays:
Can you rather add a new option to the existing image module to render multiple images?

VPavliashvili:
Hello
Unfortunately I was not able to notice your comment until now.
I can update this PR as you request when I will have some free time, hopefully in a near future

VPavliashvili:
Hello @Alexays 

Finally I was able to make a new changes as you requested in your last comment. The changes was not possible without introduction of breaking changes in configuration, for this reason I have separated old(single) and new(multiple) image rendering functionality in two strategies and implemented them separately.

I have not changed anything regarding the old implementation to avoid any kind of breaking changes, therefore configuration for a single image stays the same.

Regarding whats new: I have written new implementation for rendering multiple images, if `"multiple"=true` is passed in configuration than this changes will render one or more images from new structure of config json(i.e. it will look for `entries` array(or exec script which returns array of image file paths).
Also man page has been edited accordingly.

Let me know if any correction is necessary.

vredesbyyrd:
Thanks @VPavliashvili for this pr. Is there any reason this is not being merged? I am coming from awesomwm as a newb to wayland and wabar. I'd like to write a album cover script that runs continuously and updates on it's own, no need for polling. From what I gather that will not work with the current image module.

VPavliashvili:
@vredesbyyrd You are right, that won't work with the current implementation since it only supports only one image rendered at a time. With this PR you should be able to do that, I wanted this feature for my use case too but its been months since latest update from repo owner so idk, when he responds this pr might get merged. 

It's been months since my latest commit for this PR and I haven't opened waybar codebase since, now it shows there are merge conflicts so I will have to change some stuff as it seems, hope @Alexays  will respond in a near future and things will move forward

vredesbyyrd:
@VPavliashvili 

I much appreciate the quick response and info. I to hope @Alexays would be open to merging this. Having an image module that is more inline with the custom module would be very useful. 

> ...now it shows there are merge conflicts

I saw that as well. I was going to attempt to build waybar from source with your patch, but with conflicts it's a bit over my head. I hope it is not too much of a pain to resolve. 

Thanks for your time!

PR #3165, opened 2024-04-23
fix(#3162): hover event did not propagate causing issues

### LINKED ISSUE #3162: Drawer not working since #3139
Since the :hover capability added in #3139, drawers no longer appear to work. The lead module appears but the drawer does not open when hovered over.

haug1:
Fixes #3162.


Alexays:
Arg, thx :)

PR #3167, opened 2024-04-24
fix(battery): Fix `{health}` format replacement

kiriDevs:
Looks like I forgot to register the `{health}` replacement for the "default" (i.e. non-tooltip) replacements.

Also, some batteries report their `{health}` related data as `energy_full` and `energy_full_design` instead of `charge_full` and `charge_full_design`. This PR ~~will~~ also implements support for that.

For more information see the conversation below #3130 

---

~~Marking as draft since~~
- ~~I haven't tested this yet (quick commit from the web editor)~~
- ~~there might actually be another / more issue(s) with `{cycles}` and `{health}` in `format-alt`~~

Issues reported by @nebulosa2007 under the original PR for `{cycles}` and `{health}` (#3130)

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -712,10 +712,10 @@ auto waybar::modules::Battery::update() -> void {
     event_box_.show();
     auto icons = std::vector<std::string>{status + "-" + state, status, state};
     label_.set_markup(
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #3180, opened 2024-04-28
Update lookup for Hyprland socket location based on changes from commit a5a6480 of hyprland

JacktheDevel0per:
I tested this locally with Hyprland commit a5a6480.
I am unsure of how you manage breaking changes like this, as this is still only in git and not released.

(Sorry if I used incorrect casing/capitalization)

### DIFF
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -20,11 +20,18 @@ void IPC::startIPC() {
   std::thread([&]() {
     // check for hyprland
     const char* his = getenv("HYPRLAND_INSTANCE_SIGNATURE");
+    const char* xdgRuntimeDir = getenv("XDG_RUNTIME_DIR");
+
+    if (xdgRuntimeDir == nullptr) {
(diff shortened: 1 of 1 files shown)


sungyoonc:
https://github.com/hyprwm/Hyprland/pull/5788#issue-2267815770
> For those users, I'd recommend checking whether `$XDG_RUNTIME_DIR/hypr` exists. If yes, use that, if not, use the old `/tmp/hypr`

I think fallback to /tmp/hypr should be added as there might be cases where you're running the old version of hyprland with the new version of waybar

niksingh710:
yep i just updated today to `-git` branch and ig a fallback should be added aldo.

JacktheDevel0per:
Yeah, that is a good way to do it. I might be able to add that tonight.

zjeffer:
I didn't see this PR existed until I made [mine](https://github.com/Alexays/Waybar/pull/3183), but I implemented the fallback to `/tmp/hypr/` already, so you don't have to do it @JacktheDevel0per :)

Alexays:
Closing in favor of https://github.com/Alexays/Waybar/pull/3183

PR #3188, opened 2024-04-30
refactor(group): moving group stuff out of bar

haug1:


haug1:
I think this is better, I don't know what you think. I'd love some feedback. Just doing this as a learning exercise

haug1:
Hmm, yeah, I guess me and clang tidy fundamentally disagree about where the group code belongs. I'm not motivated to see this change through.

PR #3195, opened 2024-05-02
fix: set cursor appropriately on user event hover

haug1:
looks like a small oopsie in recent merge


x140x1n:
Is there any way to turn this off in the config?

haug1:
> Is there any way to turn this off in the config?

The setting of cursor is currently bound with whether the module has any event handlers and no other configuration options. So if you have on-click handler or anything like that in your module configuration, then no, at least not currently it looks like.

PR #3198, opened 2024-05-03
fix: custom module mediaplayer doesn't respect argument

stanly0726:
fix custom module mediaplayer which doesn't consider  --exclude argument on player appear

previously player excluded by --exclude or -x argument started after `mediaplayer.py` wont be ignored
this commit fixed it


Alexays:
Thanks!

Issue #3174, opened 2024-04-26
on-click and hover doesn't work on the group itself


itcreeperboy:
i have a group of modules and i wanted clicking any part of the group to send me to missioncenter

```
"group/hardware": {
        "on-click": "missioncenter",
        "orientation": "horizontal",
        "modules": [
            "disk",
            "memory",
            "cpu",
            "custom/gpu"
        ]
    }
```
i tried just doing this but it didn't trigger when i clicked.

not a big deal because i can just do on-click on the individual modules

but then i tried to add a border to the group on hover:

```
#hardware:hover {
  box-shadow: inset 0 0 0 4px #cba6f7;
}
``` 

and it doesn't trigger when i hover over it, unlike any normal module or the seperate modules in the group when i do this:

```
#hardware *:hover {
  box-shadow: inset 0 0 0 4px #cba6f7;
} 
```
and i havn't found a way to trigger adding a border to the whole group from hovering on one of it's modules

idk if this is intended or not, if it is, is there some other way to get the same result? sorry if i'm being stupid i'm css noob lol

here is [config](https://pastebin.com/Kt76kkkw) and [style.css](https://pastebin.com/NfL9aq88)

ordy:
Your selector for that group should be `#hardware-group` not just `#hardware`

However, the group should be a GtkBox and I don't know if they support box-shadow, while simple borders should work
```
#hardware-group: { border: 4px solid #cba6f7 } 
```
That would obviously add an extra 4px padding to your group (probably what you are trying to avoid using the box-shadow).

Now for the hover part, I couldn't make it work on the group selector, but note that `hover` for custom modules has been implemented just a week ago in #3145 so there is a chance that the group module has been left out.

`#hardware-group label:hover` works but that's just on the label part, not over the modules.

if you set `#hardware-group box` you have the border on the modules, but again, no hover support straight on `box:hover`.

If you need help targeting your css classes, run warbar with the GTK debug too:

`GTK_DEBUG=interactive waybar` then select CSS nodes in the menu that will give your id's and extra classes to work with.

haug1:
I don't see a clear and obvious reason why hover should not be working with groups.
After a quick glance a the code, I have a theory it may be a [similar problem to the one I patched](https://github.com/Alexays/Waybar/pull/3165/files): https://github.com/Alexays/Waybar/blob/master/src/group.cpp#L98.

haug1:
[...]

PR #3201, opened 2024-05-03
feat(#3174): hover for whole group

haug1:
Adds hover targeting for the whole group.

image

Not a great gif to demonstrate the fix, but without the patch, there's a transparent background instead of black.

```
#power {
  background-color: transparent;
  color: white;
}

#power *:hover {
  background-color: white;
  color: black;
}

#power:hover {
  background-color: black;
  color: white;
}
```

```
  "group/power": {
    "orientation": "inherit",
    "drawer": {
      "transition-duration": 500,
      "children-class": "not-power",
      "transition-left-to-right": false,
    },
    "modules": [
      "custom/lock",
      "custom/quit",
      "custom/suspend",
      "custom/reboot",
    ],
  },
```

```
      widget:dir(ltr)
        box#power.horizontal:dir(ltr)
          revealer.drawer:dir(ltr)
            box.horizontal:dir(ltr)
              widget.not-power:dir(ltr)
                label#custom-quit.module:dir(ltr)
              widget.not-power:dir(ltr)
                label#custom-suspend.module:dir(ltr)
              widget.not-power:dir(ltr)
                label#custom-reboot.module:dir(ltr)
          widget:dir(ltr)
            label#custom-lock.module:dir(ltr)

```

### DIFF
--- a/include/group.hpp
+++ b/include/group.hpp
@@ -19,7 +19,7 @@ class Group : public AModule {
   virtual Gtk::Box& getBox();
   void addWidget(Gtk::Widget& widget);
 
-  bool handleMouseHover(GdkEventCrossing* const& e);
+  bool handleModuleMouseHover(GdkEventCrossing* const& e);
 
  protected:
   Gtk::Box box;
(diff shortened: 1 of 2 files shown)


haug1:
Good to go, I think.

haug1:
Well, in hindsight of course that's not going to help as a hacky workaround because security reasons.. Last commit should obviously be reverted, but I don't dare to touch anything more for now in case I trigger more CI that gets stuck infinitely..

haug1:
Okay, I tried one more thing in hopes of stopping the workflow by making it the linting task, but seems it never gets that far before it gets stuck. Sorry..!

PR #3196, opened 2024-05-02
Nix Flake: Fix overlay (again)

spikespaz:
Because last time I "fixed" it, I actually broke it, I forgot a line, but this time I changed other minor stuff (added `waybar`-named overlay and package, in addition to `default`, and the package output now is properly inherited from the overlay applied on top of Nixpkgs).

PR #3208, opened 2024-05-04
flake: fix overlay not actually being applied

spitulax:
#3196 made some changes to `flake.nix`. In [this line](https://github.com/Alexays/Waybar/blob/231d6972d7a023e9358ab7deda509baac49006cb/flake.nix#L67) flake's `waybar` package was defined to be the one in `nixpkgs` because the function that was passed to `genSystems` argument was called with `nixpkgs` without the flake's overlay applied. I changed the `genSystems` function to pass `nixpkgs` that uses the flake's overlay so the package comes from the flake itself and not `nixpkgs`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,12 @@
         "x86_64-linux"
         "aarch64-linux"
       ]
-        (system: func (import nixpkgs { inherit system; }));
+        (system: func (import nixpkgs {
+          inherit system;
+          overlays = with self.overlays; [
+            waybar
+          ];
+        }));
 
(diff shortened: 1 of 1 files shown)


spikespaz:
Thanks, apologies to @Arjun31415 and others.

This PR is crucial to merge (I was [almost about to PR](https://github.com/spikespaz/waybar/commit/28a207e5f51d5a78828571816bee4d503b534580)), because I made a mistake (again!) and [this commit](https://github.com/Alexays/Waybar/pull/3196/commits/398cd8cbd42eae063d40c8051cc5d48ed6d98f9d) is borked.

Issue #3210, opened 2024-05-05
Custom modules refusing to set tooltip format.


Spelis:
i have this custom module that gets my current keyboard layout and i am trying to set the tooltip format but it isnt working.

```
	"custom/kb": {
		"format": "  {}",
		"exec": "bash ~/.config/waybar/getkb.sh",
		"interval":1,
		"tooltip-format": "Current Keyboard Layout: {}\nSwitch Keyboard Layout (󰘳 + X)"
	}
```

when i run waybar, the tooltip is the value of `format` and not `tooltip-format`.
this issue occurs with a different custom module as well.

EDIT:
tooltips do work with the `clock` module, for some reason?

PR #3213, opened 2024-05-05
fix(#3210): tooltip-format on custom modules not working in some cases

haug1:


haug1:
That worklow run (clang-tidy) looks problematic.. taking way too long

PR #3215, opened 2024-05-06
Remove listener when the window is destroyed

eyusupov:
The module was crashing when the output has been deactivated through `wlr-randr --output DP-1 --off` , for example. The module object was destroyed, but the listener was still called, resulting in a NPE.


Alexays:
Thanks!

Issue #3057, opened 2024-03-21
Add additional upower properties


REALERvolker1:
Right now, the `upower` module only supports the `percentage` and `time` properties. It would be nice if it supported other properties exposed by the device's entry in dbus. Some useful examples would be `EnergyRate` (battery usage in watts), `Temperature`, and `Capacity`.

Alexays:
Resolved on master (PR #3220): the upower module now exposes {temperature}, {model}, and {native-path} format replacements in addition to {percentage} and {time}. Documented in waybar-upower(5). Closing as resolved — open a new request if there is a specific additional property still missing.

Issue #2283, opened 2023-07-05
Core dump with the Upower module


StayPirate:
When I try to use the newly introduced Upower feature https://github.com/Alexays/Waybar/pull/2229, I get a core dump. Tagging @LukashonakV as the father of the new implemented feature.

```bash
> waybar -l trace
[2023-07-05 12:38:02.043] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/config
[2023-07-05 12:38:02.043] [debug] Found config file: $XDG_CONFIG_HOME/waybar/config
[2023-07-05 12:38:02.043] [info] Using configuration file /home/crazybyte/.config/waybar/config
[2023-07-05 12:38:02.043] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/style.css
[2023-07-05 12:38:02.043] [debug] Found config file: $XDG_CONFIG_HOME/waybar/style.css
[2023-07-05 12:38:02.043] [info] Using CSS file /home/crazybyte/.config/waybar/style.css
[2023-07-05 12:38:02.052] [debug] Output detection done: eDP-1 (Chimei Innolux Corporation 0x150C)
zsh: segmentation fault (core dumped)  waybar -l trace
```

```
Device: /org/freedesktop/UPower/devices/headphones_dev_04_52_C7_7E_67_82
  native-path:          /org/bluez/hci0/dev_04_52_C7_7E_67_82
  model:                Crazybyte QC35
  serial:               04:52:C7:7E:67:82
  power supply:         no
  updated:              Wed 05 Jul 2023 12:09:06 CEST (1593 seconds ago)
  has history:          yes
  has statistics:       no
  headphones
    warning-level:       none
    percentage:          70%
    icon-name:          'battery-missing-symbolic'
```

```json
// -*- mode: json -*-

[{
	"layer": "top",
	"position": "top",

	"modules-left": [
		"sway/workspaces",
		"custom/right-arrow-dark"
	],
	"modules-right": [
		"custom/left-arrow-dark",
		"cpu",
		"temperature#cpu",
		"custom/left-arrow-light",
		"custom/left-arrow-dark",
		"disk",
		"temperature#disk",
		"custom/left-arrow-light",
		"custom/left-arrow-dark",
		"memory",
		"custom/left-arrow-light",
		"custom/left-arrow-dark",
		"battery",
		"upower#boseqc35",
		"custom/left-arrow-light",
		"custom/left-arrow-dark",
		"pulseaudio",
[...]

PR #3220, opened 2024-05-06
Upower refactoring

LukashonakV:
Hi @Alexays , this MR is created to

1. Optimize UPower code
2. Solve the issue with **native-path** property #2283
3. Add new properties(couple of them are requested by the #3057):
   1. temperature
   2. model
4. Add more accuracy to calculate time-to-full/time-to-empty using std::chrono duration
5. Tooltip for UPower now follows general approach
6. Connection for DBUS now is asynchronous


Alexays:
Thanks, LGTM!

PR #3241, opened 2024-05-08
temperature: allow hwmon-path-abs as array

rtgiskard:
Currently, only `hwmon-path` is allowed to have multiple path, however the hwmon ID is not stable, for which `hwmon-path-abs` is better, but it's supposed to be a string. This commit extends `hwmon-path-abs` and make it works like `hwmon-path`.


Alexays:
Thanks!

rtgiskard:
> Thanks! Can you update the github wiki?

Just update by extend the type to `string/array`, might be enough

PR #3242, opened 2024-05-08
Add support of nvidia graphics using nvidia-smi

voidThread:
Hi!
I've introduced the new module which pulls data from the external 'nvidia-smi' program.
The module runs 'nvidia-smi' as a forked process and saves GPU monitor data into the exchange file. If not specified in the configuration file, the file is created in the system's temporary directory. You can define your own exchange file, but you must ensure that it exists and has proper privileges set.
Since this is my first contribution here, I'm eager to receive feedback on any areas that could be improved.

### DIFF
--- a/include/modules/nvidia_monitor.hpp
+++ b/include/modules/nvidia_monitor.hpp
@@ -1,0 +1,49 @@
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <filesystem>
+#include <string>
+
+#include "ALabel.hpp"
+#include "util/sleeper_thread.hpp"
+
+namespace waybar::modules {
+
+class NvidiaMonitor : public ALabel {
+ public:
+  NvidiaMonitor(const std::string&, const Json::Value&);
+  ~NvidiaMonitor() override;
+  auto update() -> void override;
+
+  static const std::size_t ARGUMENTS_COUNT{5};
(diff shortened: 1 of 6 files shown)


haug1:
I'm not one to comment on this really and I can't test it unfortunately, but from a quick skim of the code and the assumption that you've tested it too work on your machine; great work, looks like you've done a really good job with this. 

I can give you a heads-up about the [man pages](https://github.com/Alexays/Waybar/tree/master/man), it's probably going to get requested that you add some documentation for how to use this if it is to get merged.

Some kind of conditional include in the build step, like what's done with the cava module (if you search "HAVE_LIBCAVA" and follow the trail, you can try to do something similar) would also probably be nice. I'm not sure what's possible, should probably have lead with I'm pretty bad at C++.

voidThread:
Thanks, @haug1 !
I've been using this for a few days without any problems.
Also, I've added what you proposed.

haug1:
I didn't think of this yesterday, sorry. But this might be simply and more correctly achieved for users with a custom module instead so to manage the scope of the project and therefore not interesting to merge. Again, I'm not really one to comment on this. Just thinking out loud. :)

Alexays:
Yup, like @haug1 said, theses specific modules, as pretty simple should be done in a custom module :)

void:
lol, you're invoking me  also now, at least you're polite :)

Issue #1064, opened 2021-04-05
Create custom dropdown menus


tukusejssirs:
I have installed SwayWM and Waybar about a half a year ago. Since then I have use it as my default setup.

One thing I miss in Waybar is the ability to create a custom module with a dropdown menu. This feature request is twofold:

1. I want to output multiple rows of static information.

For example, I created a custom clock and I want to output a dropdown menu with some additional information (e.g. international days or whatever else dependent on today’s date).

2. The other use case is to put some buttons (clickable text) into the dropdown menu similar to the dropdown menu of systray icons.

---

If it is already possible, could someone give an example how I could implement this.

Thanks in advance! :wink:

mrusme:
A dedicated drop-down feature for waybar is imho not needed in order to integrate what you suggested. 

1. Tooltips can be used to output static information. 
2. For something that pops up to offer a list of options, you can simply [define the on-click handler](https://github.com/mrusme/dotfiles/blob/c41cc2e46925a33442bc4c04b8b0649f737e8918/waybar/config#L118) and run [a script that uses wofi](https://github.com/mrusme/dotfiles/blob/master/usr/local/bin/connect-openvpn) to display your options. 

However, while the tooltips and on-click handlers are *nice to haves*, if you're using Sway but still hunt around for clickable drop-down menus your workflow might not be ideal in first place. `wofi` is quite powerful in this regard and lets you set up all sorts of shortcut-triggered "drop-down menus".

tmpm697:
I want this feature also, it's not click to show drop-down like @mrusme mentioned. I want it to hove over to list drop-down menu (hove over to show like with `tooltip`), then be able to select from the list.

This sound likes overlapping whatever fuzzy menu solutions but it's not. It's a `bar` that we can nevigate `thing` to do.

For example, we can't just show a `vpn` status in the bar but not able to select a bunch of vpn profiles from drop-down menu to start/stop the `vpn`. 

`Click` on `vpn` status in waybar module is `really` not idea and then navigate again in fuzzy drop-down menu pop-up in where-who-know to able to select vpn profiles. 

It's not only inconvenient but also give users headache to understand the workflow that minus huge of productivity.

mrusme:
@tmpm697 the config and script ([new version](https://github.com/mrusme/dotfiles/blob/master/usr/local/bin/vpn)) that I mentioned in my previous comment solves **exactly** the scenario you described, using onclick and wofi.

theRealCarneiro:
[...]

PR #3247, opened 2024-05-09
Adding the ability to have dropdown menu for modules

benjamin-voisin:
Related to #1064 : Every ALabel module now can have a dropdown menu that can be spawned using any click you want. The menu is defined with a `xml` file, and the comportment of each button can be configured in the waybar/conf file. The menus can by styled in the `waybar/style.css`.

For example, you can now easily built a well-integrated power module with this config:

```json
"custom/power": {
    "format" : "⏻ ",
    "tooltip": false,
    "menu": "on-click",
    "menu-file": "~/.config/waybar/power_menu.xml",
    "menu-actions": {
        "shutdown": "shutdown",
        "reboot": "reboot",
	"suspend": "systemctl suspend",
	"hibernate": "systemctl hibernate",
    },
},
```
Where the `~/.config/waybar/power_menu.xml` look like this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<interface>
  <object class="GtkMenu" id="menu">
    <child>
		<object class="GtkMenuItem" id="suspend">
			<property name="label">Suspend</property>
        </object>
	</child>
	<child>
        <object class="GtkMenuItem" id="hibernate">
			<property name="label">Hibernate</property>
        </object>
	</child>
    <child>
        <object class="GtkMenuItem" id="shutdown">
			<property name="label">Shutdown</property>
        </object>
    </child>
    <child>
      <object class="GtkSeparatorMenuItem" id="delimiter1"/>
    </child>
    <child>
		<object class="GtkMenuItem" id="reboot">
			<property name="label">Reboot</property>
		</object>
    </child>
  </object>
</interface>
```

This gives something like this:
20240509_21h49m48s_grim

And it can by styled with some css:
```css
menu {
	border-radius: 15px;
	background: #161320;
	color: #B5E8E0;
}
menuitem {
	border-radius: 15px;
}
```
To look like this :
20240509_21h48m57s_grim

The GObject is in the AModule class, so every module could have such menu, but I figured that only the ALabel class should build it automatically reading the config. For other AModule, it could be made differently.

### DIFF
--- a/include/ALabel.hpp
+++ b/include/ALabel.hpp
@@ -27,6 +27,10 @@ class ALabel : public AModule {
 
   bool handleToggle(GdkEventButton *const &e) override;
   virtual std::string getState(uint8_t value, bool lesser = false);
+
+  std::map<std::string, GtkMenuItem*> submenus_;
+  std::map<std::string, std::string> menuActionsMap_;
+  static void handleGtkMenuEvent(GtkMenuItem* menuitem, gpointer data);
 };
 
 }  // namespace waybar
--- a/include/ALabel.hpp
+++ b/include/ALabel.hpp
@@ -28,9 +28,9 @@ class ALabel : public AModule {
   bool handleToggle(GdkEventButton *const &e) override;
   virtual std::string getState(uint8_t value, bool lesser = false);
 
-  std::map<std::string, GtkMenuItem*> submenus_;
+  std::map<std::string, GtkMenuItem *> submenus_;
   std::map<std::string, std::string> menuActionsMap_;
-  static void handleGtkMenuEvent(GtkMenuItem* menuitem, gpointer data);
+  static void handleGtkMenuEvent(GtkMenuItem *menuitem, gpointer data);
 };
 
 }  // namespace waybar
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -4,6 +4,7 @@
 #include <glibmm/markup.h>
 #include <gtkmm/eventbox.h>
 #include <json/json.h>
+#include <gtkmm.h>
 
 #include "IModule.hpp"
 
@@ -52,6 +53,10 @@ class AModule : public IModule {
   std::vector<int> pid_;
   gdouble distance_scrolled_y_;
   gdouble distance_scrolled_x_;
+  GObject* menu_;
+  std::map<std::string, GtkMenuItem*> submenus_;
+  std::map<std::string, std::string> menuActionsMap_;
+  static void handleGtkMenuEvent(GtkMenuItem* menuitem, gpointer data);
   std::map<std::string, std::string> eventActionMap_;
   static const inline std::map<std::pair<uint, GdkEventType>, std::string> eventMap_{
       {std::make_pair(1, GdkEventType::GDK_BUTTON_PRESS), "on-click"},
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -45,6 +45,7 @@ class AModule : public IModule {
   virtual bool handleMouseLeave(GdkEventCrossing *const &ev);
   virtual bool handleScroll(GdkEventScroll *);
   virtual bool handleRelease(GdkEventButton *const &ev);
+  GObject* menu_;
 
  private:
   bool handleUserEvent(GdkEventButton *const &ev);
@@ -53,10 +54,6 @@ class AModule : public IModule {
   std::vector<int> pid_;
   gdouble distance_scrolled_y_;
   gdouble distance_scrolled_x_;
-  GObject* menu_;
-  std::map<std::string, GtkMenuItem*> submenus_;
-  std::map<std::string, std::string> menuActionsMap_;
-  static void handleGtkMenuEvent(GtkMenuItem* menuitem, gpointer data);
   std::map<std::string, std::string> eventActionMap_;
   static const inline std::map<std::pair<uint, GdkEventType>, std::string> eventMap_{
       {std::make_pair(1, GdkEventType::GDK_BUTTON_PRESS), "on-click"},
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -2,9 +2,9 @@
 
 #include <glibmm/dispatcher.h>
 #include <glibmm/markup.h>
+#include <gtkmm.h>
 #include <gtkmm/eventbox.h>
 #include <json/json.h>
-#include <gtkmm.h>
 
 #include "IModule.hpp"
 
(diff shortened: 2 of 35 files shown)


benjamin-voisin:
@nevaforget My implementation uses Gtk objects in a more general way. That is why their is a `xml` file, it's not a random file, it's a Gtk ui definition. That means that the user can use all the power of Gtk menus, with documented and very used xml files. That's also why the actions need to be defined separatly : the Gtk ui definition doesn't allow this.

benjamin-voisin:
@Alexays I've added the checks asked, and I made sure that the app doesn't crash when trying to build the popup menu. If something fails during constructions, it display a warning message explaining why, and still launch the waybar.

I've also updated the manpages of all the modules supporting this behaviour, and created a dedicated `waybar-menu` manpage.

yorunoken:
This would be so so so so useful to have.
Any updates?

Alexays:
Thanks!

user880907:
I'm using the code you provided, but the GtkSeparatorMenuItem seem to not show up, there's just no spacing between items at all like it's ignored completely, what am I doing wrong?

benjamin-voisin:
> I'm using the code you provided, but the GtkSeparatorMenuItem seem to not show up, there's just no spacing between items at all like it's ignored completely, what am I doing wrong?

@user880907 Does the example I provided above work ? It has a GtkSeparatorMenuItem and it shows up correctly. You may add padding into the `style.css` file to choose the spacing.

user880907:
@benjamin-voisin It looks like this.
Adding padding sounds good but I don't know how to target particular menu item in style.css.
20240714_13h42m43s_grim

rieje:
`"menu-file": "~/.config/waybar/power_menu.xml"` doesn't seem to work as in the example--I was not able to use `~` or `$HOME` and only hardcoding the path like `/home/rieje/waybar/power_menu.xml` does (waybar says it's not able to see the file).

Heus-Sueh:
> `"menu-file": "~/.config/waybar/power_menu.xml"` doesn't seem to work as in the example--I was not able to use `~` or `$HOME` and only hardcoding the path like `/home/rieje/waybar/power_menu.xml` does (waybar says it's not able to see the file).

the same thing is happening to me

tom-gora:
Hi. Does anyone knows of any way to dynamically reload the xml template source without restarting waybar itself? I was interested in scripting myself a small dropdown with todo list for the day I could click and and tick off but that requires changing state dynamically.

MariaSolOs:
This is a great feature! Could the use of the `"menu"` properties in the `custom` module be documented in the wiki? I had to look at the default config and this PR to discover how to hook up such menu.

aaronjamesmarshallau:
I, too, have had a bit of a hunt to find information related to this. The default configuration file leverages this custom module, but (at least on NixOS) Hyprland does not generate the default `power_menu.xml` file.

If there's a guide on contributing this information to the wiki somewhere, I'd be happy to have a crack at it... once I get my own power_menu.xml working :sweat_smile:

kRHYME7:
@tom-gora Also tried to use signal in hopes that it will reload the whole module, but it is not working too.

tom-gora:
Well I did a rather inefficient hack to have a dynamic menu for choosing a model run by Ollama. It looks like this:
menu

I have a bash script that rewrites both xml template file, as well as config file:
``` bash
#!/bin/bash

menufile="$HOME/.config/waybar/ollama-model-menu.xml" >"$menufile"
modulesfile="$HOME/.config/waybar/modules.json"
names=()
while IFS= read -r name; do
	names+=("$name")
done < <(ollama ls | awk '{ print $1 }' | tail -n +2)

rewrite_menufile() {
	cat <<EOF >>"$menufile"
<?xml version="1.0" encoding="UTF-8"?>
<interface>
  <object class="GtkMenu" id="menu">
EOF
	for name in "${names[@]}"; do
		cat <<EOF >>"$menufile"
    <child>
        <object class="GtkMenuItem" id="$name">
            <property name="label">$name</property>
        </object>
    </child>
EOF
	done

	cat <<EOF >>"$menufile"
  </object>
</interface>
EOF
}

rewrite_modulesfile() {
	jq --argjson names "$(printf '%s\n' "${names[@]}" | jq -R . | jq -s .)" \
		'."custom/ollama-model" |= ( . + { "menu-actions": (."menu-actions" // {}) } | 
        ."menu-actions" += ( reduce $names[] as $name ({}; .[$name] = "ollama run " + $name ) ) )' \
		"$modulesfile" >"${modulesfile}.tmp" && mv "${modulesfile}.tmp" "$modulesfile"
}

rewrite_menufile
rewrite_modulesfile
```

`modules` file is just json in syntax but `jq` won't accept a file without .json extension for for whatever reason, so I just put `modules.json` in my config dir and symlinked it as `modules` in the same place so waybar reads `modules` and `jq` can take in the json underlying file.
The updated json object would look something like this after overwriting:
``` jsonc
  //...
  "custom/ollama-model": {
    "format": "{}",
    "exec": "$HOME/.config/hypr/scripts/utils/OllamaModel.sh",
    "tooltip": "false",
    "interval": 1,
    "menu": "on-click",
    "menu-file": "$HOME/.config/waybar/ollama-model-menu.xml",
    "menu-actions": {
      "deepseek-r1:14b": "ollama run deepseek-r1:14b",
      "phi4:latest": "ollama run phi4:latest",
      "qwen2.5-coder:14b": "ollama run qwen2.5-coder:14b"
    }
  }
  //...
  ```
  
Finally, I moved `ollama` binary to be `ollama.wrapped` and replaced it with a wrapper script to run it every time I pull or otherwise interact with ollama changing the output of `ollama ls`:

``` bash
#!/bin/bash

update_commands=("serve" "run" "pull" "cp" "rm")

/usr/local/bin/ollama.wrapped "$@"

#run hook if needed
if [[ " ${update_commands[*]} " =~  $1  ]]; then
    /home/tomeczku/.config/hypr/scripts/utils/OllamaMenuBuilder.sh
    killall waybar && waybar &

fi
```
This either calls underlying `ollama` as is or if  args match it runs the menu updating script as a post command hook.
I hope this would help you building custom solutions with what we have available.

kRHYME7:
@tom-gora Thank you for the detailed explanation.

kRHYME7:
I'm just wondering if nested  menus are posible? 

Example on tray icons they have there own native nesting of menuitems. 
image


Is it posible in this implementation?

OTAKUWeBer:
This is an **individual project** I’m working on called **[HyprMenu](https://github.com/OTAKUWeBer/HyprMenu)**. Its goal is to create a **dropdown / menu bar** that can be triggered from Waybar. I’m experimenting with adding a **custom menu UI** that opens when a module is clicked, similar to a full app menu.

**Note:** The window position is currently **not fixed** and may move depending on your setup.

PR #3267, opened 2024-05-14
Add config option to select UPower device based on device model.

luttermann:
This is useful for wireless HID devices, as they often do not have the same `native-path` when re-connected.

### DIFF
--- a/src/modules/upower.cpp
+++ b/src/modules/upower.cpp
@@ -381,12 +381,12 @@ void UPower::setDisplayDevice() {
           }
         },
         this);
-  } else { // if `nativePath_` is empty, but `model_` is not.
+  } else {  // if `nativePath_` is empty, but `model_` is not.
     g_ptr_array_foreach(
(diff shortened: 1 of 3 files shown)


src/modules/upower.cpp
```diff
@@ -379,6 +381,23 @@ void UPower::setDisplayDevice() {
           }
         },
         this);
+  } else {  // if `nativePath_` is empty, but `model_` is not.
```
LukashonakV, line 384: The current approach has already got the logic to filter necessary devices.  In order to have a general approach and make future extensions more flexible(such as filtering by model/path), it's better to implement a separate filtering function in order to avoid code duplication.

luttermann:
This is stil not using a filer function, but does away with the code duplication, if you want to have a filter function instead, I will look into that.
But I don't have a lot of time currently, so it may take a while. Anybody who have any pointers/ideas are welcome to comment or PR my code.

PR #3289, opened 2024-05-22
(hyprland) fix crash when failed to parse IPC message

yangyingchao:
IPC messages are parsed in a dedicated thread, and the thread terminates when an exception is not caught, which causes the waybar process to crash with SIGABORT.

While this issue might be related to Hyprland, it is really annoying to see waybar crash. It would be better to catch those exceptions and report errors instead of crashing.

### DIFF
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -86,7 +86,14 @@ void IPC::startIPC() {
       std::string messageReceived(buffer.data());
       messageReceived = messageReceived.substr(0, messageReceived.find_first_of('\n'));
       spdlog::debug("hyprland IPC received {}", messageReceived);
-      parseIPC(messageReceived);
+
+      try {
+        parseIPC(messageReceived);
+      } catch (std::exception& e) {
+        spdlog::warn("Failed to parse IPC message: {}, reason: {}", messageReceived, e.what());
+      } catch (...) {
+        throw;
+      }
 
       std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }


Alexays:
Thanks!

cjk:
That fixes my crashes with Hyprland, thanks for implementing
Hope this gets released soon!

Mange:
I'd also love to see this released. Waybar crashes for me every time I start a new program, so I think I've restarted Waybar around 14 times during a single workday.

Is there anything I can help you with to speed it up? :smiley:

yangyingchao:
> I'd also love to see this released. Waybar crashes for me every time I start a new program, so I think I've restarted Waybar around 14 times during a single workday.
> 

It sounds strange to crash Waybar by starting a new program. I'm not sure if this merge request will help in your case.

> Is there anything I can help you with to speed it up? 😃

This PR has already been merged, and now we just need to wait for @Alexays  to create a new release. Maybe you can get in touch with xxx and give him a little push :), or you can simply get the latest code and compile it from the source.

cjk:
Agree to what @yangyingchao said. Also, if you're on Arch-Linux, there's a Git-Package in the AUR-Repository one could use to stay on the bleeding edge.
For me, I tend to appreciacte some stability, so I'll wait for a official release :sweat_smile:

Mange:
> It sounds strange to crash Waybar by starting a new program. I'm not sure if this merge request will help in your case.

The coredump points to JSON parsing from Hyprland IPC at least. :slightly_smiling_face:
(And it only started happning recently when updating to Hyprland 0.41.1, did not happen in 0.41.0)

Note that it's only for a new program, not a new window.

> For me, I tend to appreciate some stability, so I'll wait for a official release 😅

Same here. I could probably override the package using a Nix `fetchFromGithub`, but I would prefer a release version to get it into the official binary caches since it takes time to create this override and then to clean it up again afterwards after the release is live.

Anyway, not trying to act entitled here; I have my own OSS projects that aren't released very often so I can empathize. I want to help if I can, and if not I'll just override the package anyway. :heart:

PR #3311, opened 2024-05-29
add format-icons to river/layout

AtomToast:
This PR enables the definition of `format-icons` to let you show an icon instead of the layout name.
Default behaviour stays the same.


AtomToast:
@Alexays Sorry for the ping but is there any chance of this getting a review? I've been running this patch for the last few months and it's been working well for me

AtomToast:
@rickalex21 this is unfortunately a known river limitiation atm: 
https://codeberg.org/river/river/issues/1002

It's also unlikely to change until this current rework of the layout protocol is done: https://codeberg.org/river/river/pulls/1100

rickalex21:
@AtomToast  Ok, thanks. Not a big deal since it's only an issue when switching from a tag with views to an empty tag. The empty tag has the layout of the previous tag despite it being empty. It correctly assigns the layout once a view is opened in the empty tag.

src/modules/river/layout.cpp
```diff
@@ -146,7 +146,9 @@ void Layout::handle_name(const char *name) {
     label_.hide();  // hide empty labels or labels with empty format
   } else {
     label_.show();
-    label_.set_markup(fmt::format(fmt::runtime(format_), Glib::Markup::escape_text(name).raw()));
+    std::string escaped_name=Glib::Markup::escape_text(name).raw();
```
lilly-lizard, line 149: Looks like `escaped_name` isn't used. Can this line be deleted? Are there side effects?

AtomToast: oh, don't remember really since it's almost been a year. But I think that was supposed to be used in the `getIcon()` function? I have been running this change for that same amount of time without issues. But then again I didn't change my layout names during that time.
It's probably fine to leave out?

seeschloss:
Works great for me, thanks

rickswe:
Unfortunately compiling Waybar crashed today and I don't have the time or knowledge to troubleshoot  or reproduce the issue. I don't think it was caused by the patch. The change in this pull request is so minor it's not worth compiling and risking another crash. I hope this gets merged eventually to prevent manual compiling (NixOS overlay with patch) because it has been working flawlessly for me since I started using it in February. Thanks

AtomToast:
@rickswe seems there was a [general regression](https://github.com/Alexays/Waybar/issues/4231) causing waybar to crash. I pulled the newest master changes locally but didn't push them on this branch yet. Not sure how exactly NixOS overlays work but hopefully this may help

rickswe:
@AtomToast  There's always a risk when compiling anything so not sure of the source of orgin. However, I can say that your patch is solid. I have been using it for several months up until yesterday because I'm trying to limit the amount of things that are compiled:

```diff
diff --git a/src/modules/river/layout.cpp b/src/modules/river/layout.cpp
index e938400f..589ca29f 100644
--- a/src/modules/river/layout.cpp
+++ b/src/modules/river/layout.cpp
@@ -146,7 +146,9 @@ void Layout::handle_name(const char *name) {
     label_.hide();  // hide empty labels or labels with empty format
   } else {
     label_.show();
-    label_.set_markup(fmt::format(fmt::runtime(format_), Glib::Markup::escape_text(name).raw()));
+    std::string escaped_name=Glib::Markup::escape_text(name).raw();
+    label_.set_markup(fmt::format(fmt::runtime(format_), fmt::arg("layout", name),
+                                  fmt::arg("icon", getIcon(0, name))));
   }
   ALabel::update();
 }
```

PR #3331, opened 2024-06-05
Enable using capture groups in window-rewrite

Eisfunke:
This uses `std::match_results::format` to format the output of `get` in `RegexCollection` using the match results.

The effect of this is that you can now use capture groups in the regexes in `window-rewrite` in the sway and hyprland workspaces modules, similarly to the window module, e.g. like this:

```json
"window-rewrite": {
    "title<.* - (.*) - VSCodium>": "codium $1"  // captures part of the window title and formats it into output
}
```

This example would rewrite a window titled `foo.cpp - Waybar - VSCodium` to `codium Waybar`. This is quite useful to rewrite long window titles, but keeping relevant parts of the original title, in this example the workpsace of the VSCodium window.

I have only tested this with sway. Note that I don't know C++ very well so I hope I haven't accidentally introduced a memory leak or something like that.

### DIFF
--- a/include/util/regex_collection.hpp
+++ b/include/util/regex_collection.hpp
@@ -36,7 +36,7 @@ class RegexCollection {
   std::map<std::string, std::string> regex_cache;
   std::string default_repr;
 
-  std::string& find_match(std::string& value, bool& matched_any);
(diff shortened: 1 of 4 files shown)


Alexays:
Thanks!

PR #3345, opened 2024-06-09
Organize tests and start Hyprland testing

khaneliman:
Sorry if this is too all over the place for a single PR... just kinda throwing various things together as I came across them and just wanted to fix them...

- Organized the test folder a bit so that its easier to see which test suites are failing and manage them as they scale in size. 
- Start hyprland tests, just testing the getSocketFolder change recently. 
  - This test suite caught a possible crash if someone unsets the XDG_RUNTIME_DIR before launching. Now, we continue to fallback to /tmp/hypr without seg fault. 
- Minor clang-format cleanup that was failing CI
- Change linux builders to not cancel each other if one fails
- Add python3-packaging dependency for opensuse image
  - Tested building docker image locally and then building in the opensuse container, fixed build.
  - NOTE: CI will still fail until an image is pushed with the fix.
- Bump spdlog to next minor version
  - Fixes alpine build and still works with others.
  - https://repology.org/project/spdlog/versions shows its very well distributed amongst the distributions we test against.


khaneliman:
@Alexays Did you need me to break this up or will this PR be fine in this state?

Alexays:
LGTM, thx!

PR #3370, opened 2024-06-17
allow dynamic timeout setting for idle inhibitor

torgiren:
This PR add option to enable dynamically change of idle inhibitor timeout.

my use case is I don't want to restart waybar any time I want to change a timeout.

With this feature, I'm able to change timeout dynamically using mouse scroll, start with left mouse button and stop with right button, so it's more predictable (I don't have to check if the inhibitor is enable or not, I just left click to enable, or right click to disable)

real live usecase:
I often setup timeout to 480 minutes when I start to work, so I can walk over the apartment without worrying if my screen saver will turn on or not... and later, I usually setup it to 120 when playing some movie in the evening


torgiren:
I also tried to make it backward compatible,
so that's why there are "ifs", so with this feature disabled, module should behave same as before

torgiren:
rebased to origin/master

torgiren:
rebased to origin/master

PR #3381, opened 2024-06-23
modules/battery: Deprioritize `capacity` /sys value for battery calculation if other methods are available

### LINKED ISSUE #3293: Battery module gets stuck below 100% at full battery regardless of `design-capacity`
Below is an image showing the issue with my laptop's battery:
image

This happens even if `design-capacity` is set to true, false, or not set at all

OpenSauce04:
Closes #3293

The current code for battery calculation seems to assume that the `capacity` value returns the non-design battery capacity. This is not the case, and it instead returns the design capacity of the battery.

Currently, the `capacity` value is used above all other methods if it is present in the /sys directory of the battery. This means that the design capacity of the battery will always be displayed regardless of the `design-capacity` configuration option.

To remedy this issue, I have restructured the code so that the `capacity` value is only used if the other possible options for calculation have been exhausted beforehand.

This screenshot shows two instances Waybar while the battery is plugged in at full capacity. The top instance is the current release version, and the bottom instance is the code present in this pull request:
image

This pull request maintains the functionality of the `design-capacity` configuration option. Here is the previous screenshot taken again with `design-capacity` set to `true`:
image

Please correct me if I have done anything incorrectly in this pull request, as I am not very familiar with the codebase.

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -272,13 +272,6 @@ waybar::modules::Battery::getInfos() {
       // Some battery will report current and charge in μA/μAh.
       // Scale these by the voltage to get μW/μWh.
 
-      uint32_t capacity = 0;
-      bool capacity_exists = false;
-      if (fs::exists(bat / "capacity")) {
-        capacity_exists = true;
-        std::ifstream(bat / "capacity") >> capacity;
-      }
-
       uint32_t current_now = 0;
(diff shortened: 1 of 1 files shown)


OpenSauce04:
The CI failure appears to be unrelated to the code in this pull request

Alexays:
LGTM

Issue #3388, opened 2024-06-28
crashes due to many uncaught exceptions in hyprland module


maaarghk:
```
[2024-06-28 13:25:12.496] [error] Hyprland IPC: Couldn't write (4)
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error parsing JSON: * Line 1, Column 1
  Syntax error: value, object or array expected.

[1]    2547530 IOT instruction (core dumped)  waybar
```

Seems an empty string is returned to caller here but the caller expects a valid JSON string: https://github.com/Alexays/Waybar/blob/ccc3c132124623bde5127937fe4fc9aa45a9d35d/src/modules/hyprland/backend.cpp#L199 which then fails to parse as JSON. 

the vast majority of calls to `getSocket1Reply` or `getSocket1JsonReply` ought to be wrapped in try/catch, i don't think this is really a fatal class of error, I get crashes constantly when load is high. maybe an option to increase the socket timeout would alleviate the issue?

khaneliman:
Could you try https://github.com/Alexays/Waybar/pull/3389

Issue #3365, opened 2024-06-16
Waybar crash after hyprctl dpms off/on


mattrobineau:
I started using hyprland and noticed that after my monitors turn on after dpms off, waybar had crashed. This issue is similar to #1165.

<details>
<summary>
Output
</summary>
[2024-06-15 21:28:47.554] [info] Using configuration file /home/matt/.config/waybar/config.jsonc

[2024-06-15 21:28:47.555] [info] Unable to receive desktop appearance: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop

[2024-06-15 21:28:47.555] [info] Using CSS file /home/matt/.config/waybar/style.css

[2024-06-15 21:28:47.558] [info] Hyprland IPC starting

[2024-06-15 21:28:47.562] [info] Loading persistent workspaces from Hyprland workspace rules

[2024-06-15 21:28:47.562] [error] media stopped unexpectedly, is it endless?

[2024-06-15 21:28:47.563] [warning] module backlight: Disabling module "backlight", No backlight found

[2024-06-15 21:28:47.584] [info] Loading persistent workspaces from Hyprland workspace rules

[2024-06-15 21:28:47.585] [warning] module backlight: Disabling module "backlight", No backlight found

[2024-06-15 21:28:47.586] [error] media stopped unexpectedly, is it endless?

[2024-06-15 21:28:47.828] [warning] Requested height: 30 is less than the minimum height: 43 required by the modules

[2024-06-15 21:28:47.828] [info] Bar configured (width: 1080, height: 43) for output: DP-2

[2024-06-15 21:28:47.828] [warning] Requested height: 30 is less than the minimum height: 43 required by the modules

[2024-06-15 21:28:47.828] [info] Bar configured (width: 1920, height: 43) for output: DP-1

[2024-06-15 21:30:03.469] [info] Bar removed from output: DP-2

terminate called after throwing an instance of 'std::runtime_error'
  what():  Error parsing JSON: * Line 1, Column 1
  Syntax error: value, object or array expected.

fish: Job 1, 'waybar' terminated by signal SIGABRT (Abort)
</details>

Command:

`hyprctl dispatch dpms off; sleep 12; hyprctl dispatch dpms on`

[Waybar Config](https://gist.github.com/mattrobineau/3f34048629a308272a640910f7688ce6)

Software:
1. hyprland 0.41.0-1
[...]

PR #3389, opened 2024-06-28
modules/hyprland/backend: handle empty socket response 

khaneliman:
Issues:
- Resolves https://github.com/Alexays/Waybar/issues/3388 
- Resolves https://github.com/Alexays/Waybar/issues/3365

This PR: 
- We will check for an empty response from getSocketReply before trying to parse JSON. Before the empty response check we get the JSON parse error from the JSON module. After the change, we wont try and parse the message that will cause the crash.
- Also includes some clangd cleanup. 
- Added test to verify we dont crash when we receive an empty socket response.

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -4,14 +4,11 @@
 #include <gtkmm/label.h>
 #include <json/value.h>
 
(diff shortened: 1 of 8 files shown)


Alexays:
Thanks!

Issue #3366, opened 2024-06-17
wlr/taskbar on multi-monitor: icons not displayed for apps launched before waybar


abc-nix:
Wayland compositor: sway (v 1.9) or Hyprland (0.41.1) [error happens with both]

Waybar version: 0.10.3

I have 2 monitors (output _LVDS-1_ and _HDMI-A-1_), and the exact same waybar configuration is applied to both. Specifically for wlr/taskbar:

<details>
  <summary>wlr/taskbar config</summary>

```
    "wlr/taskbar": {
      "format": "{icon}",
      "icon-size": 24,
      "on-click": "activate",
      "on-click-middle": "close",
    },
```
  
</details>

**Observations:**

1. If waybar has already started, all new apps that I open will display the correct icon on any of the waybar taskbars on both monitors.
2. If waybar starts later, with an app already open, the icon is only displayed correctly in one of the taskbars (the one in HDMI-A-1), and not on the other (there is an empty space with no icon, though it works when clicked and tooltips are displayed).

**Logs:**

By using log-level trace, I see that icons are found and "Loaded" for all open applications, but only are "now visible" on the HDMI-A-1 output. When the icon is "loaded" for the application on LVDS-1, it isn't "now visible". See examples below:

Example on HDMI-A-1:

```
Task (4) overwriting app_id 'unknown' with 'firefox-esr'
Task (4) Mozilla Firefox Private Browsing [firefox-esr] <amif> Loaded icon 'firefox-esr'
Task (4) Mozilla Firefox Private Browsing [firefox-esr] <amif> entered output 0x5d274d223ef0
Task (4) Mozilla Firefox Private Browsing [firefox-esr] <amif> now visible on HDMI-A-1
Task (4) Mozilla Firefox Private Browsing [firefox-esr] <amif> changed
```

Example on LVDS-1:

```
Task (7) overwriting app_id 'unknown' with 'thunar'
Task (7) waybar - Thunar [thunar] <amif> Loaded icon 'org.xfce.thunar'
Task (7) waybar - Thunar [thunar] <amif> entered output 0x5d274d2bc930
Task (7) waybar - Thunar [thunar] <amif> changed
```

As you can see, on LVDS-1, the line "now visible on LVDS-1" is missing. I think this could help find the issue.

I have attached the trace logs in case it is helpful.

Operations performed:

0. Run sway with 2 apps open on each monitor (4 apps in total, foot and thunar on LVDS-1, firefox and geany on HDMI-A-1).
[...]

PR #3393, opened 2024-06-30
workaround for icons not rendered for apps existing before waybar launch

RobertMueller2:
This is a workaround for #3366 .

Using gdb, I can see that `handle_app_id` can sometimes run when `tbar_->icon_themes()` has 0 elements. This happens in multi output settings and can result in empty icons for apps that have been started before waybar.

It looks to me that the Tasks can be present before `Taskbar::Taskbar` has run, I could also confirm this by temporarily setting a bool flag in the constructor. If anyone can, I would appreciate any more detailed explanation why that is even possible, because I don't understand it at all.

With an explanation there's probably better ways to do this. :) But the code in this PR addresses this by setting the app_ids for all of the taskbar's tasks again at the end of Taskbar's constructor, leading to another icon lookup with `icon_themes()` available for `handle_app_id`.

### DIFF
--- a/src/modules/wlr/taskbar.cpp
+++ b/src/modules/wlr/taskbar.cpp
@@ -788,6 +788,10 @@ Taskbar::Taskbar(const std::string &id, const waybar::Bar &bar, const Json::Valu
   }
 
   icon_themes_.push_back(Gtk::IconTheme::get_default());
+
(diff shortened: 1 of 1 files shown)


Alexays:
LGTM

PR #3395, opened 2024-06-30
add hide-empty-text option to hide module whenever output is empty but format is not

### LINKED ISSUE #3341: Custom modules: ability to hide a module?
I have a persistent module that uses JSON output to Waybar. I want it to be able to hide itself when the module thinks it has nothing useful to display — however, after reviewing [src/modules/custom.cpp](https://github.com/Alexays/Waybar/blob/master/src/modules/custom.cpp) I found no obvious way for a JSON module to hide itself (other than exiting with a non-zero exit code, being restarted and not producing any output — but I would like to avoid that for efficiency reasons).

Possible API design:
 - `{}` hides the module
 - `{"display": false}` (a new key) hides the module until a new update arrives where it is unset or set to `true`
 - empty line hides the module until a non-empty line is produced

RobertMueller2:
Resolves #3341

Custom module hides empty output, e.g. when a worker returns `{}` as json. But when using other content in the format, e.g. to add an icon before or after output, it's no longer possible to have the module hidden.

This PR addresses this by adding a config option `hide-empty-text`, when true, `text_` is checked for empty rather than `str`.

I realise I forgot the manpage, coming up in a second :P

### DIFF
--- a/man/waybar-custom.5.scd
+++ b/man/waybar-custom.5.scd
@@ -21,6 +21,10 @@ Addressed by *custom/<name>*
 	The path to a script, which determines if the script in *exec* should be executed. ++
 	*exec* will be executed if the exit code of *exec-if* equals 0.
 
+*hide-empty-text*: ++
+	typeof: bool ++
+	Disables the module when output is empty, but format might contain additional static content.
+
 *exec-on-event*: ++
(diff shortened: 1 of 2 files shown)


Alexays:
LGTM, thx

Issue #3239, opened 2024-05-08
Not keen on cursor change


Flapper01:
Would it be possible to make it optional to change the cursor to indicate module clickability?
It seems inconsistent whether or not it changes anyway.

deathtrip:
I would also like to have an option to disable it.
I think most ( if not all ) users know whether a module is clickable or not.
It should've been made optional, not forced down users' throat by changing the default behavior, with no option to go back.

ordy:
The feature was introduced in #3108. Looking at the code, I think that should be possible, would have to introduce a new setting in the config to define if we want static cursors or not, then check for the value before setting the cursor type in AModule

```
bool AModule::handleMouseEnter(GdkEventCrossing* const& e) {
  if (auto* module = event_box_.get_child(); module != nullptr) {
    module->set_state_flags(Gtk::StateFlags::STATE_FLAG_PRELIGHT);
  }
  
  if (hasUserEvents_) {
    setCursor(Gdk::HAND2);
  }
  return false;
}
```

@bananaoomarang  @haug1 
Mentioning you because you worked on this. Does that sound reasonable to you or is there maybe a better approach?

haug1:
> Looking at the code, I think that should be possible, would have to introduce a new setting in the config to define if we want static cursors or not, then check for the value before setting the cursor type in AModule

Yeah, it should be easy to hide the cursor effect behind a config flag. And I'm not sure it's necessary to set the cursor on the mouse enter/leave events. I think it can just be configured once in the constructor instead. So that's a welcome change as well.

v4u6h4n:
Good to hear that should be possible, kinda makes it look like a webpage lol :-)

ronasimi:
I get a horizontally flipped cursor in hyprland/workspaces since this was added in #3108. Personally I think this is an unecessary setting but please make it toggle-able.

haug1:
I've opened a PR that hides the cursor change behind a config flag and addresses my previous comment.

deathtrip:
Why close the pull request so early without giving the developer time to look at it?
Many pull requests wait longer to get merged.
And we need to get this fixed.

haug1:
Code is still linked, so if the issue is serious enough, that can help in the future.

v4u6h4n:
Damn, thats a pity, the hand cursor just makes waybar look so damn janky.

Flapper01:
Some of us users clearly think it is serious enough.

[...]

PR #3398, opened 2024-07-01
AModule: Cursor config option

khaneliman:
Seen a lot of discussion activity around this again... resurrecting the PR https://github.com/Alexays/Waybar/pull/3274 that resolves https://github.com/Alexays/Waybar/issues/3239. Follow up to work done in https://github.com/Alexays/Waybar/pull/3108 to make this opt in instead of always forced on users. 

One thing that might be nice is to set a global option for this... but I didn't see a way of doing that easily, atm.

### DIFF
--- a/src/AModule.cpp
+++ b/src/AModule.cpp
@@ -69,6 +69,7 @@ AModule::AModule(const Json::Value& config, const std::string& name, const std::
     event_box_.signal_scroll_event().connect(sigc::mem_fun(*this, &AModule::handleScroll));
   }
 
+  // Respect user configuration of cursor
   if (config_.isMember("cursor")) {
     if (config_["cursor"].isBool() && config_["cursor"].asBool()) {
       setCursor(Gdk::HAND2);
@@ -125,6 +126,12 @@ bool AModule::handleMouseEnter(GdkEventCrossing* const& e) {
   if (auto* module = event_box_.get_child(); module != nullptr) {
     module->set_state_flags(Gtk::StateFlags::STATE_FLAG_PRELIGHT);
   }
+
+  // Default behavior indicating event availability
+  if (hasUserEvents_ && !config_.isMember("cursor")) {
+    setCursor(Gdk::HAND2);
+  }
+
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


Alexays:
To avoid yet another breaking change, I prefer to keep the cursor change logic but make it deactivatable in the settings. WDYT?

khaneliman:
> To avoid yet another breaking change, I prefer to keep the cursor change logic but make it deactivatable in the settings. WDYT?

Yeah, i actually agree. I was just waiting for some feedback before I changed it from the original change. I can tweak the change when I get home.

khaneliman:
@Alexays Alright, implemented fallback to default behavior when someone doesn't provide a configuration.

src/AModule.cpp
```diff
@@ -64,6 +68,20 @@ AModule::AModule(const Json::Value& config, const std::string& name, const std::
...
+    if (config_["cursor"].isBool() && config_["cursor"].asBool()) {
+      setCursor(Gdk::HAND2);
+    } else if (config_["cursor"].isInt()) {
+      setCursor(Gdk::CursorType(config_["cursor"].asInt()));
+    } else {
+      spdlog::warn("unknown cursor option configured on module {}", name_);
+    }
+  } else {
+    // Default behavior indicating event availability
+    setCursor(Gdk::HAND2);
```
Alexays, line 83: Should only be set if hasUserEvents_ is true

khaneliman:
@Alexays Sorry, fixed. Also noticed the mouse enter and leave were a breaking change with tool-tips so moved it back there for the default behavior when not configured.

khaneliman:
@Alexays Whenever you get a chance

Alexays:
Thanks!

haug1:
:clap:

haug1:
I was surprised when my waybar crashed coming back from vacation. Turns out it was my own fault in the best kind of way. Had a testing configuration related to these changes left over in my config file from this:

```
  "cursor": 1231
```

As I wrote in the man page, attempting to use unavailable cursor options WILL crash your app! :joy:

PR #3401, opened 2024-07-02
Implement scrolling on the background of the waybar

JohnMH:
This patch allows user events for scrolling on the waybar itself, when the mouse is not over a module.

This experience made me hate GTK 3.0. There's some duplicate code as a result of this too, for checking scroll direction. Smooth scrolling is not handled at all for the bar itself.

This could potentially be done in a better way with the C API.

### DIFF
--- a/include/bar.hpp
+++ b/include/bar.hpp
@@ -69,6 +69,7 @@ class Bar {
   void setVisible(bool visible);
   void toggle();
   void handleSignal(int);
+  bool handleScroll(GdkEventScroll*);
 
   struct waybar_output *output;
   Json::Value config;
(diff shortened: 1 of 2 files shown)


JohnMH:
A cleaner way to do this would most likely be to split the code for scroll direction out into the util namespace.

JohnMH:
Rebased to make this easier to merge.

romuald:
:+1: for the functionality (applied locally to navigate workspaces in the bar "dead space")
```json
{
    "on-scroll-down": "swaymsg workspace next",
    "on-scroll-up": "swaymsg workspace prev",
}
```

PR #3224, opened 2024-05-06
Fix hyprland/language events not working with keyboard names with commas in them

zjeffer:
Based on the proposal by @s1syph0s (https://github.com/Alexays/Waybar/issues/2137#issuecomment-1890935492)

Should fix #2137 and #3222. Looking at https://wiki.hyprland.org/IPC/#events-list, `activelayout` has the parameters `KEYBOARDNAME,LAYOUTNAME`, so `find_last_of` should be more appropriate.

@dxrknesss @xadips Can you guys test this with your keyboard?

PR #3406, opened 2024-07-03
hyprland/language: Fix commas in keyboard names and layouts

tirsek:
Commit 29917fb0 ("Fix hyprland/language events not working with keyboard names with commas in them (#3224)") attempted to fix the problem, but introduced the opposite problem.

The event data from Hyprland is "activelayout>>KEYBOARDNAME,LAYOUTNAME", but both KEYBOARDNAME and LAYOUTNAME may contain embedded commas, so the simple `find_first_of(',')` or `find_last_of(',')` solutions will break things for different users.

This patch attempts to solve the problem by trying all combinations of keyboardname and layoutname until one makes sense, either by matching the keyboard name to the user-configured device name, or by matching the layoutname to the list of valid layouts.

### DIFF
--- a/src/modules/hyprland/language.cpp
+++ b/src/modules/hyprland/language.cpp
@@ -69,15 +69,36 @@ auto Language::update() -> void {
 
 void Language::onEvent(const std::string& ev) {
   std::lock_guard<std::mutex> lg(mutex_);
-  std::string kbName(begin(ev) + ev.find_last_of('>') + 1, begin(ev) + ev.find_first_of(','));
-  auto layoutName = ev.substr(ev.find_last_of(',') + 1);
+
+  const static std::string eventPrefix("activelayout>>");
+
+  if (!ev.starts_with(eventPrefix))
+    return;
+
+  std::string kbName;
+  std::string layoutName;
+  Layout layout;
+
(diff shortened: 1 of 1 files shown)


tirsek:
@zjeffer How about this suggestion? It works for my no-commas keyboard, but with the layout _with_ commas. I've tested the code snippet ex vivo with no issues, but perhaps @dxrknesss can confirm that it still works for your keyboard this way as well? Thanks!

zjeffer:
I feel like this should be fixed in Hyprland tbh. It doesn't make sense to use `,` as a delimiter if both keyboardname and layoutname can have comma's.

zjeffer:
There seems to already be an issue open for this: https://github.com/hyprwm/Hyprland/issues/6298

Maybe you could open a PR in Hyprland's repo to fix this?

Alexays:
Thanks for reporting and fixing this, @tirsek! Since this PR was opened, master reworked the layout parsing (last-comma-before-parenthesis heuristic), which already handles commas in *layout* names. The remaining gap — commas in the *keyboard* name breaking the `keyboard-name` filter — is now fixed in #5155 with a targeted change (matching the configured keyboard-name as a full prefix) that keeps the current parsing approach rather than replacing it with a per-comma loop. Closing in favour of #5155; credit to you for surfacing the bug. 🙏

PR #3407, opened 2024-07-03
modules/hyprland/workspace: ignore empty window-rewrite

khaneliman:
I'd like to ignore some windows from having icons or empty space taken on the bar. By filtering out empty repr we can supply rewrite rules that will ignore them from being processed and showing an empty space or default icon.

image


Alexays:
Thanks!

PR #3409, opened 2024-07-03
Adding Islamic prayer timer module

abdalrahmanshaban0:
A simple module just needs the path to file contains the index of the next pray and the time until in seconds. That file is being updated by some script and the module is just open the file in each update() and read that 2 int values.  The path to that file can be added in the configuration file.

image

### DIFF
--- a/include/modules/prayer_timer.hpp
+++ b/include/modules/prayer_timer.hpp
@@ -1,0 +1,19 @@
+#pragma once
+#include <fmt/chrono.h>
+#include "ALabel.hpp"
+#include <fstream>
+#include "util/sleeper_thread.hpp"
+namespace waybar::modules {
+
+class prayer_timer : public ALabel {
+ public:
+  prayer_timer(const std::string&, const Json::Value&);
+  virtual ~prayer_timer() = default;
+  auto update() -> void override;
+  void SecondsToString(int,char*,int);
+ private:
+  util::SleeperThread thread_;
+  std::string file_path_;
+};
+
+}  // namespace waybar::modules
--- a/meson.build
+++ b/meson.build
@@ -169,6 +169,7 @@ src_files = files(
     'src/modules/load.cpp',
     'src/modules/temperature.cpp',
     'src/modules/user.cpp',
+    'src/modules/prayer_timer.cpp',
     'src/ASlider.cpp',
     'src/main.cpp',
     'src/bar.cpp',
--- a/prayer_timer.patch
+++ b/prayer_timer.patch
@@ -1,0 +1,109 @@
+diff --git a/include/modules/prayer_timer.hpp b/include/modules/prayer_timer.hpp
+new file mode 100644
+index 00000000..3d962743
+--- /dev/null
++++ b/include/modules/prayer_timer.hpp
+@@ -0,0 +1,19 @@
++#pragma once
++#include <fmt/chrono.h>
++#include "ALabel.hpp"
++#include <fstream>
++#include "util/sleeper_thread.hpp"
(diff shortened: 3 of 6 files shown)


AlephNought0:
Did chatgpt write this?

abdalrahmanshaban0:
> Did chatgpt write this?
No, try to tell chatgpt to make it and see yourself.

abdalrahmanshaban0:
> > > Did chatgpt write this?
> > > No, try to tell chatgpt to make it and see yourself.
> 
> Oh yeah you're right, I'm sorry. Chatgpt would write it better.

Ok, let'em help you and write a better one and I'll use it.

AlephNought0:
> > > > Did chatgpt write this?
> > > > No, try to tell chatgpt to make it and see yourself.
> > 
> > Oh yeah you're right, I'm sorry. Chatgpt would write it better.
> 
> Ok, let him help you and write a better one and I'll use it.

I won't because I have no interest in Waybar. However, I'll give you some tips on how to improve it. 

First, literally don't statically link a text file into the module into which you write into multiple times. People do not have access to your home directory. When you make pull requests you also consider others. In your case that would be Islam people I assume.

Second, just use timers that you will start and stop to get how much time has passed and not whatever you're doing. 

And last, just remove the text file altogether and send the values instead straight to the wayland api. You could do this instead
```    
"prayer_timer":{ 
         prayer: "{name}",
         time: "{time}"
     },
```

or however that works in Waybar. Also remove the patch, there's no need for it in a pr. I may sound toxic but I have nothing against you. Just felt like pinching you a little bit. Enjoy coding. 

Making software may be fun, but it's hard to meet the standards.

abdalrahmanshaban0:
> > > > > Did chatgpt write this?
> > > > > No, try to tell chatgpt to make it and see yourself.
> > > 
> > > 
> > > Oh yeah you're right, I'm sorry. Chatgpt would write it better.
> > 
> > 
> > Ok, let him help you and write a better one and I'll use it.
> 
> I won't because I have no interest in Waybar. However, I'll give you some tips on how to improve it.
> 
> First, literally don't statically link a text file into the module into which you write into multiple times. People do not have access to your home directory. When you make pull requests you also consider others. In your case that would be Islam people I assume.
> 
> Second, just use timers that you will start and stop to get how much time has passed and not whatever you're doing.
> 
> And last, just remove the text file altogether and send the values instead straight to the wayland api. You could do this instead
> 
> ```
> "prayer_timer":{ 
>          prayer: "{name}",
>          time: "{time}"
>      },
> ```
> 
> or however that works in Waybar. Also remove the patch, there's no need for it in a pr. I may sound toxic but I have nothing against you. Just felt like pinching you a little bit. Enjoy coding.
> 
> Making software may be fun, but it's hard to meet the standards.

I have to use a separate program to get the prayer timings using an API. I don't configure waybar with the prayer name or time, that must be auto. I should make the module do all that stuff (connected to internet? got the json response file? parse it using a c++ lib, update the timing table and caculate the difference between now and the next pray time and setup the timer. I need to store most recent timings in case of offline and... more stuff). I've made already a notification daemon for prayer timings and I've reused it with the module to make the output of a program an input to another program.

I agree with you that it's not standard and making a module needs internet and using some other program with it not for all people. I made a fork with the module and a patch file, that's ok but I wish that pr makes anyone  intersting about making it more better.. like when the cursor on the timer label, it shows all prayer timings of the day, on click it'll show the absolute time of the next pray instead of a timer, and so on..

Alexays:
Should be a custom module instead of a native one.

Piecuuu:
based pr

PR #3421, opened 2024-07-07
pulseaudio: Add 'sink-mapping' config option

Ape:
This commit adds a new optional config to the pulseaudio modules. It maps sink names to other sink names so that if the current sink is a key, the sink named by the value is considered to be the current sink instead of it.

E.g.
```
"sink-mapping": {
  "easyeffects_sink": "speakers_sink"
}
  ```


Ape:
Any comments on this?

I have been using this for 2 weeks and it seems to work perfectly for me.

Ape:
Rebased on the current master. The clang-format failure is not related to this commit.

PR #3422, opened 2024-07-07
feat/issue 3256: Toggle drawer state

### LINKED ISSUE #3256: Feature request: Toggle drawer state
I have two group modules, one for tray icons and another for hardware related modules. They both use the drawer configuration. It would be nice if it was possible to keep them expanded (hovered state) by clicking on the group/group-leader and the other way around. Please let me know if there is already another way to achieve this behavior, thanks :)

RobertMueller2:
resolves #3256 

Adds a config to groups revealer "click-to-reveal", which allows to reveal the hidden modules by clicking rather than mouse over.

### DIFF
--- a/include/group.hpp
+++ b/include/group.hpp
@@ -25,9 +25,13 @@ class Group : public AModule {
   Gtk::Revealer revealer;
   bool is_first_widget = true;
   bool is_drawer = false;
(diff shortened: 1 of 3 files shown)


khaneliman (approved):
Just tested and works for me. Thanks!

Issue #3430, opened 2024-07-10
tray dissappears after restarting waybar "no such object path /StatusNotifierWatcher"


Shmerlard:
hi, im using wayland with archlinux,
when starting wayland everything works fine but if i will change the config file of waybar (adding module for instance) and then restart waybar i will get:
```
==> ERROR: Cannot fetch updates
[2024-07-10 13:49:46.256] [error] Host: GDBus.Error:org.freedesktop.DBus.Error.UnknownObject: No such object path '/StatusNotifierWatcher'
[2024-07-10 13:49:46.357] [error] Host: GDBus.Error:org.freedesktop.DBus.Error.UnknownObject: No such object path '/StatusNotifierWatcher'
[2024-07-10 13:49:46.370] [warning] Requested height: 30 is less than the minimum height: 35 required by the modules
[2024-07-10 13:49:46.370] [info] Bar configured (width: 1920, height: 35) for output: HDMI-A-1
[2024-07-10 13:49:46.370] [warning] Requested height: 30 is less than the minimum height: 35 required by the modules
[2024-07-10 13:49:46.370] [info] Bar configured (width: 1920, height: 35) for output: eDP-1
```

waybar start with the following script
```
#!/bin/sh

killall waybar

waybar -c ~/dotfiles/waybar/config.jsonc & -s ~/dotfiles/waybar/style.css
```
but even if i will manualy just put in the terminal `waybar &` i will get this error, i also tried killing the `nm-applet` process and the restarting waybar but it didnt work. 
thank you very much

update: i noticed that it happens when changing the css file

RobertMueller2:
Can you try if the PR mentioned above fixed the problem? It's now in git and included with 0.10.4.

Shmerlard:
yup looks like it was solved, thanks!

Issue #3310, opened 2024-05-29
waybar starting after the tray items


XxRiKxX:
in my hyprland conf, the exec-once for waybar in in the first. but still starts after the tray programs have started. so it those does not displays in the tray. the tray is working though

loki1368:
To avoid this I've made something like this in my config

exec-once = sleep 5 && qbittorrent

yorunoken:
This is very annoying when playing around with the theme or changing config file, since you  need to restart waybar, it just gets rid of all the taskbar icons.

Alexays:
Fixed on master (commit 2a748f1a): tray items are now only added to the UI after their DBus proxy is ready and has valid id/category, instead of being inserted immediately on registration. Closing as resolved — please reopen with a repro if items still fail to show on a recent build from master.

PR #3443, opened 2024-07-16
fix #3442

### LINKED ISSUE #3442: tray disappears when starting multiple instances of waybar and the close one of them
Steps to reproduce:

1. start one instance,
2. start another in terminal
3. stroke Ctrl+C to shutdown the seconds instance
4. tray in the first instance disappears with following log:
```
** (waybar:210312): WARNING **: 09:00:42.354: Status Notifier Item with bus name ':1.4' and object path '/org/ayatana/NotificationItem/nm_applet' is already registered
[2024-07-16 09:00:42.356] [error] Host: GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: Status Notifier Host with bus name ':1.203' and object path '/StatusNotifierHost/0' is already registered
```

yangyingchao:
I'm not sure if this is a proper fix; please have it reviewed before merging.

### DIFF
--- a/src/modules/sni/watcher.cpp
+++ b/src/modules/sni/watcher.cpp
@@ -67,10 +67,9 @@ gboolean Watcher::handleRegisterHost(Watcher* obj, GDBusMethodInvocation* invoca
   }
   auto watch = gfWatchFind(obj->hosts_, bus_name, object_path);
   if (watch != nullptr) {
-    g_dbus_method_invocation_return_error(
-        invocation, G_DBUS_ERROR, G_DBUS_ERROR_INVALID_ARGS,
-        "Status Notifier Host with bus name '%s' and object path '%s' is already registered",
-        bus_name, object_path);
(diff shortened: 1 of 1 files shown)


khaneliman (approved):
Worked here locally to fix the issue, thanks!

RobertMueller2:
I've tested it too and it fixes all the problems I've seen with the module. I think this might also address #3430 and #3310.

I can't help but wonder if looking at this https://github.com/Alexays/Waybar/blob/3f61df4e6634a79f047c9aa3f8c035fd970159f9/src/modules/sni/tray.cpp#L22

might avoid the situation altogether. `nb_hosts_` is a static variable that will be used for Host's `id` which ensures multi monitor trays will have unique ids but when running more than one instance of waybar of course it isn't globally unique. And other programs having a tray might also just start from 0. I don't have a suggestion how to deal with it, and I don't mean to stop the PR. I've not seen any problems with it, either. Just wanted to leave this here in case somebody picks up more work on the module in the future.

Alexays:
LGTM, thx!

PR #3447, opened 2024-07-16
hyprland: disable modules instead of rendering empty

khaneliman:
Allows us to disable modules entirely when socket connection isn't working. This is similar to how sway handles their socket connections disabling modules. This supports a single waybar config for multiple IPCs.

Previously, my waybar config would crash when Hyprland modules were in the same config when I was running under sway. In a recent PR I allowed empty responses to not crash the program. Now, I'm throwing runtime_errors when socket doesn't exist so the factory can disable the module in a similar way to the sway module. Now, a single config works properly when you have both hyprland and sway modules declared.


Alexays:
Thanks!

PR #3448, opened 2024-07-16
sway/workspaces: remove deprecated field and ignore empty rewrite rules

khaneliman:
Follow up to https://github.com/Alexays/Waybar/pull/3364 and https://github.com/Alexays/Waybar/pull/3407


Alexays:
Thanks!

PR #3457, opened 2024-07-20
Issue #3414/clock: Shift ONLY calendar

### LINKED ISSUE #3414: [clock] Printed tooltip year/month is based on currently selected calendar entry
Given a clock configuration that displays a time formatted string and a scrollable calendar in the tooltip, it looks like the printed string is based on the currently selected year/month of the calendar.

I'm not sure if this is intentional behavior, but I would expect any time formatted string to always be based on the current time (with a given timezone offset), especially since the calendar already includes the information of the currently selected year/month.

Sample config for testing the behavior:
```
 "clock": {
        ...
        "tooltip-format": "<big>{:%Y %B %d}</big>\n<tt><small>{calendar}</small></tt>",
        "actions":  {
                "on-scroll-up": "shift_up",
                "on-scroll-down": "shift_down"
        }
    },
```
When scrolling up/down the printed month/year (`%B` and `%Y`) changes, but the date stays the same (`%d`), I would expect all formatted values to be constant when scrolling (ie the current year/month/date).

RobertMueller2:
Fixes #3414

Right now, for the tooltip, all times are shifted if shift-down/shift-up actions are used. But it really only makes sense for this to apply to the {calendar} replacement, so use `shiftedNow` there and `now` for all the rest.

Because of this I didn't consider a config option. Should anyone miss the old (I'm inclined to say broken) behaviour, I'm happy to add one in the future. ;)

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -163,15 +163,16 @@ auto waybar::modules::Clock::update() -> void {
       // std::vformat doesn't support named arguments.
       m_tlpText_ =
           std::regex_replace(m_tlpFmt_, std::regex("\\{" + kTZPlaceholder + "\\}"), tzText_);
-      m_tlpText_ =
-          std::regex_replace(m_tlpText_, std::regex("\\{" + kCldPlaceholder + "\\}"), cldText_);
+      m_tlpText_ = std::regex_replace(
+          m_tlpText_, std::regex("\\{" + kCldPlaceholder + "\\}"),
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Before: tooltip uses shifted date for any date format strings outside of the calendar (maybe for historical reasons, when calendar didn't offer the possibility to include month and year etc)
satty-20240721-14:50:35

After: tooltip uses shifted date for calendar, but current date for any other date format strings
satty-20240721-14:51:35

Alexays:
LGTM, thx

Issue #605, opened 2020-02-26
pulseaudio module only detects new default device after volume change


siwyd:
I have a small script I use to quickly switch between output devices. Basically, it sets a new default device and moves all of the sink-inputs to the new sink. I've noticed that Waybar doesn't pick up the change of default device (reproducible when using Pulsemixer or something like that to pick a default device).

Waybar only seems to pick up on the new device when the volume is changed on it. So right now I have this hacky workaround in my script to have Waybar pick up the change:
```
ponymix increase 1
ponymix decrease 1
```

Would it be possible to have the module pick up the change in device without any volume changes?

mssdvd:
[i3blocks -contrib's module volume-pulseaudio](https://github.com/vivien/i3blocks-contrib/tree/master/volume-pulseaudio) has cool features like changing the default output and muting the volume.

nolanl:
I have this problem as well, using pipewire's pulse compability. waybar 0.9.17 from Debian 12. Even changing the volume didn't fix it for me, I had to restart waybar for it to detect the change in output.

xlucn:
Having this as well. Really hope this could be worked on.

What I observe:
- Sometimes the pulseaudio module can't pick up the change of current device (usually when I connect/disconnect a bluetooth earphone) at all.
- The volume change by scrolling on the module actually changes the device that the module shows at the time, not the actually used one.
- Wireplumber module detects the change successfully.
- Switch default devices back-and-forth in `pavucontrol`, and the module probably will detect it after that. Any reason device connection does not trigger the module to update?

PR #3464, opened 2024-07-22
pulseaudio: volume indicator update on default output switch

findus:
Right now the volume indicator does not update when the default output device changes.

I fixed it by saving the current default sink (and its state) and check on every sink-update-callback if the original sink is still present and active (Idle or Running). 

If that is the case, every callback from other sinks are  ignored.

Example:
Top Bar: new
Bottom Bar: current master

Fixes: https://github.com/Alexays/Waybar/issues/605

https://github.com/user-attachments/assets/0a28b77a-d530-4e44-bc1b-e08b2321307a

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,6 +1,6 @@
 name: clang-format
 
-on: [push, pull_request, workflow_dispatch]
+on: [push, pull_request]
 
 concurrency:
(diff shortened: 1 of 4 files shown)


findus:
fixed clang format, but had to touch another file unrelated to my changes to make it work

Gwenodai:
Is there a reason why this still hasn't been merged into main? It's a bit odd that you can't swap outputs with pavucontrol if you want waybar to recognise it.

Issue #3471, opened 2024-07-23
0.10.4: battery not emitting "plugged" anymore


allergicapple:
Somehow in the last update the battery status of "plugged" has gone missing for format rules and css styles.

So a plugged in laptop with a charge limit to 50%, plugged in and not charging only emits "Not Charging" but not the plugged status anymore.

The CSS rule
```
#battery:not(.charging):not(.plugged) {
// when disconnected from power
}
```
is now matching even when plugged in.

PR #3474, opened 2024-07-24
Make battery module update on plugging/unplugging again (refs #2519)

### LINKED ISSUE #2444: Battery state doesn't change unless explicitly refreshed with a click event
Battery module keeps displaying the battery icon even though it's already plugged into charger and should display plug icon.

### LINKED ISSUE #2519: Battery module no longer instantly updates
Back in version 0.9.17 (I think), the battery module would instantly update whenever I plugged or unplugged the charger on my notebook (I have it set up so that the background color for the module changes on toggling charging). However, through no fault of my own (I assume, since I didn't touch my configuration), this is no longer the case: the module now seems to update at the default interval of 60 seconds.

How do I revert this? Or, at least, change my config so that it goes back to how it was, toggling instantly on plugging/unplugging the charger *without* setting the interval to something like a few seconds (I don't want constant polling).

schmop:
When switching how you listen for battery plugging status events, the battery status will instantly update again!
Instead of inotify events on the /sys/class/.../ files, it is recommended to listen to udev netlink kernel events.

Closes #2444 
Closes #2519 

I need aid implementing this, because I'm new to this project.
I will try to make this code look more like the backlight_backend, which also uses the udev netlink interfaces.

### DIFF
--- a/include/modules/battery.hpp
+++ b/include/modules/battery.hpp
@@ -5,7 +5,10 @@
 #include <filesystem>
 #if defined(__linux__)
 #include <sys/inotify.h>
+#include "util/udev_deleter.hpp"
 #endif
+
+#include <sys/poll.h>
 
 #include <algorithm>
 #include <fstream>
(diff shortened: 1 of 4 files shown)


schmop:
So it seems, that the udev kernel events are sent very early. The battery status "Charging" might return just a bit later from the `/sys/class/powersupply/BAT0/status`  path, than the last event is received.
When implementing some `sleep(1)` this race condition does not occur anymore, but this cannot be a solution.

schmop:
Might also
Closes #3726
Closes #3603 
Closes #3471

Jajasek:
Hi, this looks awesome. @Alexays would you consider accepting this PR?

Alexays:
LGTM.

Alexays:
Reverted, FreeBSD is failing.

Issue #3498, opened 2024-08-02
Crash during CPU hotplug


gergo-salyi:
On Linux the number of CPUs can change during runtime (see [CPU hotplug](https://www.kernel.org/doc/html/latest/core-api/cpu_hotplug.html)). Users commonly use this to disable hyper-threading, E-cores, etc. For example:

```
# echo 0 > /sys/devices/system/cpu/cpu1/online
# echo 1 > /sys/devices/system/cpu/cpu1/online
```

Currently this can cause Waybar to crash:

```
$ waybar -l debug

/usr/include/c++/14.1.1/bits/stl_vector.h:1130: constexpr std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = std::tuple<long unsigned int, long unsigned int>; _Alloc = std::allocator<std::tuple<long unsigned int, long unsigned int> >; reference = std::tuple<long unsigned int, long unsigned int>&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

[1]    345703 IOT instruction (core dumped)  waybar -l debug
```

The guilt is on `waybar::modules::CpuUsage::getCpuUsage()` which is iterating through 2 cpu-number-length vectors in `src/modules/cpu_usage/common.cpp:64` unsafely assuming that the number of CPUs is unchanged and hence the length of the vectors is the same.

This results in accessing an out-of-bounds index which is caught by an assert if compiled with FORTIFY_SOURCE and aborts the program.

***I'm writing a PR to fix this...***

Backtrace:
```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x000072ed802a5463 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:78
#2  0x000072ed8024c120 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x000072ed802334c3 in __GI_abort () at abort.c:79
#4  0x000072ed804d3bb0 in std::__glibcxx_assert_fail (file=file@entry=0x5d8cf717d200 "/usr/include/c++/14.1.1/bits/stl_vector.h", line=line@entry=1130,
[...]

PR #3499, opened 2024-08-02
Handle offline CPUs and CPU hotplug

gergo-salyi:
First of all in case when the number CPUs change
prevent out-of-bound index access in
waybar::modules::CpuUsage::getCpuUsage()

Secondly on Linux when updating CPU usage
read /sys/devices/system/cpu/present
and use it to detect the offline CPUs missing from /proc/stat

For offline CPUs report 0 usage and "offline" in the tooltip

Fixes issue #3498

On Linux one can test this functionality with:
echo 0 > /sys/devices/system/cpu/cpu1/online
echo 1 > /sys/devices/system/cpu/cpu1/online

On non-Linux OSes I'm not sure how to detect offline CPUs, so I didn't add the offline CPU detection there
but at least CPU number change should not cause a crash there anymore or cause memory safety issues after this fix


gergo-salyi:
edit: fix forgot clang-format

Alexays:
LGTM, thx

PR #3504, opened 2024-08-04
Add dwl/tags option to show only active tags

Ty3uK:
Not an experienced C++ developer at all, will be glad to hear any advice 🙏
swappy-20240805_013408
Generally, it's like `sway/workspaces` behavior.

### DIFF
--- a/src/modules/dwl/tags.cpp
+++ b/src/modules/dwl/tags.cpp
@@ -198,6 +198,15 @@ void Tags::handle_view_tags(uint32_t tag, uint32_t state, uint32_t clients, uint
   } else {
     button.get_style_context()->remove_class("urgent");
   }
+
+  const bool show_only_viable = config_["show-only-viable"].asBool();
+  if (show_only_viable) {
+    if (clients > 0 || state & (TAG_ACTIVE | TAG_URGENT)) {
+      button.show();
+    } else {
+      button.hide();
+    }
+  }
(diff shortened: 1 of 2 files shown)


src/modules/dwl/tags.cpp
```diff
@@ -198,6 +198,15 @@ void Tags::handle_view_tags(uint32_t tag, uint32_t state, uint32_t clients, uint
   } else {
     button.get_style_context()->remove_class("urgent");
   }
+
+  const bool show_only_viable = config_["show-only-viable"].asBool();
+  if (show_only_viable) {
+    if (clients > 0 || state & (TAG_ACTIVE | TAG_URGENT)) {
+      button.show();
```
Ty3uK, line 205: Not sure if blindly calling `show/hide` is OK, was thinking about `button.is_visible()` checks.

bitterhalt:
Would it be possible to have similar feature in River tags?

thetwistedlogic:
> Would it be possible to have similar feature in River tags?

Just implemented this over here: https://github.com/Alexays/Waybar/pull/3823

Alexays:
Closing as a duplicate of #3925, which adds the same 'hide vacant tags' behavior for dwl/tags (hide tags with no clients that aren't active). Consolidating on #3925. Thanks! 🙏

PR #3506, opened 2024-08-05
taskbar: Send minimize geometry hints

soreau:
This allows compositors to know the minimize widget geometry so that minimize animations work properly.


soreau:
I have applied the diff from CI.

Alexays:
Thanks!

PR #3511, opened 2024-08-06
Accept "default-node-changed" signals from wireplumber, even if only the node ID is changed

hacrvlq:
This fixes an issue on my system where for some reason the initial default node ID is incorrect and the initial node name is correct. Only afterwards there is a "default-node-changed" signal with the correct ID. However, this signal is discarded as the node name is unchanged.


Alexays:
Thanks!

Issue #3302, opened 2024-05-26
CRITICAL : unhandled exception (type Glib::Error) in signal handler  g-dbus-error-quark


ashish-kus:
everyrime i run waybar on terminal  everything work fine but I see these error in std output
```
** (waybar:22634): CRITICAL **: 19:07:05.895:
unhandled exception (type Glib::Error) in signal handler:
domain: g-dbus-error-quark
code  : 41
what  : GDBus.Error:org.freedesktop.DBus.Error.UnknownObject: Unknown object '/org/freedesktop/login1/session/self'.
```
config: https://0x0.st/XZ0F.json

shimeoki:
Same issue. Appeared after I added the "backlight/slider" module.

apiraino:
I did some investigation here, I'm not sure of my findings.

I think this error is not causing anything bad, it's just that Waybar tries to connect to the wrong dbus login session. According to the following output, there is no `/org/freedesktop/login1/session/self`. Instead there is a `/org/freedesktop/login1/session/auto`:
```
$ busctl tree org.freedesktop.login1
└─ /org
  └─ /org/freedesktop
    ├─ /org/freedesktop/LogControl1
    └─ /org/freedesktop/login1
      ├─ /org/freedesktop/login1/seat
      │ ├─ /org/freedesktop/login1/seat/auto
      │ └─ /org/freedesktop/login1/seat/seat0
      ├─ /org/freedesktop/login1/session
      │ ├─ /org/freedesktop/login1/session/_33
      │ ├─ /org/freedesktop/login1/session/_34
      │ └─ /org/freedesktop/login1/session/auto
      └─ /org/freedesktop/login1/user
        ├─ /org/freedesktop/login1/user/_1000
        └─ /org/freedesktop/login1/user/self
```

I recompiled Waybar changing that:
```diff
--- a/src/util/backlight_backend.cpp
+++ b/src/util/backlight_backend.cpp
@@ -153,7 +153,7 @@ BacklightBackend::BacklightBackend(std::chrono::milliseconds interval,
   // Connect to the login interface
   login_proxy_ = Gio::DBus::Proxy::create_for_bus_sync(
       Gio::DBus::BusType::BUS_TYPE_SYSTEM, "org.freedesktop.login1",
-      "/org/freedesktop/login1/session/self", "org.freedesktop.login1.Session");
+      "/org/freedesktop/login1/session/auto", "org.freedesktop.login1.Session");
```

[...]

PR #3519, opened 2024-08-12
Fix for #3302

cooltexture1:
Issue #3302 already had a proposed solution but no pull request to implement it yet.
This implements the proposed change.


Alexays:
Can you add a fallback to self to be sure?

cooltexture1:
Uh. I actually dont know any cpp. I just implemented that one line change from the issue. I actually dont even use waybar, i just have a non-programmer friend that was bothered by the issue, and doesnt know how to use git/github.

PR #3527, opened 2024-08-18
hyprland/backend: drop unnecessary getaddrinfo call

alebastr:
Hyprland hasn't been using TCP sockets for IPC since the first release, so this getaddrinfo call and its result was never needed.

Additionally, it leaks the `aiRes`, causing test failure under ASan.

Not tested on a real Hyprland, but see https://github.com/hyprwm/Hyprland/commit/39a9980fb116108807ee68def27a6348184f282c.

CC: @khaneliman


khaneliman:
Sorry, didn't see this right away. Will verify later tonight in Hyprland.

khaneliman (approved):
LGTM, thanks!

PR #3565, opened 2024-09-03
feat: hidpi support for image module

saveman71:
The image module renders SVG poorly on an hidpi screen. This fixes that. I'm not a C expert and it was my first time using the image module so I hope this doesn't break anything.

Cheers!


saveman71:
I based my fix on https://gitlab.gnome.org/GNOME/gtk/-/issues/613 and then found that some code in other places already this (kinda) (grep for create_surface_from_pixbuf)

Alexays:
LGTM thx :)

PR #3579, opened 2024-09-08
Revert "Uses user-defined mode when toggling visibility rather than r…

### LINKED ISSUE #3578: Once hidden waybar cannot be shown
From FAQ:

> You can toggle the visibility of the bars with:
> 
> `killall -SIGUSR1 waybar`

However, the bar stays hidden when sending SIGUSR1, only SIGUSR2 shows it again.

Version: 0.10.4

hramrach:
…esetting to default mode"

But uses the invisible mode once it's set invisible.

This reverts commit f4c6dfcddc032a008f4b119edfc519fc4b38f64e.

This should fix #3578

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -406,11 +406,7 @@ void waybar::Bar::onMap(GdkEventAny* /*unused*/) {
 
 void waybar::Bar::setVisible(bool value) {
   visible = value;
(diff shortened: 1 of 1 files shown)


leon-erd:
Can be closed I think

PR #3613, opened 2024-09-15
fix: return app_identifier as itself in AAppIconLabel if it's an absolute path

### LINKED ISSUE #3612: [Window] window application icon doesn't work for absolute path
`firefox` and `com.visualstudio.code.oss` in this case just use the icon name, and `tidal-hifi` uses an absolute path. This breaks the icon shown next to the application name, despite the file existing.

image

image

image

ThatOneCalculator:
If an app's identifier is an absolute path, return said path. This is for apps like `tidal-hifi` that identify itself as `/usr/share/icons/tidal-hifi/tidal-hifi.png`. Closes #3612


ThatOneCalculator:
Tested and works

Alexays:
Thanks!

Issue #3605, opened 2024-09-14
Bell icon not appearing after system update.


DHDcc:
* Distribution: ArchLinux
* wm: Hyprland
* Kernel version: 6.10.10
* waybar version: v0.10.4

After updating my system, the bell icon on my bar just disappeared.
Here the error I got after updating my system:
```[2024-09-14 19:19:37.526] [error] custom/notify: cannot switch from manual to automatic argument indexing```

here my config for ```custom/notify```: 
```
    "custom/notify": {
      "escape": true,
      "exec": "swaync-client -swb",
      "exec-if": "which swaync-client",
      "format": "{icon} {}",
      "format-icons": {
        "dnd-inhibited-none": " <sup></sup>",
        "dnd-inhibited-notification": "<span foreground='#f38ba8'><sup></sup></span>",
        "dnd-none": " ",
        "dnd-notification": " <span foreground='#f38ba8'><sup></sup></span>",
        "inhibited-none": "<sup>󰜺</sup>",
        "inhibited-notification": "󰵙<span foreground='#f38ba8'><sup></sup></span>",
        "none": " ",
        "notification": " <span foreground='#f38ba8'><sup></sup></span>"
      },
      "on-click": "sleep 0.1 && swaync-client -t -sw",
      "return-type": "json",
      "tooltip": false
    },
```

Before update:
![1726335141_grim](https://github.com/user-attachments/assets/acb82dbb-5c68-4d73-89de-1820456d1932)

After update:
![1726335049_grim](https://github.com/user-attachments/assets/8ccde55d-efb2-482b-8026-91e61f56f8b5)

PaoloReyes:
They said updating to waybar 0.11.0 should solve it. But others say it does not. For me downgrading fmt made the trick

DHDcc:
Updating waybar to 0.11.0 does't work and downgrading it to 0.10.3 gave me this error: ```waybar: error while loading shared libraries: libfmt.so.10: cannot open shared object file: No such file or directory```

PaoloReyes:
Downgrade fmt package

PaoloReyes:
I think it will give another error with sdplog package, downgrade it and then should run flowlessly

DHDcc:
Yeah, it's working. I hope they fix this soon.

NeobinX:
- Arch/Garuda Linux
- Hyprland
- Waybar 0.11.0-3 from `extra`
- fmt 11.0.2-1 from `extra`

[...]

PR #3617, opened 2024-09-16
fix(custom): stop mixing manual and automatic arg indexing

lukasfink1:
The current documentation for the custom module suggests mixing manual (`{icon}`) and automatic (`{}`) indexing of format args. Newer versions of the fmt library seem to not support this anymore (see issue #3605).

This commit introduces a name for the `text` output of the script, so that `{text}` can now be used instead of `{}` in the configuration. The documentation in the wiki should be updated upon rolling out this change.

An alternative way would be to just replace `{}` with `{0}` in the documentation, then this pull request wouldn’t be required.

### DIFF
--- a/man/waybar-custom.5.scd
+++ b/man/waybar-custom.5.scd
@@ -55,8 +55,8 @@ Addressed by *custom/<name>*
 
 *format*: ++
 	typeof: string ++
-	default: {} ++
-	The format, how information should be displayed. On {} data gets inserted.
+	default: {text} ++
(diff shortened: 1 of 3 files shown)


alebastr:
Please, update the man (FORMAT REPLACEMENTS in waybar-custom.5.scd) and the example config.

Alexays:
To help users, can you add a warning if the user's format contains "{}"?

Alexays:
Thanks!

PR #3626, opened 2024-09-19
FIX: fix keyboard layout variants in hypland/language module

SpieringsAE:
I noticed an issue where my English (US intl., with deadkeys) layout did not show up properly in the hyprland/language module, did some debugging and this fixed it.


SpieringsAE:
nvm I just realized there is already an open pull request for this #3406 so I guess I'll close this for now. 
That pr doesn't seem to be getting a lot of attention, but it is a lot better of a fix than this one.

Issue #3619, opened 2024-09-17
compile failed with `ninja -C build`


peiwithhao:
I clone the latest git repo and just meson && ninja, but some error happened there :(

+ linux kernel: 6.10.10-arch1-1
+ ninja version: 1.12.1
+ meson version: 1.5.1
+ WM: Hyprland
```
FAILED: test/waybar_test 
c++  -o test/waybar_test test/waybar_test.p/main.cpp.o test/waybar_test.p/config.cpp.o test/waybar_test.p/.._src_config.cpp.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -Wl,--start-group -lstdc++fs /usr/lib/libCatch2.a /usr/lib/libfmt.so /usr/lib/libgtkmm-3.0.so /usr/lib/libgdk-3.so /usr/lib/libz.so /usr/lib/libharfbuzz.so /usr/lib/libpango-1.0.so /usr/lib/libglib-2.0.so /usr/lib/libgobject-2.0.so /usr/lib/libsigc-2.0.so /usr/lib/libpangocairo-1.0.so /usr/lib/libatkmm-1.6.so /usr/lib/libglibmm-2.4.so /usr/lib/libatk-1.0.so /usr/lib/libcairo.so /usr/lib/libcairo-gobject.so /usr/lib/libgio-2.0.so /usr/lib/libgdkmm-3.0.so /usr/lib/libgiomm-2.4.so /usr/lib/libgtk-3.so /usr/lib/libcairomm-1.0.so /usr/lib/libpangomm-1.4.so /usr/lib/libgdk_pixbuf-2.0.so /usr/lib/libjsoncpp.so /usr/lib/libspdlog.so -Wl,--end-group -pthread
/sbin/ld: test/waybar_test.p/config.cpp.o: in function `CATCH2_INTERNAL_TEST_21()':
config.cpp:(.text+0x287): undefined reference to `Catch::ITransientExpression::~ITransientExpression()'
/sbin/ld: config.cpp:(.text+0x361): undefined reference to `Catch::ITransientExpression::~ITransientExpression()'
/sbin/ld: config.cpp:(.text+0x454): undefined reference to `Catch::ITransientExpression::~ITransientExpression()'
/sbin/ld: config.cpp:(.text+0x568): undefined reference to `Catch::ITransientExpression::~ITransientExpression()'
[...]

PR #3635, opened 2024-09-24
Fix 3619. Bump dependency catch2

LukashonakV:
Signed-off-by: Viktar Lukashonak <myxabeer@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -586,11 +586,12 @@ endif
 
 catch2 = dependency(
     'catch2',
+    version: '>=3.7.1',
     default_options: [ 'tests=false' ],
     fallback: ['catch2', 'catch2_dep'],
     required: get_option('tests'),
 )
-if catch2.found()
+if catch2.found() and get_option('tests').enabled()
     subdir('test')
 endif
 
--- a/subprojects/catch2.wrap
+++ b/subprojects/catch2.wrap
@@ -1,10 +1,10 @@
 [wrap-file]
-directory = Catch2-3.7.0
-source_url = https://github.com/catchorg/Catch2/archive/v3.7.0.tar.gz
-source_filename = Catch2-3.7.0.tar.gz
-source_hash = 5b10cd536fa3818112a82820ce0787bd9f2a906c618429e7c4dea639983c8e88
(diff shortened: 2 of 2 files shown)


LukashonakV:
Hi @Alexays, @alebastr . Waybar meson configuration does not aware off whether tests option enabled or not. And In case when catch2 library is found forces to compile test subproject. I've reproduces case is mentioned in #3619 . By default catch2 has got 'tests' option disabled. In my opinion in such case no need to compile test subproject.

meson.build
```diff
@@ -586,11 +586,12 @@ endif
 
 catch2 = dependency(
     'catch2',
+    version: '>=3.7.1',
     default_options: [ 'tests=false' ],
     fallback: ['catch2', 'catch2_dep'],
     required: get_option('tests'),
 )
-if catch2.found()
+if catch2.found() and get_option('tests').enabled()
```
alebastr, line 594: This isn't correct, please see <https://mesonbuild.com/Reference-manual_functions.html#dependency_required>. `-Dtests=disabled` already skips the dependency lookup and sets `catch2` to an empty dependency object.
```
Dependency catch2 skipped: feature tests disabled
```

Also, no reason to enforce the version requirement. Last time I checked the test code did not use anything specific to catch2 version 3.7.1 and worked with version 2.x.

LukashonakV: Hi @alebastr , very interesting case:
1. Actually by default meson configuration for waybar provides two or even more features with the same name ```tests```. One of belongs to the project itself and others one to their dependencies. In my case ```tests``` feature is provided by the gtk-layer-shell-0 and catch2 . So when catch2 dependency is not met in the system, meson trough dependency fallback provides ```tests=false```
2. When catch2 dependency is presented in the system then ```tests``` project feature is used. In my system by default ```tests=auto``` which leads to test subproject compilation. In this case for some reason linker failes with the undefined references
----
Need to digging deeper to check why linker fails

LukashonakV: Looks like catch2 issue. I create pure small project with catch2 dependency only and not able to compile it. Will create issue to catch2

LukashonakV: Appropriate issue is created catchorg/Catch2#2912

Issue #2275, opened 2023-07-03
[Feature request] add an option to put a minimum brightness for the backlight module


beh-10257:
well I guess by default it's so easy to go to 0% accidently and well I guess if you are not as familiar with brightnessctl as a new user it would be so nice If I can just limit it to 10% or something
+ thanks

grishatop1:
Totally agree, I set it to 0%, had to buy a new laptop

ligmaSec:
agreed this is kinda annoying

TripleTrable:
For anybody interested, I just opened a PR for my local patch I used the last few weeks:
https://github.com/Alexays/Waybar/pull/3637

Drartem34:
doesn't work (put "min-brightness" into "backlight" module)

beh-10257:
this is fixed in git thanks a lot @TripleTrable

PR #3637, opened 2024-09-24
#2275  Backlight: Add minimum brightness

TripleTrable:
As currently it is possible to turn the brightness to zero which may not be desirable, this patch add a configurable brightness check.


Alexays:
Default should still be zero to avoid breaking change

LukashonakV:
> Default should still be zero to avoid breaking change 
In such case @TripleTrable you should assign 0.0

TripleTrable:
I'll work on it once im back from Work

TripleTrable:
Done. Any other requested changed?

Alexays:
LGTM thx :)

PR #3651, opened 2024-09-28
.github/workflows: don't run on forks

khaneliman:
Limits the scheduled build and push and flake update CI to not run on forks since they just create failure noise.

### DIFF
--- a/.github/workflows/nix-update-flake-lock.yml
+++ b/.github/workflows/nix-update-flake-lock.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   lockfile:
     runs-on: ubuntu-latest
+    if: github.repository == 'Alexays/Waybar'
     steps:
       - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/nix-update-flake-lock.yml
```diff
@@ -9,6 +9,7 @@ on:
 jobs:
   lockfile:
     runs-on: ubuntu-latest
+    if: github.repository == 'Alexays/Waybar'
```
alebastr, line 12: Maybe `if: github.event_name != 'schedule' || github.repository == 'Alexays/Waybar'`?

I found `workflow_dispatch` support useful when I had to update `flake.lock` without any knowledge about Nix.

khaneliman: I was gonna do that, since we do that in nixvim to allow people to manually run it. Didn't know if people would find it helpful here, too.

Issue #3344, opened 2024-06-09
Multiple waybars open after resume from dpms off.


simonm:
Hi - I use waybar, hyrpland, hypridle and hyprlock.

Most times I wake from sleep (dpms off), waybar is showing multiple copies of itself.

This can usually be fixed by a simple `pkill -SIGUSR2 waybar` but this morning I come to my workstation and I have reached a new record. 7 waybars across both monitors! 

![2024-06-09-110936_hyprshot](https://github.com/Alexays/Waybar/assets/2259/404a602d-76aa-4550-a516-7cadfbdda6d5)

Attached are the waybar logsfrom today, I woke the system (and unlocked screenlock) at 10:59:13.  (`journalctl --user-unit waybar-hypr.service --since today > waybar-log.txt`). 
[waybar-log.txt](https://github.com/user-attachments/files/15751264/waybar-log.txt)

```❯ waybar --version
Waybar v0.10.3-60-gf4da2039 (branch 'master')
```

```
❯ expac '%-30n %v' -s 'waybar|hyprland|hypridle|hyprlock'
hyprcursor                     0.1.9-1
hyprevents-git                 11.09b54e7-1
hypridle                       0.1.2-1
hyprland-git                   0.40.0.r146.a54ab301-1
hyprlock                       0.3.0-1
hyprprop-git                   16.46d12db-1
hyprshot                       1.3.0-1
hyprwayland-scanner            0.3.8-1
waybar-git                     r3484.f4da2039-1
xdg-desktop-portal-hyprland    1.3.1-6
```

Does anyone have any suggestion on how to troubleshoot?

Thanks

apiraino:
I am also experiencing this with Waybar v0.10.3 and Sway. I was hesitating reporting before I had a concrete reproducible but I didn't yet figure how to trigger this behaviour.

For now I can only confirm that:
- there is one Waybar process running at all times
- sending a `SIGUSR2` to Waybar kills them but after another DPMS off -> on they're all back
- the number of Waybars grows randomly up to 7 (why 7?)
- sending a `SIGKILL` and restarting Waybar of course wipes everything until it grows back to 7

[...]

PR #3669, opened 2024-10-07
Signal handling fixes

alebastr:
First draft. Tested on Linux with glibc only.

See individual commit descriptions for details.

TODOs:
- It should not be necessary to recreate the outputs. But the whole thing has so many races and weird things happen if I keep using the existing output list.
  Time to dig out my old work on reducing the number of `wl_display_roundtrip`s in the code.
- `CssReloadHelper` does not handle appearance changes and `.reset()` on reloads can be optimized.

Fixes: #3344

### DIFF
--- a/include/client.hpp
+++ b/include/client.hpp
@@ -15,11 +15,16 @@ struct zwp_idle_inhibit_manager_v1;
 
 namespace waybar {
 
-class Client {
+class Client : public sigc::trackable {
  public:
   static Client *inst();
   int main(int argc, char *argv[]);
-  void reset();
+
+  /* signal handlers */
+  void handleSignal(int signum);
+  void reload(int signum = 0);
+  void toggle(int signum = 0);
+  void quit(int signum = 0);
 
   Glib::RefPtr<Gtk::Application> gtk_app;
   Glib::RefPtr<Gdk::Display> gdk_display;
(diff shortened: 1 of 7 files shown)


ShellCode33:
Hi, I'm coming from #3344. Thanks for your work on this!

I see that you are spawning a new thread in the signal handler:

https://github.com/Alexays/Waybar/blob/d8a7f429e756f4dc378ed1cb9824d47ea46edab5/src/util/unix_signal.cpp#L48-L59

Even though that will probably work most of the time, you unfortunately cannot do that, `pthread_create` is not async-signal-safe. You might end up with users reporting bugs that are extremely complex to diagnose.

You can have a look at [`man 7 signal-safety`](https://man7.org/linux/man-pages/man7/signal-safety.7.html) for a list of functions you can call in a signal handler.
Some pthread functions are there:

```
pthread_kill(3)        Added in POSIX.1-2008 TC1
pthread_self(3)        Added in POSIX.1-2008 TC1
pthread_sigmask(3)     Added in POSIX.1-2008 TC1
```

But not `pthread_create`. What people usually do in signal handlers is that they set an atomic boolean (let's say `sigusr2_needs_to_be_processed`) to true, and handle it in a separate thread that is already running. Once the thread is done handling the signal, it would set `sigusr2_needs_to_be_processed` back to false.

alebastr:
> I see that you are spawning a new thread in the signal handler:

That's not what is happening here.

The signal handling thread is started from the main thread in normal context, then we mask all the signals that are interesting for us for all the application threads and listen for signals with `sigtimedwait(2)`. Not a single line of the new code runs in a signal handler context.

ShellCode33:
Oh my bad, I misunderstood what's going on then, I'm sorry for wasting your time

PR #3679, opened 2024-10-12
Add signal to hide/show group

marc24force:
Added the "signal" field in the group class to toggle the visibility of the group without requiring a mouse click if "click-to-reveal" is set to true.


marc24force:
Added a feature to also open the group with a signal if the click-to-reveal is false. Uses interval as a delay before closing the group again. Sending the signal while the the group is open resets the delay. This is only affected if the mouse cursor is not on top of the group.

PR #3694, opened 2024-10-16
Update mediaplayer.py with player actions

deigil:
- To play the specified player:
  ```bash
  python3 waybar/scripts/mediaplayer.py --player "spotify" --action play
  ```

- To pause the specified player:
  ```bash
  python3 waybar/scripts/mediaplayer.py --player "spotify" --action pause
  ```

- To toggle play/pause for the specified player:
  ```bash
  python3 waybar/scripts/mediaplayer.py --player "spotify" --action play-pause
  ```

- To skip to the next track:
  ```bash
  python3 waybar/scripts/mediaplayer.py --player "spotify" --action next
  ```

- To go back to the previous track:
  ```bash
  python3 waybar/scripts/mediaplayer.py --player "spotify" --action previous
  ```

### DIFF
--- a/resources/custom_modules/mediaplayer.py
+++ b/resources/custom_modules/mediaplayer.py
@@ -1,8 +1,7 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 import gi
 gi.require_version("Playerctl", "2.0")
 from gi.repository import Playerctl, GLib
-from gi.repository.Playerctl import Player
 import argparse
 import logging
 import sys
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks! There are a few regressions here that block merging: the shebang changes from `python3` to `python` (not always present), a broken type hint (`List[Playerctl]`), and the actions shell out to `playerctl` instead of using the already-loaded Playerctl API (redundant — users can bind `playerctl` directly). The genuinely useful part is the `--exclude` positional-arg fix; if you'd like, open a focused PR with just that and we'll take it. 🙏

Issue #3663, opened 2024-10-03
Hyprland modules not loading properly, IPC unable to connect


xiyori:
Sometimes all Hyprland modules fail to load. Happens both on startup and after hypridle. The modules are visible and respond to click and scroll actions, however, they do not update, for example, workspaces module always shows one workspace and language module shows empty string.

<details> <summary>Waybar log</summary>

```
[2024-10-03 08:11:20.463] [info] Using configuration file /home/sergej/.config/waybar/config.jsonc
[2024-10-03 08:11:20.607] [info] Unable to receive desktop appearance: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop
[2024-10-03 08:11:20.607] [info] Using CSS file /home/sergej/.config/waybar/style.css
[2024-10-03 08:11:20.613] [info] Hyprland IPC starting
[2024-10-03 08:11:20.613] [error] Hyprland IPC: Unable to connect?
[2024-10-03 08:11:20.613] [info] Loading persistent workspaces from Hyprland workspace rules
[2024-10-03 08:11:20.749] [warning] Requested height: 45 is less than the minimum height: 50 required by the modules
[2024-10-03 08:11:20.749] [info] Bar configured (width: 1900, height: 50) for output: DP-3

** (waybar:1273): WARNING **: 08:11:20.756: Status Notifier Item with bus name ':1.28' and object path '/org/ayatana/NotificationItem/nm_applet' is already registered
failed to unlink ports: No such file or directory
failed to unlink ports: No such file or directory
failed to unlink ports: No such file or directory
failed to unlink ports: No such file or directory
==> ERROR: Cannot fetch updates
```

</details>

<details> <summary>System/Version info</summary>

```
Waybar v0.11.0
```

```
Hyprland, built from branch  at commit 0f594732b063a90d44df8c5d402d658f27471dfe  (props: bump version to 0.43.0).
Date: Sun Sep 8 16:48:21 2024
Tag: v0.43.0, commits: 5196
```

</details>

Relaunching waybar always fixes the issue, but it happens again after some time. I tried a fix from https://github.com/Alexays/Waybar/issues/3230 `ln -s $XDG_RUNTIME_DIR/hypr /tmp/hypr` and it didn't help much.

vpcano:
same problem here

ninetailedtori:
Still hitting this issue even with #3695.

Issue #17, opened 2018-08-12
Avoid polling


Alexays:
**Current modules that do not using polling:**
- [ ] Workspaces
- [x] Pulseaudio
- [ ] Battery
- [x] Clock
- [ ] Network
- [ ] ~~Custom~~ (Can only use polling)
- [ ] ~~Memory~~ (Can only use polling)
- [ ] ~~Cpu~~ (Can only use polling)

Issue #16, opened 2018-08-11
When layer is set to bottom (default) tooltip shown under windows


Alexays:

Alexays:
Might be fixed with #441 can someone test?

David96:
Related: https://github.com/swaywm/sway/issues/4684 and https://github.com/swaywm/sway/pull/4781
I currently don't have the time to dig deep enough into the sway rendering code to properly fix this and I'm also not sure whether the way the pull request "solves" the issue is the best way of dealing with it.

ainola:
This is not fixed with gtk-layer-shell - Tooltips are still displayed underneath unless layer is explicitly set to top.

piu130:
@ainola How do you set it to top?

eternal-sorrow:
@piu130 `"layer": "top",`

dd23:
I had the waybar layer set to bottom, so `bemenu` would be shown on top of waybar (as recommended in #474), but since the last release also **tray icon context menus** disappeared behind windows.
Just wanted to say that this is not limited to tooltips.

L0stLink:
Exactly as dd23 said, I would have no problem using the workaround but I need to be able to use `bmenu` as well and I cannot do that if it gets hidden behind `waybar`.

teutat3s:
One possible workaround is to use `bemenu --no-overlap` and configure `waybar` to with `"layer": "top"`

David96:
This is now fixed in sway (https://github.com/swaywm/sway/pull/4781), therefore can be closed I think. Also the default layer should probably be bottom now, so things like bemenu can display over waybar.

Alexays:
Thanks @David96 🎉

Issue #15, opened 2018-08-11
Configuration of bar width


sidzwans:
HI,

I am using sway 1.0. My monitor's default resolution is 2560x1440. I force the TTY resolution to 1920x1080 by using kernel parameter like "video=HDMI-A-1:1920x1080" as recommended in sway Archwiki . Thus I am able to set the new resolution in sway's config. Command `swaymsg -t outputs' shows the current mode is 1920x1080 and there are 2 modes available.

When I run the waybar, I ended up with width of 2560. Not the width of my current resolution that I set. Thus some part of the bar is not visible.

Waybar output is like:
**Bar width configured: 1920
Bar width configured: 2560
New height: 30
New height: 30**

Can I get around this?

Thanks.

Alexays:
Weird, since the last configured width i get is normally the one using :/
I'll add a width parameter in the config to force it

Alexays:
Option to force width done in 3c66f4baa71e92331a3739f006cfdcf3cccc52a2

Alexays:
Can you confirm that your issue is fixed with last commit (without the configuration width) 8d5b61a9fddc0db9614eeb7e2a952852d36b132f ? :)

izelnakri:
@Alexays there is still a bug in this implementation as my waybar width constantly gets changed on each width change/refresh: 

```
[2023-10-21 20:21:07.720] [info] Bar configured (width: 2198, height: 34) for output: eDP-1
[2023-10-21 20:21:07.730] [info] Bar configured (width: 2214, height: 34) for output: eDP-1
[2023-10-21 20:21:17.700] [info] Bar configured (width: 2189, height: 34) for output: eDP-1
```

I'm running Waybar v0.9.22 , and setting `min-width`, `max-width` or `width` makes the waybar to throw/exit, saying they are not valid CSS property names(definitely confusing).

I dont want to set a fix width on my waybar other than 100%, let me know if I can help in anyway. Styling with CSS seem to be a challenge still for waybar in terms of `width` styling and `:hover` on waybar today and documentation and implementation needs to be improved it seems.

PR #3695, opened 2024-10-17
Fix a crash after handling SIGINT and a data race when initializing the Hyprland workspace modules

cfillion:
## Bug 1: Crash when receiving signals after `main` returns

The waybar process does not exit instantaneously.
Signals may be received after main has started freeing resources.

When a worker thread is in `fgets` this time window can last forever (until Hyprland writes to the socket).
An easy way to duplicate the crash is pressing ^C twice with a Hyprland module waiting on socket2.

    Thread 1 "waybar" received signal SIGSEGV, Segmentation fault.
    spdlog::sinks::sink::should_log (this=0x5f620b542ca5,
        msg_level=spdlog::level::info)
        at /usr/src/debug/spdlog/spdlog-1.14.1/include/spdlog/sinks/sink-inl.h:13
    13	  return msg_level >= level_.load(std::memory_order_relaxed);
    (gdb) p $_siginfo._sifields._sigfault.si_addr
    $1 = (void *) 0x5f620b542cad

## Bug 2: Random crashes and/or Hyprland socket connection failure

`Workspaces::*` and `IPC::startIPC` may both call `getSocketFolder` at the same time and are racing for the same global `socketFolder_`.

This is likely the root cause of #3663.

Typical crash A:

    [2024-10-16 07:42:09.987] [info] Hyprland IPC starting
    malloc(): unaligned tcache chunk detected
    [2024-10-16 07:42:09.987] [error] Hyprland IPC: Unable to connect?
    Thread 1 "waybar" received signal SIGABRT, Aborted.
    (gdb) bt
    #0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
    (omitted for brievety)
    #9  0x00007ffff64ae745 in operator new (sz=sz@entry=296) at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/new_op.cc:50
    #10 0x00007ffff65ab1f1 in std::filesystem::__cxx11::path::_List::_Impl::copy (this=0x555555a23350) at /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++17/fs_path.cc:249
    #11 0x00007ffff65ab3bd in std::filesystem::__cxx11::path::_List::_List (this=0x7fffffff9d30, other=<optimized out>) at /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/unique_ptr.h:454
    #12 0x00005555556f4ab1 in waybar::modules::hyprland::IPC::getSocket1Reply(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
    #13 0x00005555556f5e3d in waybar::modules::hyprland::IPC::getSocket1JsonReply(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
    #14 0x000055555571289c in waybar::modules::hyprland::Workspaces::setCurrentMonitorId() ()

Typical crash B:

    [2024-10-16 10:01:15.859] [info] Hyprland IPC starting
    [2024-10-16 10:01:15.859] [info] Loading persistent workspaces from Hyprland workspace rules
    Thread 8 "waybar" received signal SIGSEGV, Segmentation fault.
    (gdb) bt
    #0  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_S_copy
        (__d=0x5555558fbca8 "/", __s=0x2973961a26d35726 <error: Cannot access memory at address 0x2973961a26d35726>, __n=1)
        at /usr/src/debug/gcc/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:433
    (omitted for brievety)
    #15 waybar::modules::hyprland::IPC::getSocketFolder[abi:cxx11](char const*)
        (instanceSig=0x7fffffffe604 "4520b30d498daca8079365bdb909a8dea38e8d55_1729051218_1982280648") at ../src/modules/hyprland/backend.cpp:41
    #16 0x000055555564230f in waybar::modules::hyprland::IPC::startIPC()::{lambda()#1}::operator()() const ()
        at ../src/modules/hyprland/backend.cpp:70
    #17 0x00007ffff64e1c34 in std::execute_native_thread_routine (__p=0x5555558119c0) at /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/thread.cc:104
    #18 0x00007ffff62a339d in start_thread (arg=<optimized out>) at pthread_create.c:447

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -107,6 +107,10 @@ int main(int argc, char* argv[]) {
       ret = client->main(argc, argv);
     } while (reload);
 
+    std::signal(SIGUSR1, SIG_IGN);
+    std::signal(SIGUSR2, SIG_IGN);
+    std::signal(SIGINT,  SIG_IGN);
+
     delete client;
     return ret;
   } catch (const std::exception& e) {
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -18,6 +18,9 @@ namespace waybar::modules::hyprland {
 std::filesystem::path IPC::socketFolder_;
 
 std::filesystem::path IPC::getSocketFolder(const char* instanceSig) {
+  static std::mutex folderMutex;
+  std::unique_lock lock(folderMutex);
+
   // socket path, specified by EventManager of Hyprland
   if (!socketFolder_.empty()) {
     return socketFolder_;


vpcano:
On Bug 2:
I also believe that was causing issue #3663. After having a look at the return value of `getSocketFolder` function, I realized that the problem was that the function was cacheing the path of the sockets folder on the global variable `socketFolder_`, which was being raced by two different threads.

After commenting the lines:
```cpp
  if (!socketFolder_.empty()) {
    return socketFolder_;
  }
  ```
  the problem seems to have dissappeared, but I believe that the mutex is the most adequate solution. Thanks!

PR #3698, opened 2024-10-18
Add tooltip functionality on hyprland/language module

vpcano:
I added the possibility to show tooltips on hyprland/language module, as shown on the image below.

screenshot

Tooltip is enabled by default and can be disabled using `"tooltip": false` option in the config. Format can be specified in the same way as the module by preceding `tooltip-` on the same configuration keys. Format replacements such as `{short}`, `{long}`, etc. are also valid. For example:

```json
    "hyprland/language": {
        "format": "  {}",
        "format-en": "US",
        "format-es": "ES",
        "tooltip": true,
        "tooltip-format": "{long}",
    },
```

```json
    "hyprland/language": {
        "format": "  {}",
        "format-en": "US",
        "format-es": "ES",
        "tooltip": true,
        "tooltip-format": "{}",
        "tooltip-format-es": "{Español}",
        "tooltip-format-en": "{English (american)}",
    },
```


alexlnkp:
I personally would love this to be merged, but wouldn't that also mean that the same functionality should be added for `sway/language` as well?

PR #3700, opened 2024-10-20
Fix parsing issue in hyprland language module with layouts containing commas

rom4kov:
Problem: Since version 10.4, the hyprland language module had an issue with parsing keyboard layout names that contain commas. For example, the layout "English (intl., with AltGr dead keys)" would not be parsed correctly when switching layouts.

Issue: The IPC event string for a layout like "English (intl., with AltGr dead keys)" received in the onEvent function contains two commas—one separating the keyboard name from the layout name and another within the layout name. However, the function currently splits the string after the last comma, which leads to incorrect parsing.

Solution: This pull request modifies the onEvent function to split the event string at the first comma instead of the last one, allowing for accurate parsing of both the keyboard name and layout name.

### DIFF
--- a/src/modules/hyprland/language.cpp
+++ b/src/modules/hyprland/language.cpp
@@ -69,8 +69,14 @@ auto Language::update() -> void {
 
 void Language::onEvent(const std::string& ev) {
   std::lock_guard<std::mutex> lg(mutex_);
-  std::string kbName(begin(ev) + ev.find_last_of('>') + 1, begin(ev) + ev.find_first_of(','));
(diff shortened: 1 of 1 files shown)


rom4kov:
Hi! I apologize for not checking the existing discussions and the recent pull request https://github.com/Alexays/Waybar/pull/3406 related to the parsing issue with keyboard layout names. I see that there are concerns about the approach I took, particularly with handling embedded commas in both the keyboard name and layout name. Just changing from splitting on the last comma to splitting on the first comma is my solution proposes would not solve the issue for user's with layouts that have commas inside the keyboard name for example. Therefore I'm closing this PR.

PR #3706, opened 2024-10-22
Add an option to search for name in hwmon directory for the temperature module

ico277:
I added a way to specify the hwmon path by using the `name` file in the `/sys/class/hwmon/hwmonX/` and `input-filename`.

It iterates though `/sys/class/hwmon/` and checks if every folder has a `name` file then chooses the first one where the contents match the `hwmon-by-path` config variable.

Example config snippet:
```jsonc
"temperature": {
    "hwmon-by-name": "k10temp",
    "input-filename": "temp1_input",
    "critical-threshold": 75,
    "format-critical": "CRIT:{temperatureC}°C",
    "format": "{temperatureC}°C"
}
```


zen2:
Tips for the module documentation:
hwmon-by-name and input-filename can be found easily with the command
`sensors -uA` and you can have the equivalent in json format with `sensors -juA`.

PR #3723, opened 2024-10-29
Fixed a typo in power_menu.xml and reformat example

sergeyklay:
~Unified `shutdown` and `reboot` commands to use the modern systemd approach, replacing legacy SysVinit-style commands. This aligns the entire menu with the standard for contemporary distributions, enhancing consistency.~

Formatted indentation in `power_menu.xml` for readability, converting mixed spaces and tabs to 2-space indentation for better display on small screens.

Fixed a typo in `power_menu.xml` by correcting `id="hibernat"` to `id="hibernate"`.

### DIFF
--- a/man/waybar-menu.5.scd
+++ b/man/waybar-menu.5.scd
@@ -91,8 +91,8 @@ Module config :
 		"menu": "on-click",
 		"menu-file": "~/.config/waybar/power_menu.xml",
 		"menu-actions": {
-			"shutdown": "shutdown",
-			"reboot": "reboot",
+			"shutdown": "systemctl poweroff",
+			"reboot": "systemctl reboot",
 			"suspend": "systemctl suspend",
 			"hibernate": "systemctl hibernate",
 		},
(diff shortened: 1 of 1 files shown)


LukashonakV:
Not sure systemctl is a standard. It depends on login manager in chain with init manager. For example for non systemd init system like elogind - standard command starts with loginctl. See [elogind](https://wiki.gentoo.org/wiki/Elogind)
This is exact my case

LukashonakV:
> > Not sure systemctl is a standard. It depends on login manager in chain with init manager. For example for non systemd init system like elogind - standard command starts with loginctl. See [elogind](https://wiki.gentoo.org/wiki/Elogind) This is exact my case
> 
> I’m not here to argue about the spread or popularity of systemd. At the very least, I believe those users who manage their systems without systemd can do more than just copy-paste code from a guide — they can adapt it. Either way, my task was to unify the guide. As I mentioned above the goal is to enhancing consistency — half of the examples used systemd, while the other half were in a SysVinit style.

Doesn't existing solution work in systemd? At least out of the box is working well in elogind.  
If it does work well out of the box, both init systems, then such change won't impact systemd users, but will bring one more action to the rest users they should do before module usage. Am I wrong?

sergeyklay:
> If it does work well out of the box, both init systems, then such change won't impact systemd users, but will bring one more action to the rest users they should do before module usage. Am I wrong?

Alright, that sounds reasonable. Your arguments convinced me - I reverted those changes.

Issue #1561, opened 2022-05-18
How to stretch a module to fill the remaining space?


UtkarshVerma:
Currently, modules are shrunk to their content size in waybar. The behaviour I wish is to have the window module fill the remaining space, so that I can style an appropriate background color.

Is it currently possible?

vanarok:
waybar config
`    
{
"width" : 0,
}
`

UtkarshVerma:
![image](https://user-images.githubusercontent.com/31820255/170861022-bdec4e57-0456-4e78-a92f-9f99b2810d46.png)

I hope this image clarifies what I wish to achieve. If you look at the image, I have styled the window module with red background. As you can see, it doesn't fill the entire space, there is some unused space on the left of the status bar.

If you're wondering why I want this, I wish to have a different background colour for the middle region when a window is focused. Something like this:

![](https://github.com/UtkarshVerma/dwmblocks-async/blob/main/preview.png?raw=true)

Alexays:
Good news — this is possible now with the `expand` option. Set it on the module you want to grow into the leftover space:

```json
"sway/window": {
    "expand": true
}
```

From the man page: *"Enables this module to consume all left over space dynamically."* So the window module will stretch across the empty area and you can give it its own background in `style.css`. If your build predates this, just update to a recent release.

Closing as answered, but feel free to reopen if it doesn't behave as expected 🙏

PR #3730, opened 2024-10-30
Add stretching of modules and  modules-center toggling

TripleTrable:
This Patch allows the stretching of modules-{left,center,right} as well add a "expand" flag to AModule. This allows one module to consume the leftover space.
To allow the left or right modules to fully consume the center, the changes also include a way to remove the center box (center_) altogether.

This is useful for use-cases like: https://github.com/Alexays/Waybar/issues/1561

### DIFF
--- a/src/bar.cpp
+++ b/src/bar.cpp
@@ -534,13 +534,22 @@ void waybar::Bar::getModules(const Factory& factory, const std::string& pos,
 
 auto waybar::Bar::setupWidgets() -> void {
   window.add(box_);
-  box_.pack_start(left_, false, false);
-  if (config["fixed-center"].isBool() ? config["fixed-center"].asBool() : true) {
-    box_.set_center_widget(center_);
-  } else {
-    box_.pack_start(center_, true, false);
+
+  bool expand_left = config["expand-left"].isBool() ? config["expand-left"].asBool() : false;
+  bool expand_center = config["expand-center"].isBool() ? config["expand-center"].asBool() : false;
+  bool expand_right = config["expand-right"].isBool() ? config["expand-right"].asBool() : false;
+  bool no_center = config["no-center"].isBool() ? config["no-center"].asBool() : false;
+
+  box_.pack_start(left_, expand_left, expand_left);
+  if (!no_center) {
+    if (config["fixed-center"].isBool() ? config["fixed-center"].asBool() : true) {
+      box_.set_center_widget(center_);
+    } else {
+      spdlog::error("No fixed center_");
+      box_.pack_start(center_, true, expand_center);
+    }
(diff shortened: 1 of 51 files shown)


TripleTrable:
Any Interest in looking at this? The 51 come from touching nearly all man pages. 
One point to discuss, should this PR be considered, would be whether the script should execute before or after the format resolving.

src/bar.cpp
```diff
@@ -534,13 +534,22 @@ void waybar::Bar::getModules(const Factory& factory, const std::string& pos,
...
+  bool expand_center = config["expand-center"].isBool() ? config["expand-center"].asBool() : false;
+  bool expand_right = config["expand-right"].isBool() ? config["expand-right"].asBool() : false;
+  bool no_center = config["no-center"].isBool() ? config["no-center"].asBool() : false;
+
+  box_.pack_start(left_, expand_left, expand_left);
+  if (!no_center) {
+    if (config["fixed-center"].isBool() ? config["fixed-center"].asBool() : true) {
+      box_.set_center_widget(center_);
+    } else {
+      spdlog::error("No fixed center_");
```
tmccombs, line 548: Is `fixed-center: false` an error now? 

I think this changes the behavior.  Since `expand-center` defaults to false, if "fixed-center" is false, then the center box won't expand anymore, and will just be on the left.

src/bar.cpp
```diff
@@ -549,14 +558,21 @@ auto waybar::Bar::setupWidgets() -> void {
...
   getModules(factory, "modules-left");
-  getModules(factory, "modules-center");
+  if (!no_center) {
+    getModules(factory, "modules-center");
+  }
   getModules(factory, "modules-right");
+
   for (auto const& module : modules_left_) {
-    left_.pack_start(*module, false, false);
+    left_.pack_start(*module, module->expandEnabled(), module->expandEnabled());
```
tmccombs, line 567: @TripleTrable  why is expanding enabled for modules in the left, but not in the center or right? Is that just an oversight?

TripleTrable: sorry for the late reply. 
I am currently in the exam face of the University. Once I'm done with my exams I extend the functionality to the other module blocks.

This was an oversight and of course should be done for all three packs.

Thanks for noticing.

TripleTrable: I have opened a fix:
https://github.com/Alexays/Waybar/pull/3881
Do you see any other problems?

src/bar.cpp
TripleTrable: I have opened a fix:
https://github.com/Alexays/Waybar/pull/3881
Do you see any other problems?

Issue #3746, opened 2024-11-05
 Hyprland - Window module doesn't change to fullscreen style if a stacked window group is on fullscreen 


izmyname:
I have a  `window#waybar.fullscreen #window ` in my `style.css` configured in the way it changes colours, whenever a focused window is on fullscreen. However, it doesn't change its style if I put a stacked window group on fullscreen.

To demonstrate.

1. default window style, no fullscreen.
![20241106_01h27m03s_grim](https://github.com/user-attachments/assets/7672ea06-a8dc-4ff2-bdfa-59c432444d82)

2. firefox window is on fullscreen - the module changed its colours.
![20241106_01h27m32s_grim](https://github.com/user-attachments/assets/e74005db-fb7a-4a89-b34b-0af5d5d4bd8e)

3. firefox is inside a stacked group, the group is on fullscreen - the module didn't change its colours.
![20241106_01h29m44s_grim](https://github.com/user-attachments/assets/c67ef2e4-1dd3-4b3d-ba26-a0a88d3211fa)

izmyname:
Some more info from Hyprland discussions.  https://github.com/hyprwm/Hyprland/discussions/8369#discussioncomment-11165849

Aqa-Ib:
Having a quick look at waybar's code, it seems that the issue is here:
https://github.com/Alexays/Waybar/blob/5f260516af9a3e5791c26c7fc18a85a47da5b23f/src/modules/hyprland/window.cpp#L194-L198

izmyname:
Removing the code above, before building, fixes the problem.

PR #3747, opened 2024-11-06
fix hyprland's grouped window flags

Aqa-Ib:
Both flags are wrong, because:
- the active group member can be fullscreened.
- technically, a grouped window can be solo as well, because only the active group member is shown, the other members are hidden. Also, you can have a group consisting of only one member.

Fixes https://github.com/Alexays/Waybar/issues/3746


Alexays:
Thanks!

PR #3793, opened 2024-11-25
fix mpd module hang after suspend

bretello:
I've been having this annoying issue for a while, in which after waking up from a _long_ suspend the MPD module is not updating anymore.

Clicking the MPD module icon does trigger play/pause events but it does not show the active song anymore.

I think I finally manage to track it down to a broken idle connection, and this PR fixes the issue.

# How to reproduce

Run:

```bash
systemctl suspend
```

go get a beer, come back after a lot of time, and the MPD module will not update anymore.

If you don't have a lot of time, I managed to reproduce similar symptoms by killing the socket(s) using `ss`:

```bash
export MPD_HOST=<mpd host>
export MPD_PORT=<mpd port>
sudo ss -k "dst $MPD_HOST and dport $MPD_PORT" # WARNING: this will kill ALL connections to MPD_HOST:MPD_PORT
```

The mpd module will not update anymore, even though it responds to clicks.

### DIFF
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -50,7 +50,7 @@ inline int close(FILE* fp, pid_t pid) {
     ret = waitpid(pid, &stat, WCONTINUED | WUNTRACED);
 
     if (WIFEXITED(stat)) {
-      spdlog::debug("Cmd exited with code {}", WEXITSTATUS(stat));
+      spdlog::trace("Cmd exited with code {}", WEXITSTATUS(stat));
     } else if (WIFSIGNALED(stat)) {
       spdlog::debug("Cmd killed by {}", WTERMSIG(stat));
     } else if (WIFSTOPPED(stat)) {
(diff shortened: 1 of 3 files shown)


include/util/command.hpp
```diff
@@ -50,7 +50,7 @@ inline int close(FILE* fp, pid_t pid) {
     ret = waitpid(pid, &stat, WCONTINUED | WUNTRACED);
 
     if (WIFEXITED(stat)) {
-      spdlog::debug("Cmd exited with code {}", WEXITSTATUS(stat));
+      spdlog::trace("Cmd exited with code {}", WEXITSTATUS(stat));
```
bretello, line 53: I changed this (and below) to trace since they made debugging  with `-l debug` very noisy. I'll be happy to revert this and/or open another PR in case we want separation of concerns.

bretello:
Seems this is not fixing the issue, will keep investigating.

bretello:
Update: this seemingly has to to with `libmpdclient`: running `mpc idleloop` shows the same problem after a long suspend. It looks like the underlying connection dies but the client never realizes this and keeps waiting for idle events.

bretello:
Still has some hanging problems updating after a wake from sleep.

Issue #3020, opened 2024-03-17
sway/workspaces window-rewrite does not support Xwayland


mliszcz:
It is not possible to map X11 applications to icons using the window-rewrite mechanism. The implementation currently checks only for the app_id to match against the `<class>` pattern.

https://github.com/Alexays/Waybar/blob/ca17bdb214d7337c70d4af6f8f935748260a72cc/src/modules/sway/workspaces.cpp#L272

Sway does not provide this for X11 applications running with Xwayland. Instead we could use node["window_properties"]["class"] and node["window_properties"]["instance"] to determine the application.

I could send a PR if it would be accepted, but I'm not sure what would be the preferred way to solve this.
* for example: `<class>` could fall back to checking window_properties.class if app_id is not provided,
* or we could add new patterns like `<x11class>` (also it would be best to rename `<class>` to `<app_id>` to avoid any confusion but I guess it must stay for backwards compatibility).

alxdb:
Having this same issue and would also be interested in a fix. Can help contribute if I find the time

euglevi:
I think this fix would be very useful. Sorry that I do not know enough about coding to help.

cobbles:
Am also running into this problem, and for me it is not picking up the `window-rewrite-default` as the fallback so the impact is it looks like no windows are open on the workspace.

For me the application that does not show an icon is Code - OSS

output from get_tree:

```
➜ ~ swaymsg -t get_tree
    #6: workspace "2"
      #16: con "config - .dotfiles - Code - OSS" (xwayland, pid: 6049, instance: "code-oss", class: "code-oss", X11 window: 0x400004)
      #14: con "swaymsg -t get_tree" (xdg_shell, pid: 4348, app_id: "Alacritty")
```

euglevi:
For those who are interested, the workaround at the moment for me is using utilities that wrap around waybar to correctly show all icons. More specifically, I am using [Sworkstyle](https://github.com/Lyr-7D1h/swayest_workstyle).

PR #3801, opened 2024-12-01
sway/workspaces: use X11 class for XWayland windows

majutsushi:
When using `window-rewrite`, the `class<>` rule would previously only match against the `app_id` of a window. However, XWayland windows don't have an app ID, resulting in an empty string for the rewritten name.

This change falls back to checking the `class` window property if there is no app ID to support matching against XWayland windows.

Closes: #3020


Alexays:
Thanks!

PR #3806, opened 2024-12-03
Add locking for interacting with the pulseaudio thread

### LINKED ISSUE #3773: Sometimes, under complex microarchitectural conditions, Waybar gets into a state where it consumes 100% of a CPU core until I restart it
I have decided to actually try to debug it this time. I have it `SIGSTOP`'d awaiting further analysis. I have collected a `perf` trace and it shows that the CPU time is being spent in `libpulsecommon-17.0.so`'s `pa_read` function.

Here is the relevant part of the configuration file:
```
    "pulseaudio": {
        // "scroll-step": 1, // %, can be a float
        "format": "{volume}% {icon} {format_source}",
        "format-bluetooth": "{volume}% {icon} {format_source}",
        "format-bluetooth-muted": " {icon} {format_source}",
        "format-muted": " {format_source}",
        "format-source": "{volume}% ",
        "format-source-muted": "",
        "format-icons": {
            "headphone": "",
            "hands-free": "",
            "headset": "",
            "phone": "",
            "portable": "",
            "car": "",
            "default": ["", "", ""]
        },
        "on-click": "pavucontrol"
    },
```

Possibly relevant versions:

```
~$ pacman -Q waybar libpulse pipewire-pulse pipewire
waybar 0.11.0-3
libpulse 17.0-3
pipewire-pulse 1:1.2.6-1
pipewire 1:1.2.6-1
```

Let me know how you'd like me to poke the process to try and debug this further.

meithecatte:
Before this commit, Waybar would sometimes get into a state where it would consume 100% of a CPU core, and the pulseaudio widget would stop responding to volume adjustments.

In this state, the pulseaudio mainloop thread would spin, with the counter of enabled defer events at 1, but no actual enabled defer event in the list to get the counter back to zero after an iteration in the mainloop.

This could happen if the unsynchronized interactions with the mainloop thread happened to modify the list of deferred events at the same time as the mainloop.

This commit introduces locking in accordance with the PulseAudio documentation on the threaded mainloop:

> The lock needs to be held whenever you call any PulseAudio function that
> uses an object associated with this main loop. Those objects include
> pa_mainloop, pa_context, pa_stream and pa_operation, and the various event
> objects (pa_io_event, pa_time_event, pa_defer_event).

In my testing, this fixes #3773

### DIFF
--- a/src/util/audio_backend.cpp
+++ b/src/util/audio_backend.cpp
@@ -236,7 +236,9 @@ void AudioBackend::changeVolume(uint16_t volume, uint16_t min_volume, uint16_t m
   volume = std::clamp(volume, min_volume, max_volume);
   pa_cvolume_set(&pa_volume, pa_volume_.channels, volume * volume_tick);
 
+  pa_threaded_mainloop_lock(mainloop_);
   pa_context_set_sink_volume_by_index(context_, sink_idx_, &pa_volume, volumeModifyCb, this);
+  pa_threaded_mainloop_unlock(mainloop_);
 }
 
 void AudioBackend::changeVolume(ChangeType change_type, double step, uint16_t max_volume) {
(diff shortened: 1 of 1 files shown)


Alexays:
LGTM, thx!

PR #3808, opened 2024-12-04
Using 'auto' instead of 'self' in backlight module

### LINKED ISSUE #3302: CRITICAL : unhandled exception (type Glib::Error) in signal handler  g-dbus-error-quark

Nambers:
re-implementation of #3519 
fix #3302 

It works on my Arch Linux but I didn't test it comprehensively.


Alexays:
lgtm

PR #3809, opened 2024-12-05
Let network module handle ipv4 and ipv6 simultaneously

gustafullberg:
I often switch between networks that use either IPv4 or IPv6.

As the network module supports IPv4 _or_ IPv6 (but not both at the same time) I currently need to add two instances of the network module to my config and explicitly set the family to "ipv6" for one of them.

This PR makes the network module able to handle both IPv4 and IPv6 simultaneously. It is enough to have a single instance of the network module that handle both protocols and the family doesn't have to be specified. This is a bit simpler and probably what most users would expect.

### DIFF
--- a/include/modules/network.hpp
+++ b/include/modules/network.hpp
@@ -50,7 +50,6 @@ class Network : public ALabel {
   std::optional<std::pair<unsigned long long, unsigned long long>> readBandwidthUsage();
 
   int ifid_;
-  sa_family_t family_;
   struct sockaddr_nl nladdr_ = {0};
   struct nl_sock* sock_ = nullptr;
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks!

voiceroy:
I think this might have caused some bug, now I'm only shown ipv6 address by default. Even after me setting `family` to `ipv4`, I get the `ipv6` address shown to me.
screenshot_19:10:46 24-02-2025

NBAB42Bq:
Same problem here after update

MRSS02:
Same problem here after update.

(Arch Linux, Sway)

efraletti:
Same problem here after update.

(Arch Linux, Sway)

Issue #3790, opened 2024-11-25
[Feat] Add option to show fewer tags on river module.


hfytr:
Hello, I would like an option in the river/tags module which displays only tags which have views assigned to them. I would not be opposed to writing this myself.

Please confirm that such an option would be merged before I do so, as if not I will not write / use it (updating  / building from source becomes too much hassle for me). Fair warning: I am not a cpp dev, and so the code will likely be pretty bad initially.

Alexays:
Closing — this is fixed/implemented on current master and will be in the next release. Please reopen if it still occurs after updating. 🙏

Issue #2537, opened 2023-09-25
Is it possible to fully hide a class?


Co3co:
Is there a way completely hide all unoccupied tag buttons for river/tags? What I want to do is basically the same thing that the [hide vacant tags](https://dwm.suckless.org/patches/hide_vacant_tags/) dwm patch does.
I have tried setting these properties:
```
font-size: 0;
min-width: 0;
min-height: 0;
margin: 0;
padding: 0;
border: 0;
opacity: 0;
background-color: transparent;
```
But they do not fully hide the tags. I still see blank space where the tags should be, and I'm still able to click on the unoccupied tags to focus them. I also tried `display: none;` and `visibility: hidden;`, but they were not valid properties.

DreamMaoMao:
same problem ,I want to hide some tags(no window),but `visibility: hidden` did't work.

LanderMoerkerke:
A workaround can be found here: https://github.com/Alexays/Waybar/issues/2721

Alexays:
Closing — this was resolved/answered in the thread (config, support, or environment, not a Waybar bug still open). Feel free to reopen if you're still stuck. 🙏

PR #3823, opened 2024-12-12
Hide vacant tags config option on River

thetwistedlogic:
Adds the ability to hide vacant tags on river. Vacant means not occupied and not focused.
image
image

Addresses:
https://github.com/Alexays/Waybar/issues/3790
https://github.com/Alexays/Waybar/issues/2537

### DIFF
--- a/src/modules/river/tags.cpp
+++ b/src/modules/river/tags.cpp
@@ -220,7 +220,7 @@ void Tags::handle_view_tags(struct wl_array *view_tags) {
       }
       buttons_[i].get_style_context()->add_class("occupied");
     } else {
-      if (hide_vacant) {
+      if (hide_vacant && !buttons_[i].get_style_context()->has_class("focused")) {
         buttons_[i].set_visible(false);
       }
       buttons_[i].get_style_context()->remove_class("occupied");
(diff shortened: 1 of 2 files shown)


src/modules/river/tags.cpp
```diff
@@ -189,11 +189,18 @@ bool Tags::handle_button_press(GdkEventButton *event_button, uint32_t tag) {
...
+  auto hide_vacant = config_["hide-vacant"].asBool();
   for (size_t i = 0; i < buttons_.size(); ++i) {
     if ((1 << i) & tags) {
+      if (hide_vacant) {
+        buttons_[i].set_visible(true);
+      }
       buttons_[i].get_style_context()->add_class("focused");
     } else {
       buttons_[i].get_style_context()->remove_class("focused");
+      if (hide_vacant && !buttons_[i].get_style_context()->has_class("occupied")) {
```
thetwistedlogic, line 201: This is a bit hacky but I'm not sure of a better solution. 

Without this extra condition, tags will remain permanently visible as long as they have been focused.

Alexays:
LGTM, thx!

bitterhalt:
There seems to be small bug.  I see empty tags, if I enable "river/mode" or "river/window" modes.

Issue #2693, opened 2023-11-26
constant crashing from hyprland window module


Hellzbellz123:
i use hyprland git, and the last few days waybar has been crashing for me pretty randomly

the issue i am getting is this
[2023-11-25 19:55:21.419] [error] Hyprland IPC: Couldn't write (4)
waybar: ../waybar-git/src/modules/hyprland/window.cpp:138: void waybar::modules::hyprland::Window::queryActiveWorkspace(): Assertion `clients.isArray()' failed.

Waybar v0.9.24-30-g061c68ce (branch 'master')
Hyprland, built from branch main at commit 646f4bc63861e03b01fed3833bb65d1d014e1bf3 dirty (general: add workspace gaps (3877)).
Tag: v0.32.3-40-g646f4bc6

DreamMaoMao:
https://github.com/Alexays/Waybar/pull/2789

I think this problem should be the same problem as this one, trying to read data being written by another asynchronous thread

DreamMaoMao:
https://github.com/Alexays/Waybar/pull/2789

I believe they are the same problem

The original thread lock seems to have some problems, I will try to use shared_lock to fix

PR #3826, opened 2024-12-14
fix: hyprland/window get empty ipc json data

DreamMaoMao:
fix:

https://github.com/Alexays/Waybar/issues/2693

i think it is the same problem with https://github.com/Alexays/Waybar/pull/2789


DreamMaoMao:
It's working well for me, it hasn't crashed yet, I think it should have been fixed, maybe I'll test it for a while, it would be nice to have someone else to test it with

Alexays:
Thanks!

PR #3827, opened 2024-12-14
Escape markup characters in dwl/window

pope:
Without this, markup characters like [&><] will be injected directly into the Label. Escaping them makes sure that the values will be printed exactly as they appear in the window title or layout symbol.


Alexays:
Thanks!

PR #3831, opened 2024-12-17
sway: fix "Mapping is not an object" warning

### LINKED ISSUE #3763: Warning "Mapping is not an object"
```
[2024-11-13 10:50:58.164] [warning] Mapping is not an object
```

This warning appears when starting Waybar without defining the object `window-rewrite` in the configuration.

How to reproduce:
```
{
    "modules-left": ["sway/workspaces"],
    "sway/workspaces": {
      // "window-rewrite": {}
   }
}
```

Uncommenting `window-rewrite` silence this warning. I guess this value does not have a default and when it's checked [here](https://github.com/Alexays/Waybar/blob/52d582b93ad37f216df4eebe4cac40fb86852778/src/util/regex_collection.cpp#L15) it returns the warning.

Could it help assigning a default when the config is loaded (maybe [here](https://github.com/Alexays/Waybar/blob/52d582b93ad37f216df4eebe4cac40fb86852778/src/modules/sway/workspaces.cpp#L64))?

Thanks

c4rlo:
Fixes #3763.

Also a little code simplications while we're at it.

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -49,7 +49,6 @@ class Workspaces : public AModule, public sigc::trackable {
   std::vector<std::string> workspaces_order_;
   Gtk::Box box_;
   std::string m_formatWindowSeperator;
-  std::string m_windowRewriteDefault;
   util::RegexCollection m_windowRewriteRules;
   util::JsonParser parser_;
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks!

apiraino:
thanks @c4rlo !

Issue #3369, opened 2024-06-17
Sway workspace option "current-only" is empty when a window is active


ebeem:
I have the below configuration for my `sway/workspaces`.
My intention is to display current active workspace from all outputs.
I expect that there is always an active workspace (which indeed what sway shows while querying info).

```json
    "sway/workspaces": {
      "all-outputs": true,
      "current-only": true,
    },
```

The issue I am facing is that the workspace names shows up ONLY if there are no active windows in the workspace.

For example, I have two workspaces 
```
workspace#1: firefox running
workspace#2: empty
```

If I switch to `workspace#1`, waybar displays workspace empty (because firefox is focused).
If I switch to `workspace#2`, waybar displays actual workspace name (because no window is focused).

Am I missing something?

ebeem:
Was going to fix this, I built the git version and it seems like it's fixed already, just not released.
I am closing.

PR #3847, opened 2024-12-26
sway/workspaces: Made the focused check in "current-only" fully recursive

bruceblore:
I was experiencing a similar version of issue #3369. However, in my case, the issue only kicked in when a tabbed container was in use. I noticed that PR #3336 attempted to address the issues with current-only by not only checking whether the workspace itself is focused, but also whether any window within it is focused.

However, it did not check recursively. This meant that if the focused window was at the top level, it would be detected. Or if it was in a container that sway decided to also indicate as focused. (It's really not clear to me how Sway decides when to indicate a container as focused.) But if the window is in a container not indicated as focused, it will be missed. This seems to happen often with tabbed containers. Changing this check to be recursive seems to fix it.

### DIFF
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -494,14 +494,18 @@ std::string Workspaces::trimWorkspaceName(std::string name) {
   return name;
 }
 
+bool checkFocused(const Json::Value &node) {
+  return node["focused"].asBool() ||
+         std::any_of(node["nodes"].begin(), node["nodes"].end(),
+                     [](const auto &child) { return checkFocused(child); });
+}
+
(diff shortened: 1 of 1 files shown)


ebeem:
I actually noticed that as well, a fix was released earlier to check for focused nodes only one level. This has to be recursive. I just created a PR as well #4046, I think your fix is more concise. So if this gets merged please close #4046.
I also explained in my PR the problem, sway allows nesting nodes. So if the node is nested or floating, waybar will fail to display it, looking at your code I am not seeing a way to handle floating nodes. This will probably cause the workspace to be invisible again if the focused window is floating (haven't tested). Please confirm this and ensure you do the same nested check on `floating_nodes` as well so we fix all scenarios that might cause this issue.

Thanks

Issue #3631, opened 2024-09-22
implment waybar-mpd {title} // {filename}


vincentaxhe:
I have many song files, many of their metas is not recoginized by waybar-mpd, I edit their metas by foobar and show good in daphile system.
Just implment {title} // {filename}, when title is not avaiable, fall to striped filename.

krei-se:
+1 from me - mpc shows the filename when no tag is present - should be doable?!

Alexays:
Implemented on master (PR #3849): the mpd module now exposes `{filename}` and `{uri}`, so you can build a format like `{title} // {filename}` to fall back to the filename when tags are missing. See waybar-mpd(5). Closing as resolved.

PR #3849, opened 2024-12-28
mpd: add uri and filename format arguments

haansn08:
Many people have files in their music library that are not properly tagged and for such files `{title}` and `{album}` cannot be displayed by the mpd module. As a fallback the user might want to display filename or the full URI of this file instead ( see https://github.com/Alexays/Waybar/issues/3631 ).

This PR
* enables the `{filename}` format replacement to be used in tooltips (previously only available in the label, for some reason)
* documents the `{filename}` format replacement in the man page
* adds an `{uri}` format replacement which takes the value of [`mpd_song_get_uri`](https://www.musicpd.org/doc/libmpdclient/song_8h.html#a969dbd00e70ee44b573eb2feb0fb71b3)

While this does not allow to display the filename conditionally on the title being empty (format strings like `{title|filename}` don't seem to be allowed), at least one can now display the URI of the file in the tooltip and get a hint which file is missing proper tagging.

### DIFF
--- a/man/waybar-mpd.5.scd
+++ b/man/waybar-mpd.5.scd
@@ -204,6 +204,10 @@ Addressed by *mpd*
 
 *{queueLength}*: The length of the current queue.
 
+*{uri}*: The URI of the song relative to the MPD music directory.
+
+*{filename}* The last part of the URI.
+
 *{stateIcon}*: The icon corresponding to the playing or paused status of the player (see *state-icons* option)
 
 *{consumeIcon}*: The icon corresponding the "consume" option (see *consume-icons* option)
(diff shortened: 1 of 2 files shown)


ckiee:
Patch to add a `{titleOrFilename}` option, for whoever's interested..
https://github.com/ckiee/nixfiles/tree/c989a4593ba65c41d6bb0588f8b6b47b46913fac/modules/home/waybar/0001-feat-mpd-add-titleOrFilename.patch

mohammedahmed18:
you can use a script to pass the filename to the format field

```json
"mpd": {
    "format": "{stateIcon} {filename} ({elapsedTime:%M:%S}/{totalTime:%M:%S}) 🎵",
    "exec": "./scripts/mpd.sh",
    "interval": 1,
    "state-icons": {
        "playing": "",
        "paused": "",
        "stopped": ""
    }
```

and for the mpd.sh

```bash
current_played_file="$(mpc --format \"%file%\" current)"
echo -e "{\"filename\": $current_played_file}"
```

PR #3857, opened 2024-12-31
Gtk4

tmccombs:
This is based off of https://github.com/LukashonakV/Waybar/tree/gtk4 (see also https://github.com/Alexays/Waybar/pull/2956), but adds back suport for the general `menu` configuration, using a `Gio::MenuModel` and `Gtk::PopoverMenu` instead of a `Gtk::Menu`.


tmccombs:
Changed the base to master, because the gtk4 branch on this repo is very out of date. 

I also merged current master into the branch. 

For some reason I can't make a PR against the LukashonakV repo.

src/modules/sndio.cpp
```diff
@@ -140,27 +135,27 @@ auto Sndio::put_val(unsigned int addr, unsigned int val) -> void {
...
 
-  auto dir = AModule::getScrollDir(e);
+  auto dir{AModule::getScrollDir(controllScroll_->get_current_event())};
   if (dir == SCROLL_DIR::NONE) {
     return true;
   }
 
-  int step = 5;
+  int step{5};
   if (config_["scroll-step"].isInt()) {
```
tmccombs, line 154: Does it ever make sense for this to be negative? If not, we could read this as a UInt, and avoid some casting

PR #3862, opened 2025-01-03
Allow using wildcards in config include paths

emar10:
Since Waybar supports including additional configuration files using the `"include"` key, I expected to be able to use wildcards in the paths passed in. For example, a drop-in config directory:

```jsonc
{
  "include": "/etc/waybar/config.d/*.jsonc"
}
```

Currently, this *almost* works. `Config::tryExpandPath()` already expands wildcards using `wordexp()`, but only takes the first result and discards the rest.

This PR updates `Config::tryExpandPath()` to return a vector of expanded path matches instead of a single path wrapped in an optional, with an empty vector indicating no matches.

`Config::resolveConfigIncludes()` iterates over all of these matches, while other instances of path expansion (such as finding the base config path) retain their existing behavior and only use the first match.

### DIFF
--- a/include/config.hpp
+++ b/include/config.hpp
@@ -20,8 +20,8 @@ class Config {
   static std::optional<std::string> findConfigPath(
       const std::vector<std::string> &names, const std::vector<std::string> &dirs = CONFIG_DIRS);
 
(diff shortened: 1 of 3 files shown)


Alexays:
LGTM, thx!

PR #3863, opened 2025-01-03
add module wayfire/window, wayfire/workspaces

yamader:
I implemented Wayfire integration (especially for workspaces and window) since it did not seem to exist.
I have confirmed that it works in my environment, but please let me know if there is anything else I can do.

### DIFF
--- a/include/modules/wayfire/backend.hpp
+++ b/include/modules/wayfire/backend.hpp
@@ -1,0 +1,122 @@
+#pragma once
+
+#include <json/json.h>
+#include <unistd.h>
+
+#include <functional>
+#include <list>
+#include <memory>
+#include <mutex>
+#include <optional>
+#include <string>
+#include <unordered_map>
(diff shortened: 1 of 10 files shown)


bluebyt:
I just find these modules for wayfire, that should be include in the waybar wiki. Do you want me to do it?
Both modules are working great, maybe the last one missing is the language module, now I rely to the wayfire IPC interface with a special script.

yamader:
@bluebyt
Thank you for your comment.
I'm currently busy with school and can't maintain Waybar, so I would appreciate it if you could add it to the Wiki and README.
Since we CJK people use IME, I wasn't even aware of the concept of language modules ;)
If I have some free time and it hasn't been implemented yet, I'll try to contribute.

bluebyt:
It's done, thank you!
https://github.com/Alexays/Waybar/wiki/Module:-Wayfire

PR #3876, opened 2025-01-10
fix: Make signal handling safer

tmccombs:
std::signal is very restrictive about what can be done inside the signal handler. Posix adds some additional functions that are safe to call, but it is still pretty limited.

This changes waybar to use the "self pipe trick" where all the signal handler does is write a byte into a pipe, and then we have a slot run on the Gtk main loop that reads from that pipe and actually handles the signal.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,7 +1,10 @@
+#include <fcntl.h>
 #include <spdlog/spdlog.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <unistd.h>
 
+#include <cerrno>
 #include <csignal>
 #include <list>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks for tackling signal safety! Since this PR was opened, master already landed an equivalent (and more complete) fix via #4048 "Fix signal safety deadlock" — it moves signal handling to the main thread using the same async-signal-safe self-pipe technique (`writeSignalToPipe`/`catchSignals`/`handleSignalMainThread` marshalled through `SafeSignal<int>`), and additionally handles SIGCHLD child reaping and the user KillSignalAction paths. The two approaches are mutually exclusive re-implementations of the same mechanism, so this can't be merged on top without reverting the newer work. Closing as superseded by #4048 — but if you think the `Glib::signal_io` approach has a concrete advantage over the current implementation, please reopen with details and we'll take a look. 🙏

PR #3877, opened 2025-01-10
Escape tray tooltip text

pol-rivero:
Fix errors when the tooltip set by a tray app contains markup characters.

For example, [CopyQ](https://github.com/hluk/CopyQ) sets its tooltip to the last copied text, which could contain arbitrary characters, some of which (like `< > &`) can cause an error when interpreted.


Alexays:
Thanks!

Issue #3830, opened 2024-12-15
Race condition when Hyprland workspaces are renamed after creation.


matt-fff:
When a Hyprland workspace is created, it will [become orphaned](https://github.com/Alexays/Waybar/blob/e959f1d2300c5f7fcc818630c9eae59d834024ce/src/modules/hyprland/workspaces.cpp#L388) if it's renamed before [this call](https://github.com/Alexays/Waybar/blob/e959f1d2300c5f7fcc818630c9eae59d834024ce/src/modules/hyprland/workspaces.cpp#L368) to retrieve the list of workspaces from Hyprland. The workspaces are matched to monitors by name, not ID. So the newly created workspaces don't appear in the Waybar.

This is common when Waybar is used in conjunction with a tool like [hyprland-autoname-workspaces](https://github.com/hyprland-community/hyprland-autoname-workspaces).

I have a [fix I'm using locally](https://github.com/Alexays/Waybar/pull/3829), which just falls back to trying to match against the ID. Though it feels a bit hacky to me - it would be better if it was possible to extract the ID of the created workspace from the Hyprland event, itself. Then the match would be safe regardless of renaming.

matt-fff:
Looking into this more deeply, it looks like the correct approach is to update to the Hyprland v2 events, which [add support for workspace ID](https://github.com/hyprwm/Hyprland/pull/5022). Currently Hyprland sends both types.

I don't want to step on anyone's toes, but it doesn't look to me like anyone is working on this; I'm going to get a change setup to migrate the events to the updated format.

matt-fff:
Put together https://github.com/Alexays/Waybar/pull/3878

I plan on using my branch for a little while to see if I notice any issues.

PR #3878, opened 2025-01-10
Migrate Hyprland workspace events to v2

matt-fff:
The Hyprland [v2 workspace events](https://github.com/hyprwm/Hyprland/pull/5022) add the workspace ID to the event payload.

Currently Waybar relies on workspace names because that was the best the old events provided. By transitioning to IDs instead of names, Waybar can provide a more robust experience, avoid hairy issues like https://github.com/Alexays/Waybar/issues/3830.

At some point I'd like to follow up with a similar change in special workspaces, but `activespecialv2` doesn't exist yet in Hyprland.

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -322,7 +322,7 @@ void Workspaces::onEvent(const std::string &ev) {
     onSpecialWorkspaceActivated(payload);
   } else if (eventName == "destroyworkspace") {
     onWorkspaceDestroyed(payload);
-  } else if (eventName == "createworkspace") {
+  } else if (eventName == "createworkspacev2") {
     onWorkspaceCreated(payload);
   } else if (eventName == "focusedmon") {
     onMonitorFocused(payload);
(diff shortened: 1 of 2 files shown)


src/modules/hyprland/workspaces.cpp
```diff
@@ -316,13 +317,13 @@ void Workspaces::onEvent(const std::string &ev) {
...
   } else if (eventName == "activespecial") {
     onSpecialWorkspaceActivated(payload);
-  } else if (eventName == "destroyworkspace") {
+  } else if (eventName == "destroyworkspacev2") {
     onWorkspaceDestroyed(payload);
-  } else if (eventName == "createworkspace") {
+  } else if (eventName == "createworkspacev2") {
     onWorkspaceCreated(payload);
   } else if (eventName == "focusedmon") {
     onMonitorFocused(payload);
```
zjeffer, line 329: According to [the wiki](https://wiki.hyprland.org/IPC/), focusedmonv2, moveworkspacev2 and movewindowv2 are available. Perhaps you can upgrade those as well?

src/modules/hyprland/workspaces.cpp
zjeffer, line 1: general comment: can you add try/catches around the `std::stoi` calls? If those throw, they would kill the module (or the whole bar? not sure), and the exception messages for failing stoi calls are always very unclear.

src/modules/hyprland/workspaces.cpp
```diff
@@ -682,17 +699,40 @@ auto Workspaces::registerIpc() -> void {
...
+    }
+  }
+
   auto workspace =
       std::find_if(m_workspaces.begin(), m_workspaces.end(), [&](std::unique_ptr<Workspace> &x) {
-        return (name.starts_with("special:") && name.substr(8) == x->name()) || name == x->name();
+        if (name.empty()) {
+          return id == x->id();
+        }
+        return name == x->name();
```
zjeffer, line 735: Possible small improvement: make the `id` on line 711 an `std::optional`, and then check in `find_if` if that optional has value, instead of checking if the name is empty. I feel like that makes the intent clearer that the name is used as a kind of 'fallback' if the stoi fails. (but I might be nitpicking here)

src/modules/hyprland/workspaces.cpp
zjeffer: and maybe add `const` to the temporary variables you create if you don't modify or move them

zjeffer (commented):
mostly nitpicking ;)

src/modules/hyprland/workspaces.cpp
matt-fff: No problem. 

* Added `const` to the appropriate variables.
* put all the `stoi` calls in a `parseWorkspaceId` function that catches and logs the exception

src/modules/hyprland/workspaces.cpp
```diff
@@ -432,7 +440,16 @@ void Workspaces::onWorkspaceRenamed(std::string const &payload) {
...
+
+  std::string workspaceName = payload.substr(payload.find(',') + 1);
+
+  // TODO this will be in the payload when we upgrade to focusedmonv2
+  for (auto &workspace : m_workspaces) {
+    if (workspace->name() == workspaceName) {
+      m_activeWorkspaceId = workspace->id();
+      break;
+    }
+  }
```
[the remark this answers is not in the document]
matt-fff: Sure. I migrated all the v2 events that were in the hyprland docs.

src/modules/hyprland/workspaces.cpp
matt-fff: Some refactoring changed this a bit, but I think it's still aligned with making the intent clearer.

src/modules/hyprland/workspaces.cpp
```diff
@@ -908,4 +953,38 @@ int Workspaces::windowRewritePriorityFunction(std::string const &window_rule) {
...
+  
+  return {part1, part2, part3};
+}
+
+int Workspaces::parseWorkspaceId(std::string const &workspaceIdStr) {
+  try {
+    return workspaceIdStr == "special" ? -99 : std::stoi(workspaceIdStr);
+  } catch (std::exception const &e) {
+    spdlog::error("Failed to parse workspace ID: {}", e.what());
+    return INVALID_WORKSPACE_ID;
```
zjeffer, line 986: Not a big fan of this INVALID_WORKSPACE_ID. If at any point Hyprland starts allowing negative workspaces, we'll have to change this. 

If you make `parseWorkspaceId` return an `std::optional<int>`, and return std::nullopt, this implementation would be independant from any potential change from Hyprland. Then you also don't need to add the extra variable. Instead of checking whether the parsed workspace ID is equal to the `INVALID_WORKSPACE_ID`, you can simply use `if (id.has_value()) { ... }`.

matt-fff: > If at any point Hyprland starts allowing negative workspaces, we'll have to change this.

Good point. Right now workspaces only go down to -99, but optional is better. Updated.

Thanks for the suggestion.

zjeffer:
Looks great! I fixed some clang-tidy hints (mostly replacing std::find_if with std::ranges::find_if), and formatted it, here's a patch: 
[patch.txt](https://github.com/user-attachments/files/19026880/patch.txt)

@Alexays can you approve the workflow?

matt-fff:
@Alexays When you get the chance, could you approve the workflow or let me know if there's something I need to address? Thanks!

Alexays:
LGTM, thanks!

PR #3924, opened 2025-02-08
dwl/window: add .active styling class for active windows

casKd-dev:
This allows equivalent styling as of dwm's status bar when a output is highlighted, making it possible to distinguish which output is selected without the need of borders (which may be hidden by specific patches.)


mortezadadgar:
I think this make more sense to add `window#waybar.empty` class like sway and hyperland modules.

Alexays:
Closing as a subset of #3926 — that PR (same author) adds the `.active` styling class for dwl/window along with the hide-inactive/hide-empty options, superseding this one. Let's consolidate the dwl/window work there. 🙏

PR #3925, opened 2025-02-08
dwl/tags: add hide-vacant option

casKd-dev:
Equivalent to hide-vacant-tags patch for dwm, hiding non-active tags without clients


DreamMaoMao:
https://github.com/Alexays/Waybar/pull/3504

Is this the same function as this pr?

DreamMaoMao:
https://github.com/Alexays/Waybar/pull/3504

This pr seems to have been abandoned, It seemed that no further work had been done for a long time,you should continue your pr, this is a good feature.

DreamMaoMao:
He didn't have the app hidden when he started up, and needed to switch to hide the tags without a client.

DreamMaoMao:
I think you should not add all the buttons at first, but check if the current tag button is already added in the set_tag_handle, and add it if it is not

DreamMaoMao:
Sorry I couldn't wait, so I made my own...

https://github.com/Alexays/Waybar/pull/4003

PR #3926, opened 2025-02-08
dwl/window: allow hiding of title when either inactive or empty

casKd-dev:
Depends on https://github.com/Alexays/Waybar/pull/3924, allows hiding of title in some conditions


src/modules/dwl/window.cpp
```diff
@@ -118,6 +130,19 @@ void Window::handle_frame() {
   if (tooltipEnabled()) {
     label_.set_tooltip_text(title_);
   }
+  if (hide_empty_ && title_ == "") {
```
DreamMaoMao, line 133: is it better to use empty?
```cpp
  if (hide_empty_ && title_ .empty()) {

```

Issue #1957, opened 2023-01-18
Ability to define icon theme for tray


7ff04da4aa6ac37197e249299066ac30:
When configuring a taskbar for waybar, there is an option, "icon-theme", that lets the user define the theme to be used for the taskbar. Ideally, I think, this, along with other options, like "icon-size", is something that should be available for the tray as well.

I am aware that it is possible to set icon themes using GTK settings, but this is not ideal, as I'd like to be able to set an icon theme specifically to be used with the tray and another to be used with applications in general. You know, like it is currently possible with the taskbar.

luispabon:
This would be pretty handy, some icon themes are great but have really crap or unreadable tray icons and viceversa

simonwiles:
This would be really useful.  I use a "light" icon theme but have a "dark" bar.  I can supply, e.g. `GTK_THEME=Adwaita:dark waybar &` which works fine for GTK applications but QT applications will still follow what's set in `org.gnome.desktop.interface`, and I get dark icons on a dark background :(

909oce:
any update on this? some apps work fine like nm-manager and blueman which use the theme, but others like webcord dont. i have webcord-tray.svg in my theme so it should pick it up, i think being able to manually assign icons would be useful for situations like these.

Issue #1400, opened 2022-01-17
feat: custom icons in tray


jakehamilton:
Waybar allows for customizing the output of most modules. It would be awesome to let users replace applet icons with their own. This would make the bar's style even more consistent.

---

I see two possible ways people would want to use this feature. The first being with an icon font like is common with other blocks. Here's an example of what that config might look like:

```json
{
  "layer": "top",
  "modules-right": ["tray"],
  "tray": {
    "icons": {
      "discord": "icon-font-char-here"
    }
  }
}
```

The atlernative would be to use images similar to what is done today:

```json
{
  "layer": "top",
  "modules-right": ["tray"],
  "tray": {
    "icons": {
      "discord": "/path/to/image.png"
    }
  }
}
```

I suppose it's also possible to support both options if we're comfortable adding multiple entries to the tray config (eg. `fontIcons` and `imageIcons`).

---

Of course, this will only be possible if there's something (like app id) that we can match on. I'm not very familiar with how applets or the system surrounding it work, so I'm not sure if this is possible. It seems like something that _should_ be doable.

sentientmachin3:
This is exactly what I was looking for.

antidoid:
Any Updates on this?

LukashonakV:
Hi @antidoid , not sure can you check whether it [PR1978](https://github.com/Alexays/Waybar/pull/1978 ) solves current enhancement.

rickalex21:
That would be sick.

v4u6h4n:
I would love this feature.

v4u6h4n:
I would love this feature.

MrGamerDoesGames:
This would be great to see, so I can make my tray icons integrate more with my system.

Givemegitpls:
I want it! I need it!

BonfireAtNight:
This would be exactly what I was looking for.

aspauldingcode:
Is this still not possible????

bashfulrobot:
Another vote. :+1:

aspauldingcode:
I think adding a tray shade/hue overlay would be good too. 
like grayscale or sepia or a custom tint, where all icons are shaded the same.

amirbandicoot:
It is possible to achieve this.
You have to replace the icons found in **/usr/share/icons/hicolor/16x16/** with your desired icons.
Apparently waybar uses these icons for tray.

aspauldingcode:
I'll try this with NixOS, thanks.

[...]

PR #3941, opened 2025-02-16
Features/systray custom icons

ChaosInfinited:
Hello o/

I really dislike having random-looking systray icons, and I saw that others were looking for a way to customize them—so I decided to take a stab at adding some (very rudimentary) support for this.

Related issue: [#1957](https://github.com/Alexays/Waybar/issues/1957)

This isn't the cleanest solution, but it does work—at least for non-Electron apps. (Electron seems to have its own issues: [electron/electron#40936](https://github.com/electron/electron/issues/40936)).

This implementation allows users to override tray icons through their config, following the suggestion in [#1400](https://github.com/Alexays/Waybar/issues/1400).

A couple of limitations:

- I could only make this work with actual image files, not font-based icons. The latter would likely require major changes to the current tray implementation.
- The icons are set before the item's ID is known, so I had to add the override (with a check to ensure it's applicable) when the ID is received via DBus.
- Users can match icons using either a system-wide recognized icon_name or by specifying a direct file path.

To avoid unnecessary config reloading, I also introduced a per-tray icon manager singleton, since there wasn't a good place where both the config and the item ID were accessible at the same time.

Maybe it'd be useful for someone :) Let me know if you have some suggestions

### DIFF
--- a/include/modules/sni/icon_manager.hpp
+++ b/include/modules/sni/icon_manager.hpp
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <json/json.h>
+#include <spdlog/spdlog.h>
+
+#include <string>
+#include <unordered_map>
+
+class IconManager {
+ public:
+  static IconManager& instance() {
+    static IconManager instance;
+    return instance;
+  }
+
+  void setIconsConfig(const Json::Value& icons_config) {
+    if (icons_config.isObject()) {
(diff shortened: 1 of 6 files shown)


Alexays:
lgtm, thanks!

mikedebian:
Works, however, I have to kill and open up waybar again for it to start picking up the custom icons.

PR #3945, opened 2025-02-19
Hyprland IPC improvements

zjeffer:
* Fixes IPC being blocked at shutdown: previously it required Hyprland to send another IPC message over the socket (like focuswindow) before waybar would shut down.
* IPC now managed using a static instance instead of global unique ptr
* Some general improvements to the Hyprland modules


lfos:
@zjeffer - It looks like with a multi-monitor setup, disabling one monitor now results in an IPC disconnect, breaking several Hyprland modules. Could you please take a look? See #3975 for details.

PR #3947, opened 2025-02-19
Fix workspacerules not taking into account defaultName

### LINKED ISSUE #3834: Hyprland module messes up named+persistent workspaces
Here is (part of) my Hyprland config:
```
workspace = 1, monitor:eDP-1, persistent:1, defaultName:s
workspace = 2, monitor:eDP-1, default:1, persistent:1, defaultName:d
workspace = 3, monitor:eDP-1, persistent:1, defaultName:f
```
What it does it says that the workspaces with IDs 1,2,3 should have names s,d,f and should be persistent and binds them to my monitor (and 2 should be the default).

Now the waybar hyprland module messes this up by displaying 6 symbols: 1,2,3,s,d,f. Initially just 1,2,3 and d. I think 1,2,3 are extracted from the workspace rules and then s,d,f are displayed when hyprland creates them. (That's why d is the first because it's my default one so this is the (only!) workspace created in the beginning.)

Here is a video:
https://youtu.be/CUKHm4BhX-g

Intended behavior:
It should only ever show 3 symbols: s,d,f. When it scans for persistent workspaces in the beginning it should realize that they have names and that they are the same as the ones that are later focused etc.

zjeffer:
Should fix #3834

@jbirnick can you test this PR?

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -294,7 +294,8 @@ void Workspaces::loadPersistentWorkspacesFromWorkspaceRules(const Json::Value &c
     if (!rule["persistent"].asBool()) {
       continue;
     }
-    auto const &workspace = rule["workspaceString"].asString();
+    auto const &workspace = rule.isMember("defaultName") ? rule["defaultName"].asString()
+                                                         : rule["workspaceString"].asString();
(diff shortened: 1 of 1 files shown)


jbirnick:
Seems to work, at least with 1 monitor :+1: 

Thanks for fixing the hyprland stuff, I hope it the changes will get merged! :)

PR #3965, opened 2025-03-03
Add scroll bindings for MPRIS module

### LINKED ISSUE #3641: [Feature Request] Add on-scroll controls to MPRIS module
I would like to be able to go to the next/previous track on the MPRIS module by using the scroll wheel instead of right/middle click

ritze:
This add the following bindings:
* on-scroll-up
* on-scroll-down
* on-scroll-left
* on-scroll-right

This fixes #3641.

### DIFF
--- a/include/modules/mpris/mpris.hpp
+++ b/include/modules/mpris/mpris.hpp
@@ -22,6 +22,7 @@ class Mpris : public ALabel {
   virtual ~Mpris();
   auto update() -> void override;
   bool handleToggle(GdkEventButton* const&) override;
+  bool handleScroll(GdkEventScroll*) override;
 
  private:
   static auto onPlayerNameAppeared(PlayerctlPlayerManager*, PlayerctlPlayerName*, gpointer) -> void;
--- a/man/waybar-mpris.5.scd
+++ b/man/waybar-mpris.5.scd
@@ -140,6 +140,26 @@ The *mpris* module displays currently playing media via libplayerctl.
 	default: next track ++
 	Overwrite default action toggles.
 
+*on-scroll-up*: ++
+	typeof: string ++
+	default: previous track ++
+	Overwrite default action toggles.
+
+*on-scroll-down*: ++
+	typeof: string ++
+	default: next track ++
+	Overwrite default action toggles.
+
+*on-scroll-left*: ++
+	typeof: string ++
+	default: previous track ++
+	Overwrite default action toggles.
+
+*on-scroll-right*: ++
+	typeof: string ++
+	default: next track ++
+	Overwrite default action toggles.
+
 *player-icons*: ++
 	typeof: map[string]string ++
 	Allows setting _{player_icon}_ based on player-name property.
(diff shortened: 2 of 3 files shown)


Dregu:
I don't really think these new actions make a lot of sense, i.e. I at least would expect scrolling to adjust player volume by default, not tracks. Of course the module doesn't actually have volume support and everything non-default is just forwarded to the generic cmd handler anyway, so the module also already has scrolling support simply by using `playerctl ...`

(Just to rant somewhere, the whole module is actually pretty pointless if you don't like the default binds, cause it's impossible to actually use the MPRIS connection with non-default binds. Having to use generic `playerctl ...` commands for everything kinda defeats the purpose of a having direct MPRIS support...)

ritze:
I personally used just a customized script only because of the possibility to bind the mouse wheel. I think it's much more comfortable and logical to change the track than using middle and right mouse button. 

Binding the mouse wheel is not possible without this change. I guess it should be possible to bind all mouse actions for all waybar modules in general. At least there is no argument against that.

playerctl supports to change the player volume. Or we could even implement functionality to switch between focused and displayed players. See playerctld's `shift` and `unshift` option.

Dregu:
But I'm saying it is possible and I'm already using it exactly like this on v0.12.0:

```
"mpris": {
  "on-scroll-up": "playerctl volume 0.05+",
  "on-scroll-down": "playerctl volume 0.05-"
}
```

My point still being that I don't want to use `playerctl`, I want this module to handle customizable playerctl-like commands directly through the IPC socket it already has, or else it's just a shell script module with a fancy name. I should probably just make that myself instead of complaining.

ritze:
Fair enough, it works for me now too. I guess I just used an old version, when I tried it before writing this patch.

About the playerctl: I just extended the already existing code. I didn't want to reinvent the wheel. There are pros and contras implementing MPRIS with playerctl in between. But I think that this PR is the wrong place to discuss the sense of the module.

To be honest. After I saw the source code of this module, I'm thinking that writing a bash script at my own will be faster than implementing missing functionality. But I think that not everybody wants to reinvest the wheel.

ritze:
In the end this change will only add default bindings to the mouse wheel. IMHO it doesn't need to be merged.

Dregu:
I'll probably work on getting actual rebindable mpris commands soon-ish, with player volume available for formatting too. Getting the % on the bar is even more important to me than scrolling, but of course it isn't in the scope of this pr...

markand:
I've discovered that my extra back/next buttons on my mouse are doing prev/next on the mpris module by default. I was surprised (positively) that it's supported by default but without that merge request merged and not seen the appropriate events in the waybar-mpris manual page so I wonder where this feature comes from?

ritze:
@markand It's [hardcoded directly in the source file](https://github.com/Alexays/Waybar/blob/master/src/modules/mpris/mpris.cpp#L666).

Since there is no interest to bind the scroll wheel by default to, I will close this merge request.

PR #3995, opened 2025-03-22
add login-proxy option

spectrum70:
There are cases where systemd-logind is not used/running. Result is that bcklight module will not run.

Add an option that, when set to false, allows backlight module to work without systemd-logind.


Alexays:
Thanks!

PR #4001, opened 2025-03-24
feat: add display-condition configuration option

hansp27:
added a *display-condition* configuration option to `niri/workspaces` module that specifies when workspace widgets should be drawn. 

Valid options are one of the following:

*show-all*: displays widgets for all workspaces. 
*keep-named*: always displays named workspaces, irrespective of whether they contain windows or not.
*only-populated*: only displays widgets for workspaces that have windows.

The default is *show-all*.


jmickelin:
Amazing, thank you :)

greed-d:
any chance of this being merged?

jmickelin:
It looks like https://github.com/Alexays/Waybar/pull/4966 was recently merged, adding the equivalent of the `only-populated` setting (with a `hide-empty` boolean option).

Given that the other PR is now canon, could we rework this PR to extend that option to include the `keep-named` functionality? I am assuming renaming the `hide-empty` option at this point is out of the question since it would be a breaking change now, but we can work around that.

I suggest to either 1) change the type of the existing option to incorporate a third value (e.g. accept one of `true`/`false`/`"keep-named"`), or 2) add a new boolean option like `hide-empty-named`.

Or maybe call that second one `hide-empty-condition` to leave room for more conditions in the future?

Something like (in the syntax of the config man page):
```
*hide-empty-condition*:	++
	typeof: string ++
	oneof: "all", "keep-named" ++
	default: "all" ++
...
```

----

(Pinging @Alexays, @cebem1nt for input :slightly_smiling_face:  )

hansp27:
I'd be down to implement this, as long as @Alexays consents. 

In terms of UX I think the first option is cleanest and most intuitive.

For option 2. I'd suggest naming the option `hide_empty_keep_named`, and for option 3 simply `hide_condition`, nonetheless, as long as there are no other concrete conditions we have a use for, it seems that option 3 is overkill and it's better to go with option 1.

jmickelin:
Fair enough! For option 3, I was thinking of use cases like regex-based hiding (as in this PR https://github.com/Alexays/Waybar/pull/4537). But of course, there is a huge potential to overcomplicate this, so maybe it's better to just add separate options if those cases ever get added, rather than cramming all of them into one option.

One concrete situation I myself have a need for would be a hiding criterion saying "hide empty, but keep named, unless the name is `[scratch]` (the name of my scratchpad space). But that case is better addressed by other means anyway (the `ext-workspace-v1` protocol has a `hidden` attribute that can be set on workspaces. Though Niri doesn't support it yet. See here: https://github.com/niri-wm/niri/discussions/2574#discussioncomment-16651296. In any case, I agree it's completely out of scope for this PR.)

hansp27:
ty

PR #4002, opened 2025-03-24
fix(FreeBSD): Use dev.cpu temperature sysctl

lbartoletti:
dev.cpu is mentionned but not used. Moreover, you can have dev.cpu... but no hw.acpi, so include a test on dev.cpu first


Alexays:
Thanks!

PR #4025, opened 2025-04-04
flake.nix: add treefmt; treewide: clang and nix format

khaneliman:
Also format files that haven't been getting formatted.


khaneliman:
@Alexays this one should be good to. Just added a way to format quicker and fix some formatting.

khaneliman:
@Alexays removed new input and just handled configuration manually so we don't need to propagate a new input to users

PR #4035, opened 2025-04-07
meson_options: improve sndio's description

fkobi:
Correct me if I am wrong but I think that is it's only use. If that is the case the description would be less generic and thus better.


fkobi:
@Alexays ping

PR #4048, opened 2025-04-13
Fix signal safety deadlock

### LINKED ISSUE #4047: Reproducible Deadlock
For a long time now, Waybar has been occasionally deadlocking. The root cause is that Waybar's signal handlers, defined here:

https://github.com/Alexays/Waybar/blob/056295cbc6d6b8efccae51251483c6f92b3b15b2/src/main.cpp#L77-L101

are not async-signal-safe.

Very few standard library functions are async-signal-safe ([in C](https://www.man7.org/linux/man-pages/man7/signal-safety.7.html) and [in C++](https://en.cppreference.com/w/cpp/utility/program/signal#Signal_handler)). Notably, memory allocation (as used here: https://github.com/Alexays/Waybar/blob/056295cbc6d6b8efccae51251483c6f92b3b15b2/src/bar.cpp#L302) is NOT async-signal-safe.

The deadlock can be reproduced by toggling the bar (by sending `SIGUSR1`) very fast. I used this shell script:

```sh
while true; do killall -SIGUSR1 waybar; done
```

On my computer, this deadlocks Waybar within 15 seconds, and spamming my "toggle bar" hotkey generally reduced this to 5 seconds.

Potentially related to https://github.com/Alexays/Waybar/issues/348

LiterallyVoid:
Fixes #4047.

This creates a separate thread to watch for signals. I used a `pipe` to send signals to the signal thread, as pipes are one of the few signal-safe messaging options.

Because GTK isn't thread safe, I couldn't handle signals in the signal thread either, so I used a `SafeSignal` which is thread-safe (but probably not signal-safe!) so that signals were handled on the main thread.

I also moved the `SIGCHLD` handling code into the new thread. I'm unsure about this change, as it will longer mask `SIGCHLD` in the main thread. From a cursory test it seems to work and children are still being reaped correctly, but I separated this change into its own commit so it can be reverted separately.

(Incidentally, all of `reload`'s accesses were on the same thread, so I changed it from a global to a local.)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -14,64 +14,6 @@ std::mutex reap_mtx;
 std::list<pid_t> reap;
 volatile bool reload;
 
-void* signalThread(void* args) {
-  int err;
-  int signum;
-  sigset_t mask;
-  sigemptyset(&mask);
-  sigaddset(&mask, SIGCHLD);
-
-  while (true) {
-    err = sigwait(&mask, &signum);
-    if (err != 0) {
-      spdlog::error("sigwait failed: {}", strerror(errno));
-      continue;
-    }
-
-    switch (signum) {
-      case SIGCHLD:
-        spdlog::debug("Received SIGCHLD in signalThread");
-        if (!reap.empty()) {
-          reap_mtx.lock();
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #4049, opened 2025-04-13
Select backlight device same as brightnessctl logic

JunaidQrysh:
using max brightness approach for fallback device is not good for hybrid laptop users
now the fallback device is in sync with brightnessctl


JunaidQrysh:
@Alexays Its been months for this PR, its just a tiny but useful change. Can you please review it?

PR #4060, opened 2025-04-17
feat(group): add reveal-by-default option

ErrorNoInternet:


Alexays:
Thanks!
Can you add this new option to the man?

ErrorNoInternet:
Done :)

strptrk:
Hi! I see two minor issues with the current implementation:
1. If not using "click-to-reveal", "reveal-by-default" doesn't make much sense.
2. It takes 2 clicks to close the group for the first time. 

To fix, I suggest changing this:
```cpp
revealer.set_reveal_child(reveal_by_default);
```
To this:
```cpp
if (click_to_reveal && reveal_by_default) {
  box.set_state_flags(Gtk::StateFlags::STATE_FLAG_PRELIGHT);
  revealer.set_reveal_child(true);
} else {
  revealer.set_reveal_child(false);
}
```

ErrorNoInternet:
Fixes applied

PR #4068, opened 2025-04-22
Fixed one layout with multiple variant bug hyprland/language Fixes #3676

### LINKED ISSUE #3676: (hyprland/language) "format-<shortcode>" doesn't resolve if two variants of the same language is used
I'm having an issue using two english variants (qwerty and colemak-dh) and getting the customized icons and "US" and "DH" names to display.

```
        "hyprland/language" = {
          format = "{}";
          format-en = "  us";
          format-en = "  dh";
          keyboard-name = "apple-inc.-apple-internal-keyboard-/-trackpad";
          on-click = "hyprctl switchxkblayout apple-inc.-apple-internal-keyboard-/-trackpad next";
        };
```

I get the following error when I try to use the above configuration in NixOS:
`error: attribute 'format-en' already defined at /nix/store/jcw0hy1dz7lxpvrccnl0c9gdcq34qyj6-source/home/system/waybar/default.nix:174:11`

As suggested above, I tried including the variant on the second format-en (i.e. format-en_colemak_dh), but that didn't work either.

When I don't specify `format-en` at all, they appear in waybar as `English (US)` and `English (Colemak-DH)`, which is little too verbose for a toolbar. Any suggestions on how to use two variants of the same language? Thanks in advance to anyone that can help.

Rimsoo:
Hypr onfig:
```
kb_layout = us, us
kb_variant = basic, intl
```

Using find_last_of eg:
`ev = "activelayout>>kb-name-,English (US, intl., with dead keys)"`
So layout name will be = " with dead keys)" -> unknown layout !

Using find_first_of fixed it.

### DIFF
--- a/src/modules/hyprland/language.cpp
+++ b/src/modules/hyprland/language.cpp
@@ -66,7 +66,11 @@ auto Language::update() -> void {
 void Language::onEvent(const std::string& ev) {
   std::lock_guard<std::mutex> lg(mutex_);
   std::string kbName(begin(ev) + ev.find_last_of('>') + 1, begin(ev) + ev.find_first_of(','));
-  auto layoutName = ev.substr(ev.find_first_of(',') + 1);
+  
+  // Last comma before variants parenthesis, eg:
+  // activelayout>>micro-star-int'l-co.,-ltd.-msi-gk50-elite-gaming-keyboard,English (US, intl., with dead keys)
+  std::string beforParenthesis(begin(ev), begin(ev) + ev.find_last_of('('));
+  auto layoutName = ev.substr(beforParenthesis.find_last_of(',') + 1);
 
   if (config_.isMember("keyboard-name") && kbName != config_["keyboard-name"].asString())
     return;  // ignore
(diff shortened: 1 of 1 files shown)


guttermonk:
Tested and working great for me. I would recommend also updating the wiki with the following info and example.

The "format-{shortDescription}-{variant}" is working fine on my side with the following config:

Hyprland config:
```
input {
    kb_layout = us,us
    kb_variant = colemak_dh,
    kb_options = grp:win_space_toggle
    numlock_by_default = true
    follow_mouse = 1
}
```
Waybar config:
```
  "hyprland/language": {
    "format-en": "󰼭 us",
    "format-en-colemak_dh": " dh",
    "on-click": "hyprctl switchxkblayout current next",
  },
```

Alexays:
The opposite has happened here https://github.com/Alexays/Waybar/pull/3224

Rimsoo:
Indeed, however, it does not take into account the case of multiple variants. I will rework it for better detection in both cases: commas in the keyboard name and/or multiple variants

Rimsoo (commented):
Extract the layout using the parenthesis first THEN the comma,
Since there is no nested parenthesis in a kb layout/variant according to this file:
`cat /usr/share/X11/xkb/rules/xorg.xml | grep -E "\(.*\("`

Rimsoo:
Any info on whether this will/can be merged?

zeerayne:
Well, this somehow have break layout display for:
```
    kb_layout = us,ru
```

zeerayne:
I think this is related: https://github.com/Alexays/Waybar/issues/4229

mexanoz:
This breaks for strings without parentheses, hence #4229. `find_last_of` returns `-1` if there's no such char in a string, so then it tries to create `std::string(begin(ev), begin(ev) - 1)` which fails. Needs an extra check for when there's no parentheses.

PR #3868, opened 2025-01-04
[hyprland/workspaces] Implement workspace taskbars

pol-rivero:
### Summary
This PR expands the functionality of `hyprland/workspaces` to make it behave more like the `wlr/taskbar` module. See #2656 for a high-level description.

This partially closes #2656, a full fix would require porting those changes to the `sway/workspaces` module as well. I don't use Sway and I'm not familiar on how it works, but if anyone is interested in giving it a go in a separate PR feel free to ping me and I'll do my best to help.

### Config changes
Adds the following changes to the `hyprland/workspaces` config:
```jsonc
"hyprland/workspaces": {
    // EXISTING SETTINGS
    "format": "{icon}: {windows}", // Used even if "workspace-taskbar.enable" is true
    "format-window-separator": "", // Will be ignored if "workspace-taskbar.enable" is true
    "window-rewrite-default": "",  // Will be ignored if "workspace-taskbar.enable" is true
    "window-rewrite": {},          // Will be ignored if "workspace-taskbar.enable" is true
    // NEW SETTINGS
    "workspace-taskbar": {
        // Enable the workspace taskbar. Default: false
        "enable": true,
        // Format of the windows in the taskbar. Default: "{icon}". Allowed variables: {icon}, {title}
        "format": "{icon} {title:.20}",
        // Icon size in pixels. Default: 16
        "icon-size": 16,
        // Either the name of an installed icon theme or an array of themes (ordered by priority). If not set, the default icon theme is used.
        "icon-theme": "some_icon_theme",
        // Orientation of the taskbar (see screenshots below). Default: "horizontal".
        "orientation": "horizontal"
    }
}
```
Notice how `workspace-taskbar.enable` defaults to `false`, so existing configs *shouldn't* be affected by these changes without opting in.

### What's missing

I'm mainly waiting for your feedback on the config json names and structure. Once that's done I will take a look at updating the documentation with the finalized spec.

There's also a small bug that prevents the window title from being updated unless the config contains a `"window-rewrite"` element (which shouldn't be necessary). That will be fixed tomorrow.
 
### Screenshots

**Default CSS, old config file (defaults to old, text-based icons)** 

image

```json
"hyprland/workspaces": {
    "format": "{icon}: {windows}",
    "format-window-separator": ",",
    "window-rewrite-default": "?",
    "window-rewrite": {
        "terminator": "",
        "code": "󰨞"
    }
},
```

---
**Default CSS, new config file (default format is just the icon)**

image

```json
[...]

PR #4076, opened 2025-04-24
Hyperland workspace windows to use system icons

stojanov:
Added the option for the hyprland workspace module to use system icons instead of them having to be predefined.

Both options for rewrite and the system icons are available.

This also adds two new configuration options: 
"windows-system-icons": {Bool} // Sets the system icons to be used
"windows-system-icons-size": {Int} // Sets the size of the icons if not set defaults to the label's font size


stojanov:
> @stojanov Looking at the description, this looks very similar to #3868

Yes, sorry for not seeing prior.
Closing

pol-rivero:
Don't worry! Let's see if we manage to get this PR merged.

PR #4084, opened 2025-04-28
add prefer-album-artist to mpris module

PandorasFox:
image

extends PlayerInfo to include an albumartist field; reads the xesam:albumartist field into it, and substitutes it for the artist value if the appropriate bool is true (default false, so no change to existing usecases).

I also thought about adding a separate  `Mpris::getAlbumArtistStr` method, but held off since it doesn't seem strictly necessary (e.g. would be kinda clunky to use in a format string as it might not be present always). I could instead split this out to that & have it fall back to the artist tag if no albumArtist is set? idk.

(I'll add a documentation commit once this looks good, if this is all fine!)


PandorasFox:
@Alexays 🥺

PandorasFox:
🥺

PandorasFox:
ty!

PR #4092, opened 2025-05-02
Issue 3981: try and fix memory leak in privacy module

### LINKED ISSUE #3981: Massive Memory leak in privacy module
Waybar has been leaking memory quite heavily recently, sometimes going up to 20 GB of memory usage within a few minutes.
I've gotten a heaptrack profile, which shows that the memory leak is related to the privacy module:

Image

RobertMueller2:
Closes #3981

As I'm mostly dealing with languages with automatic memory management, I'm not an expert on this. :p But I think `update_tooltip` creates objects with every update and at least doesn't get rid of the childrens' children.

Attempt to address this by using `make_managed` instead of new objects.

I've tested it and at least with my limited test it worked, can't see the reported memory leak with heaptrack anymore, either.

### DIFF
--- a/src/modules/privacy/privacy_item.cpp
+++ b/src/modules/privacy/privacy_item.cpp
@@ -1,5 +1,7 @@
 #include "modules/privacy/privacy_item.hpp"
 
+#include <spdlog/spdlog.h>
+
 #include <string>
 
 #include "glibmm/main.h"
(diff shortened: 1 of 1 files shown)


JassonCordones:
Apparently you forgot to apply the clang-format to the file.

RobertMueller2:
That's not correct. I did run it on the file.
```
clang-format-19 -i --style="{BasedOnStyle: Google, AlignConsecutiveDeclarations: false, ColumnLimit: 100}" src/modules/privacy/privacy_item.cpp
```

I just didn't run it on everything else, which has been broken for a while.

If you check the details of the failed check, it failed for src/modules/sway/workspaces.cpp, which I haven't edited here.

PR #4096, opened 2025-05-03
Add idle_inhibitor style docs.

aereaux:
Copied this from another man page.

If this is correct I can try adding this to some of the other man pages.


Alexays:
Thanks!

PR #4097, opened 2025-05-04
feat: sway/window: provide {marks} format replacement

RobertMueller2:
if any check fails, that's still because of src/modules/sway/workspaces.cpp, i didn't touch that. see #4094 

EDIT: i did forget to format the header file though, should be fixed now.


Alexays:
LGTM

PR #4102, opened 2025-05-06
Use load_symbolic for gtk icon to support styling in tray icon

edwin0cheng:
Currently if an application use symblolic icon, the css styling is not applied. 
This PR use `load_symbolic` instead of `load_icon` to apply the css styling to corresponding item in tray.

Here is the example (before and after):

2025-05-07T03:12:45,638675232+08:00


edwin0cheng:
The clang-format check failed on other code, but from my side . 😅

Alexays:
Nice!

PR #4114, opened 2025-05-13
cpu: add load{1,5,15} format variables

Duncaen:
Simply exports the already gathered load averages as variables. Added `load1` as a duplicate for `load` to match `load5` and `load15`, can remove it if its not desirable.


b0ch3nski:
Hey @Alexays, could you take a look at this one? :slightly_smiling_face:

PR #4136, opened 2025-05-20
hyprland workspaces: Add sorting Special Centered

### LINKED ISSUE #4135: [hyprland/workspace] Special in the center of the list
Make `special` workspaces in center of list

Like this
Image

Roc25:
Closes #4135

### DIFF
--- a/src/modules/hyprland/workspaces.cpp
+++ b/src/modules/hyprland/workspaces.cpp
@@ -772,30 +772,39 @@ void Workspaces::setCurrentMonitorId() {
 
 void Workspaces::sortSpecialCentered() {
   std::vector<std::unique_ptr<Workspace>> specialWorkspaces;
+  std::vector<std::unique_ptr<Workspace>> hidedWorkspaces;
   std::vector<std::unique_ptr<Workspace>> normalWorkspaces;
 
   for (auto &workspace : m_workspaces) {
     if (workspace->isSpecial()) {
       specialWorkspaces.push_back(std::move(workspace));
     } else {
-      normalWorkspaces.push_back(std::move(workspace));
+      if (workspace->button().is_visible()) {
+        normalWorkspaces.push_back(std::move(workspace));
+      } else {
+        hidedWorkspaces.push_back(std::move(workspace));
+      }
     }
   }
   m_workspaces.clear();
 
   size_t center = normalWorkspaces.size() / 2;
 
-  m_workspaces.insert(m_workspaces.end(), 
-                      std::make_move_iterator(normalWorkspaces.begin()), 
+  m_workspaces.insert(m_workspaces.end(),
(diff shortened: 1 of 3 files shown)


JassonCordones:
Doing the for loop you're moving the workspace you're iterating over. Wouldn't that cause problems?

zjeffer:
> Doring the for loop you're moving the workspace you're iterating over. Wouldn't that cause problems?

I think it's fine because:
* we have a mutex lock around `doUpdate()`, which ensures m_workspaces isn't touched by another thread 
* we clear m_workspaces right after we made all workspaces nullptrs, so there's no risk of null ptr dereferencing

But I'm still not a big fan of that method, I'm trying to find a safer way to do it

EDIT:
I came up with this but I feel like it isn't much better :laughing: 
```cpp
  std::vector<size_t> normalIdx;
  std::vector<size_t> specialIdx;
  std::vector<size_t> hiddenIdx;

  for (size_t i = 0; i < m_workspaces.size(); ++i) {
    if (m_workspaces[i]->isSpecial()) {
      specialIdx.push_back(i);
    } else if (m_workspaces[i]->button().is_visible()) {
      normalIdx.push_back(i);
    } else {
      hiddenIdx.push_back(i);
    }
  }
  size_t center = normalIdx.size() / 2;

  std::vector<std::unique_ptr<Workspace>> sorted;
  sorted.reserve(m_workspaces.size());

  auto move_by_indexes = [&](const std::vector<size_t>& idxs) {
    for (size_t idx : idxs) {
      sorted.push_back(std::move(m_workspaces[idx]));
    }
  };

  move_by_indexes({normalIdx.begin(), normalIdx.begin() + center});
  move_by_indexes(specialIdx);
  move_by_indexes({normalIdx.begin() + center, normalIdx.end()});
  move_by_indexes(hiddenIdx);

  m_workspaces = std::move(sorted);
```

src/modules/hyprland/workspaces.cpp
```diff
@@ -770,7 +770,45 @@ void Workspaces::setCurrentMonitorId() {
   }
 }
 
+void Workspaces::sortSpecialCentered() {
+  std::vector<std::unique_ptr<Workspace>> specialWorkspaces;
+  std::vector<std::unique_ptr<Workspace>> hidedWorkspaces;
```
zjeffer, line 775: `hiddenWorkspaces`

zjeffer:
Can you update the docs?

JassonCordones (approved):
Looks good to me

JassonCordones:
I'll check if I can update the docs if I have some spare time during the weekend, but I don't promise anything

Roc25:
I also update wiki on github

zjeffer (approved):

PR #4149, opened 2025-05-28
Add default windows format in hyprland

Xaunther:
I have a proposal to add a default windows format to be used when there are no windows in the workspace in Hyprland. It is currently possible to default each window to an icon, but it is not possible to do so when there are no windows.

Currently, when a workspace has no windows, `{windows}` gets formatted to an empty string. With this proposal, `{windows}` may get formatted to something more useful instead, like the workspace ID.

The new JSON key is named `format-windows-default` and  does not substitute `{windows}`, only `{id}`, `{name}` and `{icon}`.

I also went a bit out of my way to add a separate method that is responsible of creating the formatted `{windows}` string, to reduce the clutter inside `hyprland::Workspace::update`.

### Example
I tested in my own setup adding `"format-windows-default": "<b>{id}</b>"` to my ´hyprland/workspaces` module with "format": "{windows}"` (top bar is with the change)
image.

### About sway
Probably this change could be ported to sway too, but since I'm not a sway user I only worked on the hyprland solution.

### DIFF
--- a/include/modules/hyprland/workspace.hpp
+++ b/include/modules/hyprland/workspace.hpp
@@ -63,6 +63,7 @@ class Workspace {
   std::optional<std::string> closeWindow(WindowAddress const& addr);
 
   void update(const std::string& format, const std::string& icon);
+  std::string windows(const std::string &icon) const;
 
  private:
   Workspaces& m_workspaceManager;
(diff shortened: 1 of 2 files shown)


Xaunther:
Closing this myself due to the lack of traction

PR #4155, opened 2025-06-01
Fix calendar extra padding if there are wide characters

### LINKED ISSUE #3132: Calendar display issue with the clock module
On my Arch Linux system, I am encountering the following issue with the calendar display, as shown in the screenshot below. 
I don't know if this is a bug or not. 

version: 0.10.0
wm: hyprland 0.38.0

image
image_2024-04-10_20-50-04

Yutsuten:
Closes #3132

Fixed the issue by counting the number of wide characters and using it when padding the calendar line.
My reference for how to check for wide characters was [src/modules/mpris/mpris.cpp L210](https://github.com/Alexays/Waybar/blob/0.12.0/src/modules/mpris/mpris.cpp#L204-L219).

### DIFF
--- a/src/modules/clock.cpp
+++ b/src/modules/clock.cpp
@@ -1,5 +1,6 @@
 #include "modules/clock.hpp"
 
+#include <glib.h>
 #include <gtkmm/tooltip.h>
 #include <spdlog/spdlog.h>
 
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #4160, opened 2025-06-03
AIconLabel.cpp: honour "rotation" property and add "swap-icon-label" propery

davidemanini:
The current behaviour for modules employing `AIconLabel` is that the icon is show on the left of the label.  This is not desirable for for vertical bars and for users with a right-to-left script.  

This PR solves this issue by:

 * making `AIconLabel` honour the `rotation` property
 * adding a `swap-icon-label` property


Alexays:
Thanks!

PR #4183, opened 2025-06-11
Memory Fix for continuous scripts in custom module

Talbrelliles:
I noticed that waybar was continually allocating more memory after I added a custom module that invoked a continuous script. After looking at it for a while, I think it is because the `ScopeGuard`—meant to clear the buffer—was capturing `buff` by value, so it would continually try and free `nullptr`.

Here's a change that should capture `buff` by reference and hopefully fix the issue (it seemed to fix it for at least my use case). 

I'm pretty rusty at C++, and newish to github in general, so feedback is appreciated.


Alexays:
Thanks!

Issue #383, opened 2019-06-12
[Feature request] call script on change battarey state


niktob560:
Call custom script on change of bat state (single script with state in params or separate user scripts for each state)

Alexays:
Implemented on master (PR #4208, commit 08da7b52): the battery module now supports an `events` config that runs commands on state changes — e.g. `on-charging`, `on-discharging`, and capacity/state-specific variants. See the waybar-battery(5) man page. Closing as resolved — please reopen if you need this for a module other than battery.

PR #4208, opened 2025-06-21
battery: Add support for battery level related events

markx86:
The user can now specify commands to be executed when a specific battery level is reached in a certain state.  
The events must be specified in the `events` object, and the keys in the object can be in the format
- `on-[status]-[state]`
- `on-[status]-[capacity]`
where:
- `[status]` can be `discharging` or `charging`
- `[state]` is the name of one of the states specified in the `states` object
- `[capacity]` is the battery percentage

For example, the following configuration will send a low battery notification when the battery reaches the warning and critical states while discharging and a full battery notification when the battery reaches 100% when charging.
```jsonc
 "battery": {
    "states": {
        "warning": 20,
        "critical": 10
    },
    "events": {
        "on-discharging-warning": "notify-send -u normal 'Low Battery'",
        "on-discharging-critical": "notify-send -u critical 'Very Low Battery'",
        "on-charging-100": "notify-send -u normal 'Battery is full'",
        },
    ...
}
```

### DIFF
--- a/src/modules/battery.cpp
+++ b/src/modules/battery.cpp
@@ -1,8 +1,9 @@
 #include "modules/battery.hpp"
-#include "util/command.hpp"
 
 #include <algorithm>
 #include <cctype>
+
+#include "util/command.hpp"
 #if defined(__FreeBSD__)
 #include <sys/sysctl.h>
 #endif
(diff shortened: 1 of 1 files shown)


Alexays:
LGTM, thanks!

Alexays:
Can you add that to the battery man? 🙏

yerlotic:
Related to #383

markx86:
Done!

markx86:
I rebased onto master and fixed the CI. I think it's ready merge, unless I missed something. @Alexays

rieje:
Users who use this feature might also want notifications for charging and uncharging but this doesn't seem to support that since it requires a state or capacity. For example, if a cable or port is problematic, it is useful to confirm on plugging the charging cable that the system is indeed charging.

Would it make sense to support this?

PR #4210, opened 2025-06-21
Feat/display systemd system state

stkth:
First part of this pull-request refactors the code of the systemd_failed_units module for two reasons:

- eases readability of code
- prepares for introduction of system state displaying feature

Users, which are not interested in the number of failed units but in the general distinguishing of the state of systemd for the user and system session can now display this information too.


Alexays:
Thanks!

PR #4211, opened 2025-06-21
Fix gentoo CI build

notpeelz:
The gentoo CI job started started failing on recent PRs. What caught my attention is that it's running a pretty old version of gcc (12.2.1). Turns out the docker image hasn't been updated in nearly 2 years (most likely because of the strict version requirement for `dev-cpp/gtkmm`).

I've reduced the frequency of the docker build workflow, as a daily rebuild of the gentoo image will most certainly bust through the free CI tier. I've also added a `workflow_dispatch` trigger, so it can be run manually if needed.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,8 +3,8 @@ name: Build and Push Docker Image
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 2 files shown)


notpeelz:
@Alexays could you run the docker workflow?

PR #4246, opened 2025-06-29
feat: Add ISO 8601 calendar to clock module

edeustua:
### Summary:
This PR adds ISO 8601 week formatting support for the calendar tooltip in Waybar's clock module. Users can now display weeks according to the ISO standard (Monday start, week 1 begins with the year's first Thursday) in the tooltip.

### Changes:
Introduced a new iso8601 configuration option to enable ISO-compliant week formatting in the tooltip.

When enabled, weeks follow {:%V} (e.g., "Week 42") per ISO 8601.

Backward-compatible—default behavior remains unchanged unless iso8601: true is set.

### Motivation:
ISO 8601 is widely used in scheduling, business, and international contexts. This change aligns Waybar with standard date/week representations (e.g., Linux’s %V), avoiding locale-specific inconsistencies.

### Example Configuration:

```jsonc
  "clock": {
    // "timezone": "America/New_York",
    "format": "W{:%V %a %b %d %I:%M %p}",
    "format-alt": "{:%Y-%m-%d}",
    "tooltip-format": "<tt><small>{calendar}</small></tt>",
    "calendar": {
      "mode": "year",
      "mode-mon-col": 3,
      "weeks-pos": "right",
      "on-scroll": 1,
      "iso": true,
      "format": {
        "months": "<span color='#ebdbb2'><b>{}</b></span>",
        "days": "<span color='#bdae93'><b>{}</b></span>",
        "weeks": "<span color='#458588'><b>W{}</b></span>",
        "weekdays": "<span color='#d79921'><b>{}</b></span>",
        "today": "<span color='#b16286'><b><u>{}</u></b></span>",
      },
    },
  },
```

### DIFF
--- a/include/modules/clock.hpp
+++ b/include/modules/clock.hpp
@@ -38,18 +38,19 @@ class Clock final : public ALabel {
     5 - tooltip-format
    */
   std::map<int, std::string const> fmtMap_;
-  uint cldMonCols_{3};                 // calendar count month columns
-  int cldWnLen_{3};                    // calendar week number length
-  const int cldMonColLen_{20};         // calendar month column length
-  WS cldWPos_{WS::HIDDEN};             // calendar week side to print
-  date::months cldCurrShift_{0};       // calendar months shift
-  int cldShift_{1};                    // calendar months shift factor
-  date::year_month_day cldYearShift_;  // calendar Year mode. Cached ymd
-  std::string cldYearCached_;          // calendar Year mode. Cached calendar
-  date::year_month cldMonShift_;       // calendar Month mode. Cached ym
-  std::string cldMonCached_;           // calendar Month mode. Cached calendar
-  date::day cldBaseDay_{0};  // calendar Cached day. Is used when today is changing(midnight)
-  std::string cldText_{""};  // calendar text to print
+  uint cldMonCols_{3};                       // calendar count month columns
+  int cldWnLen_{3};                          // calendar week number length
+  const int cldMonColLen_{20};               // calendar month column length
+  WS cldWPos_{WS::HIDDEN};                   // calendar week side to print
+  date::months cldCurrShift_{0};             // calendar months shift
+  int cldShift_{1};                          // calendar months shift factor
+  date::year_month_day cldYearShift_;        // calendar Year mode. Cached ymd
+  std::string cldYearCached_;                // calendar Year mode. Cached calendar
+  date::year_month cldMonShift_;             // calendar Month mode. Cached ym
+  std::string cldMonCached_;                 // calendar Month mode. Cached calendar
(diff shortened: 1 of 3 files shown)


include/modules/clock.hpp
```diff
@@ -38,18 +38,19 @@ class Clock final : public ALabel {
     5 - tooltip-format
    */
   std::map<int, std::string const> fmtMap_;
-  uint cldMonCols_{3};                 // calendar count month columns
```
notpeelz, line 41: The weird alignment is the result of clang-format, which is checked by the CI. Don't change the formatting or the CI will complain.

edeustua: Sorry about that. I ran `clang-format -i /include/modules/clock.hpp`. Hopefully this does the trick.

LukashonakV:
As far as I remember, the week format depends on the system locale configuration.  And it can even be changed by calling waybar with a certain locale as a separate local variable. Isn't it enough?

edeustua:
> As far as I remember, the week format depends on the system locale configuration. And it can even be changed by calling waybar with a certain locale as a separate local variable. Isn't it enough?

Yeah, you're right — the week format does depend on the system locale, and it can be overridden by launching Waybar with a different locale. However, this affects all locale-sensitive elements, including month and day names, which might not be desirable.

In addition to this, while `strftime` provides some flexibility with formats like `%V`, `%W`, and `%U`, that alone doesn't fully address the issue. For example, using `en_US` as a locale, there's no clean way to get ISO 8601 week numbers (`%V`) *without* also affecting other parts of the UI.

The goal here is to support ISO 8601-style week calendars (e.g., `%V`-like behavior) while preserving the user's locale, such as `en_US`.

To provide more context about the ISO 8601 standard, I'm attaching the NOTE in `man 3 strftime`.

>    ISO 8601 week dates
       %G, %g, and %V yield values calculated from the week-based year defined by the ISO 8601 standard.  In this system, weeks start on a Monday,  and are  numbered  from 01, for the first week, up to 52 or 53, for the last week.  Week 1 is the first week where four or more days fall within the       new year (or, synonymously, week 01 is: the first week of the year that contains a Thursday; or, the week that has 4 January in it).  When three       or fewer days of the first calendar week of the new year fall within that year, then the ISO 8601 week-based system counts those days as part of       week 52 or 53 of the preceding year.  For example, 1 January 2010 is a Friday, meaning that just three days of that calendar week fall in  2010.       Thus, the ISO 8601 week-based system considers these days to be part of week 53 (%V) of the year 2009 (%G); week 01 of ISO 8601 year 2010 starts       on Monday, 4 January 2010.  Similarly, the first two days of January 2011 are considered to be part of week 52 of the year 2010.

Alexays:
lgtm

PR #4255, opened 2025-06-30
wireplumber: add more volume formats

mBornand:
Currently waybar uses a cubic scale to show the volume in the wireplumber module. A logarithmic scale can provide more consistent volume steps.

So this changes aim at giving the user more flexibility in this regard.

- Add the possibility to show the volume in linear, cubic, db, and cubic percent (default for backward compatibility).
- Add the ability to change the scale for the scrolling action.
- Make the minimal step flexible to fit any scale.

Additional consideration
- Only the cubic percent scale uses integers, so while the new scales permit the control of decimals it creates a difference in the interface that is not obvious.
- A modification of the wiki will be needed.

configuration example
```
	"wireplumber": {
		"format": "{volume_db:.2f}dB",
		"scroll-scale": "db",
		"scroll-step": 2.5
	}
```


mBornand:
I just added entries in the man files.

mBornand:
Corrected the indentation errors in man files.

PR #4260, opened 2025-07-02
feat: add rfkill network setting

notpeelz:
This setting makes it possible to have a configuration with two network modules where one of them displays the ethernet state (disconnected, linked, ethernet), and the other, the wifi state (disabled, disconnected, linked, wifi).
Otherwise the ethernet state would show up as "disabled" (instead of "disconnected") when rfkill is active.


Alexays:
LGTM thx!

Issue #4291, opened 2025-07-12
Respect GTK color scheme variant when using GTK CSS variables


nraffp:
Currently Waybar does not seem to respect the dark/light variants of a GTK theme when using GTK CSS variables. 

For example take the following CSS in `style.css`:
```css
window#waybar {
  background-color: @theme_bg_color
}
```

I would expect the value of `@theme-bg-color`to change based on the variant (dark/light) of the current GTK theme as set by the key `color-scheme` of the schema `org.gnome.desktop.interface`. This doesn't seem to be the case and always follows the default variant.

PR #4292, opened 2025-07-12
respect gtk color scheme variant for gtk css variable

nraffp:
Addresses #4291

This might a breaking change for people who's setup expect the value of the GTK CSS variables to just always be the default variant. However I feel like this is more inline with expectations (at least mine) and better for the future.

Also I am not very familiar with the GTK C++ APIs and if this is the best way to accomplish this. But this works for me on my local build.


Alexays:
Yes!
Thanks :)

PR #4297, opened 2025-07-14
feat: language - implemented language short uppercase

csskevin:
This PR implements the argument for displaying the short display code in uppercase. Has been implemented and tested on hyprland, niri and sway :)


Alexays:
You can use pango markup for that :) 
https://docs.gtk.org/Pango/pango_markup.html

PR #4320, opened 2025-07-23
feat(hyprland/workspaces): add uniqueIcons option

### LINKED ISSUE #4318: [FR] [hyprland/workspaces] Option to hide duplicate window icons in workspace
Right now, I can have many windows groupped in a single workspace in Hyprland. When displaying the window icons, this will lead to one icon per window.
I'd like to have an option to hide duplicate icons for this use case

LoricAndre:
Adds a `unique-icons` option to the hyprland/workspaces module, which is a boolean (defaulting to false) that will make it so only one of each icon is shown.

Closes #4318


LoricAndre:
I'm not sure how to proceed for the wiki update, this line needs to be added in the corresponding section : 
```markdown
| `unique-icons` | bool | `false` | If set to true, if multiple windows that would be represented by the same icon are on the same workspace, the icon is only shown once.
```

PR #4333, opened 2025-07-29
Relative paths in included configs

Arkoniak:
This is a two part PR, it appears when I was tinkering with my `waybar` config in my ``~/.config/waybar/` directory and occasionally did 
```
includes: [
  "modules/custom.jsonc",
  "modules/groups.jsonc"
]
```
and suddenly half of my waybar just disappeared. It took me a long time to trace the issue, since there was zero messages in logs, even with `-l debug` key.

So, as a result here is the PR:

1. I've added couple of warning statements that elements description is missing. My assumption is that unlike any other elements (e.g. "cpu"), "custom" and "group" elements has no defaults to fall into, so if there is no configuration found (it can be wrong include as in my case, or just a typo) then this element will always be hidden. And so it is legit to warn user that this may happen.

2. I've added config search for include files in the same manner as regular `config` search. The only difference, I do not want for existing configs to be broken, so it is slightly more complex than just call `findConfigPath`. In this implementation if user uses absolute paths they work as usual, and this change only kicks in if user uses relative paths (which I presume mostly case of newcomers).

I've tried it in my environment and it is working as expected.

Also, I am flexible, can split this PR in two if needed or do modifications if some parts of this set up are unacceptable.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -89,19 +89,33 @@ void Config::setupConfig(Json::Value &dst, const std::string &config_file, int d
   mergeConfig(dst, tmp_config);
 }
 
+std::optional<std::string> Config::findIncludePath(const std::string name) {
+  auto match1 = tryExpandPath(name, "");
+  if (!match1.empty()) {
+    return match1.front();
+  }
+  return findConfigPath({name});
+}
+
 void Config::resolveConfigIncludes(Json::Value &config, int depth) {
   Json::Value includes = config["include"];
   if (includes.isArray()) {
     for (const auto &include : includes) {
       spdlog::info("Including resource file: {}", include.asString());
-      for (const auto &match : tryExpandPath(include.asString(), "")) {
-        setupConfig(config, match, depth + 1);
+      auto match = findIncludePath(include.asString());
+      if (match.has_value()) {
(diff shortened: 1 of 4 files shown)


src/config.cpp
```diff
@@ -89,19 +89,33 @@ void Config::setupConfig(Json::Value &dst, const std::string &config_file, int d
   mergeConfig(dst, tmp_config);
 }
 
+std::optional<std::string> Config::findIncludePath(const std::string name) {
```
notpeelz, line 92: Pass by reference to avoid the extra copy.
```suggestion
std::optional<std::string> Config::findIncludePath(const std::string& name) {
```

Arkoniak:
Also added `static` to be more aligned with `tryExpandPath`

Alexays:
Thanks!

kRHYME7:
HI @Arkoniak this PR reportedly broke wildcards * ? https://github.com/Alexays/Waybar/issues/4354 ?

Arkoniak:
> HI @Arkoniak this PR reportedly broke wildcards * ? #4354 ?

Yes, it did, I am sorry. This PR: https://github.com/Alexays/Waybar/pull/4373 addresses this issue.

PR #4334, opened 2025-07-29
feat(mpd): add playing-interval option

### LINKED ISSUE #4325: [Feature Request] mpd: reduce `Playing` state polling interval or make it configurable
I recently switched from i3/Polybar to Hyprland/Waybar and I noticed that MPD in Waybar is much less responsive than in Polybar.

After looking at the [MPD module](https://github.com/Alexays/Waybar/blob/0776e694df56c2c849b682369148210d81324e93/src/modules/mpd/state.cpp#L259-L262) source code, I determined that the problem is due to the polling happening much less frequently during the `Playing` state - every 1 second. In comparison, `Paused` and `Stopped` states use 200ms intervals.

https://github.com/Alexays/Waybar/blob/0776e694df56c2c849b682369148210d81324e93/src/modules/mpd/state.cpp#L120-L124

I run the MPD server on `localhost` and this 1-second delay is very noticeable for me coming from Polybar.

Building my own version of Waybar with the reduced interval to 200ms fixes the issue for me. This is the easy solution, but I imagine it's set to 1 second for a reason - probably to avoid too frequent requests to remote MPD servers. So I guess a better solution would be adding a configuration option for this interval.

I would be happy to open a pull request, but first I want to confirm the optimal approach to this problem with the maintainers:

1. Just reduce the interval.
2. Make the `Playing` state interval configurable.
3. Make the interval configurable for all states.
4. Something else?

Side note: not sure if this issue is related to #926.

theoratkin:
As described in #4325, the default 1-second timer for the `Playing` state is very noticeable when running MPD server on `localhost`. This pull request adds a `playing-interval` option to the `mpd` module which allows configuring this interval (in milliseconds). Defaults to 1000 milliseconds.

Closes #4325.


Alexays:
Could you add this new option to the man 🙏

theoratkin:
@Alexays done!

theoratkin:
Hey @Alexays, is there anything else blocking this from being merged? I'd love to continue using the official package instead of my fork :slightly_smiling_face: Let me know if you need me to squash the commits or do anything else.

PR #4366, opened 2025-08-11
`format-bluetooth` and `device.form-factor` support for wireplumber module

### LINKED ISSUE #3892: [Feature Request] `format-bluetooth` for wireplumber module
Would it be possible to add support for output-specific styling to the wireplumber module? This would bring it further in line with the pulseaudio module.

Also as an aside, the wireplumber module wiki page doesn't quite reflect the `format-icons` functionality it does have similar to the pulseaudio page (granted it can do less, but still portrayed very differently)

(I am well aware that for most use-cases the pulseaudio module works fine through pipewire-pulse, this is mainly a question for styling on my part - wishing to control the sink and source elements separately, which I can't find any other way of doing currently)

winkelnp:
fixes #3892

Adding (rudimentary) support for `format-bluetooth` and `format-bluetooth-muted` in the wireplumber module.

This works for me, but probably still has things that could be done substantially better, as I still have much to learn about working in c++ and with wireplumber. Any tips/help/things I need to work on are much appreciated.

Ultimately what I'd like to add is also adherence to `device.form-factor` like the pulse module does, but I couldn't quite get that working yet

Aside: I've left a couple comments in from an earlier working version, I'll remove them if we don't find any (severe) issues with the way it is now. The commented out version is an extension of what was there before, the new one is a rework taking the pulse module as an example.


winkelnp:
Just added `device.form-factor` support, should this still be the same PR or should I open a new one? I guess I can repurpose this one for both?

Edit: just realised that the way I've got it right now only works for things like bluetooth devices, not eg for wired headphones, as those are done via a "port" on the device, not by communicating a `device.form-factor` I'll see that I get that fixed (any tips welcome)

alper-han:
This can make my headphone icon work properly 😀

PR #4372, opened 2025-08-12
Fix module and tray item staying in hover state after opening menu

### LINKED ISSUE #3980: Hover style stays active when clicking a menu option or clicking anywhere other then the button for that menu.
Hi, 
New to waybar and pretty new to coding.  I've made it to the point of making some custom modules with menus.   I've made a regular style and hover style for the custom module.  Everything works with the module itself but I'm running into the issue stated in the title. 

The hover style stays active when I click a menu option or click anywhere other then the button to close the menu.  Afterwards if I hover back over the button, the style then switches back to the regular style.

rellimn:
This PR fixes #3980 where widgets stay in a hover state after using them to open a menu on click, even after closing the menu, making :hover in CSS very annoying to use with menus.

I couldn't find the reason why this happens. I assume the methods that handle the mouse leaving the widget don't get called because the menu gets focused before the pointer actually leaves the widget, but I couldn't find a way to fix this elegantly. The current fix simply unsets the prelight flag of the widget when the menu is opened, achieving the desired effect.

Also, I've had this happen to me for modules and tray icons, but it may happen in other places as well. I've tried searching the code for other classes working with menus but I couldn't find any so I left it there.

### DIFF
--- a/src/AModule.cpp
+++ b/src/AModule.cpp
@@ -172,6 +172,10 @@ bool AModule::handleUserEvent(GdkEventButton* const& e) {
       // Popup the menu
       gtk_widget_show_all(GTK_WIDGET(menu_));
       gtk_menu_popup_at_pointer(GTK_MENU(menu_), reinterpret_cast<GdkEvent*>(e));
+      // Manually reset prelight to make sure the module doesn't stay in a hover state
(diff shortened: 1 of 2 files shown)


Alexays:
lgtm, thanks!

Issue #4354, opened 2025-08-08
regression: wildcard includes are broken after v0.14.0


agorgl:
It seems that after waybar version 0.14.0 wildcard includes no longer work properly.
The observed behavior is that only the first matched configuration from the wildcard seems to be included.

Simple poc follows.

Sample configuration tree:
```
.
├── config.jsonc
└── modules
    ├── cpu.jsonc
    └── memory.jsonc
```

Sample `config.jsonc`
```
// config.jsonc
{
  "height": 26,
  "spacing": 14,
  "include": [
    "modules/*.jsonc"
  ],
  "modules-left": ["cpu", "memory"]
}
```

Sample `modules/cpu.jsonc`
```
// modules/cpu.jsonc
{
  "cpu": {
    "interval": 2,
    "format": "goo: {usage}%"
  }
}
```

Sample `modules/memory.jsonc`
```
// modules/memory.jsonc
{
  "memory": {
    "interval": 2
    "format": "foo: {used:0.1f}G",
  }
}
```

waybar 0.13.0 result:
<img width="206" height="46" alt="Image" src="https://github.com/user-attachments/assets/075df19f-dadc-4c33-bfba-9c01a28bb25b" />

waybar 0.14.0 result:
<img width="200" height="46" alt="Image" src="https://github.com/user-attachments/assets/9f4cd144-b565-4526-b0d6-8419ae11906e" />

As you can see the memory module configuration in the waybar version 0.14.0 is not applied.
Possibly caused by: https://github.com/Alexays/Waybar/pull/4333

carljustineoyales:
still broken

jhotiori:
this hasn't been fixed yet. i have recently migrated into modular configuration. wildcard loads some modules, but most of the time, essential ones do not load. the file hierarchy follows pretty much the same as the issues' sample.
 
```
├── config.jsonc
└── modules
    ├── hyprland/
        ├── window.jsonc
        ├── workspaces.jsonc
    ├── custom/ 
        ├── menu.jsonc
        ├── system.jsonc
        ├── updates.jsonc
    ├── cpu.jsonc
    └── memory.jsonc
    ├── battery.jsonc
    ├── ... etc
```

on `config.jsonc`:

```
...
"include": [
    "~/.dotfiles/waybar/modules/hyprland/*.jsonc",
    "~/.dotfiles/waybar/modules/custom/*.jsonc",
    "~/.dotfiles/waybar/modules/*.jsonc"
]
```

[...]

PR #4373, opened 2025-08-12
bugfix: expanding wildcards in config includes (#4354)

Arkoniak:
This bugfix tries to solve regression #4354

As it turns out, initial implementation used `findConfigPath` which was made with a different purpose. So, I had to update `findIncludePath` to use original `tryExpandPath` which provides full set of results instead of only one as `findConfigPath` does. This implementation produces expected result.

During manual test, one should use this file for `memory.jsonc` (there is a typo in original issue)
```
// modules/memory.jsonc
{
  "memory": {
    "interval": 2,
    "format": "foo: {used:0.1f}G",
  }
}
```


agorgl:
@Alexays can we make a bugfix release (0.14.1) with this one please?

Alexays:
LGTM, thanks!

Issue #3879, opened 2025-01-11
[MPRIS module] Widget disappears when ignored player starts playing while non-ignored player is active


Crystal4276:
## Description
When using the `ignored-players` configuration option in the MPRIS module, if one of the ignored players starts playing, the waybar display of the currently active player disappears, even if the current player is not in the ignored list.

## Steps to Reproduce
1. Configure waybar with MPRIS module using `ignored-players`
2. Example configuration:
```json
    "mpris": {
        "format": "{player_icon} {status_icon}  {artist} - {title}",
        "format-paused": "{player_icon} {status_icon} {artist} - {title}",
        "format-stopped": "{player_icon}",
        "player-icons": {
            "default": "",
            "spotify": "",
            "discord": "",
        },
        "status-icons": {
            "paused": "",
        },
         "ignored-players": ["io","chromium","firefox"],
    },
```
3. Start playing media in a non-ignored player (e.g., spotify)
4. While spotify is playing, start playing media in firefox
5. Observe that the spotify display in waybar disappears

## Expected Behavior
- When an ignored player starts playing, the current non-ignored player should remain displayed
- The widget should only disappear when:
  - Using `playerctld` and all active players are in the ignored list
  - Using a specific player that is in the ignored list

## Actual Behavior
The display disappears completely when any ignored player starts playing, regardless of whether there are active non-ignored players.

## Suggested Root Cause from Claude
In the `getPlayerInfo()` method, when using `playerctld`, the code immediately returns `nullopt` if the first active player is in the ignored list, without checking for other non-ignored active players.
<details>
<summary>Proposed fix</summary>

```cpp
auto Mpris::getPlayerInfo() -> std::optional<PlayerInfo> {
  if (!player) {
    return std::nullopt;
  }

  GError* error = nullptr;
  waybar::util::ScopeGuard error_deleter([error]() {
    if (error) {
      g_error_free(error);
    }
  });

  char* player_status = nullptr;
  auto player_playback_status = PLAYERCTL_PLAYBACK_STATUS_STOPPED;
  g_object_get(player, "status", &player_status, "playback-status", &player_playback_status, NULL);

  std::string player_name = player_;
  if (player_name == "playerctld") {
    GList* players = playerctl_list_players(&error);
    if (error) {
      throw std::runtime_error(fmt::format("unable to list players: {}", error->message));
    }
    // > get the list of players [..] in order of activity
[...]

PR #4374, opened 2025-08-12
Fix: Completely ignore ignored players

Ciubix8513:
Fix for  #3879


Alexays:
Thanks for the fix! This is already handled on master via #4846 "fix(mpris): fall back to next non-ignored player and prefer playing players" (merged), which skips ignored players, falls back to the next non-ignored one, additionally prefers a PLAYING player, remembers a last-active fallback, and adds proper unref + null-checks. On top of that, this PR's version would regress the named-player (non-playerctld) case — `found` stays false, so the widget would always be hidden for an explicitly-named player. Closing as superseded by #4846. Thanks again for reporting/fixing the underlying issue! 🙏

PR #4387, opened 2025-08-16
fix: display icons for userwide installed applications

seschwar:
Previously Waybar only displayed icons for applications which were installed systemwide.  Icons were resolved via `.desktop` files in directories specified by the environment variable `XDG_DATA_DIRS`. However the [XDG specification](https://specifications.freedesktop.org/basedir-spec/0.8/#variables) notes that this variable should only consulted **in addition** to `XDG_DATA_HOME`:

> `$XDG_DATA_DIRS` defines the preference-ordered set of base
> directories to search for data files in addition to the
> `$XDG_DATA_HOME` base directory.

This because `XDG_DATA_DIRS` contains only systemwide directories, whereas `XDG_DATA_HOME` contains the userwide directory.  Also including the latter when looking up `.desktop` files allows Waybar to display icons for applications which where installed userwide.

### DIFF
--- a/src/AAppIconLabel.cpp
+++ b/src/AAppIconLabel.cpp
@@ -63,7 +63,8 @@ std::optional<std::string> getDesktopFilePath(const std::string& app_identifier,
     return {};
   }
 
-  const auto data_dirs = Glib::get_system_data_dirs();
+  auto data_dirs = Glib::get_system_data_dirs();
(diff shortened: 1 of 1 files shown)


seschwar:
Note that I am not a C++ programmer, so I don't know if there is a better way to implement this on a technical level. But it works...

Alexays:
Thanks!

Issue #3261, opened 2024-05-11
Feature request: floating point intervals


Spelis:
integers for intervals isnt so nice so should be pretty easy to add

Spelis:
@haug1 this would be pretty simple to add right?

haug1:
I have no particular comment about this. My intuition is telling me it will be simple to implement, but the product will not very useful for most people and millisecond precision will not be ensured. 
But I may be wrong, good luck, if you choose to take it on. :slightly_smiling_face:

Spelis:
> I have no particular comment about this. My intuition is telling me it will be simple to implement, but the product will not very useful for most people and millisecond precision will not be ensured. But I may be wrong, good luck, if you choose to take it on. 🙂

💀 nah i cant I'm not confident in my C++ skills (my skills are pretty much non-existent)

RobertMueller2:
Duplicate of #537

Alexays:
Closing as a duplicate of #537. 🙏

PR #4390, opened 2025-08-16
Support millisecond precision for script intervals

### LINKED ISSUE #537: Fractional interval of module
Is it possible to set an interval of less than 1s? 0.5s, for example

aidansunbury:
Closes #537 and #3261 

Allows for "interval" and "restart-interval" to be doubles, allowing for scripts to be run a maximum frequency of every 1 ms instead of every 1 second.

Also, I don't know any C++ and mostly relied on Claude to get this PR done so apologizes in advance if I did anything stupid.

### DIFF
--- a/include/ALabel.hpp
+++ b/include/ALabel.hpp
@@ -21,7 +21,7 @@ class ALabel : public AModule {
  protected:
   Gtk::Label label_;
   std::string format_;
-  const std::chrono::seconds interval_;
+  const std::chrono::milliseconds interval_;
   bool alt_ = false;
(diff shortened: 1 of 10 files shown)


Alexays:
lgtm, thanks!

DreamMaoMao:
image

This pr caused the unit conversion module for network speed to start calculating incorrectly. As shown in the figure, it was supposed to display MB, but the module showed kb instead

cpuguy83:
Also caused https://github.com/Alexays/Waybar/issues/4521

PR #4395, opened 2025-08-19
[hyprland/workspaces] Taskbar improvements

pol-rivero:
This PR extends my work from #3868 by adding 2 config options that were requested [here](https://github.com/Alexays/Waybar/pull/3868#issuecomment-3193789696) and [here](https://github.com/Alexays/Waybar/pull/3868#issuecomment-3194003406):

```jsonc
"hyprland/workspaces": {
    "workspace-taskbar": {
        // If set to "first", the active window will be moved to the beginning of the taskbar.
        // If set to "last", it will be moved to the end.
        // It will only work if `update-active-window` is set to true.
        // Allowed values: "none" (default), "first", "last".
        "active-window-position": "first",
        // If set to true, the taskbar windows will be added in reverse order (right to left if "orientation"
        // is "horizontal", bottom to top if "vertical"). Default value: false.
        "reverse-direction": true
    }
}
```

### DIFF
--- a/include/modules/hyprland/workspace.hpp
+++ b/include/modules/hyprland/workspace.hpp
@@ -42,7 +42,6 @@ class Workspace {
   bool isPersistentConfig() const { return m_isPersistentConfig; };
   bool isPersistentRule() const { return m_isPersistentRule; };
   bool isVisible() const { return m_isVisible; };
-  bool isEmpty() const { return m_windows == 0; };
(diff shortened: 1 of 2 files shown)


dinhokusanagi:
I tested it at Nixos and it was excellent 
Amazing work


waybar

Alexays:
Thanks!

PR #4400, opened 2025-08-20
hyprland/workspaces: allow sorting by 'output'

eviefp:
Sorting workspaces by output works really well with 'active-only' and 'all-outputs' with multi-monitor setups, allowing to keep the order of the buttons while swapping workspaces.

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -148,13 +148,14 @@ class Workspaces : public AModule, public EventHandler {
   // and doesn't share windows across bars (a.k.a `all-outputs` = false)
   std::map<WindowAddress, WindowRepr, std::less<>> m_orphanWindowMap;
 
-  enum class SortMethod { ID, NAME, NUMBER, SPECIAL_CENTERED, DEFAULT };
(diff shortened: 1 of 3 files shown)


eviefp:
What do you mean by docs? The Wiki, or did I miss some docs in the repo itself?

And if yes, I assume I'll need to update the wiki after this is merged? Or?

zjeffer:
You have to update this file: https://github.com/Alexays/Waybar/blob/master/man/waybar-hyprland-workspaces.5.scd

After it's merged, you can also update this wiki page: https://github.com/Alexays/Waybar/wiki/Module:-Hyprland

eviefp:
Oh, I did edit that file actually, see https://github.com/Alexays/Waybar/pull/4400/files#diff-19fcfa11dfbb35133193b000f1f88e5c337171946f2011a44c39f9cf3741cb7dR117

Happy to add more if needed. I didn't because it doesn't feel like this option needs that much more attention than a line there.

zjeffer:
Oops, not sure how I read over that haha

LGTM

eviefp:
Rebased/fixed merge conflicts. Soft nudge to @Alexays in case that's why this hasn't gotten merged with the recent batch of merges

Issue #3975, opened 2025-03-08
waybar stops talking to hyprland ipc after a display is plugged in/out


gottaeat:
```
$ Hyprland --version
Hyprland 0.47.2 built from branch  at commit 882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff  (version: bump to 0.47.2).
Date: Sun Feb 2 00:47:17 2025
Tag: v0.47.2, commits: 5767
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.2
```

```
$ waybar --version
Waybar v0.12.0
```

`config.jsonc`:
```
{
    "layer": "top",
    "position": "top",
    "height": "",
    "width": "",
    "spacing": 4,
    "modules-left": [
        "hyprland/workspaces"
    ],
    "modules-right": [
        "hyprland/language",
        "tray"
    ],
    "hyprland/language": {
        "format": "{short}"
    },
    "tray": {
        "icon-size": 21,
        "spacing": 1
    }
}
```

hyprland display config:
```
monitor = DP-1, 3840x2160@60, 0x0, 1.5, bitdepth, 8
monitor = HDMI-A-1, 1920x1080@60, 2560x0, 1, bitdepth, 8, vrr, 0
```

full process log:
```
[2025-03-08 15:09:48.443] [info] Using configuration file /home/user/.config/waybar/config.jsonc
[2025-03-08 15:09:48.445] [info] Discovered appearance 'dark'
[2025-03-08 15:09:48.445] [info] Using CSS file /home/user/.config/waybar/style.css
[2025-03-08 15:09:48.447] [info] Hyprland IPC starting
[2025-03-08 15:09:48.447] [info] Loading persistent workspaces from Hyprland workspace rules
[2025-03-08 15:09:48.476] [info] Loading persistent workspaces from Hyprland workspace rules
[2025-03-08 15:09:48.513] [info] Bar configured (width: 2560, height: 24) for output: DP-1
[2025-03-08 15:09:48.513] [info] Bar configured (width: 1920, height: 24) for output: HDMI-A-1
[2025-03-08 15:09:48.519] [info] Bar configured (width: 1920, height: 24) for output: HDMI-A-1

# - - HDMI-A-1 DISCONNECTED HERE - - #

# - - HDMI-A-1 PLUGGED BACK IN - - #

[...]

PR #4408, opened 2025-08-22
Prevent child zombie process from tearing down Hyprland IPC

lfos:
In rare circumstances, we may fork(), e.g., as part of a custom module, and the child process may fail to exec() and exit. In those cases, the IPC destructor will be called in the child process.

Prior to this commit, this call would then result in the shared socket being closed. Prevent this by only closing the socket from the original process.

### DIFF
--- a/include/modules/hyprland/backend.hpp
+++ b/include/modules/hyprland/backend.hpp
@@ -42,6 +42,7 @@ class IPC {
   util::JsonParser parser_;
   std::list<std::pair<std::string, EventHandler*>> callbacks_;
   int socketfd_;  // the hyprland socket file descriptor
+  pid_t socketOwnerPid_;
   bool running_ = true;
 };
 
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -46,9 +46,14 @@ std::filesystem::path IPC::getSocketFolder(const char* instanceSig) {
 IPC::IPC() {
   // will start IPC and relay events to parseIPC
   ipcThread_ = std::thread([this]() { socketListener(); });
+  socketOwnerPid_ = getpid();
 }
 
 IPC::~IPC() {
+  // Do no stop Hyprland IPC if a child process (with successful fork() but
+  // failed exec()) exits.
+  if (getpid() != socketOwnerPid_) return;
+
   running_ = false;
   spdlog::info("Hyprland IPC stopping...");
   if (socketfd_ != -1) {


notpeelz:
I don't think that explanation holds up, and the proposed fix doesn't address the root cause; in a forked child, destructors shouldn't run if `exec()` fails, since glibc's `exit()` doesn't unwind the stack.

Another thing I noticed, but possibly unrelated: `command::open` should use `_exit` instead of `exit` to avoid glibc's `atexit()` callbacks.

The socket should probably be `CLOEXEC` as well, otherwise the fd gets leaked to child processes.

```diff
diff --git a/src/modules/hyprland/backend.cpp b/src/modules/hyprland/backend.cpp
index 2bd3b509..8a29edcf 100644
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -175,7 +175,7 @@ void IPC::unregisterForIPC(EventHandler* ev_handler) {
 std::string IPC::getSocket1Reply(const std::string& rq) {
   // basically hyprctl
 
-  const auto serverSocket = socket(AF_UNIX, SOCK_STREAM, 0);
+  const auto serverSocket = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
 
   if (serverSocket < 0) {
     throw std::runtime_error("Hyprland IPC: Couldn't open a socket (1)");
```

lfos:
> I don't think that explanation holds up, and the proposed fix doesn't address the root cause; in a forked child, destructors shouldn't run if `exec()` fails, since glibc's `exit()` doesn't unwind the stack.

The object is statically allocated, so I'm not sure stack unwinding is relevant. We did a fair bit of debugging in https://github.com/Alexays/Waybar/issues/3975 which shows that the destructor is called from exit() in a child process.

> The socket should probably be `CLOEXEC` as well, otherwise the fd gets leaked to child processes.

That'd likely address the issue too, as we'd eliminate the side effect of the destructor in the forked process.

I'm fine with merging this PR, merging your proposed `SOCK_CLOEXEC` change, or both.

notpeelz:
> The object is statically allocated, so I'm not sure stack unwinding is relevant. We did a fair bit of debugging in #3975 which shows that the destructor is called from exit() in a child process.

Ah I see. I didn't realize `exit()` called the destructors of static objects.

In that case isn't it better to change `exit()` to `_exit()` to avoid calling the exit handlers in the first place?

lfos:
> In that case isn't it better to change `exit()` to `_exit()` to avoid calling the exit handlers in the first place?

The problem is that due to the non-deterministic nature of the issue, we don't know exactly which exit() calls we'd need to change. There are also quite a few waybar modules/dependencies that call exit() outside the waybar code base, and it's plausible some of those are involved.

I think this PR and/or the `SOCK_CLOEXEC` change are the most reasonable solutions.

notpeelz:
> The problem is that due to the non-deterministic nature of the issue, we don't know exactly which exit() calls we'd need to change. There are also quite a few waybar modules/dependencies that call exit() outside the waybar code base, and it's plausible some of those are involved.

Ideally, as little logic as possible should run between `fork()` and `exec()` (or `_exit()` in this case). Using `exit()` instead of `_exit()` in this context can corrupt the parent's state. The responsibility lies with the code that calls `fork()`. This isn't a bug in the Hyprland IPC.

Can you try this patch and see if it fixes the issue?

```diff
diff --git a/include/util/command.hpp b/include/util/command.hpp
index 4b9decaa..4646c6a1 100644
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -113,7 +113,7 @@ inline FILE* open(const std::string& cmd, int& pid, const std::string& output_na
       setenv("WAYBAR_OUTPUT_NAME", output_name.c_str(), 1);
     }
     execlp("/bin/sh", "sh", "-c", cmd.c_str(), (char*)0);
-    exit(0);
+    _exit(0);
   } else {
     ::close(fd[1]);
   }
diff --git a/src/modules/hyprland/backend.cpp b/src/modules/hyprland/backend.cpp
index 2bd3b509..8a29edcf 100644
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -175,7 +175,7 @@ void IPC::unregisterForIPC(EventHandler* ev_handler) {
 std::string IPC::getSocket1Reply(const std::string& rq) {
   // basically hyprctl
 
-  const auto serverSocket = socket(AF_UNIX, SOCK_STREAM, 0);
+  const auto serverSocket = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
 
   if (serverSocket < 0) {
     throw std::runtime_error("Hyprland IPC: Couldn't open a socket (1)");
```

lfos:
@notpeelz - Even only the `SOCK_CLOEXEC` should be sufficient to address the original issue. As mentioned above, I'm fine with replacing the original MR with your version if preferred.

notpeelz:
> Even only the `SOCK_CLOEXEC` should be sufficient to address the original issue.

Hmm, I don't think it's sufficient. CLOEXEC will only close the fd if exec succeeds. The `fork`+failed `exec`+`exit` code path is still problematic.

Can you test the patch I posted above?

lfos:
> Can you test the patch I posted above?

Yes, that patch seems to fix the issue for me.

lfos:
@Alexays - Friendly ping, this fixes a bug that affects many Hyprland users, and it's been >5 weeks since the PR was submitted. Do you have any feedback? Do you prefer the alternative approach proposed by @notpeelz over this one?

Alexays:
Sorry for the delay, lgtm, thanks!

Issue #4184, opened 2025-06-11
[Feature request] river/language module


artyom-smushkov:
Having a way to see current keyboard layout will be very nice to have.

PR #4430, opened 2025-09-01
draft: add generic layout detection to keyboard-state

### LINKED ISSUE #4101: Keyboard layout module for labwc
Hi! Thank you for working on this bar ❤️

The only thing it's missing to be perfect for multilingual users of labwc is keyboard layout indicator. I would really appreciate it.

xi:
Fixes #4101, #4184, and potentially some others. This also has the potential to replace the language modules for sway, hyprland, and niri.

As a first step, I re-implemented the keyboard-state module to use the keyboard information provided by wayland rather than libinput. In a second step, I added layout detection based on that same information.

This works for me (on labwc), but at the same time it sounds too good to be true. So there might well be a huge downside to this approach that I am not aware of.

Currently this is a rough draft and could use some cleanup. But I wanted to get feedback on the general approach before spending too much time.

### DIFF
--- a/include/modules/keyboard_state.hpp
+++ b/include/modules/keyboard_state.hpp
@@ -1,31 +1,28 @@
 #pragma once
 
-#include <fmt/chrono.h>
-#include <gtkmm/label.h>
+#include <gdk/gdk.h>
+#include <glibmm/refptr.h>
+#include <gtkmm/box.h>
+#include <wayland-client.h>
+#include <xkbcommon/xkbcommon.h>
 
-#include <set>
-#include <unordered_map>
+#include <memory>
+#include <string>
 
 #include "AModule.hpp"
 #include "bar.hpp"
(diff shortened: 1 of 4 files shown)


ChrisHixon:
I can't get this to work in either sway or hyprland...

Caps and Num lock status are always shown as unlocked - it doesn't update at the start or when either key is toggled.

The tooltip works, sometimes.

Format codes set to show any aspect of the keyboard layout do not update at the beginning or when layout/variant is changed - it always contains the info from the first configured layout.

I don't know if I'm missing something... I merged this PR into a local copy of the master branch.

I think the stock keyboard_state module mentions adding your user to the input group, but I find that unacceptable and a security risk. I don't know if that's part of the issue or what. I'd love to have a straight up wayland way to do all of these things: num/caps lock status, current keyboard layout info.

PR #4435, opened 2025-09-02
Fix: ext/workspae hidden button was displayed incorrectly

DreamMaoMao:
fix:

https://github.com/Alexays/Waybar/issues/4434

PR #4437, opened 2025-09-03
ext/workspaces: fix appearing of hidden workspaces

### LINKED ISSUE #4434: [Bug]ext/workspace: The hidden button was wrongly displayed when a new workspace in another group was created
https://github.com/user-attachments/assets/7907257c-9609-485e-9065-865c04a50dfd

As can be seen in the video, when I inserted another monitor into my laptop (each monitor was assigned a different group), the workspace button that was already hidden in my original monitor was mistakenly displayed.

jp7677:
Do this by removing the needs-update tracking. The gain was questionable to begin with and turns out it  doesn't work correctly with multiple outputs.

While at it, I've added a "deactivate" action.

Supersedes #4435 
Closes #4434

### DIFF
--- a/include/modules/ext/workspace_manager.hpp
+++ b/include/modules/ext/workspace_manager.hpp
@@ -136,8 +136,6 @@ class Workspace {
   Gtk::Button button_;
   Gtk::Box content_;
   Gtk::Label label_;
-
-  bool needs_updating_ = false;
 };
 
 }  // namespace waybar::modules::ext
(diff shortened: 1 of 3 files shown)


jp7677:
@DreamMaoMao just to be sure, sorting is fine with your setup? It uses a similar tracking (which also doesn’t add that much) and it _might_ have the same issue.

DreamMaoMao:
The sorting is correct for me.

jp7677:
@Alexays gentle ping.

Alexays:
lgtm, thanks!

PR #4448, opened 2025-09-07
Fix: styling: set gamemode icon size and spacing to 0 if use-icon is false

### LINKED ISSUE #2206: Gamemode indicator icon takes up space even when disabled
The indicator icon for the gamemode module takes up space even if `use-icon: false` is set. This can be noticed by launching `GTK_DEBUG=interactive waybar`, finding the gamemode image in the object hierarchy, clicking on it, and noticing how there is a small blue box next to the text. Alternatively, putting a background on `#gamemode` in the CSS shows that there is more spacing to the left of the text than to the right, as that extra space is accounting for the icon (see image). Setting icon-size to 0 helps but does not fix the problem entirely (setting icon-spacing to 0 seems to have no effect).

Gamemode module with uncentered text

winkelnp:
Fixes #2206 

Quite a simple solution, might be a bit crude, is there a better way to do this?

### DIFF
--- a/src/modules/gamemode.cpp
+++ b/src/modules/gamemode.cpp
@@ -53,7 +53,6 @@ Gamemode::Gamemode(const std::string& id, const Json::Value& config)
   if (config_["icon-spacing"].isUInt()) {
     iconSpacing = config_["icon-spacing"].asUInt();
   }
-  box_.set_spacing(iconSpacing);
 
   // Whether to use icon or not
   if (config_["use-icon"].isBool()) {
@@ -64,7 +63,6 @@ Gamemode::Gamemode(const std::string& id, const Json::Value& config)
   if (config_["icon-size"].isUInt()) {
     iconSize = config_["icon-size"].asUInt();
   }
-  icon_.set_pixel_size(iconSize);
 
   // Format
   if (config_["format"].isString()) {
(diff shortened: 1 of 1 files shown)


ninetailedtori (approved):
This change will work fine, and as a patch that will function more than well enough, this could in fact be how we handle use-icon behaviour for *every* module in the project. However there may be a technically "better" way of writing this, adding the check in the label format section instead. See l221 in .../gamemode.cpp, l226 in the changes.

ninetailedtori:
Title of this one should be corrected to specify "gamemode icon size and spacing", unless this will expand to all modules.

winkelnp:
> However there may be a technically "better" way of writing this, adding the check in the label format section instead. See l221 in .../gamemode.cpp, l226 in the changes.

That is a more elegant solution, I can revert/change to that instead? It seems to work the same on my system (tbh why wouldn't it)

ninetailedtori:
Could you upload it to a separate branch in your fork and link it here? Then we can decide if to force push to this PR, perhaps :]

winkelnp:
https://github.com/winkelnp/waybar/tree/gamemode-icon-alt

Realised a smidge late that that branch is technically built off `gamemode-icon`, not off `master`, so it has both commits, I hope that's ok?

ninetailedtori:
Yes, that works well, as a format-based check! I recommend checkout to gamemode-icon, resetting the HEAD on your local gamemode-icon branch to master, then rewriting the changes on the branch in a single commit, because the other branch as you've said includes a commit then the commit that reverts it then adds a separate commit, which is a little messy :] You can force push the single format commit after, and I'll re-approve! ❤️

PR #4460, opened 2025-09-13
fix(cava): don't update immediately after silence

yerlotic:
Sometimes cava updates fft view and then gets stuck for 1 second. This prevents the initial view update

### Video of this issue:

https://github.com/user-attachments/assets/02584eed-f965-40b6-a477-7fe177bb7558

The cava should close and open with an animation, but the animation doesn't play when the fft gets updated too quickly or there is a race condition (audio started playing while the animation was closing)

### Cava config

```jsonc
    "cava": {
        "framerate": 60,
        "autosens": 1,
        "bars": 10,
        "lower_cutoff_freq": 50,
        "higher_cutoff_freq": 10000,
        "stereo": true,
        "reverse": false,
        "bar_delimiter": 0,
        "monstercat": false,
        "waves": true,
        "sleep_timer": 1,
        "format_silent": "quiet",
        "hide_on_silence": false,
        "noise_reduction": 0.77,
        "input_delay": 1,
        "format-icons" : [" ", "▁", "▂", "▃", "▄", "▅", "▆", "▇", "█"],
        "on-scroll-up": "wpctl set-volume @DEFAULT_AUDIO_SINK@ 2%+",
        "on-scroll-down": "wpctl set-volume @DEFAULT_AUDIO_SINK@ 2%-",
    },
```

### Cava style

```css
.module {
    /* ... */
    transition: min-width 1s ease-in-out,
	        width 1s ease-in-out,
	        color .4s,
	        background-color .4s;
}


#cava {
	font-family: 'Hack Nerd Font Propo';
	font-size: 25px;
	transition: min-width 1s ease-in-out,
		    color .4s,
		    background-color .4s;
}

#cava:not(.silent) {
	min-width: 150px;
	color: transparent;
}

#cava.silent {
	color: @accent;
	animation-name: show;
	animation-duration: 1s;
	animation-timing-function: steps(12);
	animation-iteration-count: 1;
	animation-direction: normal;
}

#cava.updated {
	min-width: 150px;
	color: @accent;
}

@keyframes show {
	from {
		color: transparent;
	}
}
```

### DIFF
--- a/src/modules/cava.cpp
+++ b/src/modules/cava.cpp
@@ -159,7 +159,7 @@ auto waybar::modules::Cava::update() -> void {
     // Do transformation under raw data
     audio_raw_fetch(&audio_raw_, &prm_, &rePaint_, plan_);
 
-    if (rePaint_ == 1) {
+    if (rePaint_ == 1 && !label_.get_style_context()->has_class("silent")) {
       text_.clear();
 
       for (int i{0}; i < audio_raw_.number_of_bars; ++i) {


LukashonakV:
Hi @yerlotic , can you please recheck, is this issue still exists in #4493 . This PR is transforms cava module into front/back end.. So backend now is providing silence/update signals. Frontend now using signal subscribing . So might possible now this issue can be solved.

yerlotic:
> Hi @yerlotic , can you please recheck, is this issue still exists in #4493 . This PR is transforms cava module into front/back end.. So backend now is providing silence/update signals. Frontend now using signal subscribing . So might possible now this issue can be solved.

Hi @LukashonakV, with #4493 the buggy behavior shown in the video happens constantly. It never smoothly animates from `quiet` to cava visualization, there is always a sudden jump. When going from cava to `quiet`, everything transitions smoothly

LukashonakV:
Actually on my machine it works well with my configs. See below.
In the same time with @yerlotic config it doesn't look smoothly.

@yerlotic can you recheck with my configs? 
The reason I'm asking for, I applied your patch into my branch and don't see any differences. 

My config
```sh
"cava": {
        "cava_config": "$XDG_CONFIG_HOME/cava/waybar_cava.conf",
        "input_delay": 2,
        "format-icons" : ["▁", "▂", "▃", "▄", "▅", "▆", "▇", "█" ],
        "actions": {
                   "on-click-right": "mode"
                   }
    },
```

My cava config
```sh
`## Configuration file for CAVA.
# Remove the ; to change parameters.


[general]

# Smoothing mode. Can be 'normal', 'scientific' or 'waves'. DEPRECATED as of 0.6.0
; mode = normal

# Accepts only non-negative values.
; framerate = 60

# 'autosens' will attempt to decrease sensitivity if the bars peak. 1 = on, 0 = off
# new as of 0.6.0 autosens of low values (dynamic range)
# 'overshoot' allows bars to overshoot (in % of terminal height) without initiating autosens. DEPRECATED as of 0.6.0
; autosens = 1
; overshoot = 20

# Manual sensitivity in %. If autosens is enabled, this will only be the initial value.
# 200 means double height. Accepts only non-negative values.
; sensitivity = 100

# The number of bars (0-512). 0 sets it to auto (fill up console).
# Bars' width and space between bars in number of characters.
bars = 12
; bar_width = 2
; bar_spacing = 1
# bar_height is only used for output in "noritake" format
; bar_height = 32

# For SDL width and space between bars is in pixels, defaults are:
; bar_width = 20
; bar_spacing = 5

# sdl_glsl have these default values, they are only used to calculate max number of bars.
; bar_width = 1
; bar_spacing = 0


# Lower and higher cutoff frequencies for lowest and highest bars
# the bandwidth of the visualizer.
# Note: there is a minimum total bandwidth of 43Mhz x number of bars.
# Cava will automatically increase the higher cutoff if a too low band is specified.
; lower_cutoff_freq = 50
; higher_cutoff_freq = 10000


# Seconds with no input before cava goes to sleep mode. Cava will not perform FFT or drawing and
# only check for input once per second. Cava will wake up once input is detected. 0 = disable.
sleep_timer = 5


[input]

# Audio capturing method. Possible methods are: 'fifo', 'portaudio', 'pipewire', 'alsa', 'pulse', 'sndio', 'oss', 'jack' or 'shmem'
# Defaults to 'oss', 'pipewire', 'sndio', 'jack', 'pulse', 'alsa', 'portaudio' or 'fifo', in that order, dependent on what support cava was built with.
# On Mac it defaults to 'portaudio' or 'fifo'
# On windows this is automatic and no input settings are needed.
#
# All input methods uses the same config variable 'source'
# to define where it should get the audio.
#
# For pulseaudio and pipewire 'source' will be the source. Default: 'auto', which uses the monitor source of the default sink
# (all pulseaudio sinks(outputs) have 'monitor' sources(inputs) associated with them).
#
# For pipewire 'source' will be the object name or object.serial of the device to capture from.
# Both input and output devices are supported.
#
# For alsa 'source' will be the capture device.
# For fifo 'source' will be the path to fifo-file.
# For shmem 'source' will be /squeezelite-AA:BB:CC:DD:EE:FF where 'AA:BB:CC:DD:EE:FF' will be squeezelite's MAC address
#
# For sndio 'source' will be a raw recording audio descriptor or a monitoring sub-device, e.g. 'rsnd/2' or 'snd/1'. Default: 'default'.
# README.md contains further information on how to setup CAVA for sndio.
#
# For oss 'source' will be the path to a audio device, e.g. '/dev/dsp2'. Default: '/dev/dsp', i.e. the default audio device.
# README.md contains further information on how to setup CAVA for OSS on FreeBSD.
#
# For jack 'source' will be the name of the JACK server to connect to, e.g. 'foobar'. Default: 'default'.
# README.md contains further information on how to setup CAVA for JACK.
#
; method = pulse
; source = auto

method = pipewire
source = auto

; method = alsa
; source = hw:Loopback,1

; method = fifo
; source = /tmp/mpd.fifo

; method = shmem
; source = /squeezelite-AA:BB:CC:DD:EE:FF

; method = portaudio
; source = auto

; method = sndio
; source = default

; method = oss
; source = /dev/dsp

; method = jack
; source = default

# The options 'sample_rate', 'sample_bits', 'channels' and 'autoconnect' can be configured for some input methods:
#   sample_rate: fifo, pipewire, sndio, oss
#   sample_bits: fifo, pipewire, sndio, oss
#   channels:    sndio, oss, jack
#   autoconnect: jack
# Other methods ignore these settings.
#
# For 'sndio' and 'oss' they are only preferred values, i.e. if the values are not supported
# by the chosen audio device, the device will use other supported values instead.
# Example: 48000, 32 and 2, but the device only supports 44100, 16 and 1, then it
# will use 44100, 16 and 1.
#
; sample_rate = 44100
; sample_bits = 16
; channels = 2
; autoconnect = 2


[output]

# Output method. Can be 'ncurses', 'noncurses', 'raw', 'noritake', 'sdl'
# or 'sdl_glsl'.
# 'noncurses' (default) uses a buffer and cursor movements to only print
# changes from frame to frame in the terminal. Uses less resources and is less
# prone to tearing (vsync issues) than 'ncurses'.
#
# 'raw' is an 8 or 16 bit (configurable via the 'bit_format' option) data
# stream of the bar heights that can be used to send to other applications.
# 'raw' defaults to 200 bars, which can be adjusted in the 'bars' option above.
#
# 'noritake' outputs a bitmap in the format expected by a Noritake VFD display
#  in graphic mode. It only support the 3000 series graphical VFDs for now.
#
# 'sdl' uses the Simple DirectMedia Layer to render in a graphical context.
# 'sdl_glsl' uses SDL to create an OpenGL context. Write your own shaders or
# use one of the predefined ones.
; method = noncurses

# Orientation of the visualization. Can be 'bottom', 'top', 'left', 'right' or
# 'horizontal'. Default is 'bottom'. 'left and 'right' are only supported on sdl
# and ncruses output. 'horizontal' (bars go up and down from center) is only supported
# on noncurses output.
# Note: many fonts have weird or missing glyphs for characters used in orientations
# other than 'bottom', which can make output not look right.
; orientation = bottom

# Visual channels. Can be 'stereo' or 'mono'.
# 'stereo' mirrors both channels with low frequencies in center.
# 'mono' outputs left to right lowest to highest frequencies.
# 'mono_option' set mono to either take input from 'left', 'right' or 'average'.
# set 'reverse' to 1 to display frequencies the other way around.
; channels = stereo
; mono_option = average
; reverse = 0

# Raw output target. A fifo will be created if target does not exist.
; raw_target = /dev/stdout

# Raw data format. Can be 'binary' or 'ascii'.
; data_format = binary

# Binary bit format, can be '8bit' (0-255) or '16bit' (0-65530).
; bit_format = 16bit

# Ascii max value. In 'ascii' mode range will run from 0 to value specified here
; ascii_max_range = 1000

# Ascii delimiters. In ascii format each bar and frame is separated by a delimiters.
# Use decimal value in ascii table (i.e. 59 = ';' and 10 = '\n' (line feed)).
bar_delimiter = 0
; bar_delimiter = 59
; frame_delimiter = 10

# sdl window size and position. -1,-1 is centered.
; sdl_width = 1000
; sdl_height = 500
; sdl_x = -1
; sdl_y= -1
; sdl_full_screen = 0

# set label on bars on the x-axis. Can be 'frequency' or 'none'. Default: 'none'
# 'frequency' displays the lower cut off frequency of the bar above.
# Only supported on ncurses and noncurses output.
; xaxis = none
 
# enable synchronized sync. 1 = on, 0 = off
# removes flickering in alacritty terminal emulator.
# defaults to off since the behaviour in other terminal emulators is unknown
; synchronized_sync = 0

# Shaders for sdl_glsl, located in $HOME/.config/cava/shaders
; vertex_shader = pass_through.vert
; fragment_shader = bar_spectrum.frag

; for glsl output mode, keep rendering even if no audio
; continuous_rendering = 0

# disable console blank (screen saver) in tty
# (Not supported on FreeBSD)
; disable_blanking = 0

# show a flat bar at the bottom of the screen when idle, 1 = on, 0 = off
; show_idle_bar_heads = 1

# show waveform instead of frequency spectrum, 1 = on, 0 = off
; waveform = 0

[color]

# Colors can be one of seven predefined: black, blue, cyan, green, magenta, red, white, yellow.
# Or defined by hex code '#xxxxxx' (hex code must be within ''). User defined colors requires
# a terminal that can change color definitions such as Gnome-terminal or rxvt.
# default is to keep current terminal color
; background = default
; foreground = default

# SDL and sdl_glsl only support hex code colors, these are the default:
; background = '#111111'
; foreground = '#33ffff'


# Gradient mode, only hex defined colors are supported,
# background must also be defined in hex or remain commented out. 1 = on, 0 = off.
# You can define as many as 8 different colors. They range from bottom to top of screen
; gradient = 0
; gradient_color_1 = '#59cc33'
; gradient_color_2 = '#80cc33'
; gradient_color_3 = '#a6cc33'
; gradient_color_4 = '#cccc33'
; gradient_color_5 = '#cca633'
; gradient_color_6 = '#cc8033'
; gradient_color_7 = '#cc5933'
; gradient_color_8 = '#cc3333'



[smoothing]

# Percentage value for integral smoothing. Takes values from 0 - 100.
# Higher values means smoother, but less precise. 0 to disable.
# DEPRECATED as of 0.8.0, use noise_reduction instead
; integral = 77

# Disables or enables the so-called "Monstercat smoothing" with or without "waves". Set to 0 to disable.
; monstercat = 0
; waves = 0

# Set gravity percentage for "drop off". Higher values means bars will drop faster.
# Accepts only non-negative values. 50 means half gravity, 200 means double. Set to 0 to disable "drop off".
# DEPRECATED as of 0.8.0, use noise_reduction instead
; gravity = 100


# In bar height, bars that would have been lower that this will not be drawn.
# DEPRECATED as of 0.8.0
; ignore = 0

# Noise reduction, int 0 - 100. default 77
# the raw visualization is very noisy, this factor adjusts the integral and gravity filters to keep the signal smooth
# 100 will be very slow and smooth, 0 will be fast but noisy.
; noise_reduction = 77


[eq]

# This one is tricky. You can have as much keys as you want.
# Remember to uncomment more than one key! More keys = more precision.
# Look at readme.md on github for further explanations and examples.
; 1 = 1 # bass
; 2 = 1
; 3 = 1 # midtone
; 4 = 1
; 5 = 1 # treble
```

yerlotic:
@LukashonakV, here are the videos

Here, the size snaps to wide open without animation:

https://github.com/user-attachments/assets/de9ddc74-920e-497d-8387-ccdcc824c1d8

And here it is animated with css:

https://github.com/user-attachments/assets/60b3a5ee-3fc6-4df7-b3df-ebda1c5c8e52

### Waybar configuration

```
    "cava": {
        "cava_config": "/home/user/.config/cava/waybar_cava.conf",
        "input_delay": 1,
        "format_silent": "quiet",
        "format-icons" : ["▁", "▂", "▃", "▄", "▅", "▆", "▇", "█" ],
        "actions": {
            "on-click-right": "mode"
        }
    },
```

Cava configuration is yours
Css style is the one in this PR description

Edit:

Here is an ffmpeg command for cropping the video on 1280px wide monitor
```shell
ffmpeg -i recording.mp4 -vf "crop=250:45:515:0" -ss 00:00:6 -t 15s -an output.mp4
```

Alexays:
Thanks for this fix! Unfortunately the cava module was completely rewritten since this PR (split into a backend/frontend architecture — `src/modules/cava/cava_backend.cpp`, `cavaRaw.cpp`, etc. in commit 76d3b47f "Cava back/front end transformation"). The polling `update()`/`rePaint_` code this patch targets no longer exists, and silence is now handled natively via a `signal_silence`/`onSilence` handler and a `silence_` flag. So this can't be rebased as-is. Closing as superseded by the refactor — if you still see the immediate-update-after-silence issue on current master, please open a fresh PR against the new cava files and we'll gladly take it. 🙏

PR #4474, opened 2025-09-16
fix(river): hide vacant tags on initial startup

LordMZTE:
Before this, vacant tags would show with `hide-vacant` set on initial startup, because we receive initial tag events from River before we show the bar. In that case, we won't call `set_visible(false)` on the respective buttons because they're not shown yet. This registers the output status listener after we show the bar so we won't miss any events.


Alexays:
Thanks!

PR #4477, opened 2025-09-16
[/resources/waybar.service.in] Adds detailed comments to waybar.service example

ninetailedtori:
This PR adds detailed comments to the systemd waybar.service example as well as a swaync requisite and general unit description.


notpeelz:
waybar.service isn't just an example file, it's installed to `/usr/lib/systemd/user/waybar.service`.

I don't think all of this sway-specific info belongs to this file. Have you considered documenting it on [the wiki](https://github.com/Alexays/Waybar/wiki) instead?

ninetailedtori:
I think that's a good idea tbh, lemme move the PR over to the wiki then :)

PR #4483, opened 2025-09-17
Fix clang build error

jaschiu:
On my Gentoo Linux machine where I use Clang as the default C/C++ compiler, I get this build error for some reason:

```
/usr/include/glibmm-2.4/glibmm/ustring.h:1290:11: error: invalid operands to binary expression ('StreamType' (aka 'basic_ostringstream<char>') and 'const std::_Setfill<wchar_t>')
 1290 |   stream_ << value;
      |   ~~~~~~~ ^  ~~~~~
/usr/include/glibmm-2.4/glibmm/ustring.h:1405:7: note: in instantiation of function template specialization 'Glib::ustring::FormatStream::stream<std::_Setfill<wchar_t>>' requested here
 1405 |   buf.stream(a2);
      |       ^
../Waybar-0.12.0/src/modules/clock.cpp:361:32: note: in instantiation of function template specialization 'Glib::ustring::format<std::ios_base &(std::ios_base &), std::_Setfill<wchar_t>, std::_Setw, std::basic_string<char>>' requested here
  361 |           os << Glib::ustring::format((cldWPos_ != WS::LEFT || line == 0) ? std::left : std::right,
      |
```

The above change fixes it


Alexays:
Thanks, but CI shows this change is not portable — it breaks the build on the glibmm/libstdc++ platforms (alpine, debian):

```
error: no match for 'operator<<' (operand types are 'Glib::ustring::FormatStream::StreamType' {aka 'std::__cxx11::basic_ostringstream<wchar_t>'} and 'const std::_Setfill<char>')
```

The issue is that `Glib::ustring::FormatStream::StreamType` is **`wchar_t`-based** with glibmm/libstdc++ (so it needs `std::setfill(L' ')`) but **`char`-based** with libc++ (where `L' '` fails to compile, which is what you hit locally). Switching `L' '` → `' '` unconditionally just moves the breakage to the other toolchain.

A portable fix needs to not depend on that char type — e.g. format the zero-/space-padded number into a plain `std::string` (via `fmt::format` / a `std::ostringstream`) and feed that to the ustring stream, instead of applying `setfill`/`setw` on the FormatStream directly. Could you rework it that way? Leaving open. 🙏

PR #4485, opened 2025-09-18
fix: handle null GDK window reference in surface initialization

coleleavitt:
This pull request improves the robustness of the `onMap` event handler in `bar.cpp` by adding checks to ensure the GDK window is valid before proceeding with surface initialization. This helps prevent  crashes or undefined behavior when the window is not available.

Error handling improvements:

* Added null checks for the GDK window reference and GDK window object in the `onMap` method of `Bar`, logging warnings and deferring surface initialization if either is unavailable.

fixes:
```bash
(waybar:1514241): Gdk-WARNING **: 20:57:45.295: Couldn't map window 0x560086bf4850 as subsurface because its parent is not mapped.
Segmentation fault (core dumped)
```

### DIFF
--- a/src/AIconLabel.cpp
+++ b/src/AIconLabel.cpp
@@ -36,10 +36,13 @@ AIconLabel::AIconLabel(const Json::Value &config, const std::string &name, const
   box_.set_spacing(spacing);
 
   bool swap_icon_label = false;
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks!

PR #4493, opened 2025-09-20
Cava backend. Comminication using signals

LukashonakV:
Hi @Alexays ,
This PR is a try to transform cava into backend and frontend. 
Backend now is responsible for all incoming data getting/transformation actions. Then using update/silence signals it communicates with frontend. Frontend gets backend using static inst method which gets it possible to have many frontends(because of multiples displays/bars) and single backend. This improves overall performance. 
On author machine even with the single display difference between approaches ~ 15-30% with the single display and 50-100% with 2 displays.


Alexays:
LGTM, thanks!

PR #4516, opened 2025-10-01
fix: right and middle button not work in ext/workspace module

DreamMaoMao:
When parsing the configuration, the configuration of the left button was mistakenly parsed as that of the middle button and the right button, resulting in both the middle button and the right button being able to only trigger the action of the left button


DreamMaoMao:
cc @jp7677

DreamMaoMao:
This is a very minor fix
@Alexays

PR #4530, opened 2025-10-07
[RFC] niri/window: add col/max_col parameters

### LINKED ISSUE #3653: [Feature request] Add Niri window count module
The scrolling nature of Niri means that it's common and expected to have windows (even lots of them) on the current workspace but not visible, placed to the right and left of the currently visible ones. I think that in this case it would be useful to have a little indicator in the bar showing the total number of windows on the active workspace, so one could get the idea of how many windows there are to the right and left without needing to move around. Note that the same principle applies to classic tiling wms too when windows are stacked on top of each other (the "monocle" mode of many wms): for instance, the built-in bar of dwm does exactly this if I recall correctly, showing the number of windows in the tag when in monocle mode.

I implemented this in a dirty shell script to show what kind of output I'd like to have in this module, and also to understand which events needs to be considered when updating the module (well, window opening/closing and workspace switching are kind of obvious):

```bash
#!/bin/sh

idfocused="$(niri msg -j workspaces | jq ".[] | select(.is_focused == true ) | .id")"
num="$(niri msg -j windows | jq "[.[] | select(.workspace_id == $idfocused)] | length")"
printf "%d\n" "$num"

niri msg event-stream |\
    while read -r line; do
        case "$line" in
            "Window opened"*|"Window closed"*|"Workspace focused"*)
                idfocused="$(niri msg -j workspaces | jq ".[] | select(.is_focused == true ) | .id")"
                num="$(niri msg -j windows | jq "[.[] | select(.workspace_id == $idfocused)] | length")"
                printf "%d\n" "$num"
                ;;
        esac
    done
```

Also, maybe someone finds this useful in the absence of a native module (but I don't think performance is great...)

martinetd:
Hello!

Happy waybar user here. I've been told niri now gives away some information about the workspace we could use, so here's to scratching my own itch.
I'm sure someone could do better but it's Good Enough.

This also probably kinda fixes #3653 ? as I think the idea is the same, avoiding the workspace growing too wide.
Happy to try to find a few niri users to ping for review if you'd like.

Commit message below.

--------------------

niri v25.08 added window positions to IPC, so waybar actually has the information stored in json around.

This handles the associated WindowLayoutsChanged IPC message to keep the position up to date in memory, and adds a couple of format messages to show the current column, so one can dislpay where we are in the workspace (in the hope of improving this feeling of being lost...)

Ideally a visual representation of the workspace like the one illustrated in the PR[1] would be even better but this will probably do for now.

Link: https://github.com/YaLTeR/niri/pull/1265#issuecomment-2724621758 [1]


martinetd:
Thank you!

PR #4536, opened 2025-10-09
feat: sway/workspaces: add custom workspace sorting functionality

Esensats:
Added `custom-sort` property to `sway/workspaces` that allows the user to specify their own order for each of the workspaces.

### BEFORE:
2025-10-09_06-50_BEFORE

### AFTER:

---

Helps when alphabetical sort is not ordering how you want it to.


Esensats:
Rebased the PR onto the latest master branch (b4854f96). Works as expected

PR #4539, opened 2025-10-09
bump: fmt 11.0.2 -> 12.0.0

ninetailedtori:


owenpmckenna:
Checks are failing because the hashes do not match. From logs:
```
ERROR: Subproject fmt is buildable: NO

meson.build:71:6: ERROR: Incorrect hash for source:
 6cb1e6d37bdcb756dbbe59be438790db409cdb4868c66e888d5df9f13f7c027f expected
 aa3e8fbb6a0066c03454434add1f1fc23299e85758ceec0d7d2d974431481e40 actual.
```
If those hashes are changed it should work. No idea how it compiled on fedora and debian. That seems like a problem, actually (if it's not checking hashes).

owenpmckenna:
If anyone wants to make the change, the first hash is aa3e8fbb6a0066c03454434add1f1fc23299e85758ceec0d7d2d974431481e40, and the second should be 307f288ebf3850abf2f0c50ac1fb07de97df9538d39146d802f3c0d6cada8998.

ninetailedtori:
I'll patch that, one sec.

ninetailedtori:
> If those hashes are changed it should work. No idea how it compiled on fedora and debian. That seems like a problem, actually (if it's not checking hashes).

I'll be honest, I'm not even certain how it compiled flawlessly on my own arch as well lol. That's definitely an issue worth checking out.

PR #4541, opened 2025-10-10
backlight: gracefully handle a null epoll device

aanderse:
recently i switched my laptop from using `eudev` to `mdevd` with [libudev-zero](https://github.com/illiliti/libudev-zero), a daemonless replacement for `libudev`, and `waybar` started crashing on me

i tracked it down to [this](https://github.com/Alexays/Waybar/blob/559079e9a6afda77754afaf7c8d3f588c1d6206d/src/util/backlight_backend.cpp#L204) line

apparently a [similar issue](https://codeberg.org/dnkl/yambar/issues/109)  existed in `yambar` and was fixed as well

---

i hope you will consider accepting this patch so i may continue to enjoy `waybar` without maintaining my own patch set...

i appreciate your time and consideration in reviewing this PR :bowing_man:


aanderse:
thank you very much!

PR #4543, opened 2025-10-10
feat(wlr/taskbar): add `homogeneous` option to evenly distribute button widths

mbroemme:
## Summary
This PR adds an optional `homogeneous` boolean to the `wlr/taskbar` module that evenly distributes the width of all task buttons across the module's available space. When enabled, each task button expands to take an equal share of the enclosing Gtk box (e.g., with a 1000px center box: 1 task = 1000px, 2 tasks = 500px each, 3 tasks ≈ 333px each, etc.).

## Configuration
It is off by default and enabling requires:

```jsonc
"wlr/taskbar": {
  "expand": true,        // allow the module to take available width
  "homogeneous": true    // evenly divide buttons
}
```

If the used bar supports it, also ensure the center container can grow:
```jsonc
"fixed-center": false
"expand-center": true
```

* No breaking changes. The new behavior is opt-in.
* Existing configs continue to behave the same.

## Behavior Details
* Buttons automatically reflow as tasks are created/closed, maintaining equal widths.
* Minimum sizes still apply (GTK will avoid shrinking below widget min width).
* If the surrounding layout does not let the taskbar expand (e.g., missing `"expand": true`), you won't see full-width distribution.
* Works with dynamic task counts; no explicit width math in Waybar - delegated to GTK homogeneous layout for robustness.

## Screenshots
*Before:* different sized buttons by default.
image

*After:* with `"homogeneous": true`, `"expand": true`, `"expand-center": true` and `"fixed-center": false` (visuals depend on theme and spacing).
image
image

## Docs
* Add the new `homogeneous` key to the `wlr/taskbar` module documentation with accepted values and defaults.

## Testing
1. Enable the option:
```jsonc
"wlr/taskbar": { "expand": true, "homogeneous": true }
```
2. Ensure the bar's center region can expand (e.g., `"expand-center": true`).
3. Open 1, 2, 3, … windows and verify each button takes ~1/N of the available width.
4. Close/launch apps and confirm widths update live.
5. Toggle the option off and confirm legacy behavior returns.

## Future
Very long window titles will let the bar grow beyond its configured `width` in the configuration file. It is the same as before. This will be addressed in a separate PR to ensure ellipsization/clipping to truncate text and fit the box size.

## Motivation
Users who keep many windows open want their taskbar buttons to automatically resize to fit the module width, avoiding overflow and uneven spacing. GTK provides native support for homogeneous children; this PR exposes that via a config flag for the taskbar.


mbroemme:
Any comments on this one (good/bad/neutral)?

Issue #3770, opened 2024-11-16
How can I fix the width of the waybar? When opening multiple tabs, the waybar will automatically lengthen and the module on the right will be covered. How can I solve this problem?


linux36524:
How can I fix the width of the waybar? When opening multiple tabs, the waybar will automatically lengthen and the module on the right will be covered. How can I solve this problem?

saltasatelites:
AFAIK you can only configure the minimum length. In my case for some modules I have calculated the maximum length that they need, and I have put that value as minimal length using "min-width: "

saltasatelites:
As far as I know, you can only configure the minimum length. In my case for some modules I have calculated the maximum length that they need, and I have put that value as minimal length using "min-width:"

linux36524:
> As far as I know, you can only configure the minimum length. In my case for some modules I have calculated the maximum length that they need, and I have put that value as minimal length using "min-width:"
> 
> #memory { background-color: #a3a2a2; padding: 0 5px; min-width: 75px; }

It has nothing to do with the module. What I mean is that the waybar does not have a fixed width. It will change under the influence of other modules.

PR #4545, opened 2025-10-10
feat(wlr/taskbar): add `truncate` option to ellipsize long button text

mbroemme:
## Summary
This PR adds an optional boolean `truncate` to the `wlr/taskbar` module. When enabled, task button text is ellipsized (… at the end) whenever the label would exceed the button's available width. This keeps the taskbar tidy when window titles are long or when space is tight. It will fix issue #3770 for `wlr/taskbar`.

## Configuration
It is off by default and enabling requires:

```jsonc
"wlr/taskbar": {
  "truncate": true
}
```

* No breaking changes. The new behavior is opt-in.
* Existing configs continue to behave the same.

## Screenshots
*Before:* different length of taskbar button label.
image

*After:* with `"truncate": true` (visuals depend on theme and spacing).
image

*Before:* huge amount of taskbar buttons.
image

*After:* with `"truncate": "true"`
image

## Docs
* Add the new `truncate` key to the `wlr/taskbar` module documentation with accepted values and defaults.

## Testing
1. Enable the option:
```jsonc
"wlr/taskbar": { "truncate": true }
```
2. Enforce a static width with:
```jsonc
"width": 1280
```
3. Open an app with a very long title (e.g., a browser tab with a long page title).
4. Reduce available taskbar width (resize bar/monitor or open more tasks).
5. Verify the label is a single line and ends with an ellipsis rather than wrapping or pushing layout.

## Motivation
Long window titles can cause layout jitter or overflow in the taskbar. Providing a first-class switch to ellipsize labels gives users a simple, predictable way to keep the module compact, regardless of how many tasks are present or how verbose their titles are.

### DIFF
--- a/man/waybar-wlr-taskbar.5.scd
+++ b/man/waybar-wlr-taskbar.5.scd
@@ -84,6 +84,11 @@ Addressed by *wlr/taskbar*
 	default: false ++
 	If set to true, task buttons stretch to fill the available space in the taskbar and long titles are ellipsized to fit. Only takes effect on a horizontal bar; on a vertical bar the buttons keep their content-based size. If set to false, buttons are sized to their content.
 
+*truncate*: ++
+	typeof: bool ++
+	default: false ++
+	If set to true, the task button text will be ellipsized (truncated with …) when the available button width is smaller than the label text.
+
 *on-click*: ++
 	typeof: string ++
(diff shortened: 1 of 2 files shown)


mbroemme:
Any comments on this one (good/bad/neutral)?

Issue #2684, opened 2023-11-22
[Feature Request] Add customgraph module


otaj:
Hi, thank you for the awesome bar!

I was wondering, would it be possible to add a module creating a graph such as for example [CPUGraph from QTile](https://docs.qtile.org/en/stable/manual/ref/widgets.html#cpugraph) or [Graph from eww](https://elkowar.github.io/eww/widgets.html#graph). Ideally of course with being able to feed it custom data through continuous script similarly as custom module works.

I can now simulate it with the help of [Sparks font](https://github.com/aftertheflood/sparks), but it's a bit janky and it doesn't support all possible styling option.

I can help out a bit, however my C++ is very rusty so I wouldn't be the best help :)

Thank you!

vincentaxhe:
that would be nice

Gazer:
Hello, I've started to experiment to add support for real graphs (for now line only), this is how the CPU usage looks like:

<img width="223" height="48" alt="Image" src="https://github.com/user-attachments/assets/e1bde7ab-7abb-485c-84ae-f45e7e20ce0b" />

You can check the code at https://github.com/Gazer/Waybar/tree/feat/basic-cpu-graph

imiric:
@Gazer That looks great! Would you be willing to create a PR for it?

And, more importantly, would @Alexays be interested in merging it?

Gazer:
@imiric the idea is to end with a PR, right now I´m exploring to add a custom-graph similar to the custom module so you can use an external script to draw a graph and I want to add two more types: gauge and bars. E.g. for temperature a gauge has more sense a disk usage with bars or something like that.

vincentaxhe:
@Gazer I'm interested in your work, the new module and how to configure it.

Yilmaz4:
@Gazer any progress on this?

Gazer:
I'm doing the final testing and polish but I may be able to do a PR to get feedback next week. I've working three graph types with native or custom modules. 

This config:
```
"custom-graph/memory": {
    "interval": 60,
    "graph_type": "gauge",
    "width": 52,
    "exec": "/home/gazer/mem.sh",
    "signal": 8,
    "return-type": "json"
  },
  "custom-graph/disk": {
    "interval": 2,
    "graph_type": "bar",
    "width": 10,
    "exec": "/home/gazer/disk.sh",
    "signal": 8,
    "return-type": "json"
  },
  cpu_graph": {
    "interval": 2,
    "graph_type": "line",
[...]

PR #4588, opened 2025-10-24
feat: add customgraph module

Gazer:
This PR add a graph module to be able to show data as graph and not just text as requested in https://github.com/Alexays/Waybar/issues/2684.

screenshot-2025-10-25_00-14-10
screenshot-2025-10-25_00-26-21
screenshot-2025-10-25_00-25-31

I've added:
* `AGraph` module that can render a line, bar or gauge graphs. It render values from 0 to 100 and expect the caller normalize the value.
* A `CpuGraph` module that show current CPU as a lines graph
* A `CustomGraph` that allow usage of external scripts to show data by using the percentage field of the payload. 

The module uses the css colors and modify the alpha to create color variation, making it play nice with almost all themes.

I've been using this for a couple of weeks now on Omarchy without any problem.

### DIFF
--- a/include/AGraph.hpp
+++ b/include/AGraph.hpp
@@ -24,7 +24,6 @@ class AGraph : public AModule {
   Gtk::DrawingArea graph_;
   std::deque<int> values_;
   uint16_t datapoints_ = 20;
-  uint16_t y_offset_ = 0;
   GraphType graph_type_ = GraphType::LINE;
 
   void addValue(const int n);
--- a/include/AGraph.hpp
+++ b/include/AGraph.hpp
@@ -1,0 +1,50 @@
+#pragma once
+
+#include <glibmm/markup.h>
+#include <gtkmm/label.h>
+#include <json/json.h>
+
+#include <deque>
+#include <vector>
+
+#include "AModule.hpp"
+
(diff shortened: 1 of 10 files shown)


bend-n:
setting the `datapoints` seems to cause a failure:
```c++
/usr/include/c++/15.2.1/bits/stl_deque.h:1643: void std::deque<_Tp, _Alloc>::pop_front() [with _Tp = int; _Alloc = std::allocator<int>]: Assertion '!this->empty()' failed.
```
other than that, works great!

wmvanvliet:
I don't think `graph_type` is documented in the man page

wmvanvliet:
Hi @Gazer, thanks for your great work on this graphing module! I'm now using it in my actual setup. Other then the documentation issue, I noticed that the graphs would not respect any `margin` CSS directives, which prevented it from sitting nicely among the other modules in my bar. The solution turned out to be to wrap the `GTK::DrawingArea` in a `GTK::Box` and apply the styling to the box instead of the graph. This way, it supports all the things we expect, such as a background color and a border too :) You can check out my changes here: https://github.com/wmvanvliet/Waybar/tree/feat/extend-cpu-graph

Issue #3682, opened 2024-10-12
False battery level calculation


auroraanna:
I am only noticing this issue because my laptop has two batteries **of different maximum energy** they can store. While upower correctly takes into account each battery's maximum amount of energy it can store to calculate a combined battery level, waybar's battery module just averages over the two battery percentages. This is **wrong**. Batteries should not be expected to have the same maximum amount of energy they can store.

There is no lack of data here. /sys/class/power_supply exposes each battery's **absolute** current amount of energy and maximum amount of energy.

i-lost-my-bagel:
https://github.com/Alexays/Waybar/wiki/Module:-Battery

the option `weighted-average` needs to be set to true for batteries with different capacities

why the fuck this isn't the default is beyond me

Alexays:
Closing — this is fixed/implemented on current master and will be in the next release. Please reopen if it still occurs after updating. 🙏

PR #4591, opened 2025-10-26
Improve multiple battery support

rrrnld:
- Adds information about `weighted-average` to the manpage
- Defaults `weighted-average` to true, following the principle of least astonishment

I'm not sure why the configuration option was added in the first place, because a non-weighted average would sometimes just give wrong numbers, or am I missing something? When would a non-weighted average be desirable?

See also https://github.com/Alexays/Waybar/issues/3682


Alexays:
Thanks!

PR #4595, opened 2025-10-26
fix(niri): Increase named workspace icon priority

silverdyne:
This PR aims to change the priority of workspace icons for the Niri `workspaces` module, where a **named** workspace's icon will take priority over all other types **except** the urgent one.

According to the [Niri documentation itself](https://yalter.github.io/niri/Configuration%3A-Named-Workspaces.html):
> Contrary to normal dynamic workspaces, named workspaces always exist, even when they have no windows.

Named workspaces behave like "pinned" workspaces, which are always present: I think this change reflects this behavior better than the current one, since their own icon will now always be shown (if defined).

Of course, I'd like to hear what other Niri users think of this and I'm open to make changes if necessary. Cheers!

### DIFF
--- a/src/modules/niri/workspaces.cpp
+++ b/src/modules/niri/workspaces.cpp
@@ -174,16 +174,16 @@ std::string Workspaces::getIcon(const std::string& value, const Json::Value& ws)
 
   if (ws["is_urgent"].asBool() && icons["urgent"]) return icons["urgent"].asString();
 
+  if (ws["name"]) {
+    const auto &name = ws["name"].asString();
+    if (icons[name]) return icons[name].asString();
+  }
+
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks! After review we'd rather not merge this: it moves the per-name icon lookup ahead of `active`/`focused`/`empty`, which would make niri's icon priority **inconsistent with the hyprland/workspaces module**, where `active` intentionally takes precedence over a per-name icon (see `Workspace::selectIcon`: urgent → active → special → name → …). Keeping the two modules consistent matters more here than the reordering. If you have a concrete use case that needs name-priority, let's discuss it as a configurable option — feel free to reopen with that. 🙏

PR #4643, opened 2025-11-17
(feat) add tray icon ordering

yangyingchao:
With this commit, we can place specified tray items in a fixed position.

For example, I can position the WeChat item at the far right. This way, even when other applications are launched and new items are added to the system tray, I can always simulate a mouse click on the WeChat icon correctly.


nouritsu:
This would be very useful. One of the few (VERY FEW) things that I liked about Windows was that I could remember where my system tray icons were and click to them without looking. +1 <3

Issue #4276, opened 2025-07-07
Module-specific packages are listed as required dependencies


CelDaemon:
I recently installed waybar, and noticed that multiple new desktop entries appeared in my launcher.

Most of these don't matter, but `xgps` stood out to me as I do not have a GPS in my laptop.

I understand these dependencies are required to use some of the modules provided by waybar, but they are not necessary for the core program to work when these modules are not in use.

I'd like to be able to uninstall these packages without making pacman ignore dependency checks, would it be possible to mark these dependencies are optional?

CelDaemon:
The added tags are incorrect, this issue affects all modules that depend on external packages.

Vescrity:
Mark dependencies optional or not is the job of the package's maintainer of your distribution.  

You can make a package by yourself to "provide" `gpsd`, and then you can uninstall it.  

And then you'll find that waybar refuse to launch as it needs `libgps.so.31` even if you don't use gps module.  

So I uncompressed the `gpsd` package and then copy the `libgps.so.31` to `/usr/lib/` , it can work now.

But you'll have a file won't be tracked by `pacman`, so: 

You can make a package by yourself, "provide" `gpsd` (write that in `PKGBUILD`), and put `libgps.so.31` to `usr/lib`, then install it.  

However, you may need to update it manually by yourself when `gpsd` upgrade with a new `.so.*` filename.

Bootjewolf:
Isn't it a possibility to dynamically load these libraries only when they are included in the current configuration. Or would this require a major rewrite in the way modules and their libraries are linked?

CelDaemon:
My apologies, I didn't realise it was required to actually run. I had made this issue as I thought the AUR package was handled by the actual waybar project.

Aside from that, is there a reason gpsd isn't loaded at runtime when the module is enabled? The normal gpsd package causes clutter in some launchers, which seems silly when my device doesn't support it in the first place.

qbe:
It seems to me that at least some of the modules requiring bigger or less common dependencies should be split from the waybar binary into some form of dynamically loadable module.

this  would let maintainers declare these module-specific dependencies as "optional", or even split the modules into seperate packages.

since there is already a cffi interface this should afaik at least be possible. All compiled modules derived from the waybar source could (should ?) reside in /usr/lib{arch}/waybar/ , and should be auto-loaded if present in the configuration, without specifying their path as you would for a custom cffi module.

[...]

PR #4587, opened 2025-10-23
cava nonsafe threads

LukashonakV:
Dear @Alexays , 
This PR is based on investigation of the #4567 . 
Thanks to @DaDiamondArmor and @mutoroglin I realized backend and frontend was not implemented as well.
During investigation I found non safe threading that's why it is segfaults from time to time, freezes and race conditions.
Now backend was deeply reworked. Dev tests shown that module is very stabled now. In additional it was fixed an issue when audio server for some reason is restarted. No need more to restart in that case waybar. cava will get this.
But the most surprising for me was - the significant performance benefit. Even with the single monitor CPU consumption decreased 10x times. I wasn't expected such big improvement.
In the same time it was again updated libcava because of -> karlstav/cava#698 . I also provided big rework to the upstream in order to avoid of memory leaking. So for now waybar + cava - the only one small memory leak still exists because of issue somewhere deep in the pipewire driver.
Finally this PR closes #4567 
Thank you

PR #4661, opened 2025-11-26
draft: split off gps and cava modules into run-time loaded shared objects

qbe:
this is a proof-of-concept fix for https://github.com/Alexays/Waybar/issues/4276

There is probably some clean-up work to do, but at this stage I would like some feedback.

Some motivations:

- I decided against dynamically loading libgps / libcava, as that would mean having to implement dynamic loading for each module that wants to use it. Dynamically loading the entire modules makes more sense to me.
- I tried adding the reap_mtx / reap parameters to just the AModule update() call, but this would have caused a bunch of functions aside from all update() calls to need the two parameters as well. Passing them just once in the constructors seems more reasonable to me.

### DIFF
--- a/include/AAppIconLabel.hpp
+++ b/include/AAppIconLabel.hpp
@@ -9,16 +9,16 @@ namespace waybar {
 
 class AAppIconLabel : public AIconLabel {
  public:
-  AAppIconLabel(const Json::Value &config, const std::string &name, const std::string &id,
-                const std::string &format, std::mutex& reap_mtx, std::list<pid_t>& reap,
-                uint16_t interval = 0, bool ellipsize = false,
-                bool enable_click = false, bool enable_scroll = false);
+  AAppIconLabel(const Json::Value& config, const std::string& name, const std::string& id,
+                const std::string& format, std::mutex& reap_mtx, std::list<pid_t>& reap,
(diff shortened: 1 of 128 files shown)


qbe:
The generated shared libraries, waybar-module-cava.so and waybar-module-gps.so are relatively small, 490KB resp. 482KB.

On a system without libgps / libcava or without the respective waybar-module-*.so file, loading of the module (implemented in factory.cpp) fails gracefully, and the module is disabled.

This should be enough to make libgps / libcava _optional_ dependencies in packaging systems that support this distinction.

LukashonakV:
Hi @qbe , can you please check is it aligned with the PR #4587

qbe:
> Hi @qbe , can you please check is it aligned with the PR #4587

rebasing your branch onto mine seems to work, I cannot see any failures in build or use, and the `readelf -d` output on the three elf objects is also as expected. Please check out https://github.com/qbe/Waybar/tree/cavaIssue%234567-rebase and see if it works for you.

qbe:
after this whole clang-format debacle, I rebased my work entirely. The end result is identical, but it is much more readable, simply because I started with clang-format -ing the existing code, and kept everything formatted correctly with every commit. I am hesitant to force-push over this branch, which is why I am closing this pull request and opening a new one.

PR #4667, opened 2025-11-28
split off gps and cava modules into run-time loaded shared objects

qbe:
rebased my work from https://github.com/Alexays/Waybar/pull/4661 , a fix for https://github.com/Alexays/Waybar/issues/4276 .

@Alexays , I would appreciate your input on this


qbe:
cava-glsl seems to work, whether to load the glsl code (and therefore its dependencies) when the waybar-module-cava.so is loaded is currently a compile-time option (e.g. HAVA_LIBCAVAGLSL controls whether the waybar-module-cava.so has a load-time dependency on epoxy).

This could be reworked into another use of dlfcn.h -based dynamic loading, to enable use of cava "raw" mode on systems that don't have epoxy installed with the same package that enables cava glsl usage.

@Alexays I know that this branch already contains lots of changes, but I am willing to keep it synchronized to let you merge smaller or more self-contained changes quicker. I would nevertheless be interested in your opinion on the whole runtime-loading-for-easier-packaging thing, to see if I have a chance of ever getting this (or something like it) merged.

qbe:
closed in favor of rebase ( https://github.com/Alexays/Waybar/pull/4735 )

PR #4679, opened 2025-12-04
fix(cava): dispatch GTK operations to main thread

BurgessTG:
The cava backend emits signals from a worker thread, but GTK widget
operations must run on the main thread. This causes intermittent
SIGSEGV/SIGABRT crashes with stack traces through pixman_region32_union
and gtk_label_set_markup.

This wraps all GTK operations in `onUpdate()` and `onSilence()` with
`Glib::signal_idle().connect_once()` to safely dispatch them to the
main thread.

**Additional improvements:**
- Cache style context to avoid repeated `get_style_context()` calls
- Check `has_class()` before removing classes
- Use early return pattern in `onSilence()`
- Use cached `format_silent_` instead of re-reading config

Tested on Arch Linux with 15fps cava visualization - stable with no crashes.

### DIFF
--- a/src/modules/cava/cava.cpp
+++ b/src/modules/cava/cava.cpp
@@ -1,5 +1,6 @@
 #include "modules/cava/cava.hpp"
 
+#include <glibmm/main.h>
 #include <spdlog/spdlog.h>
 
 waybar::modules::cava::Cava::Cava(const std::string& id, const Json::Value& config)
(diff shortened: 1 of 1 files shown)


LukashonakV:
@BurgessTG , why do you do the same work? See PR #4587

BurgessTG:
Tbh, I was adding cava and was experiencing crashes on my own behalf. Instead of looking for a fix or checking I found one pretty quick. Wanted to offer it out as different option 😄

LukashonakV:
@BurgessTG, thank you. I didn't realize in my PR with some enhancements dispatching was a part of backend.. where basically it should be the logic of the frontend. Original PR is fixed for now. 
Please close you PR in order do not clutter the list of incoming PR.
Thank you a lot

LukashonakV:
Hi @BurgessTG , if it's not big deal for you, can you please close the PR? Thanks.

LukashonakV:
Hi @BurgessTG , if it's not big deal for you, can you please close the PR? Thanks.

BurgessTG:
Yeah! Just curious, what other fixes are wrapped into yours?

PR #4682, opened 2025-12-07
Cava GLSL

LukashonakV:
Dear @Alexays , I'm happy to introduce one more cava frontend module which implements OpenGL frontend logic and provides new type of user experience of the waybar usage.
If this PR is confirmed and taken, then PR #4587 could be skipped. 
man for the new module will be provided further(wiki as well)

https://github.com/user-attachments/assets/0c15602a-0d3b-4899-bd42-7a5537113566


LukashonakV:
Looks like epox needs to be reordered. Will fix today

LukashonakV:
> Looks like epox needs to be reordered. Will fix today

Fixed now

LukashonakV:
Hi @Alexays , there are some improvements. 

1. It's cava frontend which now manages which frontend should be using is based on the cava configuration
2. Fixed some technical issues
3. Fully covers PR #4587 
----
Next set of waybar configurations can be used to have the same result
[waybar_configs.tgz](https://github.com/user-attachments/files/24058378/waybar_configs.tgz)

Now result is:

https://github.com/user-attachments/assets/ec4f675c-89a1-43a9-ac2c-b717f6c6a06e

LukashonakV:
One more example where the style looks quite good

https://github.com/user-attachments/assets/93381451-4a23-415a-b4ee-1b93639ee4ad

Using this waybar config: 
[Uploading waybar_config#3.tgz…]()

Alexays:
Thanks!

qbe:
@LukashonakV I am trying to rebase https://github.com/Alexays/Waybar/pull/4667 onto these changes, but this is somewhat difficult since this PR added _another_ compile-time flag.

Can you tell me how GLSL can be enabled in the config ? I could not find changes of manpages in this PR.

EDIT I checked my cava config, but no matter what output method I write in there, the cava module always fails with 'Unknown Module', on master branch and on my branch.

If you have time, I would be thankful if you could check out my branch https://github.com/qbe/Waybar/tree/dynamic-modules-cleanup

LukashonakV:
GLSL follows the same flag as usual cava. There's only 1 additional library dependency: `epoxy`. You can check wiki now. In short: which frontend is to be used is managing by the [output] section, `method` option in the cava config which is used by the waybar config

qbe:
I have epoxy installed, but no matter what I set the "method" to, it always fails with
`[warning] module cava: Disabling module "cava", Unknown module` , even when running a master build.

Can you send me a cava config file that works for you ? Or can you check out my branch if it works for you ? (you need to set LD_LIBRARY_PATH to the build directory for running a waybar build from my branch)

LukashonakV:
As I mentioned,  description,  examples,  demos you can find [wiki cava GLSL](https://github.com/Alexays/Waybar/wiki/Module:-Cava:-GLSL)
Feel free to bring corrections,  clarifications to make WIKI more useful 

----

If wiki also doesn't help you, please tag me once again

Also please provide meson configuration log. Might possible it has some highlights what's wrong

qbe:
ah, I misunderstood the cava config format (I thought only lines starting with `#` were comments). Both master branch and my branch work with your example config no. 1

I would still greatly appreciate if you could test my branch, as you are much more familiar with cava than me

LukashonakV:
What is the issue with your branch?

qbe:
do you mean Issue to mean "problem", or "thing it tries to fix" ? My branch is a rework to load gps and cava dependencies at runtime, to enable to turn these into optional dependencies in packages. See https://github.com/Alexays/Waybar/issues/4276 . I just want to make sure that my changes don't break cava usage, and you seem to be most familiar with that code.

LukashonakV:
Got you. Not sure I have time today. Let me check,  once I have time

qbe:
Excellent, thank you

(remember, LD_LIBRARY_PATH should be set to the path of the waybar build directory, where the new waybar-module-cava.so and waybar-module-gps.so reside, to enable the dynamic loader to find them at runtime without placing them into /lib or /usr/lib )

PR #4692, opened 2025-12-10
fix(hyprland/workspaces) Special Workspace Multi-Monitor Support

aaronred345:
Previously, if you were using hyprland's special workspaces with multimonitors, the display for them wouldn't update if you opened them on different monitors.
I've implemented a somewhat hacky solution by reloading hyprland everytime the special workspace event is triggered.


Alexays:
The special-workspace active-icon logic and the CSS class are reasonable, but injecting `m_ipc.getSocket1JsonReply("reload")` into `onSpecialWorkspaceActivated` reloads the **entire Hyprland config** on every special-workspace toggle — that's a heavy, surprising side effect. A local `update()` / `dp.emit()` is what's expected to refresh the module. Could you drop the `reload` call and refresh the module state directly instead? Happy to take the icon/CSS improvements once that's addressed. 🙏

aaronred345:
Hey, thanks for both the feedback and fixing it. I'm purely a hobbyist so I don't always keep up on my Git notifications.

PR #4705, opened 2025-12-18
fix: prevent duplicate layer surfaces on monitor hotplug

rwwiv:
This PR fixes duplicate waybar layer surfaces appearing after monitor hotplug events.

The `signal_monitor_added` and `signal_monitor_removed` handlers were never disconnected during SIGUSR2 reload. Each reload accumulated additional handlers, so when monitors were hotplugged, all accumulated handlers fired and created duplicate layer surfaces.

Changes:
- Store signal connections as class members
- Disconnect them before reconnecting in `bindInterfaces()`
- Clear stale `outputs_` on reload


Alexays:
Thanks!

PR #4708, opened 2025-12-19
nix: bump cava again

khaneliman:
Fix nix build after https://github.com/Alexays/Waybar/pull/4682


flyingpeakock:
Tested and it works. Without this my nix build fails.

PR #4712, opened 2025-12-21
feat: add support for bt peripheral device battery

AntoineGS:
Allows the use of {device_battery_percentage_peripheral} to show the battery level of a peripheral device, which is most commonly seen in split keyboards.


AntoineGS:
bumping this :P Been using it since December and it's been working great.

Issue #4678, opened 2025-12-02
[feature] [wireplumber] Allow selecting a source/sink by (node) name instead of the default source/sink


Omar007:
Currently it does not seem possible to have the wireplumber module operate on a specific source/sink node, only on whichever one happens to be the default.  

It would be nice if it was possible to configure the module to operate on a specific node, being able to select which node with something like a `node-name` and/or `node-alias` property in the module's config.

PR #4714, opened 2025-12-21
Implemented "only-physical" parameter in the wireplumber module.

PickNicko13:
It makes it find the first node that has adevice.id linked to the default output if the default output doesn't have it (i.e. it is a virtual output).

Useful if you want to control (and see) the volume of the actual output and not the effect sink.

May need some polishing and/or style fixing, but it works.

Somewhat related to [#4678 ](https://github.com/Alexays/Waybar/issues/4678), as it may cover some usecases of manual node selection.


PickNicko13:
I just realized that sometimes the processing nodes might actually have real outputs for processed audio (e.g. jack standalone effects). The current implementation works with the quirks of the filter-chain pipewire module, but disregards the more normal-looking jack standalone effects. I'll add it sometime tomorrow probably.

Currently it breaks really ungracefully if used with a jack effect node. It kind of... crashes the whole bar.

I'll also add some safe fallbacks and sanity checks while I'm at it.

PickNicko13:
Sorry it took so long - I was somewhat preoccupied. But it seems to be pretty robust now.

PickNicko13:
I didn't pay much notice to the amount of pull requests you have to approve at first. It appears that it might take some time and meanwhile I have already added information about the new parameter to the wiki. Should I revert it until (if ever? I haven't heard any objections, but no positive feedback either, so there's a chance you might reject it) the merge request is in the main branch?

PR #4719, opened 2025-12-24
Encode icon inside label

oisanjaya:
With this PR its possible to encode icons using label string.
This would allow modules inherits AIconLabel instead of ALabel and having icons encoded inside `format` and/or `format-icons` in config file.
Rofi’s extended dmenu protocol is used to encode icon metadata, which starts with `\0icon\1f` followed by named icon or path to an image and ended with `\n`. Because of config file will be decoded as JSON then it should written `\\0icon\\1f<icon-name>\\n` in config file strings.
Currently only one icon per label allowed, this PR will take first found pattern as icon and ignore the rest.
`icon-size` key in config will be used to determine icon size if a path to an image file being supplied. This key is ignored if a named icon being used.


luyu-wu:
Hello! I'm attempting to test out this branch.
Seems as though it may be broken? Or perhaps I am using it incorrectly.

There seems to be an issue with parsing:
image

However, I'm unsure if it is meant to be used like this, so I have attached a portion of my config.
I will try and debug this meanwhile..

oisanjaya:
> Hello! I'm attempting to test out this branch. Seems as though it may be broken? Or perhaps I am using it incorrectly.
> 

You don't need to include the < and > symbols in the icon spec. However, this PR alone won't automatically generate icons for modules unless they are updated to inherit from AIconLabel instead of ALabel.

PR #4723, opened 2025-12-27
feat(temperature): add hwmon-name option for stable sensor selection

### LINKED ISSUE #1700: Temperature sysfs paths are not stable
I find that between kernel versions, the paths in sysfs for the temperature reading change, so probably they should not be hardcoded in the waybar config.

When in one kernel version a sensor is named `/sys/class/hwmon/hwmon0/temp1_input` in a newer kernel the same sensors might be named `/sys/class/hwmon/hwmon4/temp2_input` or so. 

My suggestion would be to use the files `/sys/class/hwmon/hwmon*/name` and `/sys/class/hwmon/hwmon*/temp*_label` to identify the sensors, i.e. provide parameters in the waybar config for these instead/alternatively to `hwmon-path`.

kaosuran:
Currently, the temperature module relies on hwmon paths, which can change when new devices are added or kernel driver order changes. This makes configurations fragile; for example, adding a USB device like a new gaming controller can break the CPU or GPU temperature readings.

What this PR does:
- Introduces a new `hwmon-name` option for the temperature module.
- Allows users to select a hwmon device by its stable name (from /sys/class/hwmon/*/name) instead of a potentially shifting path.
- Requires `input-filename` to be set.
- Ensures `hwmon-name` cannot be used with `hwmon-path` or `hwmon-path-abs` to avoid conflicts.
- Updates documentation to reflect the new option.

This provides a more stable and predictable way to configure temperature sensors, preventing breakage due to dynamic hwmon ordering.

Let me know what you think!
Kind regards.

Closes #1700


kaosuran:
The nix fail seems to be coming from upstream, from cava.

kaosuran:
Hi @Alexays! Hope you’re doing well 🙂
Just wanted to follow up on this PR. It updates sensor handling and fixes issue https://github.com/Alexays/Waybar/issues/1700. Let me know if there's any additional work or changes needed on my side to get it merged. Thanks so much for your time!

PR #4724, opened 2025-12-27
Allow tooltip formatting the CPU module.

### LINKED ISSUE #2320: [Feature Request] tooltip-format for CPU module
Hi. I think It would be cool to personalize the tooltip on the CPU module, since (at least to me) individual core usage is not that relevant. I'd rather read the average frequency or the temperature.

Thanks

hallvardnmbu:
Includes handling of `tooltip-format` within the CPU module.

Closes #2320 

Currently, when tooltip is enabled for the CPU module, each core along with its usage is listed (as well as the total).

This change allows specifying any (or none) of the format replacements of [waybar-cpu(5)](https://man.archlinux.org/man/extra/waybar/waybar-cpu.5.en) in the `tooltip-format`.

It defaults to the current behaviour (i.e., when `tooltip` is `true` and no `tooltip-format` is specified).


ClairDePrince:
@Alexays 🙏

Issue #4726, opened 2025-12-27
nix/default.nix out of sync with libcava.wrap - cava build broken


xav-ie:
The nix flake build fails to include cava support because `nix/default.nix` is out of sync with `subprojects/libcava.wrap`.

```
> Run-time dependency libcava found: NO (tried pkgconfig and cmake)
> Looking for a fallback subproject for the dependency libcava
> ERROR: Subproject libcava is buildable: NO
> Message: cava is not found. Building waybar without cava
```

## Root Cause

Two commits updated cava-related files but left them inconsistent:

1. [Commit 97682a13](https://github.com/Alexays/Waybar/commit/97682a13328dba6f38f0373c1cffd101efdedcaf) (Oct 19, 2025) - "nix: bump cava"
   - Changed nix to use commit `23efcced...`
   - Copies to `subprojects/cava`

2. [Commit 13519ca5](https://github.com/Alexays/Waybar/commit/13519ca5bfa6d24e638ace7b82fe0fe7bad9bcb8) (Nov 26, 2025) - "cava. nonsafe thread.= & cava bump"
   - Renamed `cava.wrap` → `libcava.wrap`
   - Changed expected directory to `cava-0.10.7-beta`
   - Changed source to `v0.10.7-beta` tag
   - **Did not update `nix/default.nix`**

### Current state

[nix/default.nix](https://github.com/Alexays/Waybar/blob/master/nix/default.nix#L8-L16):
```nix
libcava = {
  src = pkgs.fetchFromGitHub {
    owner = "LukashonakV";
    repo = "cava";
    rev = "23efcced43b5a395747b18a2e5f2171fc0925d18";  # Old commit from Oct 12
    hash = "sha256-CNspaoK5KuME0GfaNijpC24BfALngzNi04/VNwPqMvo=";
  };
};
# ...
postUnpack = ''
  cp -R ... ${libcava.src} subprojects/cava  # Wrong directory name
'';
```

[subprojects/libcava.wrap](https://github.com/Alexays/Waybar/blob/master/subprojects/libcava.wrap):
```ini
[wrap-file]
directory = cava-0.10.7-beta  # Expects this directory name
source_url = https://github.com/LukashonakV/cava/archive/v0.10.7-beta.tar.gz
```

## Suggested Fix

Update `nix/default.nix` to match `libcava.wrap`:

```nix
libcava = {
  src = pkgs.fetchFromGitHub {
    owner = "LukashonakV";
    repo = "cava";
[...]

PR #4727, opened 2025-12-27
fix-cava-nix-sync

xav-ie:
Addresses https://github.com/Alexays/Waybar/issues/4726

- fix(nix): sync libcava source with libcava.wrap

  Update rev to v0.10.7-beta and directory name to cava-0.10.7-beta
  to match subprojects/libcava.wrap.


nouritsu:
This works on my x86_64 machines.

xav-ie:
#4708 fixed this and was merged.

Issue #4725, opened 2025-12-27
Cava module regression after PR #4682: 'Unknown module' when no cava_config specified


xav-ie:
After PR #4682 (merged Dec 18, 2025), the cava module fails with:
```
[warning] module cava: Disabling module "cava", Unknown module
```

This happens when using waybar's cava module **without** a `cava_config` file specified, which worked correctly before PR #4682.

## Root Cause

The refactoring in PR #4682 introduced a mismatch between `cava_backend.cpp` and `cava_frontend.hpp`:

### cava_backend.cpp

The output method is saved, temporarily set to RAW for initialization, then restored to the original value:

Save original output method:
https://github.com/Alexays/Waybar/blob/97eb60677e555221326009e0bc3def67e85d33f4/src/modules/cava/cava_backend.cpp#L184

Restore original output method (BUG):
https://github.com/Alexays/Waybar/blob/97eb60677e555221326009e0bc3def67e85d33f4/src/modules/cava/cava_backend.cpp#L267

### cava_frontend.hpp

The frontend only accepts `OUTPUT_RAW` or `OUTPUT_SDL_GLSL`, rejecting all other output methods:

Switch statement that throws "Unknown module":
https://github.com/Alexays/Waybar/blob/97eb60677e555221326009e0bc3def67e85d33f4/include/modules/cava/cava_frontend.hpp#L14-L25

### Cava library default

When no config file is provided, cava defaults to `OUTPUT_NONCURSES`:

Default output method is "noncurses":
https://github.com/LukashonakV/cava/blob/256cf7aa7c893fb457d54477bd20758db3d79026/config.c#L819

### Old code (pre-PR #4682)

The old implementation simply set `OUTPUT_RAW` once and never changed it:

Old cava.cpp:
https://github.com/Alexays/Waybar/blob/cbab9c9713b79e636d313819fd2f6dc55a2e9e31/src/modules/cava.cpp#L22

## Flow Analysis

1. Waybar config has no `cava_config` option → empty path passed to `load_config()`
2. Cava library uses default `output:method = noncurses` → `OUTPUT_NONCURSES`
3. `cava_backend.cpp` saves `OUTPUT_NONCURSES`, sets to `OUTPUT_RAW` for init, then **restores** to `OUTPUT_NONCURSES`
4. `cava_frontend.hpp` calls `getPrm()->output` → gets `OUTPUT_NONCURSES`
[...]

PR #4728, opened 2025-12-27
fix-cava-config

xav-ie:
Addresses https://github.com/Alexays/Waybar/issues/4725:

- fix(cava): handle all output methods in frontend

  Default to raw Cava module for non-GLSL output methods.
  Fixes "Unknown module" error when no cava_config is specified.


LukashonakV:
I'm ok with this PR with a note,  do note put here any additional commit. The only current commit with return expression by default is enough for now.

LukashonakV:
I'm ok with this PR with a note,  do note put here any additional commit. The only current commit with return expression by default is enough for now.

PR #4729, opened 2025-12-27
fix-cava-audio-height-init

### LINKED ISSUE #4244: Cava module always peaking
I currently use cava as part of my Waybar config for hyprland, and whenever I play any sort of audio (no matter the volume) all of the bars immediately start peaking.

This is my current cava config in `config.json`:
```
"cava": {
	"framerate": 30,
	"bars": 12,
	"sensitivity": 10,
	"lower_cutoff_freq": 50,
	"higher_cutoff_freq": 20000,
	"sleep_timer": 15,
	"hide_on_silence": true,
	"method": "pulse",
	"source": "auto",
	"stereo": false,
	"bar_delimiter": 0,
	"monstercat": true,
	"noise_reduction": 0.33,
	"input_delay": 1,
	"format-icons": ["▁", "▂", "▃", "▄", "▅", "▆", "▇", "█" ],
},
```
And these are my settings in `~/.config/cava/config` (though i dont think it should matter):
```
; mode = normal
; framerate = 30
; autosens = 1
; overshoot = 20
; sensitivity = 10
; lower_cutoff_freq = 50
; higher_cutoff_freq = 20000
; sleep_timer = 15
```
Even with the sensitivity at 10%, each and every bar immediately peaks even when listening to quiet asmr videos. When I set the autosens option, I find that the bars have the opposite issue where it overcompensates and the bars barely move from the off position.

I've experimented with changing the volume being provided to cava using pavucontrol, but the issue persists. Currently, the cava recording volume is at 35% within pavucontrol and the bars are still peaking. Changing the volume in this way sometimes will fix the issue, but only temporarily, as changing songs or restarting cava will ruin it again.

I would appreciate any help/insight on this topic. Thanks.

xav-ie:
This might fix #4244

The cava module bars always peak at maximum regardless of sensitivity settings.

`data_format` was changed from always being `"ascii"` to only being set when user-provided. Without an explicit value, cava defaults to `"binary"`, which calculates `height = pow(2, 16) - 1 = 65535` instead of `ascii_range`. This causes `audio_raw_fetch()` to scale bars incorrectly.

Default `data_format` to `"ascii"` while preserving user override capability. This mantains the previous default.

---

- fix(cava): default data_format to ascii for correct bar scaling

  data_format was changed from always "ascii" to user-configurable,
  but without a default. This caused cava to use "binary", calculating
  height as 65535 instead of ascii_range, making bars always peak.

### DIFF
--- a/src/modules/cava/cava_backend.cpp
+++ b/src/modules/cava/cava_backend.cpp
@@ -183,10 +183,10 @@ void waybar::modules::cava::CavaBackend::loadConfig() {
   prm_.inAtty = 0;
   auto const output{prm_.output};
   // prm_.output = ::cava::output_method::OUTPUT_RAW;
-  if (config_["data_format"].isString()) {
-    if (prm_.data_format) free(prm_.data_format);
-    prm_.data_format = strdup(config_["data_format"].asString().c_str());
-  }
+  if (prm_.data_format) free(prm_.data_format);
+  // Default to ascii for format-icons output; allow user override
+  prm_.data_format = strdup(
+      config_["data_format"].isString() ? config_["data_format"].asString().c_str() : "ascii");
   if (config_["raw_target"].isString()) {
     if (prm_.raw_target) free(prm_.raw_target);
     prm_.raw_target = strdup(config_["raw_target"].asString().c_str());


LukashonakV:
Hi @xav-ie , for now I'm not able to check. Will do , once have a time. It might possible original bug can be issued in the original cava project or libcava. If the issue is brought by the waybar cava module implementation,  then I'm ok. Otherwise it should be managed by the cava/libcava. Actually I don't remember for now, what the purpose to set max height in the waybar cava.

BTW: thank you for investigation and happy new year)

xav-ie:
Hey, @LukashonakV. Thank you for the quick response.

Your suggestion to put the change in cava led me to find out that I made a mistake.

Previously, `data_format` *always* defaulted to "ascii". Now, the new code does not default and cava defaults to "binary".

This would cause the height to now default to a large value.

We just need to restore the defaulting behavior. I updated the PR to add this back.

Happy new year!

JohnnyJumper:
settings the data_format to "ascii" for my config fixed this issue. Thanks a lot !

Alexays:
Thanks!

PR #4735, opened 2025-12-29
[rebase] feat: split off gps and cava modules into shared libraries loaded at module initialisation time

qbe:
rebase of https://github.com/Alexays/Waybar/pull/4667 , a (proposed) fix for https://github.com/Alexays/Waybar/issues/4276

justification for this PR is discussed in the issue, dynamically loading modules allows a waybar build with cava and gps capabilities to be packaged in a way that allows for libgps, libcava and epoxy to be _optional_ dependencies.

Another goal of this PR was to include as few modifications to module code as necessary. For this reason I decided to split off the cava and gps modules into shared libraries, and to load them in factory.cpp using dlopen() / dlsym() .

The first 4 commits deal with the fact that reap_mtx / reap were previously an "extern" reference deep within include/util/command.hpp , which is included in every module and thus prevents the split up. Passing reap_mtx / reap to the module constructors rectifies this, but necessarily touches lots of constructors. I recommend looking at a word-diff of these commits, and to look at the whole patch series in commit-by-commit order.


qbe:
As to why this rebase was necessary, turns out using the wrong clang-format version caused a lot of noise, while at the same time some changes regarding cava were merged.

qbe:
@Alexays I am trying to resolve the merge conflicts (again), but I am struggling with the problem of having to pass reap_mtx / reap pairs to the handleGtkMenuEvent function (src/ALabel.cpp:193) which it now needs because it calls forkExec and that function needs the reap list.

Before I take a deep dive into this, can you please take a look at my PR so far ?

I have by now sunken quite a lot of time into this series, which would be worth it if you see the value in this PR in general, but if you are generally against splitting of modules and dynamically loading them then I can go do
something else with my free time.

qbe:
Still keeping this PR up-to-date, but it feels like it is getting buried. Can I get any response from you @Alexays ?

src/modules/gps.cpp
```diff
@@ -15,13 +15,21 @@
 #include <fmt/core.h>
 #endif
 
+extern "C" {
+waybar::ALabel* new_gps(const std::string& id, const Json::Value& config, std::mutex& reap_mtx,
+                        std::list<pid_t>& reap) {
+  return new waybar::modules::Gps(id, config, reap_mtx, reap);
+}
+}
+
```
qbe: these lines should be in 631c991: "build gps module as dynamic library, load in factory using dlopen / dlsym"

I overlooked them during the rebase, and accidentally included them here. No further changes needed, comment is just for readability.

qbe:
It has been half a year now, @Alexays have you had a look at this yet ?

qbe:
should I take a look at those conflicts, or are you already on it ?

Alexays:
Hi!
Yes if you can handle the rebase, it’ll be perfect :)

qbe:
so far only meson and clang errors. Will try to fix them.

qbe:
all errors fixed, tested, cava module works with an appropriate LD_LIBRARY_PATH set

qbe:
@Alexays this can be merged now, as far as I am concerned.

qbe:
okay, I will solve the conflicts again. @Alexays , this patch touches so many files (because of the reap / reap_mtx stuff) that conflicts will always arise if another merge touched any module constructors. Please consider merging this before merging anything else.

qbe:
alternatively, I could split it up and separate the reap_mtx / reap rework into its own patch, which would not change any functionality and which would be trivial to review. The dynamic loading of modules can then be done in another patch. I just need some communication on the matter.

Issue #4741, opened 2025-12-31
[sway/workspaces] Weird behavior when using quotes in workspace names


Sayrus:
Hello,

I'm having some strange behavior when using workspaces with quotes in them. For instance:
```
# Simple example
set $example 10:"Cal"
bindsym $mod+0 workspace $example
```
When switching through Sway, the Waybar module shows `"Cal"`. When clicking the module switches to `Cal`.

```
# Adding a leading space
set $example 10:" Cal"
bindsym $mod+0 workspace $example
```
When switching through Sway, the modules shows `" Cal". However when clicking this on Waybar, it uses the workspace ` Cal ""`. Clicking this once more redirects to ` Cal`.

This matches the behavior of `swaymsg workspace "10:"Cal""`  (manually templated from https://github.com/Alexays/Waybar/blob/master/include/modules/sway/workspaces.hpp#L26).

If the module works like swaymsg, quotes in the namespace name should be escaped `swaymsg workspace "10:\"Cal\""`.

Sayrus:
```diff --git a/src/modules/sway/workspaces.cpp b/src/modules/sway/workspaces.cpp
index 182bf631..08946509 100644
--- a/src/modules/sway/workspaces.cpp
+++ b/src/modules/sway/workspaces.cpp
@@ -369,11 +369,16 @@ Gtk::Button &Workspaces::addButton(const Json::Value &node) {
                                    node["name"].asString(), node["target_output"].asString(),
                                    "--no-auto-back-and-forth", node["name"].asString()));
         } else {
-          ipc_.sendCmd(IPC_COMMAND, fmt::format("workspace {} \"{}\"",
-                                                config_["disable-auto-back-and-forth"].asBool()
-                                                    ? "--no-auto-back-and-forth"
-                                                    : "",
-                                                node["name"].asString()));
+          std::string workspaceName = node["name"].asString();
+          std::string flag = config_["disable-auto-back-and-forth"].asBool()
+                                ? "--no-auto-back-and-forth"
+                                : "";
+          // Use number if available as it is guaranteed to not have special characters
+          if (node["num"].asInt() >= 0) {
+            ipc_.sendCmd(IPC_COMMAND, fmt::format("workspace {} number {}", flag, node["num"].asInt()));
+          } else {
+            ipc_.sendCmd(IPC_COMMAND, fmt::format("workspace {} {}", flag, workspaceName));
+          }
         }
       } catch (const std::exception &e) {
[...]

PR #4742, opened 2025-12-31
 fix(sway/workspaces): use number when available to avoid parsing issues 

Sayrus:
Partial fix for https://github.com/Alexays/Waybar/issues/4741 by avoiding the name altogether. It may be possible to template using quotes if the quote type isn't present in the string, for instance:
- `my workspace` => `"my workspace`
- `my "workspace"` => `'my "spacespace'`
- `workspace's` => `"workspace's"`

But that wouldn't cover all cases. It seems sway handles this a bit differently: it handles any string but will change parsing behavior if it encounters a keyword (`output`, `prev`, ...). This is difficult to copy and maintain without using their parsing logic instead of the quoting mechanism.


Sayrus:
@Alexays Is there anything I can do to help get this MR and https://github.com/Alexays/Waybar/pull/4744 reviewed?

PR #4754, opened 2026-01-03
feat: support arbitrary command for clicking on sway workspace button

AckslD:
Instead of #4685 this allows executing an arbitrary command when clicking a workspace button. It seem the default value in the `.scd` is not working for some reason, any idea why?


AckslD:
Would probably be nice to also set `WAYBAR_OUTPUT_NAME` env var somehow for the subprocess

PR #4761, opened 2026-01-09
feat(battery): add option to smooth power for more stable time_remaining estimates

### LINKED ISSUE #4755: Battery time estimate is extremely unstable due to stateless calculation
The remaining battery time estimate reported by the Waybar battery module fluctuates wildly, often jumping between values like ~8h and ~19h within a short period of time, even under relatively stable system load.

From inspecting the code, it appears that the time remaining is calculated in a stateless manner based only on the current power draw and remaining capacity, without any form of smoothing, averaging, or historical prediction. Because instantaneous power usage varies significantly, this approach produces highly unstable and misleading estimates.

**Observed behavior:**

* Battery time remaining rapidly swings by several hours
* Small, momentary changes in power usage cause large jumps in the estimate
* The value rarely converges to something realistic during normal use

**Expected behavior:**

* Battery time should change gradually
* Short-term spikes in power usage should not dramatically affect the estimate
* Some form of averaging or prediction over time would produce more usable results

Benji-clm:
This PR adds support for an option to have power based on current and past power readings to stabilize the estimation of the time remaining on the battery.

How aggressive the smoothing/responsive to changes/noise can be changed by changing `smooth-power-time-constant` which is the half life, in seconds, of each measurement.

I've decided to not use the smooth power for charging, as battery charging curves tend to be much less noisy and more predictable/smoother.

```
"battery": {
    "smooth-power": true, // default is false if not present
    "smooth-power-time-constant": 260.0, // default is 260.0 if not present
},
```

Fixes #4755


jannickj:
@Alexays, this PR is getting stale :D

jannickj:
:O

Issue #4647, opened 2025-11-19
[Feature] Configurable layer namespace


Holbormon:
Hi,

As you may know Hyprland achieves control of layers such as waybar programmatically via [layer rules](https://wiki.hypr.land/Configuring/Window-Rules/#layer-rules). Unfortunately, these rules can currently only match the `namespace` of layers, making a granular control of individual waybars via layer rules impossible (as all waybars share the same namespace, even across different waybar processes).

Even though familiar with the docs, I could not find a way to change the namespace of a given waybar (the `name` jsonc key only affects the waybar's css class), and I think it would be neat if it were possible.

The implementation could be as simple as re-using the aforementioned `name` config value as the layer's namespace, and I think it would also make sense to do so: Indeed, if I set a name for a given waybar, then I would expect that name to show up in tools such as `hyprctl layers`.

On top of solving the Hyprland layer rule problem, it would also allow for more general scripting flexibility.

Example use case: Having **only one of** (and not all) waybars displayed above the lock screen via `layerrule abovelock`, or using the output of `hyprctl layers` in scripts.

PR #4776, opened 2026-01-16
feat: Disambiguate wayland layer namespace

Holbormon:
Provides an easy way to disambiguate several waybars so that compositors like Hyprland may apply granular layer rules to different waybars.

Closes https://github.com/Alexays/waybar/issues/4647.


Alexays:
Thanks!

PR #4789, opened 2026-01-24
fix(json): use local CharReaderBuilder for thread safety

xander1421:
## Problem

Crash with:
```
unhandled exception (type std::exception) in signal handler:
what: in Json::Value::find(begin, end): requires objectValue or nullValue
```

## Root Cause

The `JsonParser` class uses a shared `CharReaderBuilder` member that gets accessed concurrently from multiple threads:

- **IPC thread**: Event handlers call `getSocket1JsonReply()` → `parser_.parse()`
- **GTK thread**: Update functions call `getSocket1JsonReply()` → `parser_.parse()`

Each module has its own mutex, but they all share the same `IPC::inst()` singleton. The module mutexes don't protect the shared parser.

`Json::CharReaderBuilder` is not thread-safe for concurrent use without synchronization.

## Fix

Use a local `CharReaderBuilder` in `parse()` instead of a member variable. This makes each parse call independent and thread-safe.

### DIFF
--- a/include/util/json.hpp
+++ b/include/util/json.hpp
@@ -30,15 +30,16 @@ class JsonParser {
 
     std::istringstream jsonStream(modifiedJsonStr);
     std::string errs;
-    if (!Json::parseFromStream(m_readerBuilder, jsonStream, &root, &errs)) {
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #4804, opened 2026-02-02
fix: Watch new CSS files on appearance change

BBaoVanC:
If you have reload_style_on_change enabled, and use dynamic appearance styling (style-dark.css and style-light.css), then changing the appearance doesn't update the files that are being watched for reload.

Add a method to CssReloadHelper to change the CSS file being watched, and also provide the CSS file as a parameter to the callback so setupCss can be called on the right file when the file watcher triggers.

To me this API was the most obvious way to fix it, but I'm open to other ideas.

Unit tests for the watching behavior don't already exist and don't seem very feasible in the first place.


BBaoVanC:
@Alexays I just had a chance to take a look at this again. I don't really think there's a sane way to create unit tests for this. The behavior this PR adds doesn't look testable outside of making an entirely separate integration test system that makes sure file monitors are correctly updated, also meaning it has to be a full blown GIO program with a main loop and all.

Could you consider merging this without adding tests? I have tested myself by swapping between light/dark themes and confirming that it does automatically reload. I also can confirm after adding some print/cout-based debugging to the callback created in `waybar::Client::main` that it correctly stops watching the old CSS files after a scheme change.

PR #4806, opened 2026-02-03
docs: add fraudulent site notice on proj README.md

### LINKED ISSUE #4752: 'waybar.org' fraudulent site
On searching for Waybar, there seems to be a fraudulent site "waybar.org" ranking on-par with the repository & relevant wiki pages. Seen from [this issue in czkawa](https://github.com/qarmin/czkawka/issues/1583) which cites the same fake 'location', it is supposedly AI-generated. It also reuses the Hyprland logo and **provides a download link from its own servers.**

This appears to have been mentioned before in Discussions, however it was unaddressed so I leave it here in case action should be taken. Failing to find other modes of contact, I believe here is the next-best location to raise concern.

jackinfurs:
Resolves #4752.

This can be merged without attribution since it is admittedly a very minor contribution, if even one at all. Mirrors similar notices given for [czkawa](https://github.com/qarmin/czkawka?tab=readme-ov-file#officially-supported-projects) and [minimap2](https://github.com/lh3/minimap2?tab=readme-ov-file#getting-started) until the site's(') removal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,10 @@ Contributions welcome!<br>
 Have fun :)<br>
 The style guidelines are [Google's](https://google.github.io/styleguide/cppguide.html)
 
+> [!CAUTION]
+> Distributions of Waybar are only released on the [official GitHub page](https://github.com/Alexays/Waybar).<br/>
+> Waybar does **not** have an official website. Do not trust any sites that claim to be official.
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks for the notice!

PR #4819, opened 2026-02-06
fix: backlight/slider fix rounding issue on scroll up

### LINKED ISSUE #4818: backlight/slider won't increase brightness on scroll correctly
Hi! The `backlight/slider` module won't increase brightness correctly on scroll up because of a rounding issue.

cebem1nt:
bump: fixes #4818


Alexays:
Thanks!

PR #4820, opened 2026-02-06
Feature/tray order

Carltoffel:
## Deterministic and Configurable Tray Icon Ordering

This PR adds a new feature to the Waybar tray module: **stable and user-configurable ordering of tray icons.**

### Problem

Before this change, tray icons were shown in the order in which applications registered themselves via D-Bus. This caused several issues:

* The order depended on application start timing
* Icons could appear in a different order after each login
* Multiple monitors could show different tray orders
* Restarting Waybar often shuffled the tray randomly

There was no reliable way to control or stabilize this behavior.

---

#### 1. Stable default behavior

Even without any configuration, tray icons are now sorted deterministically in **alphabetical order** based on an internal key (`sort_key`).

This ensures that:

* Tray order is consistent across all monitors
* Order does not depend on application start timing
* Restarting Waybar keeps the same layout

#### 2. New user configuration options

Users can now explicitly define their preferred tray order using a simple list:

```json
"tray": {
  "order": [
    "vesktop",
    "signal desktop"
  ],
  "order-unknown": "after"
}
```

* `order` allows specifying tray icons in the exact desired order
* Items not listed are sorted alphabetically
* `order-unknown` controls whether unconfigured items appear before or after configured ones

Available keys can be easily discovered via log messages like:

```
tray: item key='<KEY>' ...
```

#### 3. Improved handling of Electron apps

Many Electron-based applications register generic tray IDs such as `chrome_status_icon_1`.
This PR improves that by:

* Deriving a meaningful key from the tooltip text
* Using that key consistently for sorting

This makes it possible to reliably sort apps like Signal, Vesktop, etc.

---

### How it works internally

* Each tray item now provides a stable `sort_key`
* The SNI Host component performs centralized reordering
* Reordering is triggered automatically when an item identity becomes available
* Sorting priority is:

  1. Items listed in the user-defined `order`
  2. All other items alphabetically
  3. Stable tie-break using insertion sequence

This guarantees deterministic results across all Waybar instances.

---

### Differences to PR #4643

This implementation differs significantly from PR #4643:

* Configuration uses a **simple ordered list** instead of numeric priorities
* Provides deterministic alphabetical ordering by default
* Sorting is triggered directly when item identity becomes known
* Includes explicit support for Electron app quirks
* Adds user-friendly logging to discover available keys

Overall, this approach focuses on simplicity and predictable behavior.

---

### Personal note

I want to mention explicitly that I have **very limited experience with C++**.
This feature was created almost entirely with the help of ChatGPT, which guided me through the design, refactoring, and implementation process step by step.

Feedback and suggestions are very welcome!

### DIFF
--- a/include/modules/sni/host.hpp
+++ b/include/modules/sni/host.hpp
@@ -6,6 +6,7 @@
 #include <json/json.h>
 
 #include <tuple>
+#include <vector>
 
 #include <unordered_map>
 
@@ -54,6 +55,9 @@ class Host {
   bool reorder_pending_{false};
   std::size_t next_seq_{0};
   std::unordered_map<Item*, std::size_t> seq_;
+  std::unordered_map<std::string, std::size_t> order_index_;
+  std::vector<std::string> order_list_;  // normalized keys in configured order
+  bool unknown_after_{true};  // configured icons first, unknown icons after (default)
 };
 
 }  // namespace waybar::modules::SNI
--- a/include/modules/sni/host.hpp
+++ b/include/modules/sni/host.hpp
@@ -7,6 +7,8 @@
 
 #include <tuple>
 
+#include <unordered_map>
+
 #include "bar.hpp"
 #include "modules/sni/item.hpp"
 
@@ -18,6 +20,8 @@ class Host {
        const std::function<void(std::unique_ptr<Item>&)>&,
        const std::function<void(std::unique_ptr<Item>&)>&);
   ~Host();
+
+  void requestReorder();
 
  private:
   void busAcquired(const Glib::RefPtr<Gio::DBus::Connection>&, Glib::ustring);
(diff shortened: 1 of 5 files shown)


Carltoffel:
I just had another idea how to implement the options:
Instead of having a list "order" and an option before/after, we could make two lists which define the tray icons before and after the alphabetical default order:

```
"tray": {
  "order-left": [
    "signal desktop",
  ],
  "order-right": [
    "vesktop",
  ],
}
```

This would move Vesktop to the right, Signal to the left, and all other icons are between them in alpabetical order.

Another thing which could be discussed is how the "reverse-direction" option should behave. Should it only affect the alphabetically ordered icons or all icons, which would mess with the naming of "left" and "right". Or maybe this option can be removed?

Carltoffel:
Closed because I made a new implementation: https://github.com/Alexays/Waybar/pull/4999

Issue #4778, opened 2026-01-17
Suspected bug found in waybar.


Folia2045:
Scrolling the mouse wheel on the group module causes Waybar to crash. Here are the logs：

Mortis1145:
 waybar
[2026-01-17 22:29:47.417] [info] Using configuration file /home/folia/.config/waybar/config.jsonc
[2026-01-17 22:29:47.417] [info] Including resource file: modules.jsonc
[2026-01-17 22:29:47.417] [info] Including resource file: modules-dividers.jsonc
[2026-01-17 22:29:47.418] [info] Discovered appearance 'dark'
[2026-01-17 22:29:47.435] [info] Using CSS file /home/folia/.config/waybar/style.css
[2026-01-17 22:29:47.444] [info] Niri IPC starting
[2026-01-17 22:29:47.446] [warning] 'swap-icon-label' must be a bool.
[2026-01-17 22:29:47.446] [warning] Hyprland is not running, Hyprland IPC will not be available.
[2026-01-17 22:29:47.446] [warning] module hyprland/window: Disabling module "hyprland/window", Hyprland IPC: HYPRLAND_INSTANCE_SIGNATURE was not set! (Is Hyprland running?)
[2026-01-17 22:29:47.450] [warning] No batteries.
[2026-01-17 22:29:47.473] [info] Bar configured (width: 1920, height: 31) for output: HDMI-A-1

** (waybar:99575): ERROR **: 22:29:48.992: 
unhandled exception (type std::exception) in signal handler:
what: in Json::Value::find(begin, end): requires objectValue or nullValue

fish: 作业 1, 'waybar' 终止于信号 SIGTRAP (跟踪或断点陷阱)

ahfriedman:
appears to be fixed by https://github.com/Alexays/Waybar/commit/68cbb507d95c37c67fc68efe223381e6359d88fc

DeleteUndoBackspace:
yep same. i did the scroll up down workaround. it worked but only inside the box. scrolling slighty outside the box going to crash the waybar. its driving me insane. haha

akrameh:
same issue

cebem1nt:
@DeleteUndoBackspace @akrameh Can you guys please make sure if it can be reproduced with the build from latest commit?

DeleteUndoBackspace:
hi @cebem1nt . the issue is still there for me unfortunately. this bug happens when i scroll slightly outside of the group but still within the panel bar.

[...]

PR #4821, opened 2026-02-06
fix: owerwrite handleScroll to remove any scrolling handling from group

cebem1nt:
Hi!

This PR completely removes execution of any command on scroll (even set specifically with "on-scroll-up" or "on-scroll-down") by overwriting handleScroll method for group module.


cebem1nt:
fixes #4778

Alexays:
Thanks!

Issue #2842, opened 2024-01-17
Feature request: filter visible tray icons + multiple modules of the same type with different configs


redpenguinyt:
I want to use nm-applet as my wifi indicator with its context menu and all, however I don't want it to sit next to my other tray items (Discord, etc.). It would be nice to be able to set some tray items to not appear in a tray (something like an exclude/include selector) and also the ability to configure multiple instances of the same module. That would allow me to have one tray with nm-applet and nothing else and another separate tray module with all other tray items

Fxxxlei:
I would like to use this kind of thing for a few things as well, as it would allow me to group nm-applet with connectivity indicators, fcitx (or other IME) with keyboard indicators, group messengers together...

Fxzzi:
I like this idea, if we were to get this it could also work as a way to exclude certain applets in tray

AlexChalk:
➕1, it would be neat to have block/allow functionality for the tray.

lucas-farias07:
+1 I don't want to have 2 wifi icons in my bar

dommcdev:
+1 I want to use nm-applet (which only works if I have the tray module enabled) but I don't want to see all the other items in my tray.

PR #4826, opened 2026-02-08
feat (tray): Ignore Tray items by service name and when dbus properties load

AntGarSil:
## Description
Implements configuration-based filtering of system tray (StatusNotifierItem) icons in Waybar. Users can now specify patterns to hide unwanted tray items via the `ignore-list` configuration option.

## Problem Statement
Some applications (e.g., blueman, network managers) show system tray icons that cannot be disabled through the application's own settings. Users need a way to filter these unwanted icons at the Waybar level.

## Solution
Added ignore-list functionality that filters SNI items based on pattern matching against service names, object paths, and item properties.

### `src/modules/sni/host.cpp`
**New Methods:**
- `checkIgnoreList()` - Filters and removes items matching ignore patterns
  - Iterates through all registered items
  - Matches patterns against: `bus_name`, `category`, `icon_name`, `id`, `title`
  - Calls `on_remove` callback and erases matched items
  - Comprehensive debug/info logging

**Modified Methods:**
- `Host::Host()` constructor - Added `ignore_list` parameter
- `itemRegistered()` - Added logging for item registration events
- `addRegisteredItem()` - Enhanced debug logging

**Constructor Signature Change:**
```cpp
Host::Host(const std::size_t id, const Json::Value& config, const Bar& bar,
           const std::vector<std::string>& ignore_list,  // NEW
           const std::function<void(std::unique_ptr<Item>&)>& on_add,
           const std::function<void(std::unique_ptr<Item>&)>& on_remove)
```

### `src/modules/sni/tray.cpp`
**New Methods:**
- `parseIgnoreList()` - Static helper to parse JSON config array
  - Validates array type
  - Extracts string patterns
  - Logging for configuration transparency
  
- `checkIgnoreList()` - Public wrapper delegating to Host's method

**New Member Variables:**
- `std::vector<std::string> ignore_list_` - Stores ignore patterns

**Modified Methods:**
- `Tray::Tray()` constructor - Initializes `ignore_list_` from config
- `onAdd()` - Calls deferred ignore check after adding item, enhanced logging
- `update()` - Periodic ignore list check to catch late-loading properties

## Configuration Usage
```json
{
  "tray": {
    "ignore-list": ["blueman", "nm-applet", "spotify"]
  }
}
```

Patterns are matched using substring search (`std::string::find()`) against multiple item fields.

## Technical Details
- **Pattern Matching**: Substring-based (not regex or wildcards)
- **Scope**: Module-level (per tray instance)

## Logging Added
- Config parsing: Shows ignore-list items being loaded
- Item registration: Logs all registered items with properties
- Filtering: Info-level logs when items are ignored with matched pattern

## Known Problems
⚠️ **Asynchronous Property Loading Issue:**

The current implementation has a timing problem:
- SNI items load D-Bus properties asynchronously
- When `onAdd()` is called, properties (`category`, `icon_name`, `title`) are empty strings
- Pattern matching against empty strings fails
- Items briefly flash in the UI before properties load

**Example from logs:**
```
[info] Tray::onAdd - item bus_name=':1.628', category='', icon_name='', title=''
```

The code now checks patterns against the `service` parameter (e.g., `:1.628/org/blueman/sni`) in `addRegisteredItem()` before Item construction, as this value is immediately available.

## Testing
- ✅ Config parsing validated with multiple patterns
- ✅ Logging confirms items are detected
- ✅ Arch
- ⚠️ Pattern matching fails due to empty properties (timing issue).  Workaround implemented.
- ⏳ Needs full fix for async property loading

## Breaking Changes
None - added functionality is opt-in via configuration.

## Related Issues
#2842


AntGarSil:
#2842 Addressing feature request

remix7531:
Using this PR on NixOS with Hyprland — works for hiding the Yubico Authenticator
tray icon (`com.yubico.yubioath`). Thanks!

One issue I ran into: SNI items register before their D-Bus properties (`Title`,
`IconName`) load, so the match at `onAdd` time misses them. The PR adds a recheck
in `Tray::update()`, but `update()` only fires on add/remove — not when an item's
properties arrive later. As a result the ignored item briefly appeared on screen
(or stayed visible until something else triggered an update).

Below is a small follow-up patch that I'm running on top of this PR:

- Periodic 100 ms recheck so the match runs once properties have populated.
- Hide each item on `onAdd` when an `ignore-list` is configured, then reveal it
  after the first successful check (using `Title` / `IconName` non-empty as the
  "properties have loaded" signal). This eliminates the brief flash on startup.

```diff
diff --git a/include/modules/sni/host.hpp b/include/modules/sni/host.hpp
--- a/include/modules/sni/host.hpp
+++ b/include/modules/sni/host.hpp
@@ -22,6 +22,7 @@ class Host {

   void checkIgnoreList(const std::vector<std::string>& ignore_list,
                        const std::function<void(std::unique_ptr<Item>&)>& on_remove);
+  void revealNonIgnored();

  private:
   void busAcquired(const Glib::RefPtr<Gio::DBus::Connection>&, Glib::ustring);
diff --git a/src/modules/sni/host.cpp b/src/modules/sni/host.cpp
--- a/src/modules/sni/host.cpp
+++ b/src/modules/sni/host.cpp
@@ -68,6 +68,16 @@ void Host::checkIgnoreList(const std::vector<std::string>& ignore_list,
   }
 }

+void Host::revealNonIgnored() {
+  for (auto& item : items_) {
+    if (!item->title.empty() || !item->icon_name.empty()) {
+      if (!item->event_box.get_visible()) {
+        item->event_box.set_visible(true);
+      }
+    }
+  }
+}
+
 void Host::busAcquired(const Glib::RefPtr<Gio::DBus::Connection>& conn, Glib::ustring name) {
   watcher_id_ = Gio::DBus::watch_name(conn, "org.kde.StatusNotifierWatcher",
                                       sigc::mem_fun(*this, &Host::nameAppeared),
diff --git a/src/modules/sni/tray.cpp b/src/modules/sni/tray.cpp
--- a/src/modules/sni/tray.cpp
+++ b/src/modules/sni/tray.cpp
@@ -49,6 +49,15 @@ Tray::Tray(const std::string& id, const Bar& bar, const Json::Value& config)
   } else if (config["icon-size"].isUInt()) {
     box_.set_spacing(config["icon-size"].asUInt() / 4);
   }
+  if (!ignore_list_.empty()) {
+    Glib::signal_timeout().connect([this] {
+      host_.checkIgnoreList(ignore_list_,
+                            std::bind(&Tray::onRemove, this, std::placeholders::_1));
+      host_.revealNonIgnored();
+      return true;
+    }, 100);
+  }
+
   dp.emit();
 }

@@ -64,6 +73,9 @@ void Tray::onAdd(std::unique_ptr<Item>& item) {
   } else {
     box_.pack_start(item->event_box);
   }
+  if (!ignore_list_.empty()) {
+    item->event_box.set_visible(false);
+  }

   spdlog::debug("Tray::onAdd deferred check - checking ignore list");
   host_.checkIgnoreList(ignore_list_, std::bind(&Tray::onRemove, this, std::placeholders::_1));
```

Might be worth folding something like this into the PR — happy to open a follow-up
if helpful.

PR #4828, opened 2026-02-08
Fix include of poll.h

hannesbraun:
This resolves the warning produced when building on Alpine Linux (or more specifically systems with musl libc).


Alexays:
Thanks!

PR #4837, opened 2026-02-09
fix: prioritize `plugged` status over `full` when battery is full and plugged in

### LINKED ISSUE #3202: Battery Module shows "format" instead of "format-plugged" when plugged in and battery is full
When the battery is at 100% and plugged, the module shows whatever is defined in "format" rather than "format-plugged".

This seems like an edge case bug, since I believe the "plugged" status should have precedence over the default one, even at full charge.

Here's my battery module configuration, in case it's helpful.

```
  "battery": {
    "states": {
      "warning": 30,
      "critical": 15
    },
    "format-icons": ["", "", "", "", ""],
    "format-plugged": "{capacity}% ",
    "format-charging": "{capacity}% "
  },
```

Cprakhar:
### Description
This PR prioritizes `plugged` state over `full` when the battery is full and plugged in

Closes #3202


Alexays:
Thanks for the fix! Two things before this can be merged:

1. **Behavior regression**: it unconditionally rewrites `Full` → `Plugged` whenever the adapter is online, so a fully-charged laptop on AC would never report `Full` — that breaks existing `format-full` / `tooltip-format-full` / `.charged` setups. The rewrite should be opt-in (config option) or only apply in the specific state you're targeting, not globally.
2. The diff contains a stray `puts(status.c_str())` debug line and the branch is behind master — please rebase and drop the debug output.

Happy to merge once the `Full` case is preserved for existing configs. 🙏

Issue #4835, opened 2026-02-09
Image in sliders stop working and consume 80% of my CPU


Maxi-gasti:
This happen me today, the waybar don't show anything and consume very high rate CPU, i was discarding the config file of waybar and found that the module image in my slider was causing all. I dont know why, i use this config since 2024, i tried to use the logs for more details but i dont find anything. 

```jsonc
    "group/arch": {
      "orientation": "vertical",
      "drawer": {
        "transition-duration": 500,
        "children-class": "not-power",
        "transition-left-to-right": true
      },
      "modules": [
        "custom/image",
        "custom/poweroff",
        "custom/reboot",
        "custom/logout"
      ]
    },

    "custom/image": {
         "path": "/home/Maximus/.config/waybar/img/bluetooth.png",
         "size": 30,
        //"tooltip": false,
        //"on-click": "rofi -show drun",
        //"format": "<span size='20000'> </span>"
    },
```

khaneliman:
Are you on release ? 

https://github.com/Alexays/Waybar/pull/4390 seems to have changed default behavior to rapid updates instead of once. can you try https://github.com/Alexays/Waybar/pull/4838

Issue #4829, opened 2026-02-08
Waybar does not show


tudorjnu:
My waybar seems to not show up. It seems that it occupies the space and the debug does not give me any errors.

<details>
<summary>Logs attached</summary>
```
[2026-02-08 16:40:14.220] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/config
[2026-02-08 16:40:14.220] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/config.jsonc
[2026-02-08 16:40:14.220] [debug] Try expanding: $HOME/.config/waybar/config
[2026-02-08 16:40:14.220] [debug] Try expanding: $HOME/.config/waybar/config.jsonc
[2026-02-08 16:40:14.220] [debug] Found config file: /home/tudorjnu/.config/waybar/config.jsonc
[2026-02-08 16:40:14.220] [info] Using configuration file /home/tudorjnu/.config/waybar/config.jsonc
[2026-02-08 16:40:14.226] [info] Discovered appearance 'dark'
[2026-02-08 16:40:14.255] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/style-dark.css
[2026-02-08 16:40:14.255] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/style.css
[2026-02-08 16:40:14.255] [debug] Try expanding: $HOME/.config/waybar/style-dark.css
[2026-02-08 16:40:14.255] [debug] Try expanding: $HOME/.config/waybar/style.css
[2026-02-08 16:40:14.255] [debug] Found config file: /home/tudorjnu/.config/waybar/style.css
[2026-02-08 16:40:14.255] [info] Using CSS file /home/tudorjnu/.config/waybar/style.css
[2026-02-08 16:40:14.261] [debug] Output detection done: eDP-1 (Samsung Display Corp. 0x4174 )
[2026-02-08 16:40:14.262] [debug] window-rewrite is not defined or is not an object, using default rules.
[2026-02-08 16:40:14.262] [info] Hyprland IPC starting
[2026-02-08 16:40:14.263] [debug] Initializing workspaces
[2026-02-08 16:40:14.263] [debug] Initializing persistent workspaces
[2026-02-08 16:40:14.263] [info] Loading persistent workspaces from Hyprland workspace rules
[2026-02-08 16:40:14.277] [debug] battery: worker interval is 30000
[...]

PR #4838, opened 2026-02-09
fix(image): treat missing interval as once

khaneliman:
PR #4390 enabled millisecond intervals but changed image interval parsing so a missing interval could resolve to 1ms. That creates a hot update loop and can spike CPU and destabilize rendering in drawer/group setups (issue #4835).

Use explicit semantics: missing, null, non-numeric, or <=0 interval is treated as "once" (max sleep), while positive numeric values still support ms precision with a 1ms floor. This keeps the intended feature (sub-second polling when requested) without defaulting to busy looping.

Also align waybar-image(5) docs with runtime behavior.

Closes https://github.com/Alexays/Waybar/issues/4835 
Closes https://github.com/Alexays/Waybar/issues/4829

### DIFF
--- a/include/util/command.hpp
+++ b/include/util/command.hpp
@@ -172,8 +172,6 @@ inline int32_t forkExec(const std::string& cmd, const std::string& output_name) 
   return pid;
 }
 
-inline int32_t forkExec(const std::string& cmd) {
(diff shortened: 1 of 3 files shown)


Alexays:
Thanks!

PR #4843, opened 2026-02-10
fix(custom): prevent UI starvation (missing tooltips) when restart-interval is 0

### LINKED ISSUE #4842: [Bug] "restart-interval": 0 on one-shot scripts creates silent render loop that kills all tooltips
**Description:**

I have discovered that setting `"restart-interval": 0` on a custom module that executes a "one-shot" script (exits immediately with code 0) creates a tight restart loop that silently breaks the UI.

Because the script exits instantly, Waybar restarts it immediately. I believe this cycle stops the event loop, stopping tooltips from displaying.

**Reproduction Steps:**
1. Create a script `repro_script.sh` that outputs JSON and exits:
```bash
#!/bin/bash
echo '{"text": "HOVER ME", "tooltip": "If you see this, the bug is fixed."}'
```
2. Run Waybar with this minimal config:
```jsonc
{
    "layer": "top",
    "modules-center": ["custom/crasher", "clock"],
    "clock": {
        "tooltip": true,
        "tooltip-format": "<big>{:%Y %B}</big>\n<tt><small>{calendar}</small></tt>"

    },
    "custom/crasher": {
        "format": "{text}",
        "return-type": "json",
        "exec": "~/repro_script.sh",
        "restart-interval": 0  // <--- THE PROBLEM
    }
}
```

3. Hover over the Clock Module.

**Expected Behavior:**
The Clock tooltip should appear. If custom/crasher is restarting too fast, Waybar should warn the user or throttle the module (either way, it should have logs).

**Actual Behavior:**
The bar renders, but tooltips are completely dead for all modules. Logs show no errors, effectively creating a "silent failure."
Commenting out the restart-interval line will bring back the tooltips.

Image

**Environment:**
- Waybar v0.15.0
- OS: Arch Linux

tteeaa:
### Bug
When a custom module script exits immediately and `"restart-interval": 0` is set, tooltips for the **entire bar** stop working.

### Cause
As documented in man/waybar-custom.5.scd, the minimum restart interval is 1ms (0.001).

"Minimum value is 0.001 (1ms). Values smaller than 1ms will be set to 1ms."

A 1ms restart loop on a script that exits instantly creates a tight loop. This prevents mouse hover events (tooltips) from processing.

### Fix
Fix: I have added a safety check in src/modules/custom.cpp. If the user sets a restart-interval lower than 1000ms (1 second), the module now:
1. Logs a warning to spdlog explaining the issue.
2. Enforces a minimum delay of 1000ms.
This breaks the tight loop, allowing the UI to remain responsive even if a user misconfigures a fast-exiting script.

### Verification
- **Before:** Hovering over any module (e.g. Clock) showed no tooltip.
- **After:** Tooltips appear instantly, and a warning is logged:
  `[warning] Custom module: restart-interval (0ms) too low. Enforcing 1000ms to prevent UI starvation.`
image

### DIFF
--- a/src/modules/custom.cpp
+++ b/src/modules/custom.cpp
@@ -91,10 +91,18 @@ void waybar::modules::Custom::continuousWorker() {
       }
       if (config_["restart-interval"].isNumeric()) {
         pid_ = -1;
-        thread_.sleep_for(std::chrono::milliseconds(
-            std::max(1L,  // Minimum 1ms due to millisecond precision
-                     static_cast<long>(config_["restart-interval"].asDouble() * 1000))));
+
+        long interval_ms = static_cast<long>(config_["restart-interval"].asDouble() * 1000);
+
+        // Minimum interval (1 second) to prevent UI starvation from tight loops
+        if (interval_ms < 1000) {
+            spdlog::warn("Custom module: restart-interval ({}ms) is too low. Enforcing 1000ms to prevent UI starvation.", interval_ms);
+            interval_ms = 1000;
+        }
+        thread_.sleep_for(std::chrono::milliseconds(interval_ms));
+
         fp_ = util::command::open(cmd, pid_, output_name_);
+
         if (!fp_) {
           throw std::runtime_error("Unable to open " + cmd);
         }


Alexays:
Thanks! A few concerns here:

1. It forces a 1000ms floor on `restart-interval`, silently overriding valid user values (e.g. `0.1`) with a warning. We shouldn't clamp a user-configured interval — if a guard is needed it should be opt-in or documented, not a silent override.
2. The 'UI starvation' rationale is weak: `continuousWorker` runs on its own `thread_`, not the GTK main thread, so a short interval there doesn't starve the UI.
3. The diff uses 4-space indentation, which will fail the project's 2-space clang-format CI.

Could you rework so user-set intervals are respected (and run clang-format)? 🙏

Alexays:
Thanks for looking into this! We'd prefer not to take this change: it silently clamps a user-configured `restart-interval` to a 1s floor, overriding an explicit setting. If someone sets `0.1`, that's their choice. The 'UI starvation' concern also doesn't really apply here — `continuousWorker` runs on its own `thread_`, not the GTK main thread, so a short interval there doesn't block the UI. Since the clamp is the whole change, closing — but if you can demonstrate an actual starvation case, please reopen with details. 🙏

Issue #4610, opened 2025-10-30
MPRIS: Use first media player that isn't ignored


RossSmyth:
Hello! Thanks for the great project!

I have my mpris module setup like:

```json
"mpris": {
      "dynamic-order": [
         "position",
          "title",
          "artist",
          "album"
        ],
        "format": "{status_icon} | {dynamic}",
        "ignored-players": [
          "firefox"
        ],
        "status-icons": {
          "paused": "⏸",
          "playing": "▶",
          "stopped": "⏹"
        }
      },
```
But if I have spotify playing at the same time I watch a video on Firefox, the MPRIS widget goes away on my bar. I assume because it just sees that firefox is there, and doesn't check any others. 

But if I run `playerctl --list-all` I get:
```
firefox.instance_1_34
spotify_player
```
So waybar should be able to scan to the first player that is not ignore, then display that.

Issue #1954, opened 2023-01-17
MPRIS: needs some improvements when more than one player is active


rodrgz:
I'm seeing some problems with the module, regarding having more than one track/video playing at the same time.

Situation 1:

1. I put a song for example in mopidy, the MPRIS module indicates mopidy playing;
2. Then, I open a video on youtube through the browser and the module stops indicating mopidy to indicate youtube, so far ok;
3. The youtube video ends, and now the module should indicate mopidy, as it is playing, but it is empty. Re-indicating mopidy only if a pause/chanee song action is performed in mopidy.

Situation 2:

1. I put a song for example in mopidy, the MPRIS module indicates mopidy playing;
2. Then, I open a video on youtube through the browser and the module stops indicating mopidy to indicate youtube, so far ok;
3. I pause the video. And there is no way for me to select mopidy to pause.

In this case, it might be interesting to order the mopidy player in the background to be the one playing at the moment. Sort by order: play  → pause.
One thing that would solve this, and I think it would look elegant, would be the possibility to switch between all the players with the mouse scroll.

siggsy:
The problem also occurs when using `ignored-players` feature. For example:
- play something on a non ignored player
- start playing something on an ignored player
- this is where the issue occurs
  - *expected:* MPRIS module does not change
  - *actual:* MPRIS module disappears as if no player is playing
  
After checking some documentation for `playerctl` it appears that this is an issue of the `playerctld` proxy since there is no way to control to which player it listens.

Taking into the consideration the original problem of multiple players, the use of the `playerctld` proxy should be avoided.

ItsFireStorm:
For anyone looking for this, a potential "fix" is this.

You can keep the player hidden, but playerctld allows you to shift up or down the current players available. So whenever it disappears, you can use the following configuration to change the selected player to the one of your choosing.

```
"on-scroll-up": "playerctld shift"
"on-scroll-down": "playerctld unshift"
```

This is still a very improvised fix, as it would be better if MPRIS just ignored the player completely instead of hiding itself when an ignored player is playing audio.

Dieterbe:
the above options don't seem to work for me, using waybar 0.14
i'm not using ignored players. i just often have multiple things playing (e.g. youtube music and youtube videos), and i often need to pause the player which is not displayed, (e.g. pausing music to watch a video or vice versa).
[...]

PR #4846, opened 2026-02-11
fix(mpris): fall back to next non-ignored player and prefer playing players

BlueManCZ:
When using `playerctld`, the MPRIS module had several issues with player selection:

- **Ignored players caused widget to disappear** — If the most recently active player was in
`ignored-players`, `getPlayerInfo()` returned `std::nullopt` with no fallback, hiding the widget
entirely even when other valid players were available.
- **Non-playing players took priority** — Opening a new player (e.g. Spotify) would immediately
take over the widget even if another player (e.g. Tidal) was actively playing, because playerctld
 orders by most recently active rather than playback state.
- **Widget flickered on player changes** — `onPlayerStop` hid the widget before triggering the
update, causing a visible flash before `update()` re-showed it with the correct player.

## Changes

- Iterate through playerctld's activity-ordered player list, skipping ignored players, and prefer
 a **playing** player over a paused/stopped one
- When the selected player differs from the playerctld proxy's top player, create a direct
`PlayerctlPlayer` connection for correct metadata
- Remove premature `event_box_.set_visible(false)` from `onPlayerStop` — `update()` already
handles visibility

## Test plan

- Configure `ignored-players` with a browser (e.g. `["firefox"]`), have another player paused
 — widget should show the paused player with correct metadata when browser plays media
- Play Tidal, then open Spotify without playing — widget should stay on Tidal
- Play Spotify while Tidal is paused — widget should switch to Spotify
- Pause all players — widget displays last active player as paused
- With all players ignored — widget should hide
- Player changes should not cause visible flicker

Fixes: #1954, fixes: #3879, fixes: #4610

### DIFF
--- a/include/modules/mpris/mpris.hpp
+++ b/include/modules/mpris/mpris.hpp
@@ -78,6 +78,7 @@ class Mpris : public ALabel {
 
   PlayerctlPlayerManager* manager;
   PlayerctlPlayer* player;
+  PlayerctlPlayer* last_active_player_ = nullptr;
   std::string lastStatus;
   std::string lastPlayer;
 
--- a/src/modules/mpris/mpris.cpp
+++ b/src/modules/mpris/mpris.cpp
@@ -178,6 +178,7 @@ Mpris::Mpris(const std::string& id, const Json::Value& config)
 }
 
 Mpris::~Mpris() {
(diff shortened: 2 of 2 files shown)


Alexays:
Thanks!

PR #4849, opened 2026-02-11
battery: power switch events

44vladimirov:
Related to https://github.com/Alexays/Waybar/pull/4208 power switch events `on-discharging` and `on-charging` have been introduced, man pages updated.

These events are executed *after* events of type `on-[status]-[state]` iff `[status]` has switched.


Alexays:
Thanks!

PR #4852, opened 2026-02-12
fix(keyboard-state): fix segfault on device hotplug removal

### LINKED ISSUE #4851: keyboard-state: segfault in libinput_device_ref due to race condition in hotplug handling
(Disclaimer: this is a real bug observed by me, but the analysis below was made by the AI and has not been reviewed or verified by me. I'm not sure if this project have an AI-policy, but feel free to close it if you don't like AI-contributions)

## Bug description

The `keyboard-state` module crashes with SIGSEGV in `libinput_device_ref` when an input device appears in `/dev/input/` that libinput cannot open. This happens on laptops with integrated keyboards when virtual input devices (power button, lid switch, etc.) are created or re-enumerated.

## Stack trace

```
Stack trace of thread 2770356:
#0  0x00007f078a47ddf4 libinput_device_ref (libinput.so.10 + 0x12df4)
#1  /usr/bin/waybar + 0x19372d   ← tryAddDevice()
#2  /usr/bin/waybar + 0x193c0b   ← hotplug_thread_ (IN_CREATE handler)
#3  /usr/bin/waybar + 0x91f16    ← thread runner
#4  execute_native_thread_routine (libstdc++.so.6 + 0xe55a4)
```

Signal: SIGSEGV (11)

## Root cause analysis

There are three related bugs in `src/modules/keyboard_state.cpp`:

### 1. Missing NULL check in tryAddDevice() (direct crash cause)

`tryAddDevice()` calls `libinput_path_add_device()` which returns NULL on failure, then passes the result directly to `libinput_device_ref()` without checking:

```cpp
auto device = libinput_path_add_device(libinput_, dev_path.c_str());
libinput_device_ref(device);  // ← segfaults when device is NULL
```

On a laptop with an integrated keyboard, `/dev/input/` devices still come and go — virtual devices for power buttons, lid switches, suspend/resume re-enumeration, etc. When the `hotplug_thread_` detects `IN_CREATE` for one of these and libinput can't open it, the NULL return causes the crash.

### 2. Missing `libinput_path_remove_device()` on hotplug removal

When `hotplug_thread_` detects `IN_DELETE`, it only erases the device from the `libinput_devices_` map but never calls `libinput_path_remove_device()`. Compare with the destructor which correctly calls it. This leaves dangling pointers in the libinput context.

### 3. No thread synchronization on `libinput_devices_`

`libinput_devices_` (`std::unordered_map`) is accessed from three threads without any mutex:

- **Main / GTK thread**: `update()` reads `libinput_devices_.begin()`
- **`hotplug_thread_`**: calls `tryAddDevice()` (inserts) and `libinput_devices_.erase()` (removes)
- **`libinput_thread_`**: indirectly depends on devices via the libinput context

## Environment

- Waybar version: 0.14.0 (all three bugs also present in 0.15.0 — no changes to keyboard_state.cpp between versions)
- libinput: 1.30.1
- OS: Arch Linux, kernel 6.18.7
- Wayland compositor: sway
- Hardware: Laptop with integrated keyboard (no USB disconnection involved)

tobixen:
## Summary

Fixes a segfault (SIGSEGV in `libinput_device_ref`) in the keyboard-state module when input devices appear or disappear in `/dev/input/`.

Three bugs fixed:

- **Missing NULL check (crash cause)**: `tryAddDevice()` passes the return value of `libinput_path_add_device()` directly to `libinput_device_ref()` without checking for NULL. On laptops, virtual input devices (power buttons, lid switch, etc.) appear and disappear; if libinput can't open one, the NULL return causes the segfault.
- **Missing cleanup on device removal**: The `IN_DELETE` handler erased devices from the map without calling `libinput_path_remove_device()`, leaving dangling pointers in the libinput context.
- **Thread safety**: Added `std::mutex` to protect `libinput_devices_` which was accessed from 3 threads without synchronization.

Fixes #4851

## Test plan

- [ ] Build waybar with this patch
- [ ] Verify keyboard-state module works normally
- [ ] Unplug/replug a USB keyboard while waybar is running — should not crash
- [ ] On a laptop: trigger device enumeration (e.g., suspend/resume) — should not crash

### DIFF
--- a/include/modules/keyboard_state.hpp
+++ b/include/modules/keyboard_state.hpp
@@ -3,6 +3,7 @@
 #include <fmt/chrono.h>
 #include <gtkmm/label.h>
 
+#include <mutex>
 #include <set>
 #include <unordered_map>
 
@@ -41,6 +42,7 @@ class KeyboardState : public AModule {
 
   struct libinput* libinput_;
   std::unordered_map<std::string, struct libinput_device*> libinput_devices_;
+  std::mutex devices_mutex_;  // protects libinput_devices_
   std::set<int> binding_keys;
 
   util::SleeperThread libinput_thread_, hotplug_thread_;
--- a/src/modules/keyboard_state.cpp
+++ b/src/modules/keyboard_state.cpp
@@ -232,9 +232,12 @@ waybar::modules::KeyboardState::KeyboardState(const std::string& id, const Bar& 
           }
           tryAddDevice(dev_path);
         } else if (event->mask & IN_DELETE) {
+          std::lock_guard<std::mutex> lock(devices_mutex_);
           auto it = libinput_devices_.find(dev_path);
           if (it != libinput_devices_.end()) {
             spdlog::info("Keyboard {} has been removed.", dev_path);
+            libinput_path_remove_device(it->second);
+            libinput_device_unref(it->second);
             libinput_devices_.erase(it);
           }
         }
@@ -245,6 +248,7 @@ waybar::modules::KeyboardState::KeyboardState(const std::string& id, const Bar& 
 }
 
 waybar::modules::KeyboardState::~KeyboardState() {
+  std::lock_guard<std::mutex> lock(devices_mutex_);
   for (const auto& [_, dev_ptr] : libinput_devices_) {
     libinput_path_remove_device(dev_ptr);
   }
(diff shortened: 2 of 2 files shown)


tobixen:
This pull request was "vibe-coded".  This is a small fix, but the efforts of reviewing it is probably bigger than the efforts I've done creating this pull request.  It's totally acceptable for me if this pull request is simply closed without any explanations given.

The crash I observed happened during heavy thrashing (but no OOM-killing - so that's not the reason why waybar crashed).  I don't have a USB keyboard nor a bluetooth keyboard so it's hard for me to verify the fix manually.  I'm not even sure if the analysis given by Claude is correct or not.

tobixen:
I prodded Claude on how I possibly can be struck by a bluetooth/usb hotplug issue when I have a laptop with an integrated keyboard, and it changed things a bit (including the issue description and pull request description).

Claude says:

>   The stack trace definitively shows the crash in tryAddDevice() called from hotplug_thread_ (an inotify IN_CREATE event), but I don't  actually know what created the device. Could have been anything — some software creating a virtual input device, a udev rule, something else entirely.                                                                                                              

Please give a hint if I should (ask Claude to) do more research into this. (note to self: `claude --resume 707df72a-e9ca-4cd0-a451-641c75061bb6` to do more research here)

Alexays:
LGTM, thx!

PR #4857, opened 2026-02-12
fix: use pango markup for consistent formatting in format and tooltip-format

Cprakhar:
### Description
This PR replaces all `format` and `tooltip-format` to use pango markup for consistent formatting across modules.


Alexays:
Thanks!

PR #4858, opened 2026-02-13
memory: use markup rather than text for memory tooltip

chrisvittal:
Seems a little strange to me that we can format the tooltip text, but can't use markup like we can elsewhere.


chrisvittal:
As an aside. It feels like the documentation around tooltips needs much improvement. And there should certainly be consistency in what can be marked up vs what is text only.

Issue #4791, opened 2026-01-26
Feature request: hyprland/workspace group/count same icon


pinusc:
In hyprland, I often have a lot of the same window open in a workspace (e.g. pdf viewer or terminal); This makes the workspace text very long, because there's one icon for every window, even if the windows are mostly all the same.

I would like an option to group icons together, e.g.

`|📄x5 💻|` instead of `|📄📄📄💻📄📄|`

This could be configurable based on number of icons to display before they are grouped. E.g. if `group_more_than_n_icons = 3`, then it would show `|📄📄 💻|` but `|📄x3💻|`, and also `group_format`, which in this example would be `x{n}`. 

This would also allow to only ever show one icon per type of window, just set `n=1` and `group_format` to empty.

Also this would probably be useful to other WMs as well, but I'm most familiar with hyprland.

PR #4861, opened 2026-02-14
feat(hyprland/workspaces): grouping icons

eonphi:
This PR allows the hyprland/workspaces to group icons of the same type, most recently suggested by #4791.
There's two new config values for the module: `window-rewrite-group-threshold` determines the threshold at which grouping happens (including `1` should the user want that.. for consistency with singular icons, maybe), while set to `0` (default) no grouping will happen; `window-rewrite-group-format` is the formatting string for the icon groups, defaults to a simple `{icon}×{count}`, and with pango's markup can do things such as `{icon}<span rise='3000' font_size='90%'>{count}</span>` (a nicer-to-the-eyes alternative to `baseline_shift=superscript`), just as an example. I've attached three screenshots with no grouping, default with threshold 2, and the above span format string as an example.

no group (3)  <br />

group default example <br /> 
group superscript example <br />


khaneliman (approved):
Tested and seems to work. Like the concept

Alexays:
Can you add the options to the man? 🙏

eonphi:
@Alexays sure! Added the man page entries

PR #4889, opened 2026-03-01
fix(util): ensure SleeperThread lifecycle safety and thread sync

khaneliman:
This PR addresses critical stability issues in the SleeperThread utility class. It fixes a
known crash during thread reassignment and resolves potential data races on
control flags.

Motivation

1. Crash on Reassignment: The operator= implementation previously replaced\
   the std::thread object without ensuring the old thread was stopped and\
   joined. In C++, destroying or reassigning a joinable std::thread calls\
   std::terminate(), causing Waybar to crash during certain configuration\
   reloads.
2. Data Races: The do_run_ and signal_ flags were accessed across multiple\
   threads (management thread and worker thread) without consistent\
   synchronization. This could lead to missed wake-up signals or threads\
   failing to exit promptly.

Proposed Changes

- Safe Reassignment: Updated SleeperThread::operator= to explicitly stop()\
  and join() any existing worker thread before starting a new one.
- Atomic Control Flags: Converted do_run_ and signal_ to std::atomic<bool>.\
  This ensures thread-safe access and visibility across threads without the\
  overhead of a mutex for simple flag checks.
- Explicit Memory Ordering: Used std::memory_order_relaxed for atomic\
  operations, which is sufficient for these lifecycle flags as they do not\
  guard non-atomic data.
- State Reset: Ensured that control flags are reset to their initial state\
  within the assignment operator before the new thread begins execution.

Testing Performed\
I performed Red-Phase testing to verify the effectiveness of the fix:

- Baseline (Broken): Reverting the changes to the master state caused the\
  new SleeperThread tests to fail with SIGABRT (due to std::terminate)\
  during reassignment.
- Fixed State: All tests passed, including:
  - SleeperThread reassignment does not terminate process: Verifies safe\
    hot-swapping of worker functions.
  - SleeperThread control flags are stable under concurrent wake and stop: A
    stress test that interleaves signals to ensure no hangs or races\
    occur.

### DIFF
--- a/test/utils/sleeper_thread.cpp
+++ b/test/utils/sleeper_thread.cpp
@@ -19,24 +19,61 @@ SafeSignal<bool>& prepare_for_sleep() {
 }  // namespace waybar::util
 
 namespace {
+int run_in_subprocess(int (*task)()) {
+  const auto pid = fork();
+  if (pid < 0) {
+    return -1;
+  }
+  if (pid == 0) {
+    alarm(5);
+    _exit(task());
+  }
+
+  int status = -1;
+  if (waitpid(pid, &status, 0) != pid) {
+    return -1;
+  }
+  if (!WIFEXITED(status)) {
+    return -1;
+  }
+  return WEXITSTATUS(status);
+}
+
 int run_reassignment_regression() {
   waybar::util::SleeperThread thread;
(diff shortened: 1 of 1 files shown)


Alexays:
LGTM, thx!

PR #4892, opened 2026-03-02
fix(command): return non-zero when child exec fails

khaneliman:
Child exec failure paths were returning success, which masked command launch
errors from callers.

I switched the child-side failure exits to _exit(127) and added errno-specific
logging so failures propagate with actionable diagnostics.


Alexays:
Thanks!

Issue #4894, opened 2026-03-02
Float Interval < 1 always creates 1ms interval for custom modules


BryceGust:
This always evaluates to 0 for any float interval between 0 and 1:
https://github.com/Alexays/Waybar/blob/fd086d0f3366206417bcf59cc18679e32154072f/src/ALabel.cpp#L29

So the std::max call evaluates to 1ms

Looks to have been introduced in https://github.com/Alexays/Waybar/commit/d8e23924106c790c11a6b8d64367fa759a20c0c1

I believe we would just need to move the `* 1000` back into the cast. Small typo.

<h2>Story:</h2>
I was playing around with a custom module and wanted it to update, so I used interval. I had a cpu monitor up and noticed that my CPU went from normally 0% usage to 8% when running waybar with my custom module. I had my interval set at 0.2 at this time. I thought my bash scripting was just terribly inefficient!

I wanted to get that CPU down though, so I bumped the interval up to 0.5 and did not notice a difference. I went to 1 and saw it went back down to 0%. I set it to 0.9 and it shot back up again.

At this point I wanted to understand what was going on so I wrote this custom module just so I could see the rate of change (Just a script that generates a random string locally):

```json
"custom/random": {
  "format": "{text}",
  "exec": "cat /dev/random | tr -dc 'a-zA-z0-9' | fold -w 8 | head -n 1",
  "interval": 0.9
}
```

When I ran waybar at this point the bar was updating extremely quickly, wayyyy faster than once every 900ms as I would have expected.

I set the interval to 0.001 and I couldn't tell the difference, which is when I looked into the code.

I can test and put up a PR if I can find the time, but I wanted to create an issue first just in case.

PR #4895, opened 2026-03-03
fix: float interval cast

BryceGust:
Fixes https://github.com/Alexays/Waybar/issues/4894

I did test the fix locally and it worked and passed all tests. I was unsure of where to add a regression test.

Apologies for lack of direct evidence. I would normally put up a video, but my linux workstation is where I tested this and that is a separate computer that I do not have this profile set up on.


Alexays:
Thanks!

BryceGust:
> Thanks!

Absolutely, and thank you!

PR #4898, opened 2026-03-03
perf(memory): optimize string operations; remove deep copies, memdup, and icon theme rescanning

khaneliman:
Bunch of different optimizations while profiling with heaptrack and valgrind to reduce memory churn. Mostly minimal gains from a lot of it but the sni optimizations do seem to help reduce a lot.


Alexays:
LGTM, thanks!

PR #4905, opened 2026-03-04
fix(mpris): disconnect GLib signals before destroying objects

sw1nn:
Waybar SEGVs in `Glib::DispatchNotifier::pipe_io_handler` when the MPRIS module is enabled. The crash is intermittent because it requires a race between signal emission and object destruction: a playerctl GLib signal callback (e.g. onPlayerPlay) calls `dp.emit()`, which writes a pointer to the Dispatcher into an internal pipe. 

If the Mpris object is destroyed before the GLib main loop reads that pipe entry, pipe_io_handler dereferences a dangling pointer. This typically occurs when a media player appears or vanishes on D-Bus (browser closing, player quitting) or during waybar shutdown/config reload.

The root cause is that `~Mpris()` calls `g_object_unref()` on the manager and player GObjects without first disconnecting the signal handlers that hold raw `this` pointers. If playerctl holds additional references to these GObjects, they survive the unref and can still fire signals targeting the already-destroyed Mpris instance.

Adopt the same cleanup pattern used by the Wireplumber module: call `g_signal_handlers_disconnect_by_data()` to sever all signal connections referencing `this` before releasing the GObjects with `g_clear_object()`. This guarantees no callbacks can enqueue stale Dispatcher notifications after teardown begins.

Additionally:
- Clean up old player in onPlayerNameAppeared before replacing it, fixing a GObject leak and accumulation of dangling signal connections
- Remove duplicate onPlayerStop signal registration (copy-paste bug)

### DIFF
--- a/src/modules/mpris/mpris.cpp
+++ b/src/modules/mpris/mpris.cpp
@@ -161,8 +161,7 @@ Mpris::Mpris(const std::string& id, const Json::Value& config)
   if (player) {
     g_object_connect(player, "signal::play", G_CALLBACK(onPlayerPlay), this, "signal::pause",
                      G_CALLBACK(onPlayerPause), this, "signal::stop", G_CALLBACK(onPlayerStop),
-                     this, "signal::stop", G_CALLBACK(onPlayerStop), this, "signal::metadata",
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

PR #4915, opened 2026-03-09
refactor(custom): move continuous exec onto GLib command stream 

khaneliman:
`custom` used two different blocking paths that could stall module updates.

  For continuous commands, the module read stdout with a blocking getline()
  loop. That worked for simple cases, but it kept long-running custom scripts
  tied to a worker thread that could sit indefinitely waiting for a newline or
  EOF.

  For interval- and signal-driven custom modules, refreshes also synchronously
  waited on previously launched event children before running the next update.
  If an older on-click/on-scroll child was still alive, the next refresh could
  be delayed behind that wait.

  This change addresses both cases:

  - continuous custom commands now run through a GLib main-loop-backed command
    stream instead of a blocking FILE*/getline() loop
  - interval refreshes now reap old event children with waitpid(..., WNOHANG)
    instead of blocking on them

Tested with some simple custom modules: 

```json
  "custom/test-github-reap": {
    "exec": "sh -c 'printf \"{\\\"text\\\":\\\"%s\\\"}\\n\" \"$(date +%T)\"'",
    "format": " (gh {text}) ",
    "interval": 5,
    "on-click": "sleep 20",
    "return-type": "json",
    "tooltip": false
  },
  "custom/test-hyprsunset-reap": {
    "exec": "sh -c 'printf \"{\\\"text\\\":\\\"%s\\\",\\\"alt\\\":\\\"%s\\\"}\\n\" \"$(date +%T)\" \"$(date +%S)\"'",
    "format": "( {text} )",
    "tooltip-format": "sec {alt}",
    "interval": 3600,
    "signal": 1,
    "on-click": "sleep 20",
    "on-scroll-up": "sleep 20",
    "return-type": "json"
  },
```

  - test-github-reap
    Click it once.
    Keep watching for 20 seconds.
    On this branch: the timestamp keeps updating every 5 seconds while the
    sleep 20 child is still running.
    Current code, the module freezes until that child exits.

  - test-hyprsunset-reap
    Click or scroll it once.
    Immediately run:

    `pkill -RTMIN+1 waybar`

    On this branch: it updates right away. Current code will not update as the signal-driven refresh waits behind the sleeping
    child.

### DIFF
--- a/include/modules/custom.hpp
+++ b/include/modules/custom.hpp
@@ -3,10 +3,12 @@
 #include <fmt/format.h>
 
 #include <csignal>
+#include <memory>
 #include <string>
 
 #include "AIconLabel.hpp"
 #include "util/command.hpp"
+#include "util/command_line_stream.hpp"
 #include "util/json.hpp"
 #include "util/sleeper_thread.hpp"
 
@@ -22,6 +24,9 @@ class Custom : public AIconLabel {
  private:
   void delayWorker();
   void continuousWorker();
+  void startContinuousProcess(bool throw_on_failure);
+  void handleContinuousProcessExit(int exit_code);
+  void scheduleContinuousRestart();
   void waitingWorker();
   void parseOutputRaw();
   void parseOutputJson();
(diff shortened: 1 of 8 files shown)


khaneliman:
Sorry, keep meaning to come back to this to figure out the freebsd path error.

khaneliman:
@Alexays alright, think i got the freebsd issue figured out

Alexays:
Thanks @khaneliman, just need a rebase and we’re good.

khaneliman:
> Thanks @khaneliman, just need a rebase and we’re good.

Rebased and resolved conflicts

PR #4925, opened 2026-03-16
Fix menu-actions GTK callback pointer handling

B2krobbery:
Fix pointer lifetime handling for menu-actions GTK callbacks.

The previous implementation passed a temporary pointer (`c_str()`) to the GTK callback, which could become invalid after the function scope ended.

This patch ensures the pointer remains valid by allocating a copy of the string using `g_strdup`, providing a stable memory reference for the callback.

Verified build locally and tested with Waybar.

### DIFF
--- a/src/ALabel.cpp
+++ b/src/ALabel.cpp
@@ -118,7 +118,7 @@ ALabel::ALabel(const Json::Value& config, const std::string& name, const std::st
         menuActionsMap_[key] = it->asString();
         g_signal_connect(submenus_[key], "activate",
                  G_CALLBACK(handleGtkMenuEvent),
-                 (gpointer)g_strdup(menuActionsMap_[key].c_str()));             
+                 (gpointer)g_strdup(menuActionsMap_[key].c_str()));          
       }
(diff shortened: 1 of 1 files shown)


staticssleever668:
@B2krobbery

> could become invalid after the function scope ended

"could" in what case exactly? This is the constructor and `menuActionsMap_` is a member variable never reassigned later.
This is just leaking memory for no gain. Try `valgrind --leak-check=yes -- ./build/waybar` and look for
`waybar::ALabel::ALabel`.
The code in this PR is also incorrectly formatted. The CI job `clang-format / lint` might be broken.
@Alexays sorry for the ping but this PR should be reverted.

B2krobbery:
The intention here was to be defensive about the data passed to the GTK callback.

While menuActionsMap_ does persist for the lifetime of the object, relying on implicit lifetime guarantees can be fragile depending on how the code evolves. The change was aimed at making the ownership explicit and avoiding potential issues if the storage pattern changes in the future.

That said, I understand your point regarding the current implementation.

Issue #4920, opened 2026-03-14
clock module - change calendar background color


e1kfL0ck:
I would like to change the calendar background color. 

I did not find in the [docs](https://github.com/Alexays/Waybar/wiki/Module:-Clock), but it seems doable as [example 2 calendar](https://github.com/Alexays/Waybar/wiki/Module:-Clock#example).

Running Fedora with hyprland.

Thanks

0x2aAs:
Since the shown calendar is just the tooltip of the clock module, you can add this to waybar/style.css
```
tooltip {
    background: rgba(30,35,35,0.85);
}
```
Of course you can specify the background with a hexadecimal value instead of rgba, but the example you mentioned has a slightly transparent background. 
BUT: This style applies to every tooltip of any other module, however. I currently don't know, if you can style the tooltip per module with the CSS file.

Where the example uses `<span color=` in lines like these:
```
                    "format": {
                              "months":     "<span color='#ffead3'><b>{}</b></span>",
                              "days":       "<span color='#ecc6d9'><b>{}</b></span>",
                              "weeks":      "<span color='#99ffdd'><b>W{}</b></span>",
                              "weekdays":   "<span color='#ffcc66'><b>{}</b></span>",
                              "today":      "<span color='#ff6699'><b><u>{}</u></b></span>"
                              }
```
You can also use `<span bgcolor=` to change the background color per calendar element like months, days etc. This also works for other modules, but with this you cannot change the background color of tooltips as a whole, only parts of it.

B2krobbery:
The calendar is rendered as part of the clock module’s tooltip, so its background is controlled by the global tooltip styling in Waybar.

You can change it using CSS like:

```css
tooltip {
    background: rgba(30,35,35,0.85);
}
```

However, this applies to all tooltips, not just the clock/calendar.

If you want more control inside the calendar, you can use `<span bgcolor=...>` in the calendar format to style specific parts (days, months, etc.), but not the full tooltip background.

As of now, Waybar does not support per-module tooltip styling, so changing only the calendar background independently isn’t possible.

nickjj:
A few weeks ago I put together [this PR](https://github.com/Alexays/Waybar/pull/4896) to solve this problem in a way that lets you use classes to control the color of each component of the calendar's tooltip. I hope someone can review / approve it. I've been using a custom fork of Waybar with it and it's working great.

Alexays:
[...]

PR #4930, opened 2026-03-18
docs: clarify global tooltip styling behavior

B2krobbery:
Related to #4920

This PR adds a clarification about tooltip styling behavior in Waybar.

Tooltip appearance is controlled globally via CSS and cannot be scoped per module. This is a common source of confusion.


nickjj:
I was able to get CSS classes to work for the clock module's tooltip in https://github.com/Alexays/Waybar/pull/4896.

It's not "really" using classes but it allows the behavior of what I imagine most folks want which is the ability to control the colors through a CSS file for the individual calendar styles.

B2krobbery:
Ahh I see, thanks for pointing that out!

From what I understand it's more of a workaround rather than true CSS scoping, so I'll update the wording to make that clearer.

PR #4946, opened 2026-03-22
fix: prevent resource leak when rebinding Wayland globals

B2krobbery:
Wayland globals (xdg_output_manager and idle_inhibit_manager) are assigned via wl_registry_bind
without checking for existing instances.

If handleGlobal is triggered multiple times, previously bound objects may be overwritten
without being released, leading to potential resource leaks.

This patch ensures that existing objects are destroyed before reassignment.


B2krobbery:
I checked the other wl_registry_bind usages they aren’t reassigned like in handleGlobal, so they don’t have the same overwrite issue.
In this path the same global can be rebound, so the previous instance needs to be destroyed before binding again.
Built and ran Waybar locally  no regressions.

PR #4960, opened 2026-03-29
feat(client): add support for 8-bit hex color codes in CSS

keepo-dot:
This allows users to use #RRGGBBAA format in their style.css. The client now detects 8-bit hex codes, transforms them into GTK-compatible rgba() syntax, and loads the modified data into the CSS provider.

- Added utility to detect 8-bit hex patterns.
- Added transformation logic to convert hex alpha to decimal.
- Intercepted CSS loading in Client::setupCss to handle the conversion.

I have tested compilation and building. As well as testing 8 bit hex values in styles.css to verify functionality.

### DIFF
--- a/include/util/hex_checker.hpp
+++ b/include/util/hex_checker.hpp
@@ -3,8 +3,15 @@
 #include <string>
 
 /**
+ * Result of transforming 8-bit hex codes to rgba().
+ */
+struct TransformResult {
+  std::string css;
+  bool was_transformed;
+};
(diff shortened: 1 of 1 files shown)


Alexays:
Thanks!

keepo-dot:
If I want to add onto this (error handling & such, though I don't think I need it since if the file doesn't exist waybar handles it), should I continue off of this PR or create a new one?

PR #4962, opened 2026-03-29
fix: prevent resource leak when rebinding dwl globals

B2krobbery:
zdwl_ipc_manager_v2 and wl_seat are assigned via wl_registry_bind without checking for existing instances.

If handle_global is triggered multiple times, previously bound objects may be overwritten without being released, leading to potential resource leaks.

Destroy existing objects before rebinding to prevent resource leaks.

Tested by building and running Waybar locally. Verified normal behavior and no regressions.
Checked with valgrind (leak-check=summary) and confirmed no definite or indirect leaks.


Alexays:
Thanks!

PR #4963, opened 2026-03-30
fix : dwl/window no waybar empty class and added current app id class

psi314:
In  `dwl/window`
Fixes : 
image

* this red thing in middle is background of #window module but there is no way to make it transparent when in workspace with no window , this pull request addresses this issue. so this `window#waybar.empty #window` selector can be used to hide it .

Adds : 
* like [this](https://github.com/Alexays/Waybar/wiki/Module:-Hyprland#style-1:~:text=This%20will%20change%20the%20color%20of%20the%20entire%20bar%20when%20either%20Chromium%20or%20kitty%20occupy%20the%20screen) , added support for window#waybar.<app_id> . 

These styles can be useful for a person using dwl & mangowc.

### DIFF
--- a/include/modules/dwl/window.hpp
+++ b/include/modules/dwl/window.hpp
@@ -30,7 +30,11 @@ class Window : public AAppIconLabel, public sigc::trackable {
   std::string title_;
   std::string appid_;
   std::string layout_symbol_;
+  std::string oldAppid_;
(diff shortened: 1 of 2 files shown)


psi314:
I dont know why this is not merged ?

Issue #4901, opened 2026-03-03
Waybar closes after I power off the PS4 controller


RovingFox:
When I hold the PS button to power the PS4 controller off, after the controller powers off, waybar also closes and I have to turn it back on from terminal. I think this started happening after the latest update.

tsuza:
Same here.

tsuza:
Here is what gets logged when it happens:

> terminate called after throwing an instance of 'std::runtime_error'
>   what():  Could not watch events for /sys/class/power_supply/ps-controller-battery-1c:96:5a:5f:3a:b6
> 
> fish: Job 1, 'waybar &' terminated by signal SIGABRT (Abort)

kimtodaro:
This happens when I turn off my non-bluetooth wireless headphones as well.

ChloeThePuppyGirl:
same issue but with ps5 dualsense, the workaround ive been using when waybar does die due to shutting off my dualsense is `waybar & quit` that way i can keep waybar and close terminal and not have to reboot... probably should make a new issue just for ps5 dualsense, but id end up linking back to this anyway so 🤷

tsuza:
I've been using `nohup waybar &`.

ChloeThePuppyGirl:
just tried it here a minute ago, does give a cleaner output and less worry for if `mod + q` will end up killing waybar when i close the terminal

Mrpaoo:
I try to reproduce in my device and check the backtrace in core dump file. It turns out that it's kind like a Time-of-check to time-of-use problem.
  The bt here


```#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007f16d0789af3 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:89
#2  0x00007f16d072f1a0 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f16d07165fe in __GI_abort () at abort.c:77
#4  0x00007f16d0a97bf6 in __gnu_cxx::__verbose_terminate_handler () at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/vterminate.cc:95
[...]

PR #4982, opened 2026-04-09
fix the io failure for hotplug-in device

Mrpaoo:
related issue #4901. When remove the device that need power-management, Waybar would crash due to uncatch exception


Mrpaoo:
fix the io failure problem mentioned at #4901. For hot-plugin device, it is inevitable to face io failure especially when you create a ifstream obj and using operator >> to access it while you remove your device, and this is the time when crash happen.
  Here operator>> may throw an exception and outer `try-catch` blk cant catch it. So just catch it and do nothing

Alexays:
Thanks!

PR #4986, opened 2026-04-12
Add "ignore-workspaces" option to sway/workspaces module

maltherd:
This is a very small change I needed personally. 

My use case is that I want to keep a bunch of workspaces open for my work time, and others for my free time. When working, I don't want to see (and cycle through) the free-time ones, and vice versa.

I wrote a little script to rename workspaces 1 → _1 and vice-versa. I just needed a way to hide the underscore-prefixed workspaces.

This is not great code, as the workspace stays hidden even when focused, but perhaps it will help searchers.

Have a nice day

### DIFF
--- a/include/modules/sway/workspaces.hpp
+++ b/include/modules/sway/workspaces.hpp
@@ -30,6 +30,8 @@ class Workspaces : public AModule, public sigc::trackable {
   static int convertWorkspaceNameToNum(const std::string& name);
   static int windowRewritePriorityFunction(std::string const& window_rule);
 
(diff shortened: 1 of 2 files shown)


maltherd:
Note: I just noticed my intended behavior can also be achieved by creating a headless output in sway (not documented officially yet, but visible in https://www.reddit.com/r/swaywm/comments/k1zl41/thank_you_devs_free_ipad_repurposed_as_a_second/) and sending the workspaces to it.

Although, it has a bad habit of stealing the 1 workspace into oblivion, I reckon that could be fixed.

PR #4997, opened 2026-04-19
feat(niri): add workspace-taskbar support

sewaustav:
**Description:**
This PR adds a workspace-taskbar mode to the niri/workspaces module, bringing functionality similar to the hyprland/workspaces module.
**Motivation:**
Currently, Niri users lack a native way to visualize windows and app icons within workspaces directly on the bar. While the wlr/taskbar module is an available option, it does not fully meet the needs for a structured taskbar; it often lacks sufficient organization of application icons and fails to maintain a logical or desired order, which is essential for efficient workflows on tiling compositors like Niri.

A native solution within niri/workspaces provides better integration, stable icon placement, and cleaner control over window management, specifically tailored for Niri's scrolling compositor.

**Key Changes:**

- **Refactoring:** I refactored the logic within the niri/workspaces module. The logic is now split into two entities: Workspaces (the controller/container) and Workspace (the entity/element). This allows for a more flexible rendering approach beyond simple button text.
- **New Feature:** Added workspace-taskbar support.
- Enables app icon display per workspace.
- Clicking an icon focuses the corresponding window.
- Middle-clicking an icon closes the window.
- **Documentation:** Updated man/waybar-niri-workspaces.5.scd with the new configuration options.

**Testing & Compatibility:**

- Backward Compatibility: Tested with existing configurations; no breaking changes detected.
- Stability: The module runs stably with no memory leaks detected during various usage scenarios.
- Icons: Verified compatibility with major icon themes, including Breeze, Adwaita, and Papirus. The module correctly retrieves and displays icons as expected.
**Configuration Example:**
```
"niri/workspaces": {
    "format": "{icon}",
    "workspace-taskbar": {
        "enable": true,
        "icon-size": 18
    }
}
```
image

### DIFF
--- a/include/modules/niri/workspace.hpp
+++ b/include/modules/niri/workspace.hpp
@@ -1,0 +1,47 @@
+#pragma once
+
+#include <gtkmm/box.h>
+#include <gtkmm/button.h>
+#include <gtkmm/image.h>
+#include <gtkmm/label.h>
+#include <json/value.h>
+
+#include <string>
+
+namespace waybar::modules::niri {
+
+class Workspaces;
+
+class Workspace {
+ public:
+  Workspace(const Json::Value& workspace_data, Workspaces& manager);
+  ~Workspace() = default;
+
+  Workspace(const Workspace&) = delete;
+  Workspace& operator=(const Workspace&) = delete;
+
+  Gtk::Button& button() { return button_; }
+  uint64_t id() const { return id_; }
+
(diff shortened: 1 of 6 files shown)


ADIX7:
Can a quick follow up be added? Currently when workspace-taskbar is true, the workspace labels are not rendered. Can a new option be added (like show-workspace-label) so that it will also be shown? It looks like this is the line: https://github.com/Alexays/Waybar/blob/4e76d7339f6bf0d7fd05219841962a3df66f087e/src/modules/niri/workspace.cpp#L130

PR #5013, opened 2026-04-29
fix(hyprland/workspaces): adapt dispatch commands for Lua IPC protocol

higorprado:
Hyprland 0.54 replaced the text-based dispatch socket protocol with a Lua-based one. Commands like "dispatch workspace 1" are now interpreted as invalid Lua (return hl.dispatch(workspace 1)), breaking workspace clicks and scroll navigation.

Add IPC::dispatch() that probes the running Hyprland on first call and routes commands through the new hl.dsp Lua API when the Lua protocol is detected, falling back to the old text format otherwise.

**Affected dispatchers:** `workspace`, `focusworkspaceoncurrentmonitor`, `togglespecialworkspace`

### DIFF
--- a/include/modules/hyprland/backend.hpp
+++ b/include/modules/hyprland/backend.hpp
@@ -4,6 +4,7 @@
 #include <filesystem>
 #include <list>
 #include <mutex>
+#include <optional>
 #include <string>
 #include <thread>
 #include <utility>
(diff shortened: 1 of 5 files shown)


zjeffer (approved):
~~window.cpp and windowcount.cpp will also need this fix~~

higorprado:
> window.cpp and windowcount.cpp will also need this fix

@zjeffer thanks for the review! I've checked `window.cpp` and `windowcount.cpp` thoroughly — neither file has any `getSocket1Reply("dispatch ...")` calls. They only use `getSocket1JsonReply()` for queries like `activeworkspace`, `clients`, `monitors`, and `workspaces`. Could you clarify what specifically needs fixing there? Are the query calls (e.g., `j/activeworkspace`) also affected by the Lua protocol change, or is there something else I'm missing?

zjeffer:
> > window.cpp and windowcount.cpp will also need this fix
> 
> @zjeffer thanks for the review! I've checked `window.cpp` and `windowcount.cpp` thoroughly — neither file has any `getSocket1Reply("dispatch ...")` calls. They only use `getSocket1JsonReply()` for queries like `activeworkspace`, `clients`, `monitors`, and `workspaces`. Could you clarify what specifically needs fixing there? Are the query calls (e.g., `j/activeworkspace`) also affected by the Lua protocol change, or is there something else I'm missing?

I misread, I thought they used getSocket1Reply. Nevermind.

---

I just noticed we have tests in test/hyprland/backend.cpp, can you add tests for the new backend functions (where possible)?

higorprado:
> I just noticed we have tests in test/hyprland/backend.cpp, can you add tests for the new backend functions (where possible)?

@zjeffer done — added 7 unit tests for the new backend functions. Had to move `buildLuaDispatch` to  public and `isLuaProtocol` + `s_luaProtocolDetected_` to protected so `IPCTestHelper` could access them. Let me know if anything else needs changing.

modarche (approved):

Akmadan23:
Hey @Alexays, since this fix is very important for hyprland users and addresses a breaking change, I was wondering if it would be possible to release v0.15.1, which would be really helpful. Thanks in advance

d3v1l-d0g:
> Hey @Alexays, since this fix is very important for hyprland users and addresses a breaking change, I was wondering if it would be possible to release v0.15.1, which would be really helpful. Thanks in advance

is anyone listening?

Pi2l:
Hey @Alexays, how about this?

> Hey @Alexays, since this fix is very important for hyprland users and addresses a breaking change, I was wondering if it would be possible to release v0.15.1, which would be really helpful. Thanks in advance

PR #5031, opened 2026-05-12
fix(hyprland/workspaces): move-to-monitor option breaking workspace buttons

emoryjgrubbs:
the dispatcher called when move-to-monitor is true was failing because of the `monitor = "current"` parameter


Alexays:
Closing as a duplicate — this same move-to-monitor fix landed via #5043 (already merged into master). Thanks for the contribution! 🙏

Issue #4426, opened 2025-08-31
[Bug] sway/language doesn't show the correct flag when a different keyboard is plugged in


mil-ad:
I mostly use GB layout keyboards but I have one keyboard with a US layout. Switching between them is well-supported in sway config:

```
input type:keyboard {
    xkb_layout "gb,ir"
    xkb_model "pc105"
    xkb_options "grp:rctrl_rshift_toggle"
}

input "12815:20565:RDR_IQUNIX_MG65_Mechanical_Keyboard" {
    xkb_layout "us,ir"
    xkb_model "pc104"
}
```

If I unplug my GB keyboard and plug in a US one sway correctly detects the change and applies the correct xkb_layout. However, my `sway/language` module doesn't deal with the change correctly. I have configured it as below and when I plug a keyboard with a different layout it doesn't show the correct flag anymore. It just shows an empty string. If I restart waybar it then detects the change and shows the correct flag:

```
"sway/language": {
    "format": "{flag}",
    "on-click": "swaymsg input type:keyboard xkb_switch_layout next",
    "tooltip": false
},
```

Issue #2079, opened 2023-03-31
wtype breaks the language module


pzant:
Any usage of the `wtype` breaks the sway language module, after using the wtype the module becomes transparent or something.

vactomas:
Similar issue with the Hyprland language module.

After using wtype, there is no language listed, until I manually change the layout to some other using my keyboard language change shortcut.

Muhammad95959:
```
"hyprland/language": {
    "keyboard-name": "$your-keyboard"
}
```

this option fixes the issue

pzant:
> ```
> "hyprland/language": {
>     "keyboard-name": "$your-keyboard"
> }
> ```
> 
> this option fixes the issue

I'm using waybar-sway-language, this module doesn't have this option.

Issue #1536, opened 2022-04-28
sway/language flag disappears with wtype


jficz:
I use this as a clipboard helper:
```
bindsym Print exec wtype $(cliphist list | wofi --dmenu | cliphist decode)
```
Every time `wtype` types the clipboard content, flag icon disappears.

Note that it's just the flag that disappears not the module or text in it - trying `format = "x{flag}x"` results in `xx` after `wtype` invocation. Switching layouts brings the flag back, including switching by `swaykbdd`. 

I'm pretty sure this is related to `wtype` since I use another helper
```
bindsym Ctrl+Print exec cliphist list | wofi --dmenu | cliphist decode | wl-copy
```
which runs the same thing except `wtype` and that doesn't cause any trouble.

Giraut:
I confirm that this still hasn't been fixed. I have the exact same problem.

discapes:
Same problem still with sway version 1.11, wtype-0.4, waybar v0.14.0

PR #5039, opened 2026-05-15
fix(sway/language): keep flag visible across hot-plug and unknown layouts

mil-ad:
Fixes the `sway/language` module silently clearing its `{flag}` (and `{short}` /
  `{shortDescription}` / `{variant}`) output until waybar is restarted.

## Root cause

Two cooperating bugs in `src/modules/sway/language.cpp`:

- **`set_current_layout` blanks `layout_` on an unknown name.**
  `std::map::operator[]` default-constructs a `Layout` on a missing key, so `country_flag()` returns `""`. Triggered by `wtype`'s transient virtual keyboard (#1536, #2079) and hot-plugged devices whose layouts weren't in the map at startup (#4426).
- **`layouts_map_` is built once, from one device.** `onCmd` only runs at startup and stores only the single device with the most layouts. New keyboards with disjoint layouts never make it into the map.

## Fix

1. Guard `set_current_layout` with `find()` and early-return on miss, preserving the last known good `layout_`.
2. Refresh `layouts_map_` from `onEvent` on `change == "added" || "xkb_keymap"` via `ipc_.sendCmd(IPC_GET_INPUTS)`.
3. Make `onCmd` re-runnable — union layouts across every keyboard input and `layouts_map_.clear()` before each rebuild to avoid compounding the duplicate-suffix pass (`"us"` → `"us1"` → `"us11"`).

`sendCmd` is synchronous and re-emits `signal_cmd` on the calling thread, so the refresh call is made after releasing `mutex_` to avoid a self-deadlock when `onCmd` re-locks it.

### DIFF
--- a/src/modules/sway/language.cpp
+++ b/src/modules/sway/language.cpp
@@ -54,21 +54,28 @@ void Language::onCmd(const struct Ipc::ipc_response& res) {
     std::lock_guard<std::mutex> lock(mutex_);
     auto payload = parser_.parse(res.payload);
     std::vector<std::string> used_layouts;
-    // Display current layout of a device with a maximum count of layouts, expecting that all will
-    // be OK
+    // Union layout names across every keyboard input so hot-plugged devices contribute their
+    // layouts to the map. Track the device with the most layouts to seed the initially displayed
+    // layout, matching the previous behaviour at startup.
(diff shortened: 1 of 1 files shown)


mil-ad:
@Alexays curious what you think about this

PR #5043, opened 2026-05-16
fix(hyprland/workspaces): fix IPC call when using move-to-monitor

tomben13:
#5013 failed to fix the issue when `move-to-monitor` is set to true.
This PR fixes this case.


niklassheth:
Just ran into this problem, thanks!

PR #5046, opened 2026-05-16
fix: mpd specify timer interval in seconds

### LINKED ISSUE #4870: Waybar cannot detect mpd if mpd is launched after waybar
System: arch linux hyprland

Issue: if I launch waybar first, then launch mpd (using systemctl --user), waybar does not detect mpd. But if mpd is already running when waybar is launched, everything works fine until I restart mpd.

Things I tried:
1. Changing interval: didn't help
2. Launched waybar with default config and default style: issue persists.
3.  Reverted to previous version of waybar (0.14.0-6): issue goes away.

Thus, this seems to be a bug in the latest release.

### LINKED ISSUE #5025: The mpd module is not shown
If the mpd demon is not working before waybar starts, then the module completely disappears and does not show format-disconnected. Just nothing.

awused:
Fixes #4870
Fixed #5025

Tested locally. This is only used in one place for the disconnected timer, where arm_timer expects seconds.

### DIFF
--- a/include/modules/mpd/state.inl.hpp
+++ b/include/modules/mpd/state.inl.hpp
@@ -1,13 +1,14 @@
 #pragma once
 
 namespace detail {
+using namespace std::literals::chrono_literals;
 
 inline bool Context::is_connected() const { return mpd_module_->connection_ != nullptr; }
 inline bool Context::is_playing() const { return mpd_module_->playing(); }
 inline bool Context::is_paused() const { return mpd_module_->paused(); }
 inline bool Context::is_stopped() const { return mpd_module_->stopped(); }
 
-constexpr inline std::size_t Context::interval() const { return mpd_module_->interval_.count(); }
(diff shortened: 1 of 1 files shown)


cpuspellcaster:
I ran into this issue myself, and patched it on my local nixos computer using the change in this PR, confirmed this resolves the issue where the reconnect timer ends up 1000× longer than the configured interval (e.g. interval: 5 in the config results in a 5000-second reconnect delay rather than 5 seconds).

One issue: 1s requires using namespace `std::chrono_literals` to be in scope, which isn't the case in `state.inl.hpp`, so this fails to compile as-is. The following achieves the same result without that dependency:

```cpp
constexpr inline std::size_t Context::interval() const { return std::chrono::duration_cast<std::chrono::seconds>(mpd_module_->interval_).count(); }
```

awused:
That's odd, it compiles for me and doesn't emit any warnings. It should be in scope when it's included, as far as I can tell, do you have different build flags set?

PR #5058, opened 2026-05-20
fix(amodule): prevent crash when cursor timeout fires after destruction

### LINKED ISSUE #5055: Crash on monitor off with cursor: true in drawer group custom modules
Hello, Waybar segfaults every time my monitor turns off. 
Took me a while to narrow it down but I got a minimal repro.

Version: 0.15.0
OS: Arch
Compositor: MangoWM

What happens:
When the output is disabled (via DPMS or wlr-randr --off), waybar crashes with a SIGSEGV.
It ~~only~~ happens if I have a drawer group containing multiple modules that have cursor: true set. 
_Or it has a_ `on-click` _parameter._

Minimal repro config:
```json
{
  "layer": "top",
  "position": "top",
  "modules-right": ["group/test"],
  "custom/a": { "format": "a", "cursor": true },
  "custom/b": { "format": "b", "cursor": true },
  "group/test": {
    "orientation": "horizontal",
    "drawer": {
      "click-to-reveal": true,
      "transition-duration": 0
    },
    "modules": ["custom/a", "custom/b"]
  }
}
```

What I checked:
<strike>
- Single custom module with cursor: true in a drawer → fine
- Multiple custom modules in a drawer but no cursor: true → fine
- Multiple custom modules with cursor: true in a drawer → crash
- Putting the same modules directly in modules-right without the group wrapper → fine
So the crash is specifically triggered by cursor: true on custom modules inside a drawer group.
</strike>

Edit: actually `on-click `also make it crashe, check https://github.com/Alexays/Waybar/pull/5058 for more details.

typedrifter:
Hey,
Waybar segfaults when my monitor turns off as stated in #5055

I narrowed it down to AModule::setCursor():
When the window isn't ready, setCursor() schedules a 1-second retry. If the output gets removed and the module is destroyed before that timeout fires it crashes.

The fix stores the timeout connection and disconnects it in the destructor, kinda same as what privacy and idle_inhibitor modules already do.

Tested with my config that was crashing 100% after the monitor was shutdown and turned back on, and the crashes have stopped.

Fixes #5055

### DIFF
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -57,6 +57,7 @@ class AModule : public IModule {
   bool hasUserEvents_;
   gdouble distance_scrolled_y_;
   gdouble distance_scrolled_x_;
+  sigc::connection cursor_timeout_conn_;
   std::map<std::string, std::string> eventActionMap_;
   static const inline std::map<std::pair<uint, GdkEventType>, std::string> eventMap_{
       {std::make_pair(1, GdkEventType::GDK_BUTTON_PRESS), "on-click"},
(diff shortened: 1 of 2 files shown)


typedrifter:
Hey sorry to ping you @Alexays, I know there is a lot of activity and probably more urgent issues to fix, but it would be great if this fix (or maybe a better one ?) could be merged :)

Issue #4783, opened 2026-01-19
Question: can taskbar items be evenly scaled like tint2 to fit within the bar?


linux36524:
When too many applications are opened, the taskbar keeps expanding and eventually pushes all right-side modules off the screen.

I have two related questions:

1，Is there any way to constrain or limit the total width of the taskbar so that it never exceeds the bar size?

2，Is it possible to make the taskbar auto-scale its items, similar to tint2 behavior — for example:

when there is only one task, it expands to fill the available taskbar space;

as more tasks are opened, each task button shrinks proportionally,

while keeping the taskbar within a fixed, non-overflowing area?

Any guidance or pointers would be greatly appreciated.

linux36524:
The same issue is mentioned here:https://github.com/TheFrankyDoll/win10-style-waybar

PR #5060, opened 2026-05-20
fix(wlr/taskbar): let task titles shrink within available space

EzequielColungaAcha:
## Summary
- Allow wlr/taskbar buttons to share the module allocation instead of each title forcing its full requested width.
- Ellipsize title labels on a single line so taskbar items shrink when many windows are open.
- Keeps neighboring modules visible while preserving existing taskbar click actions and format handling.

This addresses the overflow behavior discussed in #4783.

## Test plan
- Verified the PR diff is limited to `src/modules/wlr/taskbar.cpp`.
- Built and tested locally through the Arch package workflow before preparing the PR branch.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/man/waybar-wlr-taskbar.5.scd
+++ b/man/waybar-wlr-taskbar.5.scd
@@ -57,6 +57,11 @@ Addressed by *wlr/taskbar*
 	default: false ++
 	If set to true, group tasks by their app_id. Cannot be used with 'active-first'.
 
+*expand*: ++
+	typeof: bool ++
+	default: false ++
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks! The shrink behavior is useful, but as-is it changes the default for everyone with no opt-out, and it doesn't hold up on **vertical bars**: `content_`/`box_` follow `bar.orientation`, so `pack_start(bt, true, true)` stretches the buttons along a vertical bar, and `ellipsize END` + `width_chars(1)` truncates titles even when there's room. Could you gate this behind a config option (e.g. `shrink`/`expand`) and skip/adapt the ellipsize logic for vertical orientation? 🙏

PR #5067, opened 2026-05-23
Add native Triad IPC modules

greenm01:
Adds native Triad modules for Waybar.

Triad is a programmable Wayland window manager that runs on the River protocol. It has a small native IPC surface for shells and bars. This PR lets Waybar talk to that API directly, the same way it does for other compositor-specific modules, instead of depending on a compatibility layer whose shape is meant for another window manager.

The shared Waybar changes are small: a Meson option, module factory wiring, and manual page links. The Triad-specific code lives under its own module namespace and is only built when Triad support is enabled.

This adds:

- `triad/workspaces`
- `triad/window`
- `triad/language`

The backend connects to `$TRIAD_SOCKET`, or falls back to `${XDG_RUNTIME_DIR}/triad.sock`, reads the initial compositor state, and then follows Triad's native event stream for workspace, window, layout, and keyboard updates.

Tested on Triad with three monitors. Workspace switching, active and urgent workspace state, focused window updates, and keyboard layout updates work here.

Build and test commands run:

```sh
meson setup build -Dtriad=true -Dniri=true -Dtests=enabled
ninja -C build
meson test -C build --verbose --print-errorlogs --suite waybar
```

Triad: https://github.com/greenm01/triad

### DIFF
--- a/include/modules/triad/backend.hpp
+++ b/include/modules/triad/backend.hpp
@@ -1,0 +1,63 @@
+#pragma once
+
+#include <list>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "util/json.hpp"
+
+namespace waybar::modules::triad {
+
+class EventHandler {
+ public:
+  virtual void onEvent(const Json::Value& ev) = 0;
+  virtual ~EventHandler() = default;
+};
+
+class IPC {
+ public:
+  IPC();
+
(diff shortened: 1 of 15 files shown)


greenm01:
Hello, I'm going to close this and simply implement ext-workspace-v1 for my window manager, using a little c lib i'm working on:

https://github.com/greenm01/libext-workspace

Apologies for the noise!

PR #5068, opened 2026-05-24
Sway IPC client: robustness improvements & code cleanups

c4rlo:
Update Sway IPC client to make I/O logic and socket path discovery more robust.

- Send IPC headers and payloads with retry loops so partial `send()` calls are handled correctly.
- Handle interrupted/nonblocking `recv()` calls consistently.
- Replace unaligned `reinterpret_cast<uint32_t*>` IPC header access with `memcpy`.
- Check `popen()`/`pclose()` failures when resolving the Sway socket path, and ignore empty `SWAYSOCK` env var.
- Misc code cleanups, e.g. use designated initializers for clarity, tweak how/where we use `auto` for readability, etc.


include/modules/sway/ipc/client.hpp
```diff
@@ -41,8 +37,9 @@ class Ipc {
   static inline const std::string ipc_magic_ = "i3-ipc";
   static inline const size_t ipc_header_size_ = ipc_magic_.size() + 8;
 
-  const std::string getSocketPath() const;
-  int open(const std::string&) const;
+  static std::string getSocketPath();
+  static int open(const std::string&);
+
```
c4rlo, line 42: Made these `static` since they're not accessing any instance methods/fields.

src/modules/sway/ipc/client.cpp
```diff
@@ -29,49 +62,49 @@ Ipc::~Ipc() {
...
   std::string str;
-  {
-    std::string str_buf;
-    FILE* in;
-    char buf[512] = {0};
-    if ((in = popen("sway --get-socketpath 2>/dev/null", "r")) == nullptr) {
-      throw std::runtime_error("Failed to get socket path");
-    }
-    while (fgets(buf, sizeof(buf), in) != nullptr) {
-      str_buf.append(buf, sizeof(buf));
```
c4rlo, line 48: This was appending the entire buffer, even beyond the NUL byte, potentially writing garbage into `str_buf`. In practice it probably didn't matter because the socket path was then treated like a C-string, thereby ignoring anything beyond the embedded NUL.

PR #5073, opened 2026-05-26
[feat] `sway/workspaces`: refer to outputs by their monitor identifier in `persistent-workspaces`

ar-turek:
This PR extends the `persistent-workspaces` functionality in `sway/workspaces` with an option of pinning workspaces to displays by their output identifiers (basically, the monitor's "make and model", usually given as a vendor/model/serial triple) in addition to their output name (e.g. `DP-3`, which is unstable and can change between reboots).

That means instead of this:

```json
{
    ...
    "sway/workspaces": {
        ...
        "persistent-workspaces": {
            "1:main": ["DP-2"],
            "2:code": ["DP-2"],
            "3:secondary": ["DP-3"],
            "0:TV": ["HDMI-A-1"]
        }
    },
    ...
}
```

we can now write this:

```json
{
    ...
    "sway/workspaces": {
        ...
        "persistent-workspaces": {
            "1:main": ["Dell Inc. Dell AW2518H #ASM0cCiovard"],
            "2:code": ["Dell Inc. Dell AW2518H #ASM0cCiovard"],
            "3:secondary": ["AOC 24P4CV XL9RAHA000298"],
            "0:TV": ["Philips Consumer Electronics Company Philips FTV 0x01010101"]
        }
    },
    ...
}
```

and even if the outputs are renamed after a reboot the workspaces will still be assigned to correct displays.

### DIFF
--- a/man/waybar-sway-workspaces.5.scd
+++ b/man/waybar-sway-workspaces.5.scd
@@ -147,6 +147,7 @@ an empty list denoting all outputs.
 		"3": [], // Always show a workspace with name '3', on all outputs if it does not exist
 		"4": ["eDP-1"], // Always show a workspace with name '4', on output 'eDP-1' if it does not exist
 		"5": ["eDP-1", "DP-2"] // Always show a workspace with name '5', on outputs 'eDP-1' and 'DP-2' if it does not exist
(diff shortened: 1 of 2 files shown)


ar-turek:
Please let me know if you'd like me to make an update to the wiki as well, if/when this is merged - I'm not sure if it's possible to stage wiki changes as PRs as well.

ar-turek:
I've also noticed the potential effect this PR has on this other one: https://github.com/Alexays/Waybar/pull/5072 - I can facilitate integration of those two, if needed.

Issue #3693, opened 2024-10-16
(hyprland/languafe) [Feature request] Tooltip on the module


vpcano:
It would be great to have a tooltip that displays info about the current keyboard layout. The same format replacements used for the `format`  config option could be used to configure what the tooltip shoud show.

For example:
```js
"hyprland/language": {
	"format": "Layout: {short}",
	"tooltip" true,
	"tooltip-format": "Layout: {long}"
}
```

vpcano:
Started working on this here: #3698

guttermonk:
Curious if there's anything left to do here. Would be great to cross this off my list of bugs. If there's anything I can do to help, please let me know.

guttermonk:
I created a $50 bounty for this bug that is claimable after the PR that fixes it is merged. I'll also send a $25 donation to the project once the PR is merged.

<a href="https://www.boss.dev/issue/github/I_kwDOCJZbAc6aiU2Y"><img src="https://img.shields.io/endpoint.svg?url=https://api.boss.dev/badge/bounty%3Fgid%3Dboss%2Frvs9u1r6mugi79zfdc7lwp" /></a>

rishi-jat:
Hey @guttermonk @vpcano , I would like to work on this issue is this bounty still open to work ? Thanks!

guttermonk:
@rishi-jat sorry I already created a pull request that fixes the issue for me (see above).

yihangwu539-png:
I'll work on this bounty. Claiming it. 👨‍💻

guttermonk:
@yihangwu539-png I already created a pull request that fixes the issue, so I do not plan on paying out this bounty.

yihangwu539-png:
I'll work on this bounty. Claiming it. 👨‍💻

Taitai667:
I can work on this.

lazyGPT07:
Submitted an implementation PR: https://github.com/Alexays/Waybar/pull/5082 adds tooltip-format support for hyprland/language, with fallback to the visible label and man page documentation. Payout wallet if this helps close the bounty or saves review time: 0xB167296811544B429D17990E687548F7d552008C

18166714330cl-maker:
I can take a quick look at this.

Plan: inspect the module tooltip/language handling path first, then decide whether this can be solved as a small static code change without requiring a full Hyprland reproduction environment.

ptgram4220:
[...]

PR #5081, opened 2026-05-29
Add tooltip-format for Hyprland language module

eutv2022:
Closes #3693.

This adds `tooltip-format` support to `hyprland/language`, matching the existing `sway/language` behavior:

- reads `tooltip-format` from config
- formats the tooltip with `{short}`, `{shortDescription}`, `{long}`, and `{variant}`
- falls back to the displayed layout text when `tooltip-format` is not set
- documents the option in `waybar-hyprland-language(5)`

Verification:

- `git diff --check`

I could not run the full Meson build in this Windows environment because `meson` is not installed here.


EmteZogaf:
Duplicate of #5012

Alexays:

PR #5083, opened 2026-05-30
Add tooltip format for Hyprland language

TeedsK:
Fixes #3693

Adds `tooltip-format` support to `hyprland/language`, using the same replacement fields as `format`.

If `tooltip-format` is not configured, the tooltip falls back to the formatted label text.

Updated the hyprland language manpage with the new option and example.

Validation:
- git diff --check


EmteZogaf:
Duplicate of #5012

Alexays:

PR #5085, opened 2026-05-30
Add tooltip format for Hyprland language

xtakenotesx16:
Summary

- Add `tooltip-format` support to `hyprland/language`
- Format tooltip text with the same layout placeholders used by `format`
- Document the new option in the Hyprland language man page

 Problem

`hyprland/language` can format the visible label with `{long}`, `{short}`, `{shortDescription}`, and `{variant}`, but there is no separate tooltip format. This addresses #3693.

 Fix

The module now reads `tooltip-format` from config and applies it on update when tooltips are enabled. If `tooltip-format` is not configured, the tooltip falls back to the displayed label text, matching the behavior used by similar language modules.

Verification:
- Ran `git diff --check`.
- Full build was not run locally because the workspace does not have Meson/Ninja/C++ build dependencies.
- Runtime Hyprland check was not run locally because the workspace does not have a Hyprland session.

Risk

Low. The change is limited to `hyprland/language` tooltip rendering and documentation. Existing label formatting is unchanged.

### DIFF
--- a/include/modules/hyprland/language.hpp
+++ b/include/modules/hyprland/language.hpp
@@ -37,6 +37,7 @@ class Language : public waybar::ALabel, public EventHandler {
   util::JsonParser parser_;
 
   Layout layout_;
+  std::string tooltip_format_ = "";
 
   IPC& m_ipc;
 };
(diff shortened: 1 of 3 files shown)


EmteZogaf:
Duplicate of #5012

Alexays:

PR #5092, opened 2026-06-02
hyprland/language: add tooltip-format support

Chillzyg:
Adds tooltip support for the `hyprland/language` module so users can configure a tooltip with the same layout replacement fields used by the module format.

Example:

```json
"hyprland/language": {
  "format": "Layout: {short}",
  "tooltip": true,
  "tooltip-format": "Layout: {long}"
}
```

The module keeps the existing behavior when `tooltip-format` is not configured by using the visible label text as the tooltip.

Verification:
- `git diff --check`

I could not run the full Waybar Meson build or clang-format locally because `meson` and `clang-format` are not installed in this workspace.

Closes #3693.

### DIFF
--- a/include/modules/hyprland/language.hpp
+++ b/include/modules/hyprland/language.hpp
@@ -37,6 +37,7 @@ class Language : public waybar::ALabel, public EventHandler {
   util::JsonParser parser_;
 
   Layout layout_;
+  std::string tooltip_format_;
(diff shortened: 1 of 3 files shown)


EmteZogaf:
Duplicate of #5012

Alexays:

PR #5100, opened 2026-06-04
perf(mpd): avoid duplicate playing-state updates

SaveTheRbtz:
## Summary

- Remove the unused `queryMPD()` wrapper from the MPD state path.
- Let the playing timer reuse its existing `fetchState()` result and emit one update.

## Why

`Playing::on_timer()` already fetches MPD state before checking whether playback is still active. Calling `queryMPD()` immediately afterward fetched the same state again and emitted an extra update on every playing tick.

## Testing

- `ninja -C build`
- `meson test -C build`


SaveTheRbtz:
cc @awused @cpuspellcaster, you both have recent context on the MPD timer/state path and may be good reviewers here.

PR #5101, opened 2026-06-04
fix(mpd): avoid false no-song warning when stopped

SaveTheRbtz:
## Summary

- Only warn about a missing current song when MPD is not stopped.
- Keep the existing stopped-format fallback for missing song data.

## Why

MPD can legitimately report no current song while stopped. The old condition logged `Bug in mpd` for any missing song, including the normal stopped state, which made routine MPD idle/stopped states look like bugs.

## Testing

- `ninja -C build`
- `meson test -C build`

### DIFF
--- a/src/modules/mpd/mpd.cpp
+++ b/src/modules/mpd/mpd.cpp
@@ -124,8 +124,9 @@ void waybar::modules::MPD::setLabel() {
 
   std::string stateIcon = "";
   bool no_song = song_.get() == nullptr;
-  if (stopped() || no_song) {
-    if (no_song) spdlog::warn("Bug in mpd: no current song but state is not stopped.");
+  bool is_stopped = stopped();
+  if (is_stopped || no_song) {
+    if (no_song && !is_stopped) spdlog::warn("mpd: no current song while state is not stopped");
     format =
         config_["format-stopped"].isString() ? config_["format-stopped"].asString() : "stopped";
     label_.get_style_context()->add_class("stopped");


SaveTheRbtz:
cc @awused @2e0byo, you both have relevant MPD state/no-song context and may be good reviewers here.

2e0byo:
I suspect my old hack just had a bug in it, since the log line wasn't gated on checking it was actually stopped.

LGTM (not that I have any role here).

src/modules/mpd/mpd.cpp
```diff
@@ -125,7 +125,7 @@ void waybar::modules::MPD::setLabel() {
   std::string stateIcon = "";
   bool no_song = song_.get() == nullptr;
   if (stopped() || no_song) {
-    if (no_song) spdlog::warn("Bug in mpd: no current song but state is not stopped.");
+    if (no_song && !stopped()) spdlog::warn("mpd: no current song while state is not stopped");
```
2e0byo: nit: there's a race condition here, so it's worth assigning stopped to a var above and then using that I think.

2e0byo (changes requested):
LGTM: my original log line should have been gated behind such a check. 

(I'm just a random stranger and have no role here, just reviewing as I wrote the previous patch.)

src/modules/mpd/mpd.cpp
SaveTheRbtz: good point!

(that said, there is still race between setting these vars in the first place, but at least if's will be internally consistent)

PR #5106, opened 2026-06-05
Add tooltip-format to hyprland language module

Chenhao-XU:
Fixes #3693

This adds tooltip formatting support to `hyprland/language`, matching the behavior users already have in similar language modules.

Changes:

- read optional `tooltip-format` from the module config
- format the tooltip with `{long}`, `{short}`, `{shortDescription}`, and `{variant}`
- fall back to the displayed layout text when no custom tooltip format is configured
- document `tooltip-format` and `tooltip` in `waybar-hyprland-language(5)`

Validation:

- `git diff --check`

Notes:

- I could not run a full local build in my Windows environment because Meson and the Linux/GTK/Wayland dependency stack are unavailable here.

### DIFF
--- a/include/modules/hyprland/language.hpp
+++ b/include/modules/hyprland/language.hpp
@@ -37,6 +37,7 @@ class Language : public waybar::ALabel, public EventHandler {
   util::JsonParser parser_;
 
   Layout layout_;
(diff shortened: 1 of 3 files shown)


Alexays:

PR #5107, opened 2026-06-06
Add tooltip-format support to hyprland language

delitzer:
Fixes #3693.

Adds tooltip-format support to the hyprland/language module using the same layout replacements already available to format:

- {short}
- {shortDescription}
- {long}
- {variant}

Also falls back to the displayed layout text when tooltip-format is not configured, matching the sway/language behavior, and documents tooltip plus tooltip-format in the manpage.

Verification:
- git diff --check
- Compared the three committed fork files against the local patched files

Not run:
- meson build/test; meson is not installed on this macOS host
- clang-format; clang-format is not installed on this macOS host


Alexays:

PR #5108, opened 2026-06-06
perf(json): parse directly from string buffers

SaveTheRbtz:
## Summary

- Parse `JsonParser` input with JsonCpp `CharReader` directly from the existing string buffer.
- Keep `\x` escape compatibility, but only allocate and run `regex_replace` when that escape is present.

## Why this was needed

`baseline.perf.data` showed the JSON utility paying a large common-case regex tax before parsing:

- `std::regex` search/replace: about **1.10B child cycles**
- `std::regex::_M_dfs`: **743M self cycles**
- Json parse path: **834M child cycles**

## Change

Parse from the existing string buffer with JsonCpp's `CharReader`. Preserve the `\x` compatibility repair, but first check whether the escape is present so the common path avoids allocating a modified string and running regex replacement.

## Effect in comparison

The regex path disappears from visible hot symbols. Json parse work drops from **834M** to **188M** cycles, and the overall JsonCpp bucket drops from roughly **700M** to **186M** cycles.

## Testing

- `ninja -C build`
- `meson test -C build`

### DIFF
--- a/include/util/json.hpp
+++ b/include/util/json.hpp
@@ -7,6 +7,7 @@
 #include <codecvt>
 #include <iostream>
 #include <locale>
+#include <memory>
 #include <regex>
 
 #if (FMT_VERSION >= 90000)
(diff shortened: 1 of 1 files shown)


SaveTheRbtz:
cc @zjeffer @xander1421 @Alexays, you look like the most relevant reviewers from git history:

- @zjeffer added the `\x` -> `\u00` compatibility path and JsonParser tests in #2496.
- @xander1421 recently changed `JsonParser` to use a local `CharReaderBuilder` for thread safety in #4789.
- @Alexays reviewed/merged both related parser changes and has the original parser/module history.

PR #5109, opened 2026-06-06
feat(hyprland): add language tooltip format

sato820:
Closes #3693

This adds tooltip support to `hyprland/language`.

- supports `tooltip-format`
- uses the same replacements as `format`: `{long}`, `{short}`, `{shortDescription}`, `{variant}`
- falls back to the visible layout text when `tooltip-format` is not set
- documents `tooltip` and `tooltip-format` in `waybar-hyprland-language(5)`

I could not compile Waybar in this Windows environment because no C++ compiler is installed here, but the implementation follows the existing `sway/language` tooltip pattern and the patch is limited to the Hyprland language module plus documentation.


Alexays:

PR #5125, opened 2026-06-13
feat(wlr/taskbar): add active-only option

iamcheyan:
## Description

Add a new `active-only` configuration option to the `wlr/taskbar` module that, when set to `true`, only shows the currently active application button in the taskbar.

Other applications remain tracked internally and automatically reappear when they become active. This provides a cleaner UI for users who only want to see the focused window in their taskbar.

## Motivation

Some users prefer a minimalist taskbar that only shows the currently focused application. This feature allows hiding inactive windows while maintaining full tracking state, so windows seamlessly reappear when activated.

## Changes

- Add `visible()` getter to `Task` class for external visibility queries
- Modify `handle_output_enter()` to respect `active-only` setting when showing buttons
- Modify `handle_done()` to show/hide buttons based on active state when `active-only` is enabled
- Update man page documentation

## Configuration

```jsonc
"wlr/taskbar": {
    "active-only": true
}
```

## Testing

- Built with `meson setup -Dman-pages=enabled build`
- All existing tests pass
- Runtime tested on labwc compositor:
  - With `active-only: true`, only the focused window button is visible
  - Switching focus hides the previous window and shows the new one
  - Closing the active window shows the previously focused window
  - With `active-only: false` (default), behavior is unchanged

Signed-off-by: iamcheyan <iamcheyan@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ result-*
 .ccls-cache
 _codeql_detected_source_root
 heaptrack*
+PR_NOTES_WLR_TASKBAR_BAR_CSS_STATES.md
--- a/include/modules/wlr/taskbar.hpp
+++ b/include/modules/wlr/taskbar.hpp
@@ -102,6 +102,7 @@ class Task {
   bool minimized() const { return state_ & MINIMIZED; }
   bool active() const { return state_ & ACTIVE; }
   bool fullscreen() const { return state_ & FULLSCREEN; }
+  bool visible() const { return button_visible_; }
 
  public:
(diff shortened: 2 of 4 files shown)


Alexays:
Closing in favour of #5126 (same author): that PR already contains this `active-only` behaviour and additionally integrates ext-workspace-v1 to expose maximized/fullscreen state as CSS classes — it's a superset. Consolidating the wlr/taskbar work on #5126. Thanks! 🙏

Issue #3955, opened 2025-02-24
waybar starts automatically with KDE Plasma


K0-RR:
v0.9.24. It's not listed in the startup apps in the KDE system settings. I also tried:
systemctl --user list-units | grep waybar
cat ~/.profile |grep waybar
ls /etc/xdg/autostart/ |grep waybar
ls /usr/share/applications/ |grep waybar

blitzgneisserin:
I can confirm this issue. If waybar is used in any wayland compositor, it also start in different wayland desktop environments. The only workaround I found is creating a startup script that kills waybar. Maybe its not a waybar but a wayland issue.

kyoshiro:
Got a finding on my machine by:
`↳ systemctl --user list-units | grep waybar`

disabled it:
`↳ systemctl --user disable waybar.service`

steveo314:
Waybar does in fact start when you log into Plasma(Wayland). I'm on Debian Sid logged into Plasma(Wayland) looking at waybar how I have it configured for hyprland. I did not configure waybar to open with Plasma.

KD0BPV:
@steveo314 disabling it via systemd will only make it so systemd doesn't automatically start it. As long as you have hyprland configured to start Waybar via an `exec-once` statement in your config file, you should be fine, as I understand it.

Akirapearl:
Hi, I also have this issue with waybar on Debian testing, running KDE and Niri, even after disabling it with `sudo systemctl --user stop waybar.service && sudo systemctl --user disable waybar.service` i tried without the --user flag but it returns a unknown service waybar, also I did run this command during Niri's [initial config](https://github.com/YaLTeR/niri/wiki/Example-systemd-Setup) `systemctl --user add-wants niri.service waybar.service`

tomsen-san:
In my case the waybar.service was in global scope. To disable I had to 
`systemctl --global disable waybar.service`

Vladimir-csp:
`ExecCondition=/lib/systemd/systemd-xdg-autostart-condition "" "KDE"`

Alexays:
Closing — this was resolved/answered in the thread (config, environment, or upstream — not a Waybar bug still open). Feel free to reopen if you're still stuck. 🙏

PR #5135, opened 2026-06-17
waybar.service: Use sway-session.target instead of graphical-session.target

leigh123linux:
graphical-session.target is now activated by multiple desktop environments
including GNOME and Cinnamon. waybar only functions under wlroots-based
Wayland compositors so WantedBy=graphical-session.target causes it to be
started unconditionally in any graphical session.

Use WantedBy=sway-session.target to scope waybar to Sway sessions only.
Users of other wlroots compositors can enable it manually against their
own session target.

### DIFF
--- a/resources/waybar.service.in
+++ b/resources/waybar.service.in
@@ -11,4 +11,4 @@ ExecReload=kill -SIGUSR2 $MAINPID
 Restart=on-failure
 
 [Install]
-WantedBy=graphical-session.target
+WantedBy=sway-session.target


leigh123linux:
@alebastr

RobertMueller2:
should solve issue #3955

Alexays:
Thanks! We can't take this as-is: switching `WantedBy` to `sway-session.target` makes the unit Sway-specific, but Waybar targets "Sway and Wlroots based compositors" — river/Hyprland/other wlroots setups rely on `graphical-session.target`, so this would break autostart for them. The `[Unit]` block also still references `graphical-session.target` (PartOf/After), making the result inconsistent. If you need a Sway-specific target, that's better done in a Sway-specific drop-in on your side rather than the shipped unit. 🙏

leigh123linux:
It targets everything including places it isn't welcomed eg: gnome, cinnamon and kde
You need to make this your number one priority.

Issue #2882, opened 2024-02-03
WirePlumber integration doesn't handle PipeWire/WirePlumber restart


misuzu:
When I restart PipeWire (`journalctl --user -u pipewire.service`) the WirePlumber integration stops working without logging anything.

It also says `segmentation fault` on ^C:
```
^C[2024-02-03 10:11:42.019] [info] Quitting.
^C[1]    10856 segmentation fault (core dumped) 
```

Here's my waybar config:
```json
{
    "layer": "top",
    "position": "top",
    "height": 24,

    "modules-left": ["hyprland/workspaces"],
    "modules-center": ["hyprland/window"],
    "modules-right": ["network", "memory", "wireplumber", "battery", "clock", "tray"],
    "clock": {
        "format": "{:%a %b %d %T}",
        "interval": 1,
        "tooltip": false
    },
    "memory": {
        "interval": 30,
        "format": "RAM: {used:0.2f} / {total:0.2f} GiB",
        "tooltip": false
    },
    "wireplumber": {
        "format": "VOL: {volume}% {icon}",
        "format-bluetooth": "{volume}% {icon}",
        "scroll-step": 2,
        "on-click": "pavucontrol",
        "tooltip": false
    },
    "network": {
        "interface": "wlp1s0",
        "format": "D: {bandwidthDownBytes}, U: {bandwidthUpBytes}",
        "interval": 1
    },
    "battery": {
        "format": "BAT: {capacity}%",
    },
    "tray": {
        "icon-size": 16
    }
}
```

```shellsession
% waybar -v    
Waybar v0.9.24
```

Fxzzi:
Is this also possibly related to my issue which is that the wireplumber module sometimes doesn't work when I first login to hyprland? I have to restart waybar for the module to show properly (otherwise it's stuck on 0%)

RamyFerhat59:
I have the exact same problem on described by @Fxzzi

megalinee:
I have the same issue. I autologin to arch and start Hyprland on a launch script within getty so I'm assuming the quick start isn't giving enough time for wireplumber to start.

rtgiskard:
+1

megalinee:
If you change the pipewire default device after starting waybar it fixes the issue. I just put `pactl set-default-sink <default-device>` to execute after waybar automatically and the issue is gone. To get the device name you can use the command `pactl list`

Myned:
[...]

PR #5168, opened 2026-07-04
fix(wireplumber): reconnect when PipeWire/WirePlumber restarts (#2882)

Alexays:
**Please review carefully — this refactors the wireplumber module's connection lifecycle, and I couldn't runtime-test it.**

**wireplumber: no reconnect when PipeWire/WirePlumber restarts (#2882).** The module's connection was one-shot: the constructor called `wp_core_connect()` once, connected only to the object manager's `installed` signal, and had no `WpCore` `disconnected` handler or connect-failure recovery. So when PipeWire/WirePlumber restarted or crashed, the core silently disconnected, `om_`/`mixer_api_`/`def_nodes_api_` went dead, and the module stayed stale/blank until Waybar was restarted.

This splits the one-shot init into a reusable `setupConnection()` / `teardownConnection()` pair (used by both the constructor and reconnect), connects the `WpCore` `disconnected` signal, and adds a bounded `Glib::signal_timeout` (2s, no busy loop) that tears down the dead connection and rebuilds it, retrying until PipeWire is back and re-arming on a later disconnect. It preserves the recent async-callback use-after-free guard (`isModuleAlive()`/registry) and cancels the reconnect timer + tears down cleanly in the destructor; it also clears the cached default node-name out-params on teardown to avoid a per-reconnect leak. `handleScroll` already no-ops while `mixer_api_` is null, so the reconnect window is safe.

### DIFF
--- a/src/modules/wireplumber.cpp
+++ b/src/modules/wireplumber.cpp
@@ -3,6 +3,7 @@
 #include <spdlog/spdlog.h>
 
 #include <cmath>
+#include <memory>
 #include <string>
 #include <unordered_set>
 
@@ -25,6 +26,17 @@ static constexpr unsigned kReconnectIntervalMs = 2000;
 bool waybar::modules::Wireplumber::isModuleAlive(waybar::modules::Wireplumber* self) {
   return std::find(modules.begin(), modules.end(), self) != modules.end();
 }
+
+namespace {
+// user_data for the async load/activate callbacks. Pairs the module with the connection generation
+// the call was scheduled under so a completion belonging to a torn-down connection can be dropped
+// (see Wireplumber::connection_generation_ and #2882). Heap-allocated per call; the callback takes
+// ownership and frees it.
+struct AsyncCall {
+  waybar::modules::Wireplumber* self;
+  uint32_t generation;
+};
+}  // namespace
 
 waybar::modules::Wireplumber::Wireplumber(const std::string& id, const Json::Value& config)
     : ALabel(config, "wireplumber", id, "{volume}%"),
@@ -56,6 +68,12 @@ waybar::modules::Wireplumber::Wireplumber(const std::string& id, const Json::Val
                                                    : "Audio/Sink");
   only_physical_ = config_["only-physical"].isBool() ? config_["only-physical"].asBool() : false;
 
+  // Wire the scroll handler once, here, rather than in onMixerApiLoaded: the latter now re-runs on
+  // every reconnect and would accumulate duplicate handlers. handleScroll no-ops while mixer_api_
+  // is null, so wiring it before the first successful connect is safe.
+  event_box_.add_events(Gdk::SCROLL_MASK | Gdk::SMOOTH_SCROLL_MASK);
+  event_box_.signal_scroll_event().connect(sigc::mem_fun(*this, &Wireplumber::handleScroll));
+
   if (!setupConnection()) {
     spdlog::error("[{}]: Could not connect to PipeWire: '{}'", name_, type_);
     throw std::runtime_error("Could not connect to PipeWire\n");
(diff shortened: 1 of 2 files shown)


src/modules/wireplumber.cpp
```diff
@@ -46,32 +51,51 @@ waybar::modules::Wireplumber::Wireplumber(const std::string& id, const Json::Val
...
 
   type_ = g_strdup(config_["node-type"].isString() ? config_["node-type"].asString().c_str()
                                                    : "Audio/Sink");
   only_physical_ = config_["only-physical"].isBool() ? config_["only-physical"].asBool() : false;
 
+  if (!setupConnection()) {
+    spdlog::error("[{}]: Could not connect to PipeWire: '{}'", name_, type_);
+    throw std::runtime_error("Could not connect to PipeWire\n");
+  }
+}
```
khaneliman, line 81: suggestion (non-blocking): `onMixerApiLoaded` (unchanged by this diff, but now reachable repeatedly) ends with:

https://github.com/Alexays/Waybar/blob/0b88b4ef0c5c7e1e04ebd43846baf3a593cafa83/src/modules/wireplumber.cpp#L574-L575

Before this PR that ran once, from the constructor. Now `setupConnection()` re-runs the same async chain (`asyncLoadRequiredApiModules` → ... → `onMixerApiLoaded`) on every successful reconnect, and `sigc::signal::connect` doesn't dedupe — `event_box_` isn't recreated by `setupConnection()`/`teardownConnection()`, so a duplicate scroll handler is added per reconnect.

Behavior stays correct: GTK's `scroll-event` uses the boolean *handled* accumulator, so emission stops at the first handler returning `TRUE` and the duplicates never run — they just accumulate as dead connections (plus a redundant `add_events` call) for the life of the module.

Since this wiring only needs to happen once per module instance, move it here and delete it from `onMixerApiLoaded` (linked lines above). `handleScroll` already no-ops safely while `mixer_api_` is null, so wiring it before the first successful connect is safe.

```suggestion
  event_box_.add_events(Gdk::SCROLL_MASK | Gdk::SMOOTH_SCROLL_MASK);
  event_box_.signal_scroll_event().connect(sigc::mem_fun(*this, &Wireplumber::handleScroll));
}
```

src/modules/wireplumber.cpp
```diff
@@ -81,12 +105,62 @@ waybar::modules::Wireplumber::~Wireplumber() {
...
+  teardownConnection();
+  spdlog::info("[{}]: attempting to reconnect to PipeWire...", name_);
+  if (setupConnection()) {
+    spdlog::info("[{}]: reconnected to PipeWire", name_);
+    return false;
+  }
+  teardownConnection();
+  spdlog::debug("[{}]: reconnect failed; retrying in {} ms", name_, kReconnectIntervalMs);
+  return true;
+}
```
khaneliman, line 178: issue (blocking): stale async completions from a superseded connection can corrupt the new generation's state

`onReconnectTimeout` rebuilds `wp_core_`/`om_`/`apis_`/`pending_plugins_` in place on the same `self`, but `wp_core_load_component()` / `wp_object_activate()` are called with no `GCancellable`, and `isModuleAlive()` only checks that the C++ object still exists — not which connection generation a callback belongs to. If PipeWire drops again while the async chain (`onDefaultNodesApiLoaded` → `onMixerApiLoaded` → `onPluginActivated`) is still in flight — likely with a crash-looping service, the exact case this PR targets — the stale callback fires against the rebuilt connection's fields:

https://github.com/Alexays/Waybar/blob/0b88b4ef0c5c7e1e04ebd43846baf3a593cafa83/src/modules/wireplumber.cpp#L488-L490

A stray decrement corrupts the new generation's `pending_plugins_` (never reaches 0, or hits it early), and `wp_core_install_object_manager` can run on the new `om_`/`wp_core_` out of sequence — reproducing the stale/blank symptom of #2882 via a second disconnect during the reconnect window.

Fix: tag each generation (epoch counter bumped in `setupConnection()`, captured and compared in the callbacks), or pass a per-generation `GCancellable` cancelled in `teardownConnection()`, so completions from a torn-down generation are no-ops.

Alexays:
Thanks for the thorough review — both are addressed in the latest push.

**Generational aliasing (blocking):** you're right that `isModuleAlive()` only proves `self` still exists, not that a completion belongs to the current connection. Each async call now carries an `AsyncCall{self, generation}` as its user_data; `connection_generation_` is bumped in `setupConnection()`, and every one of the three async callbacks (`onDefaultNodesApiLoaded`, `onMixerApiLoaded`, `onPluginActivated`) drops out when its captured generation no longer matches — checked *after* the `isModuleAlive` short-circuit so a freed `self` is never dereferenced. A stale completion from a torn-down connection can no longer touch the rebuilt one's `pending_plugins_`/`om_`, so the second-disconnect-during-reconnect window you described is a no-op now.

**Duplicate scroll handler (non-blocking):** moved the `add_events`/`signal_scroll_event().connect` to the constructor and removed it from `onMixerApiLoaded`, since `handleScroll` already no-ops while `mixer_api_` is null.

Issue #3851, opened 2024-12-28
[FEATURE_REQUEST] static width and units for the network#speed module


ipuustin:
Currently the network#speed module has the bandwidth values length changing -- would it be possible to add padding to the values to prevent distracting resizing?

For example, go from `△ 4.0B/s  ▽ 128.0B/s` to `△   4.0B/s  ▽ 128.0B/s`

While at it, could we configure the lower and upper limit of the units used? I would like to have a kB (or MB) as the base unit to make the network traffic speed recognizable by a glance. So go from `△ 4.0B/s  ▽ 128.0B/s` to `△   0.0kB/s  ▽   0.1kB/s`.

michaeltraxler:
Yes, I agree, that feature would be very helpful. It is quite irritating, that the text in the waybar moves left to right when the used bandwidth changes. It should be a static text format, only the numbers should change, not the position. Otherwise it distracts too much, as the whole text moves around. 
Or can you point to the position in the source code, where I can change the format of the output myself?

theatischbein:
Isn't it already there?

I had the same problem and looked into it.

I change my config to this

```
    "network": {
       ...
        "format-wifi": "{essid:.15} {signalStrength}% {bandwidthTotalBytes:>3} ",
       ...
    },
```

The magic happens here `{bandwidthTotalBytes:>3}`.

With `{var:>3}` I tell the formater to reverse space for 3 digits in front or do a right align.

The code base is here: https://github.com/Alexays/Waybar/blob/master/include/util/format.hpp#L65C1-L79C6

michaeltraxler:
Very strange. I don't get any difference, also with ":>20".
I have "Waybar v0.12.0" on my system.

theatischbein:
here is my full network configuration:

```
    "network": {
        "format-wifi": "{essid} {signalStrength}% {bandwidthTotalBytes:>3} ",
//        "format-wifi": "{essid:.40} Down: {bandwidthDownBytes:>10} / Up: {bandwidthUpBytes:>10}",
        "format-ethernet": "{ipaddr}/{cidr} ",
        "tooltip-format": "{essid} on {ifname} ({frequency}) via {gwaddr} ",
        "format-linked": "{ifname} (No IP) ",
        "format-disconnected": "Disconnected ⚠",
        "format-alt": "{ifname}: {ipaddr}/{cidr}",
        "interval": 3,
    },
```

Maybe some other option does something strange?

michaeltraxler:
[...]

PR #5170, opened 2026-07-04
feat(format): expand pow_format modifiers

### LINKED ISSUE #2162: Network module: compact bandwidth formatting
For the network module, it would be nice to format the bandwidth speeds in a more compact, and also non-varying manner.

```
Current      Desired
420.1MB/s    420 M
420.1kB/s    420 k
420.1B/s     420 B
42.0B/s       42 B
4.2B/s         4 B
```

This shaves off a few unimportant characters for valuable waybar real estate, and prevents the module from varying in width (when wrapped in a monospace font, e.g. `<span font='NotoSansMono-Regular'>{bandwidthDownBytes}</span>`).

layus:
**What does this PR do?**

Extend the pow_format spec parser beyond alignment so configs can shape the rendered number, not just pad it. A forced scale (#, k, M, G, T, P) pins the SI prefix instead of auto-selecting it and, since the author then knows both scale and unit, hides the prefix and unit by default; U brings the unit back, u hides it in auto mode, and i forces integer display. b and B override the decimal or binary base independently of the call site, and a trailing width now fixes the coefficient field when a scale is forced, overflowing to '#' when it does not fit.

The old '>' and '<' branches rendered via fmt::format("{}", s), which discarded every new field on the recursive call; they now build from a single render path that reads the current formatter, so modifiers survive alignment. The auto path is byte-identical to before, verified against the previous implementation.

Adds a Catch2 suite covering each modifier alone and combined, plus the backward-compatible alignment cases, and documents the modifiers in the network and disk man pages. See test cases for examples.

**Related issues**
Closes #2162 #3851 

**Checklist**
- [X] Code is formatted with `clang-format`
- [X] Builds locally (`ninja -C build`)
- [X] Man page updated for any new/changed user-facing option (`man/`)
- [X] Tested against the affected module(s)

### DIFF
--- a/include/util/format.hpp
+++ b/include/util/format.hpp
@@ -23,30 +23,51 @@ class pow_format {
 namespace fmt {
 template <>
 struct formatter<pow_format> {
-  char spec = 0;
-  int width = 0;
+  char spec = 0;           // alignment: '>', '<', '=' (0 = none)
+  int width = 0;           // width digits; enforced only when scale_spec != 0
+  char scale_spec = 0;     // forced scale: 0 = auto, else one of '#','k','M','G','T','P'
+  char unit_pref = 0;      // unit tri-state: 0 = default, 'u' = hide, 'U' = show
+  char base_pref = 0;      // base tri-state: 0 = call-site, 'b' = decimal, 'B' = binary
+  bool force_int = false;  // 'i': force integer display
 
   template <typename ParseContext>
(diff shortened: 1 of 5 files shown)


layus:
Not sure all these options will ever be used, but now they are in the code.
Also, the only think I will not do is omit the "i" in the scale, so I will never render ki as k. Use b to switch off binary base if you want plain k/M/G

Issue #5186, opened 2026-07-06
possible memory leak with minimal Hyprland-only config


WXZhao7:
**Describe the bug**

Waybar appears to grow memory steadily over time even with a very small configuration.

<img width="1920" height="1280" alt="Image" src="https://github.com/user-attachments/assets/b13ff62a-2120-4994-8a5d-f4c2bc812f60" />

In this test run, CSS was disabled and only two built-in Hyprland modules were enabled:

- `hyprland/workspaces`
- `hyprland/window`

All center/right modules were commented out.

**To reproduce**

`systemctl --user start waybar.service`

```jsonc
{
  "layer": "top",
  "position": "top",
  "height": 34,
  "spacing": 0,
  "modules-left": [
    "hyprland/workspaces",
    "hyprland/window"
  ],
  "hyprland/workspaces": {
    "format": "{icon} {windows}",
    "format-window-separator": "",
    "sort-by": "id",
    "show-special": true,
    "special-visible-only": true,
    "on-scroll-up": "hyprctl dispatch 'hl.dsp.focus({ workspace = \"e+1\" })'",
    "on-scroll-down": "hyprctl dispatch 'hl.dsp.focus({ workspace = \"e-1\" })'",
    "format-icons": {
      "urgent": "",
      "special": "󰊠"
    },
    "workspace-taskbar": {
      "enable": true,
      "update-active-window": true,
      "format": "{icon}",
      "icon-size": 20,
      "on-click-window": "hyprctl dispatch 'hl.dsp.focus({ window = \"address:{address}\" })'"
    }
  },
  "hyprland/window": {
    "format": "{title}",
    "icon": true,
    "icon-size": 16,
    "max-length": 54,
    "separate-outputs": true,
    "rewrite": {
      "^\\s*$": "Desktop"
    }
  }
}
```

**Expected behaviour**
Waybar memory usage should remain mostly stable over time after startup.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Environment**

- Waybar: `v0.15.0`
- OS/kernel: `Linux arch 7.1.2-arch3-1 x86_64`
- Installed from Arch Linux official package: extra/waybar 0.15.0-2
- Compositor: Hyprland 0.55.4

**Debug output**

```
❯ waybar -l debug
[2026-07-06 19:15:32.807] [debug] Try expanding: $XDG_CONFIG_HOME/waybar/config
[...]

PR #5194, opened 2026-07-08
fix(hyprland/workspaces): skip taskbar rebuild when content is unchanged

tomasz-feliksik:
## Summary

First, thanks for Waybar — it's what I run every day. I went digging into the
memory growth reported in #5186 and think I found where a lot of it comes from,
so I wanted to share what I have along with a small change that helped a lot on
my setup. I'm not a Waybar/GTK expert, so please read the analysis below as
"here's what I measured" rather than the final word — I'm very happy to rework
the approach into whatever fits the codebase best.

## What seems to be happening

`Workspace::updateTaskbar()` rebuilds the whole per-window taskbar subtree
(buttons, labels, icons) on every call, and with `update-active-window` it is
called for every workspace on every `activewindowv2` event. From what I can
tell, most focus changes leave a given workspace's taskbar identical, so many of
those rebuilds don't change anything on screen.

Profiling with heaptrack (master `98b2a56`, 3 monitors, `update-active-window:
true`, driving 500 focus changes) pointed at the recreated widgets: each one,
when shown, appears to make GTK compute a `GtkCssStaticStyle` that isn't
released. ~75 MB accumulated over the 500 events (~143 KB each), almost all via:

    gtk_css_static_style_new_compute
      ... gtk_widget_show / gtk_widget_map / gtk_widget_realize ...
      Workspace::updateTaskbar()
      Workspace::update()
      Workspaces::updateWorkspaceStates()
      Workspaces::doUpdate()

(I could be misreading the GTK side of this — but the widget churn from the
rebuilds is clearly what triggers it.)

## What the change does

Compute the window list first, then skip the teardown+rebuild when a signature
of what the function renders (per-window class/title/address/active-state, plus
the workspace name and icon) hasn't changed since the last render. When nothing
visible changed, the existing widgets are already correct. Genuine changes
rebuild exactly as before. I tried to keep it small and match the surrounding
style, but I'm glad to take a different approach if this isn't the shape you'd
want.

## What it helps with

The redundant rebuilds — a workspace's taskbar being torn down and rebuilt when
its content didn't actually change. With `update-active-window` that was
happening on every focus event, for every workspace, on every monitor.

| | accumulated / 500 events | per event |
|---|---|---|
| before | ~73 MB | ~143 KB |
| after | ~0.36 MB | ~0.7 KB |

On my machine (NixOS, Hyprland, 3 monitors) waybar used to reach ~2 GB in
~7.5 h; with this change it stays around ~130 MB over a comparable stretch.

## What it does not solve (being upfront)

I don't think this addresses the underlying allocation — GTK still seems to hold
onto a `GtkCssStaticStyle` per widget shown; the change only avoids showing new
widgets when nothing changed. So genuine taskbar changes still rebuild and still
grow memory by their share:

- a window opening, closing, or moving between workspaces
- the active-window highlight moving within a strip
- a window title changing (browser tabs, terminals, editors) — titles are part
  of the rendered content, so they count as a change

So growth should now track real taskbar activity rather than the raw focus-event
rate — a big drop for most workloads, though not zero, and a title-churn-heavy
session would still creep up slowly. A complete fix probably lives either in
GTK's style lifetime, or in reusing the taskbar widgets in place instead of
recreating the subtree; both felt bigger than I wanted to attempt without your
input. I'd be happy to take a run at the widget-reuse version if you think
that's the better direction.

## Testing

- heaptrack before/after on master with the reported config
- A/B on an identical 500-focus-change run (with an idle control): +73 MB → +0.36 MB
- `clang-format` clean; builds with `meson setup` + `ninja -C build`
- Running it live (NixOS, Hyprland, 3 monitors); the taskbar still renders and
  updates correctly on window open/close and focus changes

Refs #5186 — I left it as a reference rather than an auto-close, since there's
still the residual above. Totally your call whether to close it or keep it open
to track the GTK side.

## Checklist

- [x] Man page updated for any new/changed user-facing option — N/A, no
      user-facing option added or changed
- [x] Tested against the affected module(s) — `hyprland/workspaces` taskbar]

### DIFF
--- a/include/modules/hyprland/workspace.hpp
+++ b/include/modules/hyprland/workspace.hpp
@@ -100,6 +100,10 @@ class Workspace {
   Gtk::Label m_labelBefore;
   Gtk::Label m_labelAfter;
 
+  // Signature of the last taskbar render, used to skip redundant rebuilds that
+  // otherwise leak GtkCssStaticStyle on every refresh (Alexays/Waybar#5186).
+  std::string m_lastTaskbarSignature;
+
   bool isEmpty() const;
   void updateTaskbar(const std::string& workspace_icon);
   bool handleClick(const GdkEventButton* event_button, WindowAddress const& addr) const;
--- a/src/modules/hyprland/workspace.cpp
+++ b/src/modules/hyprland/workspace.cpp
@@ -521,13 +521,7 @@ bool Workspace::isEmpty() const {
 }
 
 void Workspace::updateTaskbar(const std::string& workspace_icon) {
-  for (auto child : m_content.get_children()) {
-    if (child != &m_labelBefore) {
-      m_content.remove(*child);
-    }
-  }
-
-  // Build a list of windows to display, removing duplicates by window_class
+  // Build the list of windows to display, removing duplicates by window_class
   // and respecting max-icons limit
   std::vector<const WindowRepr*> windowsToShow;
   std::set<std::string> seenClasses;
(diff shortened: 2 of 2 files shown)


CtByte:
Hi, 

I also want to say thanks for Waybar. It made my migration to Linux a very pleasant one :slightly_smiling_face:

I hope it's okey to post this here, as I spent my fair share of digging into this before finding this PR.

---

Independent black-box confirmation, plus a quantified before/after for exactly the case this PR targets.

**Setup:** waybar 0.15.0 (Arch), single monitor, Hyprland. I was chasing a long-standing "waybar memory keeps growing, have to restart it periodically" problem, bisected it to the taskbar, then found #5186 / this PR.

**Baseline control** — a module-less bar held dead flat at ~49 MB for 10 min (0 KB growth), so this isn't core/GTK baseline.

**Focus-churn A/B** (the scenario this PR fixes): 8 windows on one workspace, 500 `hyprctl dispatch cyclenext` focus changes, taskbar on (`format:"{windows}"`, `icon-size:18`), RSS sampled every 100:

| `update-active-window` | RSS start → end | growth |
|---|---|---|
| **true** | 66 → 153 MB | **+87 MB**, dead linear (~17 MB / 100 changes, ~174 KB/event) |
| **false** | 64 → 64 MB | **+0 MB**, flat at every sample |

The ~174 KB/event lines up with your heaptrack ~143 KB/event, and `update-active-window: false` removes the growth entirely — consistent with your finding that the per-focus taskbar rebuilds are the source.

**Window open/close A/B** (a separate path, for completeness): open 6 windows, let the taskbar render icons, close all 6, ×8 (48 opens). Taskbar on: 61 → 85 MB (+23 MB, ~0.5 MB/window). Taskbar off (`format:"{name}"`): 55 → 55 MB (+0). So there's also a smaller rebuild-on-content-change leak on open/close, which the "skip rebuild when content unchanged" approach should help too.

Interim mitigation for anyone landing here: setting **`update-active-window: false`** eliminates the focus-change leak (measured +0 MB above) while keeping the per-window icons — leaving only the much rarer open/close churn — until this lands.

Thanks for the patch and the clear write-up. 👍

Issue #5205, opened 2026-07-17
Bad timing when initialized obj


Mrpaoo:
## Description
Currently, we just put `timer_thread={....}` in constructor, which may be call instantly and at this time, the obj wasn't ready yet, and i may cause some timing issue.

## Workaround
Since we use factory to create obj, it's okay do the init stuff that rely fully constructed obj right after `factory.Makemodule()`.

https://github.com/Alexays/Waybar/blob/7f732f055316d3b0ab89f57e4347d24c1cc96167/src/bar.cpp#L598-L604

Issue #5200, opened 2026-07-14
crash when screen close


iamveryverystupid:
**Describe the bug**
when output closes, with wlr-randr --output eDP-1 --off or suspending on my laptop, it segfaults

**To reproduce**

**Environment**
- Waybar v0.15.0-934-gcf19c836 (branch 'HEAD')
- gentoo llvm profile, with libcxx
- mango

**Debug output**

```
[2026-07-14 11:57:35.487] [debug] Output eDP-1 mapped (DPMS on), resuming modules
[2026-07-14 11:57:35.488] [debug] GTK widget tree:
window#waybar.background.left.eDP-1.topbar.mode-default:dir(ltr)
  decoration:dir(ltr)
  box.vertical:dir(ltr)
    box.vertical.modules-left:dir(ltr)
      widget:dir(ltr)
        box#gentoo.vertical:dir(ltr)
          widget:dir(ltr)
            box#custom-gentoo.horizontal.module:dir(ltr)
              image:dir(ltr)
              label:dir(ltr)
          revealer.drawer:dir(ltr)
            box.vertical:dir(ltr)
              widget.drawer-child:dir(ltr)
                box#custom-fuzzel.horizontal.module:dir(ltr)
                  image:dir(ltr)
                  label:dir(ltr)
              widget.drawer-child:dir(ltr)
                box#custom-foot.horizontal.module:dir(ltr)
                  image:dir(ltr)
                  label:dir(ltr)
              widget.drawer-child:dir(ltr)
                box#custom-qalculate.horizontal.module:dir(ltr)
                  image:dir(ltr)
                  label:dir(ltr)
              widget.drawer-child:dir(ltr)
                box#custom-thunar.horizontal.module:dir(ltr)
                  image:dir(ltr)
                  label:dir(ltr)
      widget:dir(ltr)
        box#custom-spacer.vertical.module:dir(ltr)
          label:dir(ltr)
          image:dir(ltr)
      widget:dir(ltr)
        box#window.vertical:dir(ltr)
          widget:dir(ltr)
            box#workspaces.vertical.module:dir(ltr)
              button.flat.text-button:dir(ltr)
                label:dir(ltr)
          revealer.drawer:dir(ltr)
            box.vertical:dir(ltr)
              widget.drawer-child:dir(ltr)
                box#custom-spacer.vertical.module:dir(ltr)
                  label:dir(ltr)
                  image:dir(ltr)
              widget.drawer-child:dir(ltr)
[...]

PR #5208, opened 2026-07-18
refactor: rm direct dp.emit() call in constructor.

Mrpaoo:
**What does this PR do?**
Remove the `dp.emit()` call in constructor, at this time, the `dp` isn't connect to callback, so it make no sense.

**Related issues**
N/A

**Checklist**


Mrpaoo:
However, `dp.emit()` still call before the `dp.connect` according to the https://github.com/Alexays/Waybar/issues/5200#issuecomment-5001892596, which `thread` could call `dp.emit()` before connected. I would track this down in #5205

Issue #4801, opened 2026-02-01
Tooltip of custom module doesn't sync


LolPopGames:
I try to make custom indicator module, but his tooltip doesn't sync unless I rerun the process or send `-USR2` signal (`pkill -USR2 waybar`)

This is the implementation:
+ `config.jsonc`:
```jsonc
// There is only what is needed for this issue
{
    "modules-right": [
        "custom/indicators",
    ],
    "custom/indicators": {
        "format": "{text}",
        "interval": 1,
        "return-type": "json",
        "exec": "$HOME/.config/waybar/custom/indicators.sh"
    }
}
```

+ `custom/indicators.sh`:
```sh
#!/bin/sh

# input4 is my keyboard
NUMLOCK=$(cat /sys/class/leds/input4::numlock/brightness)
CAPSLOCK=$(cat /sys/class/leds/input4::capslock/brightness)
# The value for $TOUCHPAD is defined correctly, just ignore this line
TOUCHPAD=$([ -f "$XDG_RUNTIME_DIR/hypr-scripts/touchpad-off" ] && echo 0 || echo 1)

# making JSON
# {"text": "CHARACTERS_IF_NEEDED", "tooltip": "□ NumLock\n□ CapsLock\n□ Touchpad"}
echo -n '{"text": "'
[ $NUMLOCK -eq 1 ]  && echo -n '󰎤' # this is the chars from nerd fonts
[ $CAPSLOCK -eq 1 ] && echo -n '󰬈'
[ $TOUCHPAD -eq 0 ] && echo -n '󰍾'
echo -n '", "tooltip": "'
[ $NUMLOCK -eq 1 ]  && echo -n '󰎤' || echo -n '󰎦'; echo -n ' NumLock\n'
[ $CAPSLOCK -eq 1 ] && echo -n '󰬈' || echo -n '󰯫'; echo -n ' CapsLock\n'
[ $TOUCHPAD -eq 1 ] && echo -n '󰍽' || echo -n '󰍾'; echo -n ' Touchpad'
echo '"}'
```

What is displayed and synchronized correctly in format, but if we, for example, press NumLock and hover over this module, the tooltip will show that it is still turned off

And the question is:
> How to correctly sync this tooltip?

P.S. I work on Arch Linux, Hyprland v0.53.3 (Wayland), Waybar v0.14.0

Issue #4519, opened 2025-10-02
All tooltips stop rendering across the bar when user includes a custom module with a menu, tooltip enabled, and custom tooltip-format.


uhs-robert:
## Issue
**All tooltips stop rendering across the bar** when there is at least one `custom/*` module that simultaneously has:
- `"tooltip": true`
- `"tooltip-format": "<any value, including empty string>"`
- `"menu": "<any value>"`

### Notes
- Reproducible **only** with `custom/*` modules; built-in modules are unaffected.
- A menu file or `menu-actions` are **not required** to trigger the bug.
- Occurs **with or without** custom CSS.
- Console/trace logs show **no errors and no hover events** when the bug is active.

### Suspected regression window 
Tooltips worked in a source build from late Sep 2025. The issue appeared after rebuilding from source on Oct 1–2, 2025.

## Minimal Steps to Reproduce

1. Add the `custom module config` below
2. Launch waybar
3. Hover over ANY tooltip on the bar, and observe that it doesn't display
4. Remove tooltip-format, reload bar, and observe that tooltips work again

### Custom Module Config
```
"custom/power": {
	"format" : "⏻ ",
	"tooltip": true,
    "tooltip-format": "",
	"menu": "on-click",
},
```

For a full real-world setup, use the example from [man(5)](https://github.com/Alexays/Waybar/blob/18a9835b7cf03ce116c0da03939e1cfef7796f7e/man/waybar-menu.5.scd) but enable tooltip and set tooltip-format to any value.

## Expected Result

Tooltips continue to work across all modules when the user adds a custom/module with tooltip: true, tooltip-format: "", and a menu.

## Video

https://github.com/user-attachments/assets/4f868ea0-b3cb-4df7-aa53-699814a3ba60

## Debug Logs

[waybar-trace-no-css.log](https://github.com/user-attachments/files/22663571/waybar-trace-no-css.log)

## System Environment
```
Waybar: Waybar v0.14.0
Kernel: 6.16.8-200.fc42.x86_64
OS: "Fedora Linux 42 (KDE Plasma Desktop Edition)"
GTK: 3.24.49
gtkmm: 3.24.10
Wayland: 1.24.0
Hyprland: Hyprland 0.51.1 built from branch  at commit 71a1216abcc7031776630a6d88f105605c4dc1c9  ([gha] Nix: update inputs).
Date: Mon Sep 22 20:54:03 2025
Tag: v0.51.1, commits: 6436
built against:
[...]

PR #5212, opened 2026-07-20
fix(tooltips): keep dynamic tooltips from resetting hover delay

bjnobre:
Cache ALabel tooltip markup and provide it through query-tooltip instead of repeatedly setting GtkWidget's tooltip-markup property. This prevents frequently updating modules from restarting GTK's display-wide tooltip timer.

Keep the active tooltip updated directly so dynamic tooltip contents continue refreshing while the pointer remains stationary.

## Problem

  Setting `tooltip-markup` queues a display-wide GTK tooltip query. A
  module updating more frequently than GTK's hover delay continually
  restarts that delay, preventing its own tooltip—and potentially other
  modules' tooltips—from appearing.

  ## Testing

  Tested with a custom module updating its displayed text and tooltip every
  250 ms, alongside a static control module.

  Verified that:

  - Both tooltips appear normally.
  - The rapidly updating tooltip refreshes while the pointer remains still.
  - The rapidly updating module no longer prevents the control tooltip
    from appearing.
  - The patched production build compiles and runs with the existing
    Waybar configuration.

  ## Related issues

  Related to #4801, #4519, and #4842.

  #4812 restored dynamic tooltip synchronization by repeatedly setting
  `tooltip-markup`, but was reverted because doing so resets GTK's hover
  state. #4900 subsequently avoided redundant property assignments by
  deduplicating identical markup.

  That does not cover genuinely dynamic tooltip content: when the value
  changes on every update, every assignment is non-redundant and GTK's
  display-wide tooltip delay is still continually restarted. This change
  provides tooltip content through `query-tooltip` and updates an already
  active tooltip directly, preserving both initial display and live
  updates.

  **Checklist**
  - [x] N/A, no user-facing option
  changed

### DIFF
--- a/include/ALabel.hpp
+++ b/include/ALabel.hpp
@@ -4,6 +4,7 @@
 #include <fmt/format.h>
 #include <glibmm/markup.h>
 #include <gtkmm/label.h>
+#include <gtkmm/tooltip.h>
 #include <json/json.h>
 
 #include <optional>
@@ -92,6 +93,7 @@ class ALabel : public AModule {
   // no collation weight, so two different icons compare equal.
   std::optional<std::string> last_label_markup_;
   std::optional<std::string> last_tooltip_markup_;
+  Glib::RefPtr<Gtk::Tooltip> active_tooltip_;
 };
 
 }  // namespace waybar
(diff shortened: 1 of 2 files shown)


Alexays:
Thanks!

PR #5216, opened 2026-07-23
util/enum: make header-only and drop mandatory Hyprland dependency

LukashonakV:
Hi @Alexays , 
While working on GTK4, I found a mandatory dependency that makes it impossible to build Waybar core without a Hyprland dependency, which was a bit surprising to me. I asked Kimi to prepare a proper justification, and this is what we came up with )))

----

The generic `util::EnumParser` was implemented in `src/util/enum.cpp` with explicit template instantiations for Hyprland-specific enums (`Workspaces::SortMethod`, `Workspaces::ActiveWindowPosition`) and unconditionally included `modules/hyprland/workspaces.hpp`. This architectural violation forced all builds to compile Hyprland-specific headers, making Hyprland a mandatory dependency even when its module was disabled.

The patch refactors `EnumParser` into a header-only free function `parseStringToEnum`, deletes `src/util/enum.cpp` from the build, and makes the lookup maps `static inline const`. This:

* **Removes the hard dependency** of generic utility code on a specific compositor module.
* **Enables builds without Hyprland**, since the utility no longer references Hyprland types in a compiled translation unit.
* **Eliminates boilerplate** (ctor/dtor/explicit instantiations) by letting the template be instantiated implicitly at call sites.
* **Avoids ODR issues** by marking shared lookup tables as `inline`.


Alexays:
LGTM, thx

Issue #2656, opened 2023-11-09
wlr/workspaces: wlr/taskbar: group taskbar buttons under workspaces [sway] [hyprland]


BrianCArnold:
Note: The feature I'm kind of talking about here would be kind of a combination of taskbar and workspaces.

Given the following workspaces and windows:
![image](https://github.com/Alexays/Waybar/assets/38593283/cebaab2a-e8d6-4ad6-8a97-73741f62e768)

This is how my workspaces and taskbar show up in my waybar:
- Workspaces
  - 1
  - 2
  - 3
- Taskbar
  - Window A
  - Window B
  - Window M
  - Window N
  - Window Y
  - Window Z

Which makes it difficult to see how I have things organized.

It would be nice to have a unified module that presents that information in a somewhat hierarchical nature, either like this:

| |
| --- |
|**1** |
| Window A |
| Window B |
|**2** |
| Window M |
| Window N |
|**3** |
| Window Y |
| Window Z |

Or

![image](https://github.com/Alexays/Waybar/assets/38593283/5bc28616-382d-4d66-95e1-728343c71971)

Vaisakhkm2625:
![1700672020_grim](https://github.com/Alexays/Waybar/assets/68694876/7e3379e6-aa6a-4082-8c27-a9f2f62715a4)

i am also looking for this exact functionality, and this is actually possible with "hyprland/workspaces", the biggest issue is, hyprland/workspaces don't set icons automatically and we have to set icons for each applications according to classname.... which suchs...

NikkSaan:
nwg-panel supports this... I've tried it, and it's truly a better UX! I would also include the ability to drag and drop icons to enable moving windows between workspaces using the mouse only...

ps. Very nice presentation of the feature OP!

jgregoire:
I really only want to see windows in the taskbar for the currently active workspace.

WammKD:
Would like to, also, second this; I used to use this with `tint2` all the time.

![Screenshot from 2022-03-08 17-02-23](https://github.com/Alexays/Waybar/assets/10533934/1d594cf7-3cec-4812-b931-a41f8dca2d61)
(only screenshot I could find from before I switched to Wayland though some more icons would help make it clearer what's going on here…)

[...]

Issue #1043, opened 2021-03-04
Group icons in wlr-taskbar for windows in a container


kushraj:
Feature request : Group the icons in wlr-taskbar for windows in a single container

For those using the wlr-taskbar module with icons only, this feature will make switching focus a lot handy and practical.

I will really like to see something close to this getting implemented

phi-mah:
I would second this. In particular, it would be nice to identify the windows living on the scratchpad. This would remind me that I will not find these by cycling through my workspaces (Thus, kind of related to #1034).

I did some research on this, and as far as I understood these things it might be difficult to achieve: The taskbar module solely relies on the the ` wlr-foreign-toplevel-management-unstable-v1` protocol, which does not know the particular sway window properties.

By contrast, e.g. the `sway` module uses `ipc` to connect to the running sway instance.   Feature requests like this would imply to pull in information from there also in the taskbar module and thus sacrificing generality...

PR #5223, opened 2026-07-28
feat: Sway taskbar

### LINKED ISSUE #762: [Taskbar] Add option "all_workspaces"
I have started to implement feature show tasks only for current workspace. Feature will be controlled by option "all_workspaces": bool

### LINKED ISSUE #1278: [wlr/Taskbar] Add an option to sort opened application by workspace
Hi, actually, applications in taskbar spawn only in the order of opening, so if one opens a lot of application on different workspaces, it can be difficult to idenitfy them by the workspace.

Another option would be to group them by workspace

Any feedback on this idea?

Thanks!

### LINKED ISSUE #1997: how to show only opened apps in the current workspace in wlr/taskbar?
wlr/taskbar shows all open applications regardless of the current workspace I'm using

### LINKED ISSUE #3591: Taskbar - Limit to current workspace
Hello,

A feature request that would be very nice to have is to limit taskbar to show apps running only in the current workspace.

Thanks!

### LINKED ISSUE #5084: "wlr/taskbar" module doesn't work properly on new swaywm 1.12 release
Even though I have a lot of apps open, it refuses to list those apps unless that app is currently on active workspace.. 

> This is working normally back in sway 1.11

https://github.com/user-attachments/assets/50891c92-2e16-44cb-ab22-8f8fd21af232

OS: `Arch linux`
waybar: `waybar-git 0.15.0.r173.g0594574-1` and `0.15`
WM: Swaywm `1.12`

Samarkin:
**What does this PR do?**
Adds an alternative to `wlr/taskbar` that uses Sway IPC instead of a generic Wayland protocol.
The new module's configuration is a superset of `wlr/taskbar`'s, so it works as a drop-in replacement.

Immediately available benefits:
1. **Ability to display applications in tabbed containers.**
Fixes #5084, where Sway 1.12 / wlroots 0.20 started sending `output_leave` for windows outside the active workspace, so the toplevel protocol stopped reporting them. Also addresses the limitation diagnosed in #1043.
`wlr/taskbar`: wlr-taskbar `sway/taskbar`: sway-taskbar

2. **Separation of workspaces and outputs.**
Implements the six-year-old request in #762, and gives Sway the per-workspace filtering that #1997 and #3591 were closed without (`workspace-taskbar` covers Hyprland and niri only). It also removes the need for the `"all-outputs": true` workaround from #5084, which trades away per-output filtering to get per-workspace behavior, since `all-workspaces` is now a separate option.
3. **New CSS classes for windows on the visible workspace and current output** (`.visible`, `.current_output`), so windows you aren't looking at can be de-emphasized in CSS rather than hidden outright.
4. **Per-workspace ordering.**
Tasks follow Sway's tree order, grouped by output and workspace, rather than the order they were opened – what #1278 asked for. Visible with `all-workspaces: true` and the default `sort-by-app-id: false`.

**Related issues**
Closes #5084
Closes #762
Closes #1997 _(for Sway too this time)_
Closes #3591 _(for Sway too this time)_
Closes #1278
Refs #2656
Refs #1043

**Checklist**
- [x] Tested ~~against the affected module(s)~~ the newly added module thoroughly and expect no impact on other modules

**Note**
The PR is split into several commits that follow the natural progression of the module from a simple `wlr/taskbar` clone to a genuinely useful expansion.
A good way to review is to explore the diff between the first commit and wlr/taskbar, and then treat every next commit as a separate change.

### DIFF
--- a/.github/wiki/mapping.json
+++ b/.github/wiki/mapping.json
@@ -35,7 +35,7 @@
   "Module:-PulseAudio-Slider": ["waybar-pulseaudio-slider"],
   "Module:-River": ["waybar-river-tags", "waybar-river-mode", "waybar-river-window", "waybar-river-layout"],
   "Module:-Sndio": ["waybar-sndio"],
-  "Module:-Sway": ["waybar-sway-workspaces", "waybar-sway-window", "waybar-sway-mode", "waybar-sway-scratchpad"],
+  "Module:-Sway": ["waybar-sway-workspaces", "waybar-sway-window", "waybar-sway-mode", "waybar-sway-scratchpad", "waybar-sway-taskbar"],
   "Module:-Systemd-failed-units": ["waybar-systemd-failed-units"],
   "Module:-Taskbar": ["waybar-wlr-taskbar"],
   "Module:-Temperature": ["waybar-temperature"],
--- a/include/modules/sway/taskbar.hpp
+++ b/include/modules/sway/taskbar.hpp
@@ -1,0 +1,196 @@
+#pragma once
+
+#include <gdk/gdk.h>
+#include <giomm/desktopappinfo.h>
+#include <glibmm/refptr.h>
+#include <gtkmm/box.h>
+#include <gtkmm/button.h>
+#include <gtkmm/image.h>
+#include <gtkmm/label.h>
+
+#include <cstdint>
+#include <map>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <string_view>
+#include <unordered_set>
+#include <vector>
+
+#include "AModule.hpp"
+#include "bar.hpp"
+#include "modules/sway/ipc/client.hpp"
+#include "util/icon_loader.hpp"
+#include "util/json.hpp"
+
+namespace waybar::modules::sway {
+
+// Plain data snapshot describing a single view (window), parsed off the sway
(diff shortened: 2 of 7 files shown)


Samarkin:
@AlieeLinux, @anmolmathias – would you be willing to try my branch?

Samarkin:
@Alexays – did you get a chance to take a look yet?
It seems like a useful addition (judging by the number of referenced issues), it solves a real problem, works fine in all the cases I tested, and I think the risk of merging it is pretty low even if it's not perfect – it's not malware and the impact is limited to people who will explicitly enable the new module.

AlieeLinux:
@Samarkin sorry for late reply, I will try it!
I update you latar

AlieeLinux:
update: with all-output; true, it works normally and intended, but if oyu remove the all-output to default, the issue exist again
image

Still, I think i can close my issue now

Samarkin:
@AlieeLinux – thanks for the test! Try `all-workspaces: true`, I added it specifically for your scenario.

larpedd:
!! need

larpedd:
hmm seem to notice that this patch only work on sway,,, not wlroots DE's :(((

Samarkin:
@Alexays – is there any issue with the code that you want me to address?

PR #5231, opened 2026-07-30
fix(hyprland): detect dispatch protocol via configProvider

### LINKED ISSUE #5198: hyprland/workspaces: Workspace switching buttons are unclickable on version >= 0.55 with traditional config format
**Describe the bug**
workspace clicking is broken for the users who still use traditional `hyprland.conf` configs.
[Hyprland only loads](https://github.com/hyprwm/Hyprland/blob/v0.55.0/src/config/ConfigManager.cpp#L45-L51) the Lua configuration manager if the user's config file explicitly ends with `.lua`. Otherwise, it falls back to the traditional config parser.
As a result, Waybar sends the Lua format to a traditional Hyprland instance, resulting in an  Invalid dispatcher error and breaking clicks.

**To reproduce**
Use traditional config on hyprland v0.55.
Use waybar workspace module.

CristianMz21:
Fixes #5198.

## Problem

`IPC::isLuaProtocol()` decides between the legacy and the Lua dispatch protocol from the Hyprland version alone:

```cpp
luaProto = major > 0 || (major == 0 && minor >= 54);
```

That is not the rule Hyprland uses. It [only loads the Lua config manager when the config file name ends in `.lua`](https://github.com/hyprwm/Hyprland/blob/v0.55.0/src/config/ConfigManager.cpp#L45-L51); an instance started with a traditional `hyprland.conf` keeps the legacy parser regardless of its version.

So on Hyprland >= 0.54 with a `.conf`, Waybar sends Lua-formatted dispatches, Hyprland answers `Invalid dispatcher`, and clicking a workspace button does nothing.

## Fix

Hyprland already reports which config manager it loaded, in the `systeminfo` reply:

```
$ hyprctl systeminfo | grep configProvider
configProvider: lua
```

This reads that field first and falls back to the existing version heuristic when it is absent, which is the case on versions predating the Lua config manager. The query is read-only, so detection stays free of the side effects that motivated replacing the earlier `dispatch workspace __waybar_probe__` approach.

Behaviour by case:

| `configProvider` | Result |
| --- | --- |
| `lua` | Lua protocol |
| anything else | legacy protocol |
| field absent (older Hyprland) | falls back to the version heuristic |

The parsing sits in a pure `parseConfigProvider()` helper, separate from the socket call, so it is covered by unit tests without needing a live compositor.

## Tests

Added to `test/hyprland/backend.cpp`:

- **`parseConfigProvider reads the config manager Hyprland loaded`** — a realistic full `systeminfo` reply, bare `lua`, bare `legacy`, `legacy` substituted into the full reply (the exact #5198 case), and an unrecognised value that must not be treated as Lua.
- **`parseConfigProvider tolerates formatting variations`** — tab separator, multiple spaces, CRLF line ending, value on the last line with no trailing newline, and an empty value.
- **`parseConfigProvider returns nullopt when the field is absent`** — an older reply and an empty reply, both of which must fall through to version detection rather than assume legacy.
- **`luaProtocolFromSystemInfo returns nullopt when Hyprland is not running`** — `getSocket1Reply` throws; detection has to degrade to the version fallback instead of propagating and breaking the click.

Result on this branch:

```
$ meson test -C build --print-errorlogs
1/3 waybar   OK  0.04s
2/3 hyprland OK  0.02s
3/3 utils    OK  0.71s
Ok: 3   Fail: 0

$ ./build/test/hyprland/hyprland_test
All tests passed (64 assertions in 16 test cases)
```

Also verified end to end on Hyprland 0.56.1 with a Lua config: `systeminfo` over the IPC socket reports `configProvider: lua`, and the resulting `/dispatch hl.dsp.focus({ workspace = "3" })` returns `ok` and switches workspace.

Caveat worth stating plainly: I could not exercise the legacy branch against a live pre-0.54 instance, only the parsing that selects it. Confirmation from someone running a traditional `hyprland.conf` would be welcome.

### DIFF
--- a/include/modules/hyprland/backend.hpp
+++ b/include/modules/hyprland/backend.hpp
@@ -50,14 +50,10 @@ class IPC {
   /// Resolved once from the config manager it reports, then cached.
   static bool isLuaProtocol();
 
-  /// Ask Hyprland which config manager it loaded, via the "systeminfo" reply.
-  /// Returns nullopt when the query fails or the field is not reported.
-  static std::optional<bool> luaProtocolFromSystemInfo();
-
-  /// Extract the "configProvider:" field from a "systeminfo" reply.
-  /// Returns true for the Lua manager, false for anything else, and nullopt
-  /// when the field is absent, as on versions predating the Lua config manager.
-  static std::optional<bool> parseConfigProvider(const std::string& systemInfo);
+  /// Whether a "systeminfo" reply reports the Lua config manager. An absent
+  /// "configProvider:" field means the instance predates the Lua config
+  /// manager (< 0.55) and therefore speaks the legacy protocol.
+  static bool isLuaConfigProvider(const std::string& systemInfo);
 
   static std::optional<bool> s_luaProtocolDetected_;  // cached detection result
 
--- a/include/modules/hyprland/backend.hpp
+++ b/include/modules/hyprland/backend.hpp
@@ -47,8 +47,17 @@ class IPC {
   static std::filesystem::path socketFolder_;
 
   /// Detect whether the running Hyprland uses the Lua-based IPC protocol.
-  /// Returns true for Hyprland >= 0.54 (Lua config), false for older versions.
+  /// Resolved once from the config manager it reports, then cached.
   static bool isLuaProtocol();
+
+  /// Ask Hyprland which config manager it loaded, via the "systeminfo" reply.
+  /// Returns nullopt when the query fails or the field is not reported.
+  static std::optional<bool> luaProtocolFromSystemInfo();
+
+  /// Extract the "configProvider:" field from a "systeminfo" reply.
+  /// Returns true for the Lua manager, false for anything else, and nullopt
+  /// when the field is absent, as on versions predating the Lua config manager.
+  static std::optional<bool> parseConfigProvider(const std::string& systemInfo);
 
   static std::optional<bool> s_luaProtocolDetected_;  // cached detection result
 
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -10,7 +10,6 @@
 #include <unistd.h>
 
 #include <array>
-#include <cctype>
 #include <cerrno>
 #include <cstring>
 #include <filesystem>
@@ -19,6 +18,7 @@
 #include <string_view>
 
 #include "util/scoped_fd.hpp"
+#include "util/string.hpp"
 
 namespace waybar::modules::hyprland {
 
@@ -294,42 +294,29 @@ Json::Value IPC::getSocket1JsonReply(const std::string& rq) {
   return parser_.parse(reply);
 }
 
-std::optional<bool> IPC::parseConfigProvider(const std::string& systemInfo) {
+bool IPC::isLuaConfigProvider(const std::string& systemInfo) {
   // Hyprland reports which config manager it actually loaded in "systeminfo",
   // as a "configProvider: lua" / "configProvider: legacy" line. That is the
   // authoritative signal: the version alone is not enough, because Hyprland
   // only uses the Lua manager when the config file name ends in ".lua", so a
   // >= 0.54 instance started with a traditional hyprland.conf still speaks the
-  // legacy dispatch protocol.
+  // legacy dispatch protocol. The field landed together with the Lua config
+  // manager in 0.55, so its absence means the instance predates Lua support
+  // entirely and necessarily speaks the legacy protocol.
   static constexpr std::string_view key = "configProvider:";
 
   const size_t keyPos = systemInfo.find(key);
   if (keyPos == std::string::npos) {
-    return std::nullopt;
-  }
-
-  size_t valuePos = systemInfo.find_first_not_of(" \t", keyPos + key.size());
-  if (valuePos == std::string::npos) {
     return false;
   }
 
-  const size_t end = systemInfo.find_first_of("\r\n", valuePos);
-  std::string provider = systemInfo.substr(valuePos, end - valuePos);
-  while (!provider.empty() && std::isspace(static_cast<unsigned char>(provider.back()))) {
-    provider.pop_back();
-  }
+  const size_t valuePos = keyPos + key.size();
+  const size_t lineEnd = systemInfo.find('\n', valuePos);
+  const std::string value = lineEnd == std::string::npos
+                                ? systemInfo.substr(valuePos)
+                                : systemInfo.substr(valuePos, lineEnd - valuePos);
 
-  spdlog::debug("Hyprland IPC: systeminfo reports configProvider '{}'", provider);
-  return provider == "lua";
-}
-
-std::optional<bool> IPC::luaProtocolFromSystemInfo() {
-  try {
-    return parseConfigProvider(getSocket1Reply("systeminfo"));
-  } catch (const std::exception& e) {
-    spdlog::warn("Hyprland IPC: could not read configProvider from systeminfo ({})", e.what());
-    return std::nullopt;
-  }
+  return trim(value) == "lua";
 }
 
 bool IPC::isLuaProtocol() {
@@ -337,12 +324,15 @@ bool IPC::isLuaProtocol() {
     return *s_luaProtocolDetected_;
   }
 
-  // configProvider landed together with the Lua config manager in 0.55, so its
-  // absence means the instance predates Lua support entirely and necessarily
-  // speaks the legacy protocol. That makes the field sufficient on its own, and
-  // it is read-only, so detection has none of the side effects of an actual
-  // dispatch probe.
-  const bool luaProto = luaProtocolFromSystemInfo().value_or(false);
+  // The query is read-only, so detection has none of the side effects of an
+  // actual dispatch probe.
+  bool luaProto = false;
+  try {
+    luaProto = isLuaConfigProvider(getSocket1Reply("systeminfo"));
+  } catch (const std::exception& e) {
+    spdlog::warn("Hyprland IPC: could not read systeminfo ({}), assuming legacy protocol",
+                 e.what());
+  }
 
   if (luaProto) {
     spdlog::info("Hyprland IPC: detected Lua-based dispatch protocol");
--- a/src/modules/hyprland/backend.cpp
+++ b/src/modules/hyprland/backend.cpp
@@ -296,7 +296,7 @@ Json::Value IPC::getSocket1JsonReply(const std::string& rq) {
 
 bool IPC::isLuaConfigProvider(const std::string& systemInfo) {
   // Hyprland reports which config manager it actually loaded in "systeminfo",
-  // as a "configProvider: lua" / "configProvider: legacy" line. That is the
+  // as a "configProvider: lua" / "configProvider: hyprlang" line. That is the
   // authoritative signal: the version alone is not enough, because Hyprland
   // only uses the Lua manager when the config file name ends in ".lua", so a
   // >= 0.54 instance started with a traditional hyprland.conf still speaks the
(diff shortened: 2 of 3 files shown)


CristianMz21:
Now running in production on a real machine, not just in tests.

Backported onto the 0.14.0 packaging and installed system-wide (Fedora 43, Hyprland 0.56.1, Lua config). Clicking a workspace button switches, and the detection resolves through the new path on the first dispatch:

```
$ journalctl --user -u waybar.service | grep 'dispatch protocol'
[info] Hyprland IPC: detected Lua-based dispatch protocol (configProvider)
```

No `Invalid dispatcher` in the log afterwards. The module in use is `hyprland/workspaces#rw` with `persistent-workspaces` and `window-rewrite`, so this covers the taskbar-style configuration too, not only plain numbered buttons.

Still the same caveat as above: the legacy branch is covered by the unit tests but I have no pre-0.54 instance to confirm it against.

src/modules/hyprland/backend.cpp
```diff
@@ -333,9 +386,9 @@ bool IPC::isLuaProtocol() {
   }
 
   if (luaProto) {
-    spdlog::info("Hyprland IPC: detected Lua-based dispatch protocol (Hyprland >= 0.54)");
+    spdlog::info("Hyprland IPC: detected Lua-based dispatch protocol (version >= 0.54)");
   } else {
```
Mrpaoo: Is the code here make no sense any more, can we just remove them

CristianMz21: You are right, and it turned out to be stronger than "no longer needed" — the fallback was actively wrong. Removed in bfd0cfe.

Checked when each piece landed in Hyprland:

| | |
| --- | --- |
| v0.54.0 released | 2026-02-27 |
| `config/lua`: init lua config manager ([#13817](https://github.com/hyprwm/Hyprland/pull/13817)) | 2026-04-26 |
| v0.55.0 released | 2026-05-09 |

`src/config/lua/` does not exist in the v0.54.0 tree; it appears in v0.55.0. `configProvider` follows exactly the same line — absent in v0.54.0, present from v0.55.0 on.

Two consequences:

1. There is no version that speaks Lua but does not report `configProvider`, so the field is sufficient on its own and the fallback is dead code.
2. `minor >= 54` was off by one. On a real 0.54 instance, which has no Lua config manager at all, it claimed Lua and broke exactly the clicks this is meant to fix. Falling back to it kept that bug alive.

So absent field now means legacy, which is correct for every release predating 0.55.

Net effect is smaller than before: -36 lines from the old heuristic, and one detection path instead of two.

CristianMz21:
One more data point, since #5198 raised the question of what happens when Hyprland drops legacy config support ([hyprwm/Hyprland#15539](https://github.com/hyprwm/Hyprland/pull/15539), merged 2026-07-22): the field this PR relies on survives that removal.

On Hyprland `main`, after legacy support is gone, `systeminfo` still reports it:

```cpp
// src/helpers/SystemInfo.cpp
configProvider: {}
// ...
Config::typeToString(Config::mgr()->type())
```

```cpp
// src/config/ConfigManager.cpp
const char* Config::typeToString(eConfigManagerType t) {
    switch (t) {
        case CONFIG_LUA: return "lua";
        default: return "error";
    }
}
```

So the three cases this PR handles stay correct across that transition:

| Reply | Detected as | Correct for |
| --- | --- | --- |
| `configProvider: lua` | Lua | 0.55+ with a `.lua` config, and every post-removal build |
| any other value | legacy | 0.55/0.56 started with `hyprland.conf` |
| field absent | legacy | anything predating 0.55, which has no Lua manager |

Worth contrasting with the version heuristic this replaces: it would have kept drifting, since "new enough" stops implying anything about the parser once the config file name is no longer the deciding factor.

Also note `main` gained a JSON form of the reply (`"configProvider": "lua"`). This PR queries the plain-text `systeminfo`, which is unchanged, so nothing breaks — but it is there if a future change prefers parsing JSON.

include/modules/hyprland/backend.hpp
```diff
@@ -47,9 +47,18 @@ class IPC {
...
   static bool isLuaProtocol();
 
+  /// Ask Hyprland which config manager it loaded, via the "systeminfo" reply.
+  /// Returns nullopt when the query fails or the field is not reported.
+  static std::optional<bool> luaProtocolFromSystemInfo();
+
+  /// Extract the "configProvider:" field from a "systeminfo" reply.
+  /// Returns true for the Lua manager, false for anything else, and nullopt
+  /// when the field is absent, as on versions predating the Lua config manager.
+  static std::optional<bool> parseConfigProvider(const std::string& systemInfo);
```
Bart97: The name `parseConfigProvider` suggests that the function returns a parsed config provider, whereas it actually answers the question "is the config provider Lua?". Since the return type is `std::optional<bool>` it would be better to use a name similar to `isLuaProtocol`, where the boolean states are clearly understandable.

src/modules/hyprland/backend.cpp
```diff
@@ -292,48 +294,58 @@ Json::Value IPC::getSocket1JsonReply(const std::string& rq) {
...
-      if (dot != std::string::npos && dot + 1 < versionStr.size()) {
-        minor = std::stoi(versionStr.substr(dot + 1));
-      }
-      luaProto = major > 0 || (major == 0 && minor >= 54);
-    } else {
-      spdlog::warn("Hyprland IPC: could not parse version '{}', assuming legacy protocol",
-                   versionStr);
-    }
+  const size_t end = systemInfo.find_first_of("\r\n", valuePos);
+  std::string provider = systemInfo.substr(valuePos, end - valuePos);
```
Bart97: Consider adding a check for end == std::string::npos. This code will *usually* work (unless someone tries to build it on some weird platform), but I'm not sure if it's considered as a good pattern.

std::string::npos is defined as `static constexpr size_type npos = size_type(-1);`, where size_type is defined as an unsigned integer type. So npos actually equals to the maximum positive value of size_type.
This line relies on end being a huge number in such case, and valuePos will usually have a significantly lower value, so the result of the subtraction will still exceed the remaining length of the string.
Technically this is fine, because [according to cppreference substr](https://cppreference.com/cpp/string/basic_string/substr)

> Returns a substring [pos, pos + count). If the requested substring extends past the end of the string, i.e. the count is greater than size() - pos (e.g. if count == npos), the returned substring is [pos, size()).

src/modules/hyprland/backend.cpp
```diff
@@ -292,48 +294,58 @@ Json::Value IPC::getSocket1JsonReply(const std::string& rq) {
...
+  while (!provider.empty() && std::isspace(static_cast<unsigned char>(provider.back()))) {
```
Bart97: include/util/string.hpp contains an rtrim function, would that be sufficient?

src/modules/hyprland/backend.cpp
```diff
@@ -292,48 +294,58 @@ Json::Value IPC::getSocket1JsonReply(const std::string& rq) {
...
+  if (s_luaProtocolDetected_.has_value()) {
+    return *s_luaProtocolDetected_;
   }
 
+  // configProvider landed together with the Lua config manager in 0.55, so its
+  // absence means the instance predates Lua support entirely and necessarily
+  // speaks the legacy protocol. That makes the field sufficient on its own, and
+  // it is read-only, so detection has none of the side effects of an actual
+  // dispatch probe.
+  const bool luaProto = luaProtocolFromSystemInfo().value_or(false);
```
Bart97: Is it worth it to have those functions return an optional bool considering that in the end nullopt follows the same path as false?

src/modules/hyprland/backend.cpp
```diff
@@ -292,48 +294,58 @@ Json::Value IPC::getSocket1JsonReply(const std::string& rq) {
...
+  try {
+    return parseConfigProvider(getSocket1Reply("systeminfo"));
   } catch (const std::exception& e) {
-    spdlog::warn("Hyprland IPC: version detection failed ({}), assuming legacy protocol", e.what());
+    spdlog::warn("Hyprland IPC: could not read configProvider from systeminfo ({})", e.what());
+    return std::nullopt;
+  }
+}
+
+bool IPC::isLuaProtocol() {
```
Bart97, line 322: Are you sure that this function needs to be split into 3? In this case I feel like it increases complexity and makes code harder to read. I suppose it makes it slightly easier to write UTs, but imo it's not a significant difference.

Bart97 (changes requested):

include/modules/hyprland/backend.hpp
CristianMz21: Agreed — renamed to `isLuaConfigProvider` in 92742e4, so the bool it returns is the question the name asks.

src/modules/hyprland/backend.cpp
CristianMz21: Made it an explicit branch in 92742e4: `lineEnd == npos ? substr(valuePos) : substr(valuePos, lineEnd - valuePos)`. You are right that the clamping behaviour is guaranteed by the standard, but the explicit form reads as intent instead of as a coincidence that happens to be well-defined.

src/modules/hyprland/backend.cpp
CristianMz21: Better than sufficient — switched to `trim()` from `util/string.hpp` in 92742e4. It also absorbs the leading separator whitespace and the trailing CR, which removed the hand-rolled `find_first_not_of` step and the pop_back loop entirely.

src/modules/hyprland/backend.cpp
CristianMz21: You are right, and the history explains how it got there: the tri-state existed to route "field absent" to the version fallback. Once Mrpaoo's comment removed that fallback, nullopt and false converged on the same path and the optional answered nothing the caller could act on. Dropped in 92742e4 — plain bool, absent field documented as "pre-Lua instance, hence legacy".

src/modules/hyprland/backend.cpp
CristianMz21: Collapsed to two in 92742e4: `isLuaProtocol` (socket + cache + logging) and `isLuaConfigProvider` (pure parsing). I kept that one seam deliberately: CI has no compositor, so the parsing paths — absent field, legacy value, formatting variations — are only reachable through a function that takes the reply as a string. Folding it into `isLuaProtocol` would make those branches untestable rather than simpler. If you still prefer a single function I will fold it, but this is the trade-off as I see it.

Bart97:
Changes look fine to me now, for some reason github doesn't let me mark my comments as resolved.

CristianMz21:
Thanks @Bart97 — resolved all six threads. You could not do it yourself because resolving requires either write access or PR authorship, and you have neither here; as the author I can, so that is on me rather than a GitHub glitch.

One correction went in on top, found while re-verifying the branch rather than reported by anyone: the fixtures asserted on `configProvider: legacy`, which Hyprland never emits. `Config::typeToString` returns:

```cpp
case CONFIG_LUA:    return "lua";
case CONFIG_LEGACY: return "hyprlang";
```

Runtime behaviour was unaffected — the check is `trim(value) == "lua"`, so every non-lua value already fell through to legacy — but the test covering the exact case this PR exists to fix, a >= 0.55 instance started with a traditional `hyprland.conf`, asserted on a string that cannot occur. So the headline regression was not actually exercised. Fixed in 0f12de9, verified against `ConfigManager.cpp` at v0.55.0, v0.56.0 and v0.56.1. Tests still pass (64 assertions, 15 cases).

Two things that need someone with write access:

1. **Workflow approval.** All five runs on head `0f12de9` are parked at `action_required` under the first-time-contributor gate. The previous head was green across all 10 checks, so this is purely the per-push gate, not a regression.
2. **@Bart97**, if you are willing — converting your "changes look fine" into a formal Review → Approve would clear the red *Changes requested* badge. Your `CHANGES_REQUESTED` from 2026-08-04 is still your latest formal review, so the PR renders as blocked even though `reviewDecision` is `null`.

No further pushes planned from my side, so the next approval should be the last one needed.

Bart97 (approved):

TanishqPokharia:
any updates on this?

PR #5238, opened 2026-08-07
fix(tray): fall back when dbus menu is unavailable

nanomatters:
**What does this PR do?**
Some SNI items advertise a menu object path without exporting the com.canonical.dbusmenu interface. Creating a GTK dbus menu for those items prevents their native context menu from being used.

Probe the mandatory dbusmenu "Version" property and only create the GTK menu after the probe succeeds. Until then, keep the existing ContextMenu fallback so startup remains non-blocking and ordinary tray actions are unchanged.

This example below is when running Windows native Steam which renders its own menu.

image

**Related issues**

**Checklist**
- [X] Code is formatted with `clang-format`
- [X] Builds locally (`ninja -C build`)
- [ ] Man page updated for any new/changed user-facing option (`man/`)
- [X] Tested against the affected module(s)

### DIFF
--- a/include/modules/sni/item.hpp
+++ b/include/modules/sni/item.hpp
@@ -94,6 +94,8 @@ class Item : public sigc::trackable {
   Glib::RefPtr<Gdk::Pixbuf> getIconByName(const std::string& name, int size);
   double getScaledIconSize();
   static void onMenuDestroyed(Item* self, GObject* old_menu_pointer);
+  void validateMenu();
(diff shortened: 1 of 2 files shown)


nanomatters:
The clang-format is failing but it is not a file I modified.

./src/modules/backlight.cpp:127:55: error: code should be clang-formatted [-Wclang-format-violations]
      backend.set_scaled_brightness(preferred_device_, static_cast<int>(std::round(min_brightness)));

PR #5249, opened 2026-08-12
privacy: disconnect pending timeouts on destruction

shivankgarg98:
**What does this PR do?**

Disconnect the privacy module's pending GeoClue and visibility timeout
connections when the module is destroyed.

The timeout callbacks are attached to GLib's main context. During a
`SIGUSR2` reload, the `Privacy` object can be destroyed while the GeoClue
timeout remains pending. If it runs afterwards, it dereferences freed
memory and Waybar crashes in `Privacy::locationTimeout()`.

Add a destructor that disconnects both timeout connections before the
object is torn down.

The unpatched binary crashed after four reloads spaced 1.5 seconds apart.
The patched binary survived ten reloads without a new coredump.

**Related issues**

None.

**Checklist**
- [x] Man page updated for any new/changed user-facing option (`man/`) — N/A, no user-facing change

### DIFF
--- a/include/modules/privacy/privacy.hpp
+++ b/include/modules/privacy/privacy.hpp
@@ -17,6 +17,7 @@ namespace waybar::modules::privacy {
 class Privacy : public AModule {
  public:
   Privacy(const std::string&, const Json::Value&, Gtk::Orientation, const std::string& pos);
+  ~Privacy() override;
   auto update() -> void override;
 
  private:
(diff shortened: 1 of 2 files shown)


shivankgarg98:
`./src/modules/backlight.cpp:127:55: error: code should be clang-formatted [-Wclang-format-violations]`

this issue seem unrelated as I'm touching the Privacy Module only.

PR #5258, opened 2026-08-17
feat: add num-tags for mango/workspaces

cebem1nt:
**What does this PR do?**
Adds num-tags option to mango/workspaces similar to dwl/tags 

**Related issues**
None

**Checklist**

~~Little note. It also removes previous buttons sorting based on tags because while testing, tags were always pre-sorted.~~


cebem1nt:
Turns out there was an option for it in mango...

PR #5264, opened 2026-08-21
ext/worksapces: sort buttons when workspace changes group

llyyr:
handle_workspace_enter/leave never called set_needs_sorting. As a result, moving a workspace to a different output's group appended new buttons to the end of the box instead of placing them in sorted order.

**Checklist**


llyyr:
Ping on this

PR #5274, opened 2026-08-24
feat/hyprland set workspace click action

### LINKED ISSUE #2832: Feature Request: Disable on-click on Workspaces
Hey there :-)

I would like an option to disable the default on-click behaviour of the workspaces modules (I'm using the Hyprland one), as I control all my workspace switching via hotkeys and am worried about accidently switching to a different workspace with a click whilst I'm streaming on Twitch.

Also, thanks to all the devs for this project, best bar available imo.

daher13:
**What does this PR do?**
Add option to set click action on Hyprland workspaces
Just set on hyprland/workspaces:
"on-click": "disabled" to disable the workspaces click action
Default means no action is executed

**Related issues**
Closes #2832

**Checklist**

### DIFF
--- a/include/modules/hyprland/workspaces.hpp
+++ b/include/modules/hyprland/workspaces.hpp
@@ -48,6 +48,10 @@ class Workspaces : public AModule, public EventHandler {
   auto taskbarWithIcon() const -> bool { return m_taskbarWithIcon; }
   auto barScroll() const -> bool { return m_barScroll; }
 
+  std::string onClick() const { return m_onClick; }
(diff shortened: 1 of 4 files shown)


opennomad:
would it make sense to re-use the  the `on-click` and set it to `disabled` rather than introducing a new parameter? this could then be mirrored to `on-click-middle` and `on-click-right`.

daher13:
> would it make sense to re-use the the `on-click` and set it to `disabled` rather than introducing a new parameter? this could then be mirrored to `on-click-middle` and `on-click-right`.

Good idea. Done! Thanks!

PR #5287, opened 2026-08-31
sni/watcher: accept combined "bus_name/object_path" item registrations (Electron 43.4.1+)

### LINKED ISSUE #5240: tray: "Invalid Status Notifier Item" for Electron 43+ SNI items using well-known bus names
> **Update 2026-08-28 — Upstream Electron bug, not Waybar**
>
> This is **not** a Waybar/GDBus bug as originally diagnosed — thanks to @marcioflaviof and @boredland for narrowing it down.
>
> **Root cause:** Electron 43.3+ registers SNI under `org.freedesktop.StatusNotifierItem-PID-N` but its `Properties.GetAll/Get` handler only answers on that well-known name, not on its unique name `:1.XXX`. Waybar's `Gio::DBus::Proxy` (correctly) resolves to the unique name -> `get_cached_property_names() == []` -> `Invalid Status Notifier Item`.
>
> Upstream: `electron/electron#52674` — last good `43.2`. Waybar broken from `43.3.0` through `43.4.x` (e.g. Slack `43.4.0` fails on Waybar, see https://github.com/Alexays/Waybar/issues/5240#issuecomment-5450332913). On my Omarchy/Quickshell system `43.4.0-1` is still fine, latest `43.4.1-1` breaks — Quickshell's tray proxy apparently doesn't hit the unique-name path the same way.
>
> Verified: `Ping`/`Introspect` OK on both names, `GetAll` FAIL on `:1.XXX` only.
>
> Keeping open as tracker for Waybar workaround (`Item::proxyReady()` fallback/raw call or keep proxy on well-known name). See comments for workarounds.
>
> ---
> *Original report follows:*

**Environment:**
- Waybar: `0.15.0.r959.g084d874-1` (waybar-git)
- Electron: `43.3.0`
- Hyprland: `0.56.2`
- Arch Linux, kernel `7.1.6`

**Reproduction:**
1. Install Obsidian (`obsidian` package, uses Electron 43) with the "Tray" community plugin v0.3.5 by dragonwocky
2. Enable the tray icon in the plugin settings
3. Observe: no tray icon appears in Waybar

**What happens:**
Waybar logs:
```
[error] Invalid Status Notifier Item: org.freedesktop.StatusNotifierItem-<PID>-<N>, /StatusNotifierItem
```

The item is registered on D-Bus under a **well-known** bus name (`org.freedesktop.StatusNotifierItem-PID-N`), which is Electron 43+'s new SNI naming convention. `g_dbus_proxy_get_cached_property_names()` returns **empty** for this well-known name, so `proxyReady()` sees no `Id` or `Category` and invalidates the item.

**Why it's a Waybar/GDBus issue (not Electron):**
The properties are fully accessible via raw D-Bus calls — `busctl get-property` and `dbus-send GetAll` both return valid `Id`, `Category`, `IconPixmap`, etc. The problem is specifically that GDBus's proxy caching fails for well-known names of the form `org.freedesktop.StatusNotifierItem-*`.

**Why other tray icons work:**
Electron ≤39 and other apps use unique connection names like `:1.724/StatusNotifierItem`. GDBus caches properties fine for those. Electron 43+ changed to well-known names, exposing this GDBus proxy bug.

**Possible fix:**
`proxyReady()` in `src/modules/sni/item.cpp` should fall back to fetching `Id` and `Category` via raw `g_dbus_connection_call_sync` when `get_cached_property_names()` returns empty.

teivng:
Fixes #5240.

## Problem

Electron ≥ 43.4.1 (Chromium 150.0.7871.224) changed how its tray icon registers with the StatusNotifierWatcher: instead of a bare object path it now sends a combined `bus_name/object_path` string. Captured on the bus (`busctl --user monitor org.kde.StatusNotifierWatcher`), Signal-Desktop 8.25.0:

```
Member=RegisterStatusNotifierItem
  STRING "org.freedesktop.StatusNotifierItem-36777-1/StatusNotifierItem/1";
```

`Watcher::handleRegisterItem` only understands two shapes — a string starting with `/` (object path, resolved against the sender) or a plain bus name — so it runs `g_dbus_is_name()` on the whole combined string and rejects it:

```
ErrorName=org.freedesktop.DBus.Error.InvalidArgs
ErrorMessage="D-Bus bus name 'org.freedesktop.StatusNotifierItem-36777-1/StatusNotifierItem/1' is not valid"
```

On rejection Electron releases its bus name and destroys the tray icon permanently (it does not retry, not even when the watcher restarts), so the app shows no tray icon for the rest of its lifetime. Side-by-side comparison of the same app across the Electron bump:

| | Signal 8.23.0 (Electron 43.0.0) | Signal 8.25.0 (Electron 43.4.1) |
|---|---|---|
| Registers as | `/org/chromium/StatusNotifierItem/1` | `org.freedesktop.StatusNotifierItem-36777-1/StatusNotifierItem/1` |
| Result | accepted | rejected → icon torn down |

Slack (Electron 43.4.0) still registers with a bare bus name and works, so this will progressively hit every Electron app as they update past 43.4.1. KDE's StatusNotifierWatcher accepts the combined form.

## Fix

When the service string contains a `/` but does not start with one, split at the first `/` into bus name and object path. The two existing forms are handled exactly as before. As a bonus, `g_bus_watch_name()` now watches the actual bus name for these items, so they are pruned correctly when the owner exits.

## Testing

Running this patch on waybar 0.15.0 under Hyprland: Signal 8.25.0's icon registers, appears in the tray, and survives repeated window close/reopen (close-to-tray) cycles; Slack, Telegram, Bitwarden, udiskie and nm-applet items are unaffected.

### DIFF
--- a/src/modules/sni/watcher.cpp
+++ b/src/modules/sni/watcher.cpp
@@ -87,10 +87,16 @@ gboolean Watcher::handleRegisterItem(Watcher* obj, GDBusMethodInvocation* invoca
                                      const gchar* service) {
   const gchar* bus_name = service;
   const gchar* object_path = "/StatusNotifierItem";
+  g_autofree gchar* bus_name_buf = nullptr;
 
   if (*service == '/') {
     bus_name = g_dbus_method_invocation_get_sender(invocation);
     object_path = service;
+  } else if (const gchar* slash = g_strstr_len(service, -1, "/")) {
+    // Electron >= 43.4.1 registers with a combined "bus_name/object_path" string
+    bus_name_buf = g_strndup(service, slash - service);
+    bus_name = bus_name_buf;
+    object_path = slash;
   }
   if (g_dbus_is_name(bus_name) == FALSE) {
     g_dbus_method_invocation_return_error(invocation, G_DBUS_ERROR, G_DBUS_ERROR_INVALID_ARGS,


felipelube:
Confirming this fixes Slack's missing tray icon on my machine.

Environment:
- OS: Debian GNU/Linux 13 (trixie), kernel 6.12.107
- Compositor: Hyprland 0.55.2
- Waybar: v0.12.0 + this fix (built locally)
- Slack: 4.52.155 (Electron 44.0.0)

danielkec:
• Thanks for the fix! I can confirm this restores Slack’s tray icon on my setup:

  - Ubuntu 26.04.1 LTS
  - Hyprland 0.53.3
  - Waybar 0.15.0 with this patch
  - Slack 4.52.155 (Electron 44.0.0)

PR #5293, opened 2026-09-01
fix(tray): create the dbus menu once the probe succeeds

### LINKED ISSUE #5285: Context menu in tray module renders empty on first click, works normally on second click
**Describe the bug**
When attempting to open an application's context menu from the tray module, the menu opens but appears broken with no visible options inside.
If I open it a second time, the menu renders perfectly fine with all options visible. However, if I restart the tray application, the bug returns and the menu breaks on the first click again.

**To reproduce**
1. Start Waybar with the tray module enabled.
2. Launch an application that uses a tray icon.
3. Right click tray icon to open its context menu. Observe that the menu is broken.
4. Close the menu and click it a second time. Observe that it now renders correctly.

```jsonc
{
    "layer": "top",
    "modules-right": ["tray"],
    "tray": {
        "spacing": 10
    }
}
```

**Expected behaviour**
The context menu should render at full size with all options visible.

**Screenshots**
`nm-applet`
Image
`telegram-desktop`
Image

**Environment**
- Waybar v0.15.0-1004-g6d60c8e0 (branch 'master')
- AUR `waybar-git` 
- Hyprland 0.56.2

**Debug output**
Run `waybar -l debug` and paste any relevant output:

```
(waybar:26688): LIBDBUSMENU-GLIB-CRITICAL **: 16:38:40.314: dbusmenu_menuitem_send_about_to_show: assertion 'DBUSMENU_IS_MENUITEM(mi)' failed
*** BUG ***
Set a breakpoint on '_pixman_log_error' to debug

*** BUG ***
Set a breakpoint on '_pixman_log_error' to debug

*** BUG ***
Set a breakpoint on '_pixman_log_error' to debug
```

Motaphe:
**What does this PR do?**

Tray context menus open empty on the first click and only work on the second.

b58018bc swapped the `makeMenu()` call in the `Menu` property handler for an async probe, but `menuProbeReady()` never calls `makeMenu()`. `handleClick()` is now the only call site, so the menu gets created and popped in the same block. Dbusmenu loads its layout asynchronously, so the first click gets a menu with no root item yet:

```
LIBDBUSMENU-GLIB-CRITICAL: dbusmenu_menuitem_send_about_to_show: assertion 'DBUSMENU_IS_MENUITEM(mi)' failed
*** BUG *** In pixman_region32_init_rect: Invalid rectangle passed
```

That empty menu is the small blank popup users see. The second click works because the layout has arrived.

It is a re-regression of 01ad3d96 ("fix(tray): pre-create dbusmenu for tray items", 2019), which fixed the same assertion by creating the menu at property-set time.

Calling `makeMenu()` when the probe succeeds restores that timing. Items that do not export `com.canonical.dbusmenu` still fail the probe and still fall back to `ContextMenu`.

The diff is @awsms's from #5285, credited in the commit. They also suggested reverting b58018bc outright. I have no strong view, happy to switch this to a revert if you prefer.

**Related issues**

Closes #5285

**Checklist**
- [ ] Man page updated. Not applicable, no user-facing option changed.

**Testing**

Hyprland, three SNI items (KDE Connect, Nextcloud, an Electron app).

Before: empty popup on the first click, full menu on the second, with the two log lines above.
After: full menu on the first click, log clean.

### DIFF
--- a/src/modules/sni/item.cpp
+++ b/src/modules/sni/item.cpp
@@ -593,6 +593,9 @@ void Item::menuProbeReady(Glib::RefPtr<Gio::AsyncResult>& result, const std::str
   try {
     proxy_->get_connection()->call_finish(result);
     has_dbus_menu_ = true;
+    // Dbusmenu synchronizes its layout asynchronously. Create the GTK menu as soon as the probe
+    // succeeds so it has time to populate before the first click.
+    makeMenu();
   } catch (const Glib::Error&) {
   }
 }


etherb0x:
Experienced this myself just now, and it looks like a few people have hit the same path and had Claude or somesuch land on the same fix.

Apps I saw the slow (empty-then-populated) right-click menu on: fcitx5, qpwgraph, SVP Manager, which all use a Qt `/MenuBar` dbusmenu.

Build is `Waybar v0.15.0-1004-g6d60c8e0` (master).

bokicoder:
> They also suggested reverting https://github.com/Alexays/Waybar/commit/b58018bc64ad3248149afc1c959ecf69c3173901 outright. I have no strong view, happy to switch this to a revert if you prefer.

This really needs to be reverted. It also causes `blueman‑tray`'s right‑click menu to fail to show up, an issue this PR does not resolve. `blueman‑tray` is likely not the only affected tray application; other tray programs may be broken in the same way.

Motaphe:
Closing this out. #5327 landed the same fix and #5315 went further by dropping the Version probe entirely, which also covers the blueman-tray case raised here. Confirmed working on my current build. Thanks all.

Issue #3475, opened 2024-07-25
Execute custom command on entering/exiting waybar area 


cmihail:
Hello. I have the following setup with 2 overlay waybar bars:

```
// -*- mode: jsonc -*-
[{
    "name": "minimal_bar",
    "include": "~/.config/waybar/modules.jsonc",
    "layer": "top",
    "mode": "top",
    "exclusive": false,
    "passthrough": false,
    "height": 30,
    "width": 10, // Force minimum width
    "spacing": 2, // Gaps between modules
    "reload_style_on_change": true,
    "modules-center": [
        "clock",
        "custom/notifications",
        "battery"
    ]
}, {
    "name": "full_bar",
    "include": "~/.config/waybar/modules.jsonc",
    "layer": "top",
    "mode": "top",
    "exclusive": false,
    "passthrough": false,
    "height": 30, // Waybar height (to be removed for auto height)
    "width": 750, // Waybar width
    "spacing": 2, // Gaps between modules
    "reload_style_on_change": true,
    "start_hidden": true,
    "ipc": true,
    "modules-left": [
        "hyprland/workspaces"
    ],
    "modules-center": [
        "clock",
        "custom/notifications",
        "battery"
    ],
    "modules-right": [
        "tray",
        "pulseaudio#microphone",
        "pulseaudio",
        "backlight",
        "power-profiles-daemon",
        "custom/power"
    ]
}]
```

The full bar starts hidden and it is displayed when clicking on the clock in minimal bar or via a keyboard binding by executing `killall -SIGUSR1 waybar`.

While this works, I would prefer to not have to click on clock for this action and have the full bar displayed when I enter with the mouse in the minimal bar area, and have the full bar hidden when I exit the full bar area.

From what I see these events are already handled at https://github.com/Alexays/Waybar/blob/master/src/AModule.cpp#L125 in order to change the cursor type, so adding custom actions shouldn't be too hard, maybe with a config in the bar like:

```
{
  "name": <bar_name>,
  .......
  "on-mouse-enter": <custom_action>,
  "on-mouse-exit": <custom_action>
}
```

Does this sound good? I'd be willing to implement this myself if it is fine.

PR #5312, opened 2026-09-08
feat(modules): add on-enter and on-leave events

### LINKED ISSUE #2002: [on-hover] Addition of on-hover event
Is it possible to have on-hover Event added?
as this will open many possibility.

someonewithpc:
**What does this PR do?**

Adds `on-enter` and `on-leave` command hooks to `AModule`, so any module can run something when the pointer arrives and leaves.

`AModule` already receives GTK's enter- and leave-notify events. Today they drive the prelight state and the pointer cursor, and nothing else can hang off them. Issue #2002 has asked for this since 2023.

Dispatch follows `on-scroll-*`. Module actions go first through `doAction()`, then the user command. A module that only wants the CSS `:hover` state sees no change, since nothing runs unless the option is set.

Both options resolve in the constructor rather than on every crossing. Beyond avoiding repeated work, that matters to anyone backporting this to a build older than ae119543, where a `Group` holds a dangling `config_` and a JSON lookup from a crossing handler becomes a use-after-free. jsoncpp's type assert then throws inside a GTK signal handler and takes the bar down. I hit exactly that developing against 0.15.0.

A value naming a built-in action has already run through `doAction()`, so it does not also run as a shell command. `handleUserEvent` has carried that guard since #3284, and `handleCrossingEvent` mirrors it.

Two kinds of spurious crossing need handling. When the pointer moves between the event box and a child that owns a `GdkWindow`, such as a workspace button or a group's box, GTK sends a leave/enter pair carrying `GDK_NOTIFY_INFERIOR`, and `group.cpp` already guards its reveal logic against those. Pointer grabs, a menu popping up for instance, add further pairs. Inferior crossings are ignored and the rest latched on a `hovered_` flag, so each command runs once per hover.

A throw from a crossing handler would abort the bar, so the dispatch logs the module name and carries on, for the same reason `handleUserEvent` catches `fmt::format_error`.

`Group` overrides both handlers without chaining up to `AModule`'s, so it calls the shared helper directly.

One thing I left alone: `pid_children_` gains an entry per forked command and is never reaped, which `on-click` and `on-update` already do today. Hover fires far more often than clicking does, so if you want that bounded, say the word and I will do it here or in a separate PR.

**Naming, worth deciding before this merges**

This uses `on-enter` and `on-leave`, matching the existing per-module vocabulary (`on-click`, `on-scroll-up`, `on-update`) and GTK's own enter-notify and leave-notify names.

PR #5280 and issue #3475 instead propose `on-mouse-enter` and `on-mouse-exit` for the bar-level equivalent. If both land, Waybar ends up with two names for one idea and config files carry both spellings indefinitely. I would rather you pick than have me quietly match an unmerged PR. Renaming is small and I am happy to do it either way.

**Man pages**

The docs commit touches 33 files because the option applies to every module and there is no shared page for options common to all of them. Happy to trim that to a smaller set, or to add a shared page and point the module pages at it, if you would prefer either.

**Testing**

No automated test. Crossing events come from GTK, and the suite covers utils and Hyprland IPC.

Verified by hand in a nested headless sway driving a synthetic pointer. Commands fire once per hover across many modules, through two levels of group nesting, and across rapid enter and leave transitions, with no duplicate or missed runs. The #3284 guard was checked against a control on the same pointer path: 40 shell runs with the `actions` entry removed, 0 with it in place.

**Related issues**

Closes #2002
Related: #3475, #5280

**Checklist**

---

### DIFF
--- a/include/AModule.hpp
+++ b/include/AModule.hpp
@@ -92,6 +92,11 @@ class AModule : public IModule {
   virtual bool handleToggle(GdkEventButton* const& ev);
   virtual bool handleMouseEnter(GdkEventCrossing* const& ev);
   virtual bool handleMouseLeave(GdkEventCrossing* const& ev);
+  // Called from AModule's handleMouseEnter and handleMouseLeave. A subclass
+  // that overrides those without chaining up gets no on-enter or on-leave
+  // unless it calls this itself, as Group does.
+  void handleCrossingEvent(GdkEventCrossing* const& ev, bool entering);
+
   virtual bool handleScroll(GdkEventScroll*);
   virtual bool handleRelease(GdkEventButton* const& ev);
 
@@ -110,9 +115,14 @@ class AModule : public IModule {
   const bool isTooltip;
   const bool isExpand;
   bool hasUserEvents_;
+  bool hovered_{false};
   gdouble distance_scrolled_y_;
   gdouble distance_scrolled_x_;
   sigc::connection cursor_timeout_conn_;
+  // Resolved in the constructor: jsoncpp throws on a non-object value, and an
+  // exception escaping a GTK signal handler aborts the bar.
+  const std::string on_enter_cmd_;
+  const std::string on_leave_cmd_;
   static const inline std::map<std::pair<uint, GdkEventType>, std::string> eventMap_{
       {std::make_pair(1, GdkEventType::GDK_BUTTON_PRESS), "on-click"},
       {std::make_pair(1, GdkEventType::GDK_BUTTON_RELEASE), "on-click-release"},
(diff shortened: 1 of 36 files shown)


rolfen:
I would like this to take actions (such as 'focus', 'maximize', etc.) in the `wlr-taskbar` module, similarly to the 'on-click' events.

I would also like it to have a configurable optional delay before being activated.

The aim would be, in stacked desktop environments such as labwc, to drag a file icon (for example) to a button on the taskbar and after hovering there for a short while, it would restore (if minimized) and focus the corresponding window, so that the file icon can be dragged there.

PR #5315, opened 2026-09-09
fix(tray): accept dbusmenus without the Version property

khaneliman:
**What does this PR do?**

Fixes tray menus that provide a working `GetLayout` method but do not export the `Version` property, including `blueman-tray`.

#5327 fixed the ordinary empty-first-click case by calling `makeMenu()` after the `Version` probe succeeds. This PR is rebased onto master `4c495a41`, which includes that fix. The remaining problem is the probe itself: if `Version` is unavailable, Waybar never creates a GTK menu, even when `GetLayout` works.

A fresh live `busctl` check of `blueman-tray` confirmed that `GetLayout` works, `Get com.canonical.dbusmenu Version` fails, and the SNI object has no `ContextMenu` method. The probe therefore rejects its usable menu, and the fallback cannot help. This is the gap @bokicoder reported on #5293.

This change:

- Creates the menu when the `Menu` property arrives, retaining early asynchronous layout loading without the `Version` gate.
- Checks `dbusmenu_client_get_root()` before displaying the GTK menu. If no layout root has arrived, it calls the item's `ContextMenu` instead of opening an empty GTK menu.
- Preserves the #5238 fallback for items that advertise a `Menu` path without serving `com.canonical.dbusmenu`, including the Steam case that motivated it.
- Removes `validateMenu()`, `menuProbeReady`, and `has_dbus_menu_`.

**Related issues**

Follow-up to #5327 and #5238. Addresses the remaining compatibility problem reported on #5293. The original first-click report is #5285; #5327 now covers that ordinary case.

**Testing**

Fresh checks after rebasing:

- `ninja -C build -j 4`: passed.
- `make test`: passed, 3 Meson tests (`waybar`, `utils`, `hyprland`), 0 failures.
- A focused libdbusmenu client probe loaded the live Blueman menu with a root and 14 children despite the missing `Version` property.
- The same probe returned no root for a nonexistent menu path, matching the fallback condition. This does not replace a visual click test.
- Live `blueman-tray`: `GetLayout` succeeds; the `Version` query fails with "No such property"; SNI introspection lists no `ContextMenu` method.
- `clang-format --dry-run --Werror include/modules/sni/item.hpp src/modules/sni/item.cpp`: passed.
- Both changed files are byte-for-byte identical to the previously tested PR head `632e39e6`; the rebase only resolves the probe callback changed by #5327.

The runtime results below are from the original validation, not a fresh run against the rebased branch. In this table, "this PR" means the original validation.

Headless harness: sway (headless backend) + private session bus + a pygobject fake StatusNotifierItem serving a two-entry libdbusmenu server. Clicks used a `zwlr_virtual_pointer_v1` client; screenshots used grim.

| Build | Item | Click 1 | Click 2 | waybar log |
|---|---|---|---|---|
| master 6d60c8e0 | dbusmenu | empty popup | full menu | assertion + pixman errors |
| master 6d60c8e0 | Menu path, no dbusmenu | ContextMenu called | ContextMenu called | clean |
| this PR | dbusmenu | full menu | full menu | clean |
| this PR | Menu path, no dbusmenu | ContextMenu called | ContextMenu called | one libdbusmenu "Getting layout failed" warning |

<details>
<summary>Original first-click regression evidence, before #5327</summary>

The original #5285 regression started when b58018bc (#5238) replaced `makeMenu()` in the `Menu` property handler with an async `Version` probe. The callback did not create the menu. `handleClick()` became the only call site. GTK displayed the menu while libdbusmenu was still fetching its layout over D-Bus:

```
LIBDBUSMENU-GLIB-CRITICAL: dbusmenu_menuitem_send_about_to_show: assertion 'DBUSMENU_IS_MENUITEM(mi)' failed
*** BUG *** In pixman_region32_init_rect: Invalid rectangle passed
```

This repeated the regression fixed by 01ad3d96 (2019), which created the menu at property-set time. The same assertion was observed on the first right click of a real tray item with the then-running master build on Hyprland 0.56.2.

#5293 proposed restoring `makeMenu()` in the probe callback. #5327 has now merged that approach. Neither removes the `Version` requirement addressed here.

</details>

**Checklist**

- [ ] Man page updated for any new/changed user-facing option (`man/`) - not applicable, no user-facing option changed
- [x] Tested against the affected module(s), with historical runtime evidence and fresh checks distinguished above


khaneliman:
@Alexays if you get a chance. Finally looked into this bug I was seeing recently.

daviesjamesdaniel:
Independently root-caused the same bug (before finding this PR or #5293), so sharing what I gathered in case it's useful corroboration.

Real `WAYLAND_DEBUG=1` trace against unpatched master, same tray item, first click (glitched) vs. second click (fine):

```
# First click - positioner created/configured for an essentially empty menu
-> xdg_positioner#45.set_size(50, 8)
   xdg_popup#47.configure(1326, 22, 50, 8)

# Second click, moments later - correctly sized
-> xdg_positioner#48.set_size(98, 56)
   xdg_popup#46.configure(1326, 21, 98, 56)
```

Waybar's own stderr at the exact moment of the glitched (50x8) open:

```
*** BUG ***
```

This correlated exactly with the visible glitch across every test run I did - present on every bad open, absent on every good one - alongside the `dbusmenu_menuitem_send_about_to_show` assertion already noted in this PR's description.

Also independently confirmed the specific case this PR's fallback addresses matters in practice: Steam's tray `Menu` object does answer `com.canonical.dbusmenu` introspection (`GetLayout`, `AboutToShow`, etc. all present via `busctl introspect`), but its menu (~16 items) took noticeably longer than a trivial 2-item menu to actually populate after creation - so whatever fix lands here needs to tolerate that variance, not just a fixed short delay. Sounds like early creation (this PR) sidesteps that by giving it a much longer head start than click time ever would.

I had a different (more defensive, generic-polling) fix up as #5322, closing that now in favor of this one - it's the more precise fix since it addresses the actual regression (missing `makeMenu()` after b58018bc) rather than working around the race generically.

PR #5324, opened 2026-09-14
fix(hyprland): Support addressable workspace IPC

### LINKED ISSUE #5316: hyprland/workspaces won't work on the next release of Hyprland
**Describe the bug**

Since hyprwm/Hyprland#16140, IPC format has changed, and Waybar no longer recognizes workspaces-related messages. (see https://github.com/hyprwm/Hyprland/commit/e9e2f64ccfd53c5e0729f8266608c937bffebf7c#diff-fe6d64bbfaa55edfaed2a1b19e085593a2050d07805c33d12159e7670701067b)

Other hyprland modules may be affected too.

**To reproduce**

Launch Hyprland and try switching between workspaces.

```jsonc
// -*- mode: jsonc -*-
{
    "modules-left": [
        "hyprland/workspaces"
    ]
}
```

**Environment**

- Waybar version (`waybar --version`): `Waybar v0.15.0-1005-gf9db22f4 (branch 'makepkg')`
- Installed from (distro package / source / other): source, #5284 
- Compositor and version (Sway, Hyprland, River, …): `Hyprland 0.56.0 built from branch main at commit 599ff9042c866dcad4bb458979f7be68a6d2ffef clean`

tonybutt:
Fixes #5316.

Hyprland removed the numeric workspace `id` from its IPC in hyprwm/Hyprland#16140, replacing it with a string `address` plus an explicit `type` (`"numbered"`, `"special"`, `"named"`). Waybar still reads `["id"].asInt()`, which now returns 0 for every workspace, so every workspace compares equal to every other:

- persistent workspaces are never created
- every client matches every workspace, so all window icons render on all of them

This matches the reports in #5316 ("both of space show as 1", "`No workspace with id 0`", and the `hyprctl reload` workaround that stops working on the next workspace switch).

## Does this overlap #5013 / #5043 / #5231?

No. Those fixed the *dispatch* path — how waybar sends commands to Hyprland under the Lua IPC protocol. This fixes the *parse* path — how waybar reads workspace identity out of the IPC payload. Different layers.

To confirm rather than assume, I built master at `2a12740b`, which contains all three, and ran it against Hyprland `main`:

| | master @ `2a12740b` | with this PR |
|---|---|---|
| `No workspace with id 0` warnings | 11 | 0 |
| workspaces collapsed onto id 0 | 7 | 0 |
| distinct workspaces created | 2 | 7 |

Master creates only `0` and `1`; with this PR the same session creates `1`, `6` and five special workspaces by name.

## Approach

A single pure function, `parseWorkspaceIdentity()`, normalises both payload shapes into `{address, kind, number}`. Detection is by field presence — `address` when present, otherwise the legacy `id` — so **no call site branches on compositor version and older Hyprland keeps working**.

Workspaces are then keyed by `address`, and their kind comes from `type`. That last part also removes the arithmetic in `sortWorkspaces()` that inferred kind from the id's sign and range (`-99` special, `-98..-1` named special, `<= -1377` named): the compositor now states kind explicitly, so waybar no longer has to reverse-engineer it.

Two commits: the parser (additive, with tests), then the migration. Both build and pass tests.

## Behaviour changes worth flagging

- **`{id}` renders the address.** For numbered workspaces the output is byte-identical. For named and special ones it replaces a negative sentinel (`-99`) with something meaningful. Man page updated.
- **`sort-by: id` groups numbered → named → special**, rather than ordering by id sign. Man page updated.
- **The generic `special` workspace no longer pins last** among specials — that behaviour was the `-99` magic number this change removes.

## Also fixed along the way

- A persistent workspace named literally `special` threw `std::out_of_range` from `substr(8)`.
- A persistent workspace whose selector matched an existing workspace by name (e.g. `special`, or a named workspace) was duplicated instead of deduped, and never got `persistent-config` applied.

## Testing

`test/hyprland/workspace_identity.cpp` and `test/hyprland/workspace_order.cpp` cover the parser on both schemas and the ordering axioms. The ordering functions live in `workspace_identity.cpp` rather than inline in `workspaces.cpp` specifically so they are reachable from the test target.

The orderings are strict weak orderings, which `std::ranges::sort` requires — violating that is UB. Neither assumes a `Numbered` workspace has a usable `number`, since an address can be labelled `numbered` and still not fit in an `int`; comparing such a workspace numerically against some peers and by name against others is what produces a comparison cycle.

Verified at runtime against Hyprland `main` (post-#16140): correct per-workspace icons, persistent workspaces created, no errors. The legacy path is covered by unit tests only, since the machine runs `main`.

`sortWorkspaces()` itself is still not directly unit-testable — it needs a live `Workspaces&` and `Gtk::Button`, so the extracted orderings are covered but the call site is not.

## Out of scope

`SortMethod::NUMBER` is also not a strict weak ordering — `std::stoi` throws on a non-numeric name and falls through to a name comparison. That predates this change and is untouched here, but it is in the same `switch` so it is worth naming.

Happy to rework any of this — the identity abstraction, the commit split, or the behaviour changes above. If there is a constraint or a use case I have not considered, or you would rather this were shaped differently, say so and I will adjust.

cc @kolayne @edeustua — you both mentioned in #5316 that you were planning to look at this, so flagging it to avoid duplicated effort.


src/modules/hyprland/workspace.cpp
```diff
@@ -55,9 +55,11 @@ Workspace::Workspace(const Json::Value& workspace_data, Workspaces& workspace_ma
...
   if (m_name.starts_with("name:")) {
     m_name = m_name.substr(5);
-  } else if (m_name.starts_with("special")) {
-    m_name = m_id == -99 ? m_name : m_name.substr(8);
+  } else if (m_identity.kind == WorkspaceKind::Special) {
     m_isSpecial = true;
+    if (m_name.starts_with("special:")) {
+      m_name = m_name.substr(8);
+    }
   }
```
kolayne, line 63: I could also have a special workspace whose _name_ is `special:123` (i.e., `hyprctl dispatch 'hl.dsp.workspace.toggle_special("special:123")'`). This will then cut the prefix off, won't it?

(same for `name:`)

src/modules/hyprland/workspace.cpp
```diff
@@ -230,10 +232,25 @@ bool Workspace::handleClicked(GdkEventButton* bt) const {
...
+    m_identity.address = value;
+  }
+}
+
 void Workspace::initializeWindowMap(const Json::Value& clients_data) {
   m_windowMap.clear();
   for (const auto& client : clients_data) {
-    if (client["workspace"]["id"].asInt() == id()) {
+    const auto clientWorkspace = parseWorkspaceIdentity(client["workspace"]);
+    if (clientWorkspace.has_value() && clientWorkspace->address == address()) {
```
kolayne, line 253: Should we issue a warning here if `!clientWorkspace.has_value()`?

src/modules/hyprland/workspace_identity.cpp
```diff
@@ -0,0 +1,147 @@
...
+int workspaceKindRank(WorkspaceKind kind) {
+  switch (kind) {
+    case WorkspaceKind::Numbered:
+      return 0;
+    case WorkspaceKind::Named:
+      return 1;
+    case WorkspaceKind::Special:
+      return 2;
+  }
+  return 1;
```
kolayne, line 113: Should be impossible? Log error here?

src/modules/hyprland/workspace_identity.cpp
```diff
@@ -0,0 +1,147 @@
...
+const char* workspaceTypeName(WorkspaceKind kind) {
+  switch (kind) {
+    case WorkspaceKind::Numbered:
+      return "numbered";
+    case WorkspaceKind::Special:
+      return "special";
+    case WorkspaceKind::Named:
+      break;
+  }
+  return "named";
```
kolayne, line 75: Log if the given `kind` is invalid?
```suggestion
    case WorkspaceKind::Named:
      return "named";
  }
  // ------> Log error here? <--------
  return "named";
```

src/modules/hyprland/workspace.cpp
kolayne: Even if I'm wrong with this one and you handle this in some other place, I'd argue:
1. This is a tricky corner case, so we should probably have tests for this (a special namespace whose name starts with `special:`, a named workspace whose name starts with `name:`, a special workspace whose name starts with `name:`, etc).
2. The fact that I read this piece of code and had this question indicates that these checks are confusing. Would be nice if we could get rid of these checks on workspace name completely in the main module code. Should we shift the workspace parsing logic to the workspace_identity's area of responsibility?

vp1981:
I gave it a try, and in my case waybar with this PR works as expected on Hyprland:
```
Hyprland 0.56.0 built from branch main at commit c26dbf9337998ba9afcfe84050e36969bb787598 clean (main: check XDG_RUNTIME_DIR later ( 16256)).
Date: Mon Sep 14 18:20:52 2026
Tag: v0.56.0-191-gc26dbf933, commits: 7816

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.2, system has 0.14.2
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.15.0, system has 0.15.0

Version ABI string: c26dbf9337998ba9afcfe84050e36969bb787598_aq_0.15_hu_0.14_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```

I'm not familiar with the codebase, but I think @kolayne's comments are worth considering.

src/modules/hyprland/workspace.cpp
tonybutt: Checked this, and the two halves land differently.

`special:` is fine as written — `substr(8)` removes exactly one prefix, so `special:special:123` becomes `special:123`, which is the name you'd want. (`isDoubleSpecial()` rejects those at `createworkspacev2` anyway, but `initializeWorkspaces()` doesn't, so it is reachable at startup.)

`name:` is a real bug, and worse than a stripping-too-much bug: it shouldn't be in this path at all. Hyprland never prepends `name:` to a reported name — it's selector syntax. The only thing producing a `name:`-prefixed name here is `createMonitorWorkspaceData()`, i.e. a `persistent-workspaces` entry. So a workspace genuinely called `name:foo` was being truncated to `foo` by selector handling that had leaked into the IPC path.

Both fixed in dc2c2c56 by the split you suggest below.

tonybutt: Agreed on both counts, and done in dc2c2c56.

`workspace_identity` now owns the mapping in both directions:

- `workspaceDisplayName(rawName, kind)` — for names Hyprland reports. Strips exactly one `special:`, and nothing else. Never touches `name:`.
- `parseWorkspaceSelector(selector)` / `workspaceRawName(selector)` — for selectors the user writes in `persistent-workspaces`, split into `{kind, name}` and back.

`createMonitorWorkspaceData()` runs the selector through that pair, so the placeholder payload is IPC-shaped and the `Workspace` constructor no longer knows selector syntax exists. The four remaining `starts_with("special")` / `substr(8)` sites in `workspaces.cpp` went the same way.

Writing the round-trip test surfaced something I'd otherwise have missed: `special` and `special:special` are different workspaces that display the same name, so the selector has to carry an `isGenericSpecial` flag or the mapping isn't invertible. That also replaced the `name() != "special"` sentinel in `handleClicked()`, which would have misrouted a special workspace the user named `special`.

Tests in `test/hyprland/workspace_identity.cpp` cover the cases you listed — `special:special:123`, `name:foo` as a real name, `special:name:foo`, `specialfoo` (not special), bare `special:` / `name:` prefixes, and the selector round-trip.

src/modules/hyprland/workspace.cpp
tonybutt: Yes — a client with no parseable workspace means malformed IPC and shouldn't be silent. Added in dc2c2c56, at `debug` rather than `warn`: `initializeWindowMap()` runs once per workspace, so one malformed client would otherwise be reported once per workspace.

src/modules/hyprland/workspace_identity.cpp
tonybutt: Both of these are gone in dc2c2c56, but by deleting the function rather than logging in it.

`workspaceKindRank()` was a hand-written table mapping `Numbered/Named/Special` to `0/1/2` — which is exactly `WorkspaceKind`'s own declaration order. Two copies of one fact, free to drift. It's now `constexpr int workspaceKindRank(WorkspaceKind kind) { return static_cast<int>(kind); }` in the header, with the enum documented as load-bearing. No table, no unreachable branch, no magic numbers.

I'd have pushed back on logging here regardless: this runs inside a sort comparator, so a log on that branch would fire O(n log n) times if it ever did execute.

src/modules/hyprland/workspace_identity.cpp
tonybutt: Took your shape for `workspaceTypeName()` — every case returns, so `-Wswitch` now fails the build if an enumerator is added without a case, which catches the mistake before it ships rather than at runtime. The trailing `return` stays only to satisfy `-Wreturn-type`, with a comment saying so.

Left the log out deliberately: the compiler is the stronger guard, and a log on a branch that can't execute is noise in the source for no runtime benefit. Happy to add one if you'd rather have belt and braces.

src/modules/hyprland/workspaces.cpp
```diff
@@ -84,28 +85,45 @@ Json::Value Workspaces::createMonitorWorkspaceData(std::string const& name,
...
+
+  if (workspace == m_workspaces.end()) {
+    // A persistent-workspaces entry is addressed by the selector the user
+    // wrote, which on legacy Hyprland never matches the live workspace's
+    // numeric address (a configured "special" against address "-99").
+    const auto selector = workspace_data["name"].asString();
+    workspace =
+        std::ranges::find_if(m_workspaces, [&selector](std::unique_ptr<Workspace> const& w) {
+          return workspaceSelectorMatchesName(selector, w->name());
+        });
```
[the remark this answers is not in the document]
tonybutt: Confirmed and fixed in dc2c2c56. `special:spotify` and a named `spotify` both display `spotify`, and the fallback compared display names only.

The match now requires equal `WorkspaceKind`, which separates them (selector `spotify` classifies Named, `special:spotify` Special), and is restricted to placeholders that haven't yet adopted a live identity, so it can't fire for two live workspaces at all.

src/modules/hyprland/workspaces.cpp
```diff
@@ -360,33 +382,28 @@ void Workspaces::onSpecialWorkspaceActivated(std::string const& payload) {
...
+  const auto [workspaceAddress, _] = splitDoublePayload(payload);
 
   auto const workspaceRules = m_ipc.getSocket1JsonReply("workspacerules");
   auto const workspacesJson = m_ipc.getSocket1JsonReply("workspaces");
 
   for (auto workspaceJson : workspacesJson) {
-    const auto currentId = workspaceJson["id"].asInt();
-    if (currentId == *workspaceId) {
+    const auto currentIdentity = parseWorkspaceIdentity(workspaceJson);
+    if (currentIdentity.has_value() && currentIdentity->address == workspaceAddress) {
```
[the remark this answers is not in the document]
tonybutt: The conclusion holds but not via this mechanism. `createworkspacev2` emits the numeric id, not the literal `special` — the pre-change code parsed it with `std::stoi` (`parseWorkspaceId`, `workspaces.cpp:1219` at 2a12740b). The literal `special` case in that helper existed for `persistent-workspaces` selectors, which `createWorkspace()` and `removeWorkspace()` already handled via the name fallback.

Where it did bite is `onWorkspaceRenamed()` and `onWorkspaceIdChanged()`, which matched address-only and had no fallback. dc2c2c56 routes all four through one `findWorkspace()` that tries the address and then the selector, so the legacy mapping is restored everywhere rather than in three of four places.

tonybutt:
Pushed dc2c2c56 addressing this round. Summary, including the finding that had no inline thread to reply to.

## Persistent placeholder kept its synthetic address

The one I'd call blocking, and the finding greptile raised outside the diff. Confirmed: `createMonitorWorkspaceData("name:web")` builds address `name:web`; when the live workspace appeared at address `web`, `createWorkspace()` missed on address, matched by display name, and returned early without adopting the live identity. Every later address-keyed lookup — clients, visibility, window counts, destroy — then missed the button on screen.

Bounded to named and special persistent entries: a numbered one (`"3"` → address `3`) matches by address and was never affected. It now adopts the live identity on match.

## Named and special workspaces conflated

`special:spotify` and a named `spotify` display the same name, and the fallback compared display names only. It now requires equal `WorkspaceKind` and is restricted to placeholders that have not yet adopted an identity, so it cannot fire between two live workspaces.

## Legacy `special` → `-99`

Real, though not by the route greptile described — `createworkspacev2` emits a numeric id, not the literal `special`. The gap was that `onWorkspaceRenamed()` and `onWorkspaceIdChanged()` matched address-only with no selector fallback. All four sites now share one `findWorkspace()`.

## Name parsing moved into `workspace_identity`

@kolayne's structural point, which turned out to be load-bearing rather than cosmetic. Separating "names Hyprland reports" from "selectors the user writes" made three latent bugs fall out:

- `name:` is selector syntax Hyprland never prepends to a reported name, so a workspace genuinely called `name:foo` displayed as `foo`. That check should never have been in the IPC path.
- A workspace merely named `specialfoo` was classified special.
- The generic special workspace was identified as "the one displaying `special`" — which is also what a special workspace the user *named* `special` displays. It now carries an explicit flag, which also replaced the `name() != "special"` sentinel in `handleClicked()`.

`special:` itself was already correct: `substr(8)` strips one prefix, so `special:special:123` → `special:123`.

## Magic numbers

`workspaceKindRank()`'s `0/1/2` table duplicated `WorkspaceKind`'s own declaration order — two copies of one fact, free to drift. It is now `static_cast<int>(kind)` with the enum documented as load-bearing, which also removes the unreachable branch @kolayne flagged. `workspaceTypeName()` keeps a switch, now exhaustive so `-Wswitch` catches a new enumerator at build time.

I left the logging out of both rather than adding it: `workspaceKindRank()` runs inside a sort comparator, and the compiler is the stronger guard than a runtime log on a branch that cannot execute. Say the word if you'd rather have both.

## Tests

`test/hyprland/workspace_identity.cpp` covers the corner cases @kolayne asked for: `special:special:123`, `name:foo` as a real name, `special:name:foo`, `specialfoo`, bare `special:` / `name:` prefixes, and a selector round-trip. That round-trip is what surfaced `special` vs `special:special` being distinct workspaces with identical display names — I would not have found it by inspection.

Full suite: 37 cases, 646 assertions, passing. Still verified at runtime against Hyprland `main`; the legacy path remains unit-test-only.

edeustua (approved):
Looks good to me.

include/modules/hyprland/workspace.hpp
```diff
@@ -34,11 +35,18 @@ class Workspace {
...
+  WorkspaceKind kind() const { return m_identity.kind; };
+  std::optional<int> number() const { return m_identity.number; };
   std::string name() const { return m_name; };
   std::string output() const { return m_output; };
   bool isActive() const { return m_isActive; };
   bool isSpecial() const { return m_isSpecial; };
+  // The special workspace with no name -- `togglespecialworkspace` with an
+  // empty argument. Not the same as a special workspace the user named
+  // `special`, which displays the same name but toggles by name.
+  bool isGenericSpecial() const { return m_isGenericSpecial; };
```
kolayne: > Not the same as a special workspace the user named `special`

Is it not, indeed?

At least on my version (built from https://github.com/hyprwm/Hyprland/commit/7ebf13), if I do `hyprctl dispatch 'hl.dsp.workspace.toggle_special("special")'` and `hyprctl dispatch 'hl.dsp.workspace.toggle_special()'`, the two dispatchers bring me to the same workspace.

<hr>

While we're at it, I also find this name a bit confusing. Maybe should be, e.g., AnonymousSpecial/NamelessSpecial/UnnamedSpecial?

src/modules/hyprland/workspace_identity.cpp
kolayne: Agreed, a logger in the comparator probably does not make sense.

src/modules/hyprland/workspace_identity.cpp
kolayne: The compiler is a stronger guard for sure, but the thing is that C++ does not require that the set of values of `WorkspaceKind` be the variants you defined (unlike, e.g., Rust enums). For example, both `WorkspaceKind::Special | WorkspaceKind::Named` and `123` [are](https://en.cppreference.com/cpp/language/enum) valid values of type `WorkspaceKind` in C++.

In particular, the compiler was complaining about `-Wreturn-type` before you added a catch-all return exactly because there are some valid values of `WorkspaceKind`, for which the function would end without a return (which would have been undefined behavior).

Now, your current code has well-defined behavior. Still, should something bad happen, it'll be pretty hard to debug, given that the catch-all return returns `"named"`, which matches one of the previous values. It should at least be something like `"unknown"`. We may also log a warning. We may alternatively abort waybar (🥲).

Notably, if these were Rust enums, it would have been exactly like you described: the compiler would have proven that no cases other than the three are possible at runtime, and would give you the opposite warning, telling that your catch-all return at the end of the function is unreachable.

include/modules/hyprland/workspace_identity.hpp
```diff
@@ -0,0 +1,105 @@
...
+std::optional<WorkspaceIdentity> parseWorkspaceIdentity(const Json::Value& workspace);
+
+// The literal Hyprland IPC `type` name for a kind. Inverse of the `type`
+// mapping `parseWorkspaceIdentity` applies.
+const char* workspaceTypeName(WorkspaceKind kind);
+
+// Classifies a bare address for the one path that has no IPC `type` field to
+// read: the `workspacev2`-style rename that moves a workspace between the
+// numbered and named namespaces.
+WorkspaceKind workspaceKindForAddress(const std::string& address);
```
kolayne, line 22: Should `workspaceKindForAddress` and `parseWorkspaceSelector` become just constructors of the corresponding classes?

Maybe some other functions can be converted to methods, too.

include/modules/hyprland/workspace_identity.hpp
```diff
@@ -0,0 +1,105 @@
...
+// The declaration order is load-bearing: it is the display order the
+// workspaces module groups by, which workspaceKindRank() reads off directly.
+enum class WorkspaceKind { Numbered, Named, Special };
+
+// A workspace's stable key plus its kind, normalized across Hyprland IPC
+// schema versions. `number` is set only for Numbered workspaces.
+struct WorkspaceIdentity {
+  std::string address;
+  WorkspaceKind kind = WorkspaceKind::Named;
+  std::optional<int> number;
```
kolayne: One thing to consider is if we can change this structure such that it does not contain redundancy. For instance, store all the information that is needed to build the address, and then only generating address when requested, rather than storing it.

It would be ideal if the structure was also such that any representable value is logically valid... We can probably do this with `std::variant` (to represent that every workspace has either a name or a number). Is the project's C++ standard recent enough to use `std::variant`?

include/modules/hyprland/workspaces.hpp
```diff
@@ -190,7 +193,7 @@ class Workspaces : public AModule, public EventHandler {
 
   bool m_withIcon;
   uint64_t m_monitorId;
-  int m_activeWorkspaceId;
+  std::string m_activeWorkspaceAddress;
```
kolayne: Seems like the workspace identity can completely abstract parsing details, including the concept of address. Should we have the workspace identity object here and in the methods?

src/modules/hyprland/workspaces.cpp
```diff
@@ -888,39 +909,39 @@ void Workspaces::removeWorkspacesToRemove() {
...
-    }
-    return name == x->name();
+  // Not every identifier that reaches us is an address. A persistent-workspaces
+  // placeholder is keyed by its selector until it adopts a live identity, and
+  // legacy Hyprland identifies the generic special workspace as the literal
+  // string `special` in some payloads rather than by its id.
+  const auto selector = parseWorkspaceSelector(addressOrSelector);
+  return std::ranges::find_if(m_workspaces, [&](std::unique_ptr<Workspace> const& w) {
+    return w->kind() == selector.kind && w->isGenericSpecial() == selector.isGenericSpecial &&
+           w->name() == selector.name;
```
kolayne: This should likely be a method on either workspace or selector class.

src/modules/hyprland/workspace_identity.cpp
```diff
@@ -0,0 +1,179 @@
...
+std::string workspaceDisplayName(const std::string& rawName, WorkspaceKind kind) {
+  if (kind == WorkspaceKind::Special && rawName.starts_with(kSpecialPrefix)) {
+    return rawName.substr(kSpecialPrefix.size());
+  }
+  return rawName;
+}
+
+bool isGenericSpecialName(const std::string& rawName, WorkspaceKind kind) {
+  return kind == WorkspaceKind::Special && rawName == kGenericSpecial;
+}
```
kolayne: This is suspicious that you check for both. Should the first condition always imply the second one?

If so, I would write this as
```cpp
if (kind == WorkspaceKind::Special) {
    assert(rawName.starts_with(kSpecialPrefix));
    // ...
```

If there is no such implication (i.e., we may receive a `rawName` that has already been stripped of the `special:` prefix), then there should be a bug when the function is called for `special:special:123` if its prefix is stripped: the function will strip it again.

src/modules/hyprland/workspace_identity.cpp
```diff
@@ -0,0 +1,179 @@
...
+    return true;
+  }
+  // Hyprland's workspace events carry the identifier the workspace was created
+  // with, not the address `workspaces` reports. They agree for numbered and
+  // special workspaces, but a named one created as `name:web` is announced as
+  // `name:web` and reported at address `web`, so the raw comparison misses it.
+  const auto selector = parseWorkspaceSelector(identifier);
+  return identity.kind == selector.kind &&
+         isGenericSpecialName(rawName, identity.kind) == selector.isGenericSpecial &&
+         workspaceDisplayName(rawName, identity.kind) == selector.name;
```
kolayne: Should this equality check be a method of `WorkspaceIdentity`?

kolayne (commented):
Thank you for your fixes!

I had a fresh look and left some more comments.

include/modules/hyprland/workspace.hpp
tonybutt: You are right, and it is worse than a naming problem -- the distinction was not real, and asserting it was breaking a live case.

Checked it two ways. On 0.56.0 here, `toggle_special()` and `toggle_special("special")` both land on address `special:special`. And in Hyprland's source, `getWorkspaceIDNameFromString` has named the generic special workspace `special:special` since v0.41 (v0.30-v0.35 called it a bare `special`, which is where my claim came from -- two years and fifteen releases ago).

So `isGenericSpecial` is gone in 919939aa, along with the flag on `WorkspaceSelector` and the branch in `handleClicked()`. Clicking any special workspace now dispatches `togglespecialworkspace <display name>`; for the generic one that is `togglespecialworkspace special`, which resolves back to `special:special`.

Removing it fixes a bug I had shipped with it. `persistent-workspaces: { "special": [...] }` built its placeholder with the raw name `special` and the flag set, the live workspace arrives as `special:special` with the flag clear, and the kind+flag+name match never fired -- so the placeholder kept its synthetic address and every later address-keyed lookup missed the button. Verified against 0.56.0 before and after: `Persistent workspace special adopting live address special:special`, one button, correct active state.

The only behaviour lost is on Hyprland <= 0.40, where the two really were separate workspaces. Supporting both shapes needs version detection, and I would rather not carry that for a release that old -- happy to add it if you disagree.

src/modules/hyprland/workspace_identity.cpp
tonybutt: Correct, and the distinction matters here: `-Wreturn-type` fires precisely because `-Wswitch` exhaustiveness is not a proof about the value set. I had the guarantee backwards.

919939aa returns `"unknown"` and logs the numeric value:

```cpp
  spdlog::error("Unknown WorkspaceKind {}", static_cast<int>(kind));
  return "unknown";
```

`"unknown"` was the part I had wrong independently of the reasoning -- returning `"named"` makes a corrupt kind indistinguishable from a real named workspace at every downstream reader. Not aborting: the value feeds a `type` field that `parseWorkspaceIdentity` maps back to `Named`, so a bad cast degrades to a mislabelled button rather than anything unsafe.

include/modules/hyprland/workspace_identity.hpp
tonybutt: Split the two in 919939aa, on whether the parse can fail.

`parseWorkspaceSelector` became `explicit WorkspaceSelector(const std::string&)`. Every string names some workspace, so the parse is total and a constructor is the honest spelling.

`parseWorkspaceIdentity` stayed a free factory: it returns `std::optional`, because a payload can carry neither schema's identity fields, and a constructor cannot decline to construct without throwing. The header now says which rule each one is under.

`workspaceKindForAddress` I left free as well -- it returns a `WorkspaceKind`, so there is no object for it to be a constructor of.

Other functions did move onto the types, prompted by your later comments: `WorkspaceSelector::rawName()`, `WorkspaceIdentity::number()`, `asSelector()` and the two `matches()` overloads.

include/modules/hyprland/workspace_identity.hpp
tonybutt: Yes to the redundancy, no to the variant -- and the first half turned out to matter more than I expected.

C++20 (`cpp_std=c++20` in meson.build), so `std::variant` is available.

`number` is gone as a field; it is now `WorkspaceIdentity::number()`, read off the address. The two can no longer disagree. That also made the ordering tests honest: they used to fabricate a Numbered identity with `std::nullopt` for its number, which the type now refuses, so they construct addresses that genuinely do not parse.

On the variant: I do not think `variant<int, string>` is the right shape. It splits along the wrong axis -- a numbered workspace's number is derivable from its address, whereas named and special are both string-keyed and are distinguished by `kind`, which the variant cannot carry. The address is the one field that is a key in both IPC schemas, so it is the natural single source of truth, and `kind` is genuinely independent of it (a legacy payload's address is `-99` and tells you nothing).

What I did add is `name`, the display name. It is not derivable: modern Hyprland reports address `special:spotify` and name `special:spotify`, but legacy reports address `-99` and name `special:special`. It used to live on `Workspace` and be passed alongside the identity everywhere, which is how the comparators ended up with four parameters. `Workspace::m_name` is gone and the comparators take two identities.

include/modules/hyprland/workspaces.hpp
tonybutt: Renamed and routed through the identity in 919939aa, and it was hiding a bug.

The field never held an address despite its name: `onWorkspaceActivated` and `onMonitorFocused` store whatever identifier the event carried, which for a named workspace created as `name:web` is `name:web`, while `workspaces` reports its address as `web`. Comparing that string to `address()` in `updateWorkspaceStates()` could not match. It is now `m_activeWorkspaceIdentifier`, compared with `workspace->identity().matches(...)`, which tries the address and then the selector.

The same function had a second name-shaped comparison next to it: `workspace->name() == currentWorkspaceName` compared a display name against a raw IPC name, so it never fired for a special workspace. That payload now goes through `parseWorkspaceIdentity` and is matched as `identity().matches(activeIdentity->asSelector())`.

I stopped short of making the field a `WorkspaceIdentity`. What arrives is an identifier, not an identity -- there is no kind, and manufacturing one would mean guessing. Keeping it a string and letting the identity interpret it seemed the smaller lie.

src/modules/hyprland/workspaces.cpp
tonybutt: Moved in 919939aa. The lambda body is now `w->identity().matches(selector)`, and `WorkspaceIdentity::matches(const WorkspaceSelector&)` is the single place that knows kind and display name are what identify a workspace across the selector boundary.

Kept the two passes: address across the whole list first, then selector. Otherwise iteration order decides between a workspace whose address matches and an earlier one whose selector does, which is a coin flip rather than a rule.

src/modules/hyprland/workspace_identity.cpp
tonybutt: The implication does not hold, so the assert would fire -- but you are right that the code did not say why, and one of the two reasons it exists has now gone away.

Two callers reach it with `Special` and no prefix. Hyprland before 0.41 reported the generic special workspace as a bare `special` (`getWorkspaceIDNameFromString` set `outName = "special"` through v0.35). And `onSpecialWorkspaceActivated` passes `WorkspaceKind::Special` with the monitor's `specialWorkspace.name`, which is the empty string when no special workspace is open.

On your second branch: there is no double-strip, because the function is never applied twice. It maps a raw IPC name to a display name once, at the boundary. 919939aa makes that structural rather than a convention -- `WorkspaceIdentity` now stores the display name, so every later reader takes it from there and nothing re-derives it. The header comment says so explicitly, and says why the kind check is load-bearing rather than belt-and-braces:

```
// This is not idempotent, and is not meant to be: it maps a raw IPC name to a
// display name once. Nothing calls it on a name that has already been through
// it -- WorkspaceIdentity stores the display name, so every later reader takes
// it from there. The kind is still checked alongside the prefix because a
// Hyprland older than 0.41 reported the generic special workspace as a bare
// `special`, with no prefix to strip.
```

src/modules/hyprland/workspace_identity.cpp
tonybutt: Done in 919939aa. `workspaceMatchesIdentifier(identifier, identity, rawName)` is now two overloads on the type:

```cpp
bool matches(const WorkspaceSelector& selector) const {
  return kind == selector.kind && name == selector.name;
}
bool matches(const std::string& identifier) const {
  return address == identifier || matches(WorkspaceSelector{identifier});
}
```

The free function needed `rawName` as a third argument only to recover the display name and the generic-special flag. With the flag gone and the display name stored on the identity, there is nothing left to pass, which is what let it become a method.

tonybutt:
Pushed 919939aa for this round. The headline is that one of your questions was a bug report.

## `special` vs a special workspace named `special`

They are the same workspace, and claiming otherwise was breaking `persistent-workspaces: ["special"]`.

Two checks. On 0.56.0, `toggle_special()` and `toggle_special("special")` both land on address `special:special`. In Hyprland's source, `getWorkspaceIDNameFromString` has named it `special:special` since v0.41; v0.30-v0.35 used a bare `special`, which is where my claim came from and is two years stale.

The placeholder for a `special` entry was built with the raw name `special` and the generic flag set; the live workspace arrives as `special:special` with the flag clear; the match never fired, so the placeholder kept its synthetic address and every later address-keyed lookup missed the button on screen. `isGenericSpecial` is gone from `Workspace`, `WorkspaceSelector` and `handleClicked()`.

## Naming moved into the identity

`WorkspaceIdentity` now carries the display name next to the address and kind, derives `number()` from the address instead of storing it, and owns the matching callers used to open-code -- `matches(selector)`, `matches(identifier)`, `asSelector()`. `Workspace::m_name` and `m_isSpecial` are gone; the sort comparators take two identities instead of four arguments.

That consolidation surfaced a second bug. `m_activeWorkspaceAddress` never held an address: events carry the identifier a workspace was created with, so a named workspace created as `name:web` is announced as `name:web` while `workspaces` reports its address as `web`. It is now `m_activeWorkspaceIdentifier` and is compared through `matches()`. The `workspace->name() == currentWorkspaceName` check beside it compared a display name against a raw IPC name and so never fired for a special workspace; that payload now goes through `parseWorkspaceIdentity`.

## Smaller things

- `parseWorkspaceSelector` is now `explicit WorkspaceSelector(const std::string&)`. `parseWorkspaceIdentity` stays a factory because it returns `optional` -- the line I drew is whether the parse can fail.
- `workspaceTypeName()` returns `"unknown"` and logs. You were right that `-Wswitch` exhaustiveness is not a proof about the value set, and `"named"` made a corrupt kind indistinguishable from a real one.
- The `kind == Special && starts_with("special:")` pair both matter: a pre-0.41 generic special has no prefix, and `onSpecialWorkspaceActivated` passes an empty name when nothing is open. There is no double-strip, because the identity stores the display name and nothing re-derives it -- now stated in the header rather than left as a convention.

## Verification

40 cases, 673 assertions, passing. Runtime on Hyprland 0.56.0 with `persistent-workspaces: { "special", "name:scratch", "9" }`: both non-numbered placeholders adopt their live address (`Persistent workspace special adopting live address special:special`), the `special` button activates in place with no duplicate, ordering stays numbered -> named -> special. The legacy path remains unit-test-only.

Still open from your list: whether to keep supporting Hyprland <= 0.40, where the generic special workspace and one named `special` really were distinct. Doing both shapes needs version detection. I left it out; say the word if you would rather have it.

vp1981:
I don't know the exact reason, but now WS#10 is after WS#1 and before WS#2. It used to be the last one.

Current Hyprland version is:
```
$ hyprctl version -j
{
    "branch": "main",
    "commit": "83cf6a6ed540dc37808434259c6a3ba663de9616",
    "version": "0.56.0",
    "dirty": false,
    "commit_message": "ci/discussions: regex tweak ( 16307)",
    "commit_date": "Sat Sep 19 19:29:53 2026",
    "tag": "v0.56.0-198-g83cf6a6ed",
    "commits": "7823",
    "buildAquamarine": "0.15.1",
    "buildHyprlang": "0.6.8",
    "buildHyprutils": "0.14.2",
    "buildHyprcursor": "0.1.13",
    "buildHyprgraphics": "0.5.1",
    "systemAquamarine": "0.15.1",
    "systemHyprlang": "0.6.8",
    "systemHyprutils": "0.14.2",
    "systemHyprcursor": "0.1.13",
    "systemHyprgraphics": "0.5.1",
    "abiHash": "83cf6a6ed540dc37808434259c6a3ba663de9616_aq_0.15_hu_0.14_hg_0.5_hc_0.1_hlg_0.6",
    "flags": []
}
```

List of WSes:
```
$ hyprctl workspaces -j
[{
    "address": "1",
    "id": 1,
    "type": "normal",
    "name": "1",
    "monitor": "DP-1",
    "monitorID": 0,
    "windows": 0,
    "hasfullscreen": false,
    "lastwindow": "0x0",
    "lastwindowtitle": "",
    "ispersistent": true,
    "tiledLayout": "dwindle"
},{
    "address": "2",
    "id": 2,
    "type": "normal",
    "name": "2",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 2,
    "hasfullscreen": false,
    "lastwindow": "0x563a87ebac50",
    "lastwindowtitle": "vladimir@smoon2:~",
    "ispersistent": true,
    "tiledLayout": "dwindle"
},{
    "address": "special:stash",
    "type": "special",
    "name": "special:stash",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 1,
    "hasfullscreen": false,
    "lastwindow": "0x563a87c44a80",
    "lastwindowtitle": "Telegram",
    "ispersistent": false,
    "tiledLayout": "dwindle"
},{
    "address": "special:magic",
    "type": "special",
    "name": "special:magic",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 2,
    "hasfullscreen": false,
    "lastwindow": "0x563a87c4da10",
    "lastwindowtitle": "vladimir@smoon2:/media/storage/musics",
    "ispersistent": false,
    "tiledLayout": "dwindle"
},{
    "address": "5",
    "id": 5,
    "type": "normal",
    "name": "5",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 1,
    "hasfullscreen": false,
    "lastwindow": "0x563a87d6a960",
    "lastwindowtitle": "run-m4-main.sh (~/projects/neutrino/magexp/bin) - VIM",
    "ispersistent": false,
    "tiledLayout": "dwindle"
},{
    "address": "10",
    "id": 10,
    "type": "normal",
    "name": "10",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 1,
    "hasfullscreen": false,
    "lastwindow": "0x563a87dfadd0",
    "lastwindowtitle": "vladimir@smoon2:~",
    "ispersistent": false,
    "tiledLayout": "dwindle"
},{
    "address": "6",
    "id": 6,
    "type": "normal",
    "name": "6",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 1,
    "hasfullscreen": false,
    "lastwindow": "0x563a87ddfc60",
    "lastwindowtitle": "vladimir@smoon7:~",
    "ispersistent": false,
    "tiledLayout": "monocle"
},{
    "address": "special:qterm-alacritty",
    "type": "special",
    "name": "special:qterm-alacritty",
    "monitor": "HDMI-A-1",
    "monitorID": 1,
    "windows": 1,
    "hasfullscreen": false,
    "lastwindow": "0x563a87cf53f0",
    "lastwindowtitle": "vladimir@smoon2:~",
    "ispersistent": false,
    "tiledLayout": "dwindle"
}]
```

test/hyprland/workspace_order.cpp
```diff
@@ -0,0 +1,202 @@
...
+TEST_CASE("sort-by default is a strict weak ordering", "[workspace_order]") {
+  requireStrictWeakOrdering(&workspaceLessByDefault, mixedPopulation());
+}
+
+TEST_CASE("default ordering groups numbered then named then special", "[workspace_order]") {
+  std::vector<WorkspaceIdentity> entries{
+      special("special:spotify", "spotify"),
+      named("web"),
+      numbered(10),
+      numbered(2),
```
kolayne: Let's also add `1` to make sure it is ordered before 10 (@vp1981's issue).

src/modules/hyprland/workspace.cpp
```diff
@@ -230,10 +224,26 @@ bool Workspace::handleClicked(GdkEventButton* bt) const {
   return false;
 }
 
+void Workspace::setAddress(std::string const& value) {
+  // The kind must come from the new address, not from the identity being
+  // replaced: a renumbering event can move a workspace between the numbered and
+  // named namespaces. The display name is unaffected and carries over.
+  m_identity.address = value;
+  m_identity.kind = workspaceKindForAddress(value);
+}
```
kolayne: Shouldn't this also update `m_identity.name`?

Is `WorkspaceIdentity::name` redundant when we have `WorkspaceIdentity::address`?

tonybutt:
Thanks for the report @vp1981 and the hyprctl dump [hyprwm/Hyprland#16269](https://github.com/hyprwm/Hyprland/issues/16269) (merged Sep 17) changed the workspace type from numbered/named/special to normal/special, and now tells numbered and named workspaces apart by whether id is present. The parser only knew numbered, so every numbered workspace was read as named and sorted as text (1 < 10 < 2). The latest push reads both formats, and your output is now a test case. Could you check it again on your build?

test/hyprland/workspace_order.cpp
tonybutt: Added 1, and the test now expects 1, 2, 10, web, spotify. That said, this comparator was already numeric and wouldn't have caught vp1981's issue. The actual bug was in parsing: [hyprwm/Hyprland#16269](https://github.com/hyprwm/Hyprland/issues/16269) renamed numbered to normal, so numbered workspaces were read as named and compared by name. That's fixed in parseWorkspaceIdentity, with a test.

src/modules/hyprland/workspace.cpp
tonybutt: Hyprland's CHLWorkspace::changeID sets the display name to the new number unless the workspace was renamed (if (!m_wasRenamed) m_name = ...), so we now do the same: if the name still equals the old address it follows the new one, otherwise it's kept. This lives in WorkspaceIdentity::renumber() so it's tested. The same code shows changeworkspaceid only fires for numbered workspaces.

 On name vs address: they're not redundant. address is the key we match IPC events against, and name is what's displayed. They differ for a renamed numbered workspace (renameworkspace 3 code gives address 3, name code), for special workspaces (special:magic vs magic), and on Hyprland before addressable workspaces, where the address is the stringified id (e.g. -1377). Neither can be derived from the other, which is why number() is derived from address but name is stored.

src/modules/hyprland/workspace_identity.cpp
```diff
@@ -0,0 +1,191 @@
...
+  spdlog::error("Unknown WorkspaceKind {}", static_cast<int>(kind));
+  return "unknown";
+}
+
+WorkspaceKind workspaceKindForAddress(const std::string& address) {
+  if (isSpecialName(address)) {
+    return WorkspaceKind::Special;
+  }
+  if (!address.empty() &&
+      std::ranges::all_of(address, [](unsigned char c) { return std::isdigit(c) != 0; })) {
```
kolayne: Alternatively,

```suggestion
  if (parse_number(address).has_value()) {
```

kolayne (approved):
Overall LGTM!
(I did not review tests)

